PR #198, opened 2019-03-04
Support CSV loader

darionyaphet:
#114 [[tools] Support CSV Format Loader](https://github.com/vesoft-inc/nebula/issues/114)

syntax：

* `load vertex [vertex-csv-file]`

* `load edge [edge-csv-file]`


dangleptr:
We have importer now.  The pr is useless now.

PR #208, opened 2019-03-11
add spark sst file generator job to generate sst files

spacewalkman:
A spark job which does the following things:

1. parsing an input mapping file to map a hive table to a tag/edge, in which the table's PK(logically) should be identified
2. use nebula native client to encode a tag's key and values
3. define a custom hadoop OutputFormat and RecordWriter, which should generate a sub dir for one partition per worker  in specified sst file output dir


dutor:
@darionyaphet @dangleptr @sherman-the-tank 
Please take a look at this, to end the long journey.

darionyaphet:
I have already commented.It seems need some modify before proceeding to the next step. @dutor

src/tools/spark-sstfile-generator/README.md
```diff
@@ -0,0 +1,159 @@
+Generate sst files from hive tables datasource, guided by a mapping file, which maps hive tables to vertexes and edges,
```
dangleptr: Well done for the doc.

src/tools/spark-sstfile-generator/README.md
```diff
@@ -0,0 +1,159 @@
...
+      }
+    ]
+  }
+}
+```
+
+# FAQ
+## How to use libnebula-native-client.so under CentOS6.5(2.6.32-431 x86-64)
+
+1. build rocksdbjava natively under CentOS6.5, which will generate librocksdbjni-linux64.so
```
dangleptr: The format is missed up.

src/tools/spark-sstfile-generator/README.md
dangleptr: Use period instead of comma after "edges"

src/tools/spark-sstfile-generator/README.md
```diff
@@ -0,0 +1,159 @@
...
+ -mi,--mapping_file_input <arg>              hive tables to nebula graph schema mapping file
+ -pi,--date_partition_input <arg>            table partitioned by a date field
+ -ri,--repartition_number_input <arg>        repartition number
+ -so,--sst_file_output <arg>                 where the generated sst files will be put, must be local file
+ -ti,--datasource_type_input <arg>           data source types support, support only hive in so far, default=hive
+```
+
+# Mapping file format
+
+When used in production, should be modified according to your scenario, and strip the comment, for the json parser used here lacks the capability of handling comments
```
dangleptr: it should be

spacewalkman:
https://github.com/vesoft-inc/nebula/pull/208#discussion_r264566791

Yes, in convention, a short option is just a single letter, but there are so many of them in this spark job. I want to use suffix _i_ and _o_ to distinguish INPUT and OUTPUT  option respectively, which is clearer,  WDYT?

spacewalkman:
https://github.com/vesoft-inc/nebula/pull/208#discussion_r264562151

IMO, sbt is the defacto-standard build tool is scala world. It bring extra inconveniences to change to maven.

spacewalkman:
About the comments which recommends using 3rd party lib [scala-FNV](https://github.com/philipsoutham/scala-fnv) , After thinking twice, i think that maybe it is not a good choice. The cons are as following:
1. It is maintained by a single person and claims a personal copyRight, would introduce some LICENSE conflict in future.
2. The  [core code](https://github.com/philipsoutham/scala-fnv/blob/master/src/main/scala/com/philipsoutham/hash/FNV.scala) is much the same as what we do now

spacewalkman:
I'll reopen a new PR after nebula repo change to PUBLIC

spacewalkman:
New PR#420  is [HERE](https://github.com/vesoft-inc/nebula/pull/420)

PR #256, opened 2019-04-03
Make query bound related interfaces asynchronously

dangleptr:
Process vertices requested simultaneously


dangleptr:
Please refer #392

Issue #284, opened 2019-04-18
Force meta client to refresh its local cache


dangleptr:
To improve the performance,  we use local cache inside meta-client.  
It is refreshed at intervals.   So we can't ensure the timeliness. 

We should figure out one mechanism to force the meta client  refresh the cache.

PR #303, opened 2019-04-24
refresh cache

steppenwolfyuetong:
#284 Force meta client to refresh its local cach

 implement `refresh cache`
 It can be improved in the future.


dutor:
How is this going?

@dangleptr @sherman-the-tank

steppenwolfyuetong:
> How is this going?
> 
> @dangleptr @sherman-the-tank

I think there are two ways to choose:
1) Implement the `refresh local meta cache `first, and implement the `refresh global meta cache` in the future.
2) directly implement the `refresh  (global)  meta cache`

dangleptr:
We need to redesign the feature.  Let's close it firstly.

PR #313, opened 2019-04-28
Support scan key and value in meta

darionyaphet:
* Support scan key and value processor.
* fix scan to support key and value .
* add some small improve


src/meta/test/ProcessorTest.cpp
```diff
@@ -4,7 +4,6 @@
  *  (found in the LICENSE.Apache file in the root directory)
  */
 
-#include "base/Base.h"
```
dutor: Why did you remove this?

dutor:
How is going?

darionyaphet:
> How is going?

I have rebased this one and reopen it later :)  thanks @dutor  🎉

src/meta/test/ProcessorTest.cpp
darionyaphet: duplicate included, `base/Base.h` have imported by `src/meta/processors/BaseProcessor.h`

PR #316, opened 2019-04-29
move the KeyUtils from storage to common/base named NebulaKeyUtils

zhangguoqing:
The reason for doing this refactor is that other modules will use it
in the future, so it is more suitable as a basic object.
In addition, the getTagId function has been added.


dangleptr:
Jenkins go

PR #330, opened 2019-05-07
Support Time-to-live in query(syntax)

steppenwolfyuetong:
1.  support Time-to-live  in  syntax
2. also extend to support other schema options
3. improve create/alter edge for using  TTL


steppenwolfyuetong:
Close this PR, use new PR #403.

Issue #327, opened 2019-05-07
Implement the unreserved keywords in parser module.


boshengchen:
We should support unreserved keywords in parser module. To fix the syntax error of the keyword as LABEL. for example: 
A keyword KW_EMAIL exist. below statement will be return a syntax error .
`CREATE TAG person(name string, EMAIL string)`

PR #360, opened 2019-05-13
Support unreserved keywords.

boshengchen:
refer to issue #327

### DIFF
--- a/src/parser/parser.yy
+++ b/src/parser/parser.yy
@@ -177,15 +177,7 @@ unreserved_keyword
     | KW_TAG                { $$ = new std::string("TAG"); }
     | KW_VERTEX             { $$ = new std::string("VERTEX"); }
     | KW_EDGE               { $$ = new std::string("EDGE"); }
-    | KW_INT                { $$ = new std::string("INT"); }
-    | KW_BIGINT             { $$ = new std::string("BIGINT"); }
-    | KW_DOUBLE             { $$ = new std::string("DOUBLE"); }
-    | KW_STRING             { $$ = new std::string("STRING"); }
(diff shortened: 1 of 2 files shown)


dutor:
> >   **I actually thought about it，The yy_destroy_  garbage collector reclaims memory used by objects that can never again be accessed.  And the yy_destroy_  will be executed when the program completes or if there is parse error. I didn't find other memory leaks that might have caused. Could help me analyze it ?**
> 
> ```
>  inline
>   void
>    GraphParser ::yy_destroy_ (const char* yymsg, basic_symbol<Base>& yysym) const
>   {
>     if (yymsg)
>       YY_SYMBOL_PRINT (yymsg, yysym);
> 
>     // User destructor.
>  switch (yysym.type_get ())
>     {
> ......
>       case 118: // name_label
>         { delete (yysym.value.strval); }
>         break;
>       case 119: // unreserved_keyword
>         { delete (yysym.value.strval); }
>         break;
> .......
> }
> ```

The so-called destructor are generated by the `destructor` directives. If I were correct, that will not cover cases here, but I will verify this.

Please hold on for a while.

dutor:
Sorry for the long latency.

I have verify the memory leaks problem mentioned above. It's definitely true.

BTW. This PR is obsolete now because your own fork had been detached with upstream. Could you reopen a new one?

boshengchen:
Sure, let me reopen a PR instead of this one. Wait a moment please.

boshengchen:
Opened a new PR #369. Close this one.

boshengchen:
Opened a new PR #396. Close this one.

PR #363, opened 2019-05-13
Cut-off the relationship between dataMan and Meta-client. (Fix #234)

### LINKED ISSUE #234: Cut-off the relationship between RowReader and Meta-client.
There is another problem about module dependency.  
Normally, we don't want RowReader has any relationship with Meta client.   But due to the static methods in RowReader,  RowReader has to depend on meta client module.  It is no sense.  

Let's refactor it in another issue.

_Originally posted by @dangleptr in https://github.com/vesoft-inc/nebula/pull/233#issuecomment-476105966_

dangleptr:


darionyaphet:
Jenkins go

dangleptr:
Please refer #390

Issue #283, opened 2019-04-18
Discover storage service automatically on meta server-side.


dangleptr:
Currently,  we use "addHost" to register all storage daemons after init the whole cluster. 
We should support discover storage service automatically in meta-server side.  

Subtask of #182

sherman-the-tank:
Is this still valid?

The current solution is to use "ADD HOST" manually maintain the cluster and use the heartbeat to automatically detect the live storage service

PR #379, opened 2019-05-15
Adjust meta client ctor  && implement heartbeat logic inside.

dangleptr:
1. Force meta client to reuse IOThreadPool outside.
2. Move meta_server_addrs flag outside, so we don't check the flag in MetaClient ctor any more.
3. Implement heartbeat logic.  

Subtask of #283

### DIFF
--- a/src/daemons/StorageDaemon.cpp
+++ b/src/daemons/StorageDaemon.cpp
@@ -24,6 +24,9 @@ DEFINE_string(local_ip, "", "Local ip speicified for NetworkUtils::getLocalIP");
 DEFINE_bool(mock_server, true, "start mock server");
 DEFINE_bool(daemonize, true, "Whether to run the process as a daemon");
 DEFINE_string(pid_file, "pids/nebula-storaged.pid", "");
+DEFINE_string(meta_server_addrs, "", "list of meta server addresses,"
+                                     "the format looks like ip1:port1, ip2:port2, ip3:port3");
+DEFINE_int32(io_handlers, 10, "io handlers");
 
 using nebula::Status;
 
@@ -72,7 +75,7 @@ int main(int argc, char *argv[]) {
     }
 
     if (FLAGS_data_path.empty()) {
-        LOG(FATAL) << "Storage Data Path should not empty";
+        LOG(ERROR) << "Storage Data Path should not empty";
         return -1;
     }
     LOG(INFO) << "Starting the storage Daemon on port " << FLAGS_port
@@ -88,7 +91,15 @@ int main(int argc, char *argv[]) {
     uint32_t localIP;
     CHECK(NetworkUtils::ipv4ToInt(result.value(), localIP));
 
-    auto metaClient = std::make_unique<nebula::meta::MetaClient>();
+    if (FLAGS_meta_server_addrs.empty()) {
+        LOG(ERROR) << "meta_server_addrs flag should be set!";
+        return -1;
+    }
+    auto ioThreadPool = std::make_shared<folly::IOThreadPoolExecutor>(FLAGS_io_handlers);
+    auto addrs = nebula::network::NetworkUtils::toHosts(FLAGS_meta_server_addrs);
+    auto metaClient = std::make_unique<nebula::meta::MetaClient>(ioThreadPool,
+                                                                 std::move(addrs),
+                                                                 true);
     metaClient->init();
 
     nebula::kvstore::KVOptions options;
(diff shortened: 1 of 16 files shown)


src/daemons/StorageDaemon.cpp
```diff
@@ -88,7 +91,15 @@ int main(int argc, char *argv[]) {
     uint32_t localIP;
     CHECK(NetworkUtils::ipv4ToInt(result.value(), localIP));
 
-    auto metaClient = std::make_unique<nebula::meta::MetaClient>();
+    if (FLAGS_meta_server_addrs.empty()) {
+        LOG(ERROR) << "meta_server_addrs flag should be set!";
+        return -1;
+    }
+    auto ioThreadPool = std::make_shared<folly::IOThreadPoolExecutor>(FLAGS_io_handlers);
```
dutor: Could we share the same pool with the hosting service?

src/daemons/StorageDaemon.cpp
```diff
@@ -114,7 +125,10 @@ int main(int argc, char *argv[]) {
 
     auto handler = std::make_shared<StorageServiceHandler>(kvstore.get(), std::move(schemaMan));
     gServer = std::make_unique<apache::thrift::ThriftServer>();
-    CHECK(!!gServer) << "Failed to create the thrift server";
+    if (!!gServer) {
```
dutor: By default, `make_unique` could never return a nullptr, instead it will throw bad_alloc on memory allocation failure.

src/daemons/StorageDaemon.cpp
dangleptr: Hosting service?  If it means the thrift service,  yes of course.  Please note line 143.

src/storage/test/StorageClientTest.cpp
```diff
@@ -28,26 +28,29 @@ TEST(StorageClientTest, VerticesInterfacesTest) {
...
     std::string dataPath = folly::stringPrintf("%s/data", rootPath.path());
-    auto sc = TestUtils::mockServer(dataPath.c_str(), localIp, localDataPort);
+    auto sc = TestUtils::mockServer(mClient.get(), dataPath.c_str(), localIp, localDataPort);
 
     LOG(INFO) << "Add hosts and create space....";
-    auto mClient = std::make_unique<meta::MetaClient>();
-    mClient->init();
-    auto r = mClient->addHosts({HostAddr(localIp, localDataPort)}).get();
+
+        auto r = mClient->addHosts({HostAddr(localIp, localDataPort)}).get();
```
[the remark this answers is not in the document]
dangleptr: Good catch

src/daemons/StorageDaemon.cpp
dangleptr: Thanks for your reminding.    Good point.  👍

src/meta/client/MetaClient.cpp
```diff
@@ -10,25 +10,18 @@
 #include "meta/NebulaSchemaProvider.h"
 
 DEFINE_int32(load_data_interval_second, 2 * 60, "Load data interval, unit: second");
-DEFINE_string(meta_server_addrs, "", "list of meta server addresses,"
-                                     "the format looks like ip1:port1, ip2:port2, ip3:port3");
-DEFINE_int32(meta_client_io_threads, 3, "meta client io threads");
+DEFINE_int32(heartbeat_interval_sec, 10, "Heartbeat interval, unit: second");
```
dutor: secs or seconds? Maybe we should unify them, e.g. the above line.

src/meta/client/MetaClient.cpp
```diff
@@ -824,5 +843,20 @@ SchemaVer MetaClient::getNewestEdgeVerFromCache(const GraphSpaceID& space,
...
+    CHECK_NOTNULL(listener_);
+    auto localHost = listener_->getLocalHost();
+    cpp2::HBReq req;
+    nebula::cpp2::HostAddr thriftHost;
+    thriftHost.set_ip(localHost.first);
+    thriftHost.set_port(localHost.second);
+    req.set_host(std::move(thriftHost));
+    return getResponse(std::move(req), [] (auto client, auto request) {
+        return client->future_heartBeat(request);
+    }, [] (cpp2::HBResp&& resp) -> decltype(auto) {
```
dutor: FYI. I sincerely suggest not to use `decltype(auto)` if not necessary while `auto` suffices, because you may return a reference to local storage by accident. Consider this:
```cpp
decltype(auto) get() {
    int var = 0;
    int &ref = var;
    return ref;
    // or return (num);
    // or return (num &= 0xF)
}
```

dangleptr: You are right.  We should be careful about it.

dangleptr:
Please refer #383

PR #380, opened 2019-05-15
Fixed unstable tests on ProcessUtils

dutor:
As title.


src/common/process/test/ProcessTest.cpp
```diff
@@ -43,12 +43,15 @@ TEST(ProcessUtils, isPidAvailable) {
     }
     {
         // pid file which contains pid of current process
-        auto pidFile = "/tmp/maybe-non-existing-dir/process_test.pid";
+        auto pidFile = folly::stringPrintf("/tmp/non-existing-dir-%d/process_test.pid", ::getpid());
```
dangleptr: Why not TempDir?

dutor: I had considered it.

But TempDir for TempFile will create the directory or file and be capable of deleting them automatically. But what I need here is a temp name and auto deletion, but not the file creation.

dutor:
> Unit testing passed.

What the hell.

src/common/process/test/ProcessTest.cpp
dangleptr: Fair enough

PR #381, opened 2019-05-15
Use ephemeral port to bind on in tests

### LINKED ISSUE #260: Use ephemeral port to bind on in tests
We should not allow any obvious unstable cases in our testing.

steppenwolfyuetong:
Use ephemeral port to bind on in tests #260

Sometimes you need a program to bind to a port that can't be hard-coded. Generally this is when you want to run several of them in parallel; if they all bind to port a specified port, only one of them can succeed.

so use system unused ephemeral port 

This will close #260

### DIFF
--- a/src/interface/storage.thrift
+++ b/src/interface/storage.thrift
@@ -148,7 +148,7 @@ struct AddVerticesRequest {
     1: common.GraphSpaceID space_id,
     // partId => vertices
     2: map<common.PartitionID, list<Vertex>>(cpp.template = "std::unordered_map") parts,
-    // If true, it equals an upset operation.
+    // If true, it equals an upsert operation.
     3: bool overwritable,
 }
 
@@ -156,7 +156,7 @@ struct AddEdgesRequest {
     1: common.GraphSpaceID space_id,
     // partId => edges
     2: map<common.PartitionID, list<Edge>>(cpp.template = "std::unordered_map") parts,
-    // If true, it equals an upset operation.
+    // If true, it equals an upsert operation.
     3: bool overwritable,
 }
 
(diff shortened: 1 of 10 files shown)


steppenwolfyuetong:
> This is a better way than the one in the repo. But it's not the perfect one which I've been expecting.
> 
> We should pass 0 as the port number to bind on, in which case the OS would choose an available port for us. Then we could retrieve the actually bound port through other interfaces(fortunately, both proxygen and thrift provide such interfaces).
> 
> You could investigate such examples in the UTs of the graph and webservice module.

Excellent. I will study it.

steppenwolfyuetong:
> This is a better way than the one in the repo. But it's not the perfect one which I've been expecting.
> 
> We should pass 0 as the port number to bind on, in which case the OS would choose an available port for us. Then we could retrieve the actually bound port through other interfaces(fortunately, both proxygen and thrift provide such interfaces).
> 
> You could investigate such examples in the UTs of the graph and webservice module.

👍

steppenwolfyuetong:
Fix and rebase,  please review, thx.

src/interface/storage.thrift
```diff
@@ -148,15 +148,15 @@ struct AddVerticesRequest {
     1: common.GraphSpaceID space_id,
     // partId => vertices
     2: map<common.PartitionID, list<Vertex>>(cpp.template = "std::unordered_map") parts,
-    // If true, it equals an upsert operation.
+    // If true, it equals an upset operation.
```
dutor: *upsert* means update or insert.

steppenwolfyuetong: My bad.   👍👍

laura-ding:
jenkins go

steppenwolfyuetong:
Fix conficts  and rebase master.

dutor:
Jenkins go!

PR #383, opened 2019-05-15
Adjust meta client ctor  && implement heartbeat logic inside.

dangleptr:
1. Force meta client to reuse IOThreadPool outside.
2. Move meta_server_addrs flag outside, so we don't check the flag in MetaClient ctor any more.
3. Implement heartbeat logic.

Subtask of #283

After the repo to be public,  changes could NOT be rebased on the original pr #379 
So Let's use the new one to track it.

### DIFF
--- a/src/daemons/StorageDaemon.cpp
+++ b/src/daemons/StorageDaemon.cpp
@@ -85,31 +85,37 @@ int main(int argc, char *argv[]) {
                    << ", status:" << result.status();
         return EXIT_FAILURE;
     }
-    uint32_t localIP;
-    if (!NetworkUtils::ipv4ToInt(result.value(), localIP)) {
-        LOG(ERROR) << "Convert ip failed, ipstr:" << result.value();
+    auto hostRet = nebula::network::NetworkUtils::toHostAddr(result.value(), FLAGS_port);
+    if (!hostRet.ok()) {
+        LOG(ERROR) << "Bad local host addr, status:" << hostRet.status();
+        return EXIT_FAILURE;
+    }
+    auto& localHost = hostRet.value();
+    auto metaAddrsRet = nebula::network::NetworkUtils::toHosts(FLAGS_meta_server_addrs);
+    if (!metaAddrsRet.ok() || metaAddrsRet.value().empty()) {
+        LOG(ERROR) << "Can't get metaServer address, status:" << metaAddrsRet.status()
+                   << ", FLAGS_meta_server_addrs:" << FLAGS_meta_server_addrs;
         return EXIT_FAILURE;
     }
 
-    if (FLAGS_meta_server_addrs.empty()) {
-        LOG(ERROR) << "meta_server_addrs flag should be set!";
-        return EXIT_FAILURE;
-    }
     std::vector<std::string> paths;
     folly::split(",", FLAGS_data_path, paths, true);
     std::transform(paths.begin(), paths.end(), paths.begin(), [](auto& p) {
         return folly::trimWhitespace(p).str();
     });
+    if (paths.empty()) {
+         LOG(ERROR) << "Bad data_path format:" << FLAGS_data_path;
+         return EXIT_FAILURE;
+    }
 
     auto ioThreadPool = std::make_shared<folly::IOThreadPoolExecutor>(FLAGS_io_handlers);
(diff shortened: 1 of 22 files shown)


src/daemons/StorageDaemon.cpp
```diff
@@ -72,23 +75,37 @@ int main(int argc, char *argv[]) {
...
+    }
+    uint32_t localIP;
+    if (!NetworkUtils::ipv4ToInt(result.value(), localIP)) {
+        LOG(ERROR) << "Convert ip failed, ipstr:" << result.value();
+        return EXIT_FAILURE;
     }
-    LOG(INFO) << "Starting the storage Daemon on port " << FLAGS_port
-              << ", dataPath " << FLAGS_data_path;
 
+    if (FLAGS_meta_server_addrs.empty()) {
```
laura-ding: We should add a regular to check format of FLAGS_meta_server_addrs？

src/graph/ExecutionEngine.cpp
```diff
@@ -22,7 +24,11 @@ ExecutionEngine::~ExecutionEngine() {
 
 
 Status ExecutionEngine::init(std::shared_ptr<folly::IOThreadPoolExecutor> ioExecutor) {
-    metaClient_ = std::make_unique<meta::MetaClient>();
+    if (FLAGS_meta_server_addrs.empty()) {
```
laura-ding: ditto

src/daemons/StorageDaemon.cpp
dangleptr: Good proposal.  We could check the format of server_address inside NetWorkUtils::toHosts

dutor: FYI. gflags supports some kind of *validator* that will be check against whenever a specific flag is being modified. We could utilize that at the first place for most of our flags. We could do that in future.

src/daemons/StorageDaemon.cpp
```diff
@@ -72,27 +75,47 @@ int main(int argc, char *argv[]) {
...
-    CHECK(result.ok()) << result.status();
-    uint32_t localIP;
-    CHECK(NetworkUtils::ipv4ToInt(result.value(), localIP));
+    if (paths.empty()) {
+         LOG(ERROR) << "Bad data_path format:" << FLAGS_data_path;
+         return EXIT_FAILURE;
+    }
 
-    auto metaClient = std::make_unique<nebula::meta::MetaClient>();
+    auto ioThreadPool = std::make_shared<folly::IOThreadPoolExecutor>(FLAGS_io_handlers);
```
dutor: We could share the same thread pool with the one in ThriftServer

src/meta/test/MetaClientTest.cpp
```diff
@@ -24,7 +26,7 @@ using nebula::cpp2::ValueType;
 using apache::thrift::FragileConstructor::FRAGILE;
 
 TEST(MetaClientTest, InterfacesTest) {
-    FLAGS_load_data_interval_second = 1;
+    FLAGS_load_data_interval_secs = 1;
```
dutor: As for these tests, I think there are two ways to make them faster. First, we could have shared the same MockServer instead of launch one per test. Second, seconds is appropriate for a polling interval, but too long for unit testing, while I cannot come up with a better way. But you still could impove it by `usleep(FLAGS_load_data_interval_secs * 1000000 + 100000/*100ms*/)` to save 0.9s per sleep.

Of course, we could leave these work to the next PR.

src/daemons/StorageDaemon.cpp
dangleptr: I have done it.  Please note line 151

dutor: Excellent.

dangleptr:
Jenkins go

src/meta/test/MetaClientTest.cpp
dangleptr: Sounds good

dangleptr:
Jenkins go

dangleptr:
Damn it!  Conflicts again.

dangleptr:
Jenkins go

dangleptr:
Jenkins go

PR #384, opened 2019-05-15
Support semicolon at the end of "exit" sentence

### LINKED ISSUE #375: Support semicolon at the end of "exit" sentence.
This is mysql style.  Try to follow it.  
At least,  it should not be error when i type the semicolon.

darionyaphet:
This PR fixed #375

### DIFF
--- a/src/console/CmdProcessor.cpp
+++ b/src/console/CmdProcessor.cpp
@@ -312,7 +312,8 @@ void CmdProcessor::printData(const cpp2::ExecutionResponse& resp,
 
 bool CmdProcessor::processClientCmd(folly::StringPiece cmd,
                                     bool& readyToExit) {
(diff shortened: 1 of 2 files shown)


src/console/CmdProcessor.cpp
```diff
@@ -381,6 +382,14 @@ bool CmdProcessor::process(folly::StringPiece cmd) {
     return true;
 }
 
+void CmdProcessor::normalize(folly::StringPiece &command) {
+    command  = folly::trimWhitespace(command);
+    if (command.endsWith(";")) {
```
dutor: Take care of multiple ending `;`s, e.g. `quit;;`

darionyaphet: Nice, I will using `while` to determine the loop ending.

PR #387, opened 2019-05-16
Reorganized all linters into the same directory

dutor:
As title. Now the C++ and Java linter or config are grouped into the `.linters` directory.


darionyaphet:
LGTM

Issue #348, opened 2019-05-10
To unify the naming of keywords


dutor:
Such as
  * Replace `OVERWRITE` with `IF NOT EXISTS`
  * In admin queries, use `DROP` instead of `REMOVE`, `CHANGE` instead of `SET`
  * Replace `FIND` with `LOOKUP`

sherman-the-tank:
```register/unregister``` +1

dutor:
`register/unregister` -1

dutor:
> `Host` is a special structure to represent the storage instance and differ from the other.
> 
> `register/unregister` could explain for join and leave the cluster, it's semantics more suitable.
> 
> I like it :) @dangleptr Do you have any suggestions? @dutor

I don't see any specialness in *Host*, please enlighten me. For me, register/unregister have no different meanings with add/drop, neither better nor worse.

sherman-the-tank:
Honestly I don't see big YES or NO here. I'll say let's use ADD/DROP HOST. Here are the reasons

1) They are short, so easy to type
2) Consistent with ADD/DROP USER

dangleptr:
> > `Host` is a special structure to represent the storage instance and differ from the other.
> > `register/unregister` could explain for join and leave the cluster, it's semantics more suitable.
> > I like it :) @dangleptr Do you have any suggestions? @dutor
> 
> I don't see any specialness in _Host_, please enlighten me. For me, register/unregister have no different meanings with add/drop, neither better nor worse.

Actually,  in our system,  the "add/remove hosts " command is not really add/remove a host into our cluster.  It is just register/unregister it in whitelist.  It will be used to filter hosts which send heart beat to Meta service.

So i think the accurate command name for it should be "Register/Unregister host in whitelist" or remove the whitelist and trust our users.





> Honestly I don't see big YES or NO here. I'll say let's use ADD/DROP HOST. Here are the reasons
> 
> 1. They are short, so easy to type
> 2. Consistent with ADD/DROP USER

1. I don't think there is any difference for typing between DROP and REMOVE.  
2. For USER commands,  we use CREATE/DROP now.

steppenwolfyuetong:
I think we can refer to the corresponding sql of the current popular database.

sherman-the-tank:
First we need to make the concept clear. We DO NOT have a white list. What we have is called CLUSTER. The host in the cluster might not serve due to all kinds of reason.

[...]

PR #391, opened 2019-05-16
[Reopen] Unify remove to drop in schema 

darionyaphet:
subtask of #348


src/graph/CMakeLists.txt
```diff
@@ -29,7 +29,7 @@ add_library(
     InterimResult.cpp
     VariableHolder.cpp
     AddHostsExecutor.cpp
-    RemoveHostsExecutor.cpp
+    DropHostsExecutor.cpp
```
dangleptr: Keep it as it be until we have a conclusion.

darionyaphet: I will keep `RemoveHosts` and update if there is progress.

src/meta/processors/DropTagProcessor.h
```diff
@@ -4,29 +4,30 @@
  * attached with Common Clause Condition 1.0, found in the LICENSES directory.
  */
 
-#ifndef NEBULA_GRAPH_REMOVETAGPROCESSOR_H
-#define NEBULA_GRAPH_REMOVETAGPROCESSOR_H
+#ifndef META_REMOVETAGPROCESSOR_H
```
laura-ding: DROP

src/meta/processors/DropTagProcessor.h
```diff
@@ -4,29 +4,30 @@
...
             : BaseProcessor<cpp2::ExecResp>(kvstore) {}
 
     StatusOr<std::vector<std::string>> getTagKeys(GraphSpaceID id, const std::string& tagName);
 };
 
 }  // namespace meta
 }  // namespace nebula
-#endif  // NEBULA_GRAPH_REMOVETAGPROCESSOR_H
+
+#endif  // META_REMOVETAGPROCESSOR_H
```
laura-ding: ditto

src/meta/processors/DropTagProcessor.h
darionyaphet: Good catch ~

src/meta/processors/DropEdgeProcessor.h
```diff
@@ -31,3 +31,4 @@ class RemoveEdgeProcessor : public BaseProcessor<cpp2::ExecResp> {
 }  // namespace nebula
 
 #endif  // META_REMOVEEDGEPROCESSOR_H_
```
laura-ding: DROP

PR #392, opened 2019-05-16
Make query bound related interfaces asynchronously

dangleptr:


src/storage/QueryBaseProcessor.cpp
```diff
@@ -0,0 +1,15 @@
+/* Copyright (c) 2018 - present, VE Software Inc. All rights reserved
```
darionyaphet: append new file's copyright should be 2019

dangleptr:
Jenkins go

dangleptr:
Any suggestions about this pr?  I need it to go on next one.   @dutor @sherman-the-tank

whitewum:
is this pr still on going? need any review?

dangleptr:
> is this pr still on going? need any review?

Let me rebase it.

src/storage/test/QueryBoundBenchmark.cpp
```diff
@@ -0,0 +1,186 @@
+/* Copyright (c) 2018 - present, VE Software Inc. All rights reserved
+ *
+ * This source code is licensed under Apache 2.0 License
+ *  (found in the LICENSE.Apache file in the root directory)
```
whitewum: don't forget to renew the license.

src/storage/test/QueryBoundBenchmark.cpp
```diff
@@ -0,0 +1,186 @@
...
+    req.set_return_columns(std::move(tmpColumns));
+    return req;
+}
+
+}  // namespace storage
+}  // namespace nebula
+
+void run(int32_t iters, int32_t handlerNum) {
+    FLAGS_max_handlers_per_req = handlerNum;
+    auto req = nebula::storage::buildRequest();
```
whitewum: buildRequest seems time consuming. separate to timing.

src/storage/test/QueryBoundBenchmark.cpp
```diff
@@ -0,0 +1,186 @@
...
+    run(iters, 10);
+}
+/*************************
+ * End of benchmarks
+ ************************/
+
+
+int main(int argc, char** argv) {
+    folly::init(&argc, &argv, true);
+    nebula::fs::TempDir rootPath("/tmp/QueryBoundBenchmarkTest.XXXXXX");
```
whitewum: I'm not sure this benchmark need to IO (I can't read it from context). if it is, HDD or SSD matters.

dangleptr: It needs to read data from rocksdb.  And the benchmark is based on SSD.

src/storage/test/QueryBoundBenchmark.cpp
dangleptr: Good point.

src/storage/test/QueryBoundBenchmark.cpp
dangleptr: Good catch.

src/storage/test/QueryBoundBenchmark.cpp
dangleptr: By the way, as the test data is not large enough,  all requests would hit the cache.

whitewum: a better way is to warm cache in a prepare step. and then do the benchmarks.

dangleptr: We just test the same request with single thread VS multi threads,  so no really needs with warming step.

src/storage/QueryBaseProcessor.inl
```diff
@@ -274,21 +323,31 @@ void QueryBaseProcessor<REQ, RESP>::process(const cpp2::GetNeighborsRequest& req
...
+                     this,
+                     returnColumnsNum] (auto&& t) mutable {
+        CHECK(!t.hasException());
+        std::unordered_set<PartitionID> failedParts;
+        for (auto& bucketTry : t.value()) {
+            CHECK(!bucketTry.hasException());
+            for (auto& r : bucketTry.value()) {
+                auto& partId = std::get<0>(r);
+                auto& ret = std::get<2>(r);
+                if (ret != kvstore::ResultCode::SUCCEEDED
```
whitewum: if anything fails, go on or break?
do you need to mock a failure?

src/storage/QueryBaseProcessor.inl
```diff
@@ -255,16 +254,66 @@ kvstore::ResultCode QueryBaseProcessor<REQ, RESP>::collectEdgeProps(
...
+template<typename REQ, typename RESP>
+int32_t QueryBaseProcessor<REQ, RESP>::getBucketsNum(int32_t verticesNum, int handlerNum) {
+    if (verticesNum < FLAGS_min_vertices_num_mp) {
+        return 1;
+    }
+    return std::min(verticesNum, handlerNum);
+}
+
+template<typename REQ, typename RESP>
+std::vector<Bucket> QueryBaseProcessor<REQ, RESP>::genBuckets(
```
whitewum: the benchmark and ut do not directly cover different vertexnum and max_handlers_per_req. 
the code looks fine. add a ut is better.

src/storage/test/QueryBoundBenchmark.cpp
```diff
@@ -0,0 +1,192 @@
...
+req_parts=1, vrpp=100, handler_num=10
+Notes: all requests hit the cache.
+
+============================================================================
+QueryBoundBenchmark.cpprelative                           time/iter  iters/s
+============================================================================
+query_bound_1                                              122.30ms     8.18
+query_bound_3                                               41.18ms    24.29
+query_bound_10                                              13.69ms    73.04
+============================================================================
```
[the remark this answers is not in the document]
dangleptr: Yes, the performance is linear under light load.  Under heavy load,  to avoid a big request hold all processors,  i have set the max handlers used limit with one flag.

src/storage/QueryBaseProcessor.inl
dangleptr: Currently,  when one point failed,  the failure will be kept inside and go on to deal with the other points.  In the end,  we cluster the failures by part and return it back to client.

src/storage/QueryBaseProcessor.inl
dangleptr: Good point.  I will add more factors for genBuckets method,  for example,  current io load, cpu usages etc.

src/storage/test/QueryBoundBenchmark.cpp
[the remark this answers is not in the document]
whitewum: We don't have enough stress tests. maybe a TODO to remember this.

src/storage/QueryBaseProcessor.inl
whitewum: this requires more performance counters. a TODO is enough now.

src/storage/test/QueryBoundBenchmark.cpp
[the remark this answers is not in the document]
dangleptr: Fair enough!

src/storage/QueryBaseProcessor.inl
dangleptr: Fair enough!

dangleptr: I have added some UTs for genBucket method.

src/storage/Collector.h
```diff
@@ -123,40 +123,48 @@ class StatsCollector : public Collector {
...
+    void collectDouble(ResultType ret, double v, const PropContext& prop) override {
         if (ret == ResultType::SUCCEEDED) {
+            std::lock_guard<std::mutex> lg(lock_);
             prop.sum_ = boost::get<double>(prop.sum_) + v;
             prop.count_++;
         }
     }
 
-    void collectString(ResultType ret, folly::StringPiece& v, PropContext& prop) override {
+    void collectString(ResultType ret, folly::StringPiece& v, const PropContext& prop) override {
```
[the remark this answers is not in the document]
dangleptr: That's because i don't want other fields be changed during collectXXX,   and the mutable fields will be moved to other places in the future.

src/storage/QueryBaseProcessor.inl
```diff
@@ -274,21 +325,31 @@ void QueryBaseProcessor<REQ, RESP>::process(const cpp2::GetNeighborsRequest& req
...
+        std::unordered_set<PartitionID> failedParts;
+        for (auto& bucketTry : t.value()) {
+            CHECK(!bucketTry.hasException());
+            for (auto& r : bucketTry.value()) {
+                auto& partId = std::get<0>(r);
+                auto& ret = std::get<2>(r);
+                if (ret != kvstore::ResultCode::SUCCEEDED
+                      && failedParts.find(partId) == failedParts.end()) {
+                    failedParts.emplace(partId);
+                    this->pushResultCode(this->to(ret), partId);
```
critical27: Maybe we can add a break here? Cause once a vertex in a part failed, the whole part would return failed.

dangleptr: We can't break here,  because for one bucket it may contains vertices from different parts.

dangleptr:
@critical27 Any concerns?

PR #393, opened 2019-05-16
modify cpplint.py's path

laura-ding:


dutor:
what a shame on me.

thanks for fixing.

dutor:
jenkins run tests

PR #396, opened 2019-05-17
Support unreserved keywords

boshengchen:
Instead of PR #360

### DIFF
--- a/src/parser/parser.yy
+++ b/src/parser/parser.yy
@@ -99,6 +99,7 @@ class GraphScanner;
 %token <doubleval> DOUBLE
 %token <strval> STRING VARIABLE LABEL
 
+%type <strval> name_label unreserved_keyword
 %type <expr> expression logic_or_expression logic_and_expression
 %type <expr> relational_expression multiplicative_expression additive_expression
 %type <expr> unary_expression primary_expression equality_expression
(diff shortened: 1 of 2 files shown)


src/parser/parser.yy
```diff
@@ -167,6 +168,31 @@ class GraphScanner;
...
+name_label
+     : LABEL { $$ = $1; }
+     | unreserved_keyword { $$ = $1; }
+     ;
+
+unreserved_keyword
+     : KW_SPACE              { $$ = new std::string("SPACE"); }
+     | KW_TAG                { $$ = new std::string("TAG"); }
+     | KW_VERTEX             { $$ = new std::string("VERTEX"); }
+     | KW_EDGE               { $$ = new std::string("EDGE"); }
```
dutor: IMO. We should reserve `TAG`, `VERTEX` and `EDGE`, since they are widely used in our data plane queries, in which case we care about performance.

boshengchen: Yeah, that makes sense.

dutor:
What I thought was wrong about the memory leaks. There is no leaks for the way you do this, because the memory allocation only happens when an unreserved keyword is reduced to a `name_label`.

Great one!

dutor:
Jenkins, please run tests.

boshengchen:
Thank you very much for your analysis @dutor .

boshengchen:
Rebased.

boshengchen:
> @boshengchen Please let me take care of the rebasing after approval next time.

I understand. I mistakenly thought that only I can rebase .

laura-ding:
jenkins go

laura-ding:
jenkins go

PR #399, opened 2019-05-17
[Reopen] Support scan key and value in meta

darionyaphet:
* Support scan key and value processor.
* Fix scan to support key and value .
* Add some small improve


src/interface/meta.thrift
```diff
@@ -277,9 +277,15 @@ struct ScanReq {
 }
 
 struct ScanResp {
-    1: ErrorCode code,
-    2: common.HostAddr  leader,
-    3: list<string> values,
+    1: ErrorCode                                                 code,
+    2: common.HostAddr                                           leader,
+    3: map<string, string>(cpp.template = "std::unordered_map")  values,
```
dangleptr: rename it "Result" ?

src/interface/meta.thrift
```diff
@@ -321,6 +327,8 @@ service MetaService {
     ExecResp remove(1: RemoveReq req);
     ExecResp removeRange(1: RemoveRangeReq req);
     ScanResp scan(1: ScanReq req);
+    KeyOrValueScanResp scanKey(1: ScanReq req);
```
dangleptr: Do we really need the two interfaces?

src/client/cpp/GraphClient.cpp
```diff
@@ -4,7 +4,6 @@
  * attached with Common Clause Condition 1.0, found in the LICENSES directory.
  */
 
-#include "base/Base.h"
```
dangleptr: Why remove it?

src/common/base/Status.h
```diff
@@ -98,10 +98,7 @@ class Status final {
     STATUS_GENERATOR(SyntaxError);
 
     // TODO(dangleptr) we could use ErrorOr to replace SpaceNotFound here.
-    STATUS_GENERATOR(SpaceNotFound);
```
dangleptr: Why?

src/interface/meta.thrift
darionyaphet: Actually, I considered use a flag to distinguish scan the keys and the values.

But I find it‘s confuse.

src/interface/meta.thrift
darionyaphet: fixed

src/common/base/Status.h
darionyaphet: We have remove them from `meta.thrift`'s `ErrorCode` and replace with `E_NOT_FOUND`

src/client/cpp/GraphClient.cpp
darionyaphet: duplicate imported from another `.h`

src/interface/meta.thrift
dangleptr: I mean do we really need to scan key or value only? 
There is no efficient problems for the interfaces in meta.

src/common/base/Status.h
dangleptr: Why remove them?  Any benefits?

whitewum:
is it still on going?

PR #257, opened 2019-04-03
user management in Meta Server

boshengchen:
related issue #180.

Issue #190, opened 2019-02-26
Access rights control


sherman-the-tank:
Need to establish the entire access rights control mechanism. Here is the current thoughts

1) There is one **God** user ```admin```, like `root` of Linux systems. ```admin``` is the **only** person who can create a Graph Space 

2) All other access rights are based on Graph Space. There are three types of rights for each Graph Space: ```Admin```, ```User```, ```Guest```

```Admin``` is an administrator for a given Graph Space. An ```Admin``` can delete the Graph Space, can manage the schema in the Graph Space, and can access (read and write) the data in the Graph Space

```User``` is a normal user for a given Graph Space. A ```User``` can access (read and write) the data in the Graph Space

```Guest``` is a read-only role for a given Graph Space. A ```Guest``` **cannot** modify the data in the Graph Space

3) A normal user will associate with a default Graph Space, which is the current Graph Space as soon as the user logs on

4) A normal user can have different access right on different Graph Spaces

boshengchen:
Below is operation and permission define. Any suggestions?

### Operation and permission define:

Operation | GOD | ADMIN | USER | GUEST
------------ | ------------- | ------------- | ------------- | -------------
kGo|:heavy_check_mark:|:heavy_check_mark:|:heavy_check_mark:|:heavy_check_mark:
kSet|:heavy_check_mark:|:heavy_check_mark:|:heavy_check_mark:|:heavy_check_mark:
kPipe|:heavy_check_mark:|:heavy_check_mark:|:heavy_check_mark:|:heavy_check_mark:
kUse|:heavy_check_mark:|:heavy_check_mark:|:heavy_check_mark:|:heavy_check_mark:
kMatch|:heavy_check_mark:|:heavy_check_mark:|:heavy_check_mark:|:heavy_check_mark:
kAssignment|:heavy_check_mark:|:heavy_check_mark:|:heavy_check_mark:|:heavy_check_mark:
kCreateTag|:heavy_check_mark:|:heavy_check_mark:||
kAlterTag|:heavy_check_mark:|:heavy_check_mark:||
kCreateEdge|:heavy_check_mark:|:heavy_check_mark:||
kAlterEdge|:heavy_check_mark:|:heavy_check_mark:||
kDescribeTag|:heavy_check_mark:|:heavy_check_mark:|:heavy_check_mark:|:heavy_check_mark:
[...]

PR #400, opened 2019-05-17
User manager in meta

boshengchen:
Instead of PR #257 .
1. Addressed comments.
2. storage the data using dataman .

### DIFF
--- a/src/graph/PermissionManager.h
+++ b/src/graph/PermissionManager.h
@@ -1,0 +1,47 @@
+/* Copyright (c) 2019 vesoft inc. All rights reserved.
+ *
+ * This source code is licensed under Apache 2.0 License,
+ * attached with Common Clause Condition 1.0, found in the LICENSES directory.
+ */
+
+
+#ifndef GRAPH_PERMISSIONMANAGER_H
+#define GRAPH_PERMISSIONMANAGER_H
+
+// Operation and permission define:
+// Operation        | GOD           | ADMIN         | USER          | GUEST
+// ---------------- | ------------- | ------------- | ------------- | -------------
+// kGo              | Y             | Y             | Y             | Y
+// kSet             | Y             | Y             | Y             | Y
+// kPipe            | Y             | Y             | Y             | Y
+// kUse             | Y             | Y             | Y             | Y
+// kMatch           | Y             | Y             | Y             | Y
+// kAssignment      | Y             | Y             | Y             | Y
+// kCreateTag       | Y             | Y             |               |
+// kAlterTag        | Y             | Y             |               |
+// kCreateEdge      | Y             | Y             |               |
+// kAlterEdge       | Y             | Y             |               |
+// kDescribeTag     | Y             | Y             | Y             | Y
+// kDescribeEdge    | Y             | Y             | Y             | Y
+// kRemoveTag       | Y             | Y             |               |
+// kRemoveEdge      | Y             | Y             |               |
+// kInsertVertex    | Y             | Y             | Y             |
+// kInsertEdge      | Y             | Y             | Y             |
+// kShow            | Y             | Y             | Y             | Y
+// kDeleteVertex    | Y             | Y             | Y             |
+// kDeleteEdge      | Y             | Y             | Y             |
+// kFind            | Y             | Y             | Y             | Y
+// kAddHosts        | Y             |               |               |
+// kRemoveHosts     | Y             |               |               |
+// kCreateSpace     | Y             |               |               |
+// kDropSpace       | Y             | Y             |               |
+// kYield           | Y             | Y             | Y             | Y
+// kCreateUser      | Y             |               |               |
+// kDropUser        | Y             |               |               |
+// kAlterUser       | Y             | Y             | Y             | Y
+// kGrant           | Y             | Y             |               |
+// kRevoke          | Y             | Y             |               |
+// kChangePassword  | Y             | Y             | Y             | Y
+
+
+#endif  // GRAPH_PERMISSIONMANAGER_H
--- a/src/interface/meta.thrift
+++ b/src/interface/meta.thrift
@@ -31,6 +31,9 @@ enum ErrorCode {
     E_STORE_FAILURE          = -31,
     E_STORE_SEGMENT_ILLEGAL  = -32,
 
+    E_INVALID_PASSWORD       = -41,
+    E_INPROPER_ROLE          = -42,
+
     E_UNKNOWN        = -99,
 } (cpp.enum_strict)
 
@@ -43,10 +46,19 @@ enum AlterSchemaOp {
 } (cpp.enum_strict)
 
 
+enum RoleType {
+    GOD    = 0x01,
+    ADMIN  = 0x02,
+    USER   = 0x03,
+    GUEST  = 0x04,
+} (cpp.enum_strict)
+
+
 union ID {
     1: common.GraphSpaceID  space_id,
     2: common.TagID         tag_id,
     3: common.EdgeType      edge_type,
+    4: common.UserID        user_id,
 }
 
 struct IdName {
@@ -98,6 +110,19 @@ enum HostStatus {
 struct HostItem {
     1: common.HostAddr      hostAddr,
     2: HostStatus           status,
+
+struct UserItem {
+    1: string account,
+    2: string first_name,
+    3: string last_name,
+    4: string email,
+    5: string phone,
+}
+
+struct RoleItem {
+    1: string account,
+    2: string space,
+    3: RoleType RoleType,
 }
 
 struct ExecResp {
@@ -311,6 +336,63 @@ struct HBReq {
     1: common.HostAddr host,
 }
 
+struct CreateUserReq {
+    1: UserItem user,
+    2: string encoded_pwd,
+    3: bool missing_ok,
+}
+
+struct DropUserReq {
+    1: string account,
+    2: bool missing_ok,
+}
+
+struct AlterUserReq {
+    1: UserItem user_item,
+}
+
+struct GrantRoleReq {
+    1: RoleItem role_item,
+}
+
+struct RevokeRoleReq {
+    1: RoleItem role_item,
+}
+
+struct GetUserReq {
+    1: string account,
+}
+
+struct GetUserResp {
+    1: UserItem user_item,
+}
+
+struct ListUsersReq {
+}
+
+struct ListUsersResp {
+    1: list<UserItem> users,
+}
+
+struct ListRolesReq {
+    1: string space,
+}
+
+struct ListRolesResp {
+    1: list<RoleItem> roles,
+}
+
+struct ChangePasswordReq {
+    1: string account,
+    2: string new_encoded_pwd,
+    3: string old_encoded_pwd,
+}
+
+struct CheckPasswordReq {
+    1: string account,
+    2: string encoded_pwd,
+}
+
 service MetaService {
     ExecResp createSpace(1: CreateSpaceReq req);
     ExecResp dropSpace(1: DropSpaceReq req);
@@ -343,5 +425,17 @@ service MetaService {
     ScanResp scan(1: ScanReq req);
 
     HBResp           heartBeat(1: HBReq req);
+
+    ExecResp createUser(1: CreateUserReq req);
+    ExecResp dropUser(1: DropUserReq req);
+    ExecResp alterUser(1: AlterUserReq req);
+    ExecResp grantRole(1: GrantRoleReq req);
+    ExecResp revokeRole(1: RevokeRoleReq req);
+    GetUserResp getUser(1: GetUserReq req);
+    ListUsersResp listUsers(1: ListUsersReq req);
+    ListRolesResp listRoles(1: ListRolesReq req);
+    ExecResp changePassword(1: ChangePasswordReq req);
+    ExecResp checkPassword(1: CheckPasswordReq req);
+
 }
 
--- a/src/interface/meta.thrift
+++ b/src/interface/meta.thrift
@@ -45,6 +45,13 @@ enum AlterSchemaOp {
     UNKNOWN = 0x04,
 } (cpp.enum_strict)
 
+/**
+** GOD is A global senior administrator.like root of Linux systems.
+** ADMIN is an administrator for a given Graph Space.
+** USER is a normal user for a given Graph Space. A User can access (read and write) the data in the Graph Space.
+** GUEST is a read-only role for a given Graph Space. A Guest cannot modify the data in the Graph Space.
+** Refer to header file src/graph/PermissionManager.h for details.
+**/
 
 enum RoleType {
     GOD    = 0x01,
@@ -112,16 +119,23 @@ struct HostItem {
     2: HostStatus           status,
 
 struct UserItem {
-    1: string account,
-    2: string first_name,
-    3: string last_name,
-    4: bool   is_lock,
+    1: string account;
+    // Disable user if lock status is true.
+    2: bool   is_lock,
+    // The number of queries an account can issue per hour
+    3: i32    max_queries_per_hour,
+    // The number of updates an account can issue per hour
+    4: i32    max_updates_per_hour,
+    // The number of times an account can connect to the server per hour
+    5: i32    max_connections_per_hour,
+    // The number of simultaneous connections to the server by an account
+    6: i32    max_user_connections,
 }
 
 struct RoleItem {
-    1: string account,
-    2: string space,
-    3: RoleType role_type,
+    1: common.UserID        user_id,
+    2: common.GraphSpaceID  space_id,
+    3: RoleType             role_type,
 }
 
 struct ExecResp {
@@ -376,11 +390,11 @@ struct ListUsersResp {
     1: ErrorCode code,
     // Valid if ret equals E_LEADER_CHANGED.
     2: common.HostAddr  leader,    
-    3: list<UserItem> users,
+    3: map<common.UserID, UserItem>(cpp.template = "std::unordered_map") users,
 }
 
 struct ListRolesReq {
-    1: string space,
+    1: common.GraphSpaceID space_id,
 }
 
 struct ListRolesResp {
--- a/src/interface/meta.thrift
+++ b/src/interface/meta.thrift
@@ -117,6 +117,7 @@ enum HostStatus {
 struct HostItem {
     1: common.HostAddr      hostAddr,
     2: HostStatus           status,
+}
 
 struct UserItem {
     1: string account;
--- a/src/interface/meta.thrift
+++ b/src/interface/meta.thrift
@@ -115,14 +115,13 @@ struct UserItem {
     1: string account,
     2: string first_name,
     3: string last_name,
-    4: string email,
-    5: string phone,
+    4: bool   is_lock,
 }
 
 struct RoleItem {
     1: string account,
     2: string space,
-    3: RoleType RoleType,
+    3: RoleType role_type,
 }
 
 struct ExecResp {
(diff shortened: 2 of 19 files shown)


src/meta/processors/AuthenticationProcessor.cpp
```diff
@@ -0,0 +1,310 @@
+/* Copyright (c) 2018 vesoft inc. All rights reserved.
```
darionyaphet: please update to `2019`

boshengchen: Sorry, my bad, let me check all new files.

src/interface/meta.thrift
```diff
@@ -77,6 +89,22 @@ struct EdgeItem {
...
 }
 
+
+
+struct UserItem {
+    1: string account,
+    2: string first_name,
+    3: string last_name,
+    4: string email,
+    5: string phone,
```
darionyaphet: why supported `phone` ?

src/meta/processors/AuthenticationProcessor.cpp
boshengchen: I have checked the license head, and all the files are 2018 currently. Do we need to be changed them to 2019?

src/interface/meta.thrift
boshengchen: > why supported `phone` ?

It's just a user property. It is optional. More such options may be added in the future.

src/meta/processors/AuthenticationProcessor.cpp
boshengchen: Changed to 2019 with all new files as below :
src/meta/processors/AuthenticationProcessor.cpp
src/meta/processors/AuthenticationProcessor.h
src/meta/test/AuthProcessorTest.cpp

src/interface/meta.thrift
darionyaphet: marked as `optional` will be better ?

boshengchen: I think It is safe if set the optional to default (i.e. remove the optional keyword). What are the advantages of setting this parameter to `optional`?

darionyaphet: I don't think these attributes are necessary.

dangleptr: optional and default are not the same. 
default means the field must be written out, but we could read it or not as needs.  If the filed is not set,  we will read a default value for it. 
optional means if the field is not set,  we will read a nullptr.

In this case,  default and optional both are ok.  All of them are backward compatible. 
But for email and phone,  i think we have no needs for them.  But there are some other options we could support.   See https://dev.mysql.com/doc/refman/8.0/en/create-user.html

src/interface/meta.thrift
```diff
@@ -291,6 +319,63 @@ struct HBReq {
     1: common.HostAddr host,
 }
 
+struct CreateUserReq {
+    1: UserItem user,
+    2: string encoded_pwd,
+    3: bool missing_ok,
```
dangleptr: What's this?

src/meta/MetaServiceUtils.cpp
```diff
@@ -301,5 +306,129 @@ cpp2::ErrorCode MetaServiceUtils::alterColumnDefs(std::vector<nebula::cpp2::Colu
...
+    std::string key;
+    key.reserve(64);
+    key.append(kUsersTable.data(), kUsersTable.size());
+    key.append(reinterpret_cast<const char*>(&userId), sizeof(userId));
+    return key;
+}
+
+std::string MetaServiceUtils::userVal(const std::string& password,
+                                      const cpp2::UserItem& userItem,
+                                      std::shared_ptr<const SchemaProviderIf> schema) {
```
dangleptr: Don't use dataman in Meta module.  It is too heavy.

src/interface/meta.thrift
boshengchen: > What's this?

missing_ok？ If the syntax contains `IF NOT EXISTS` ,  missing_ok should be true, else should be false.

boshengchen:
Thanks for your response @darionyaphet @dangleptr , I have two questions to discuss as below :
1, About user options,  In fact, my original design and implementation was based on mysql ，There are options for default roles at that time. It was changed to what it is now after a couple of iterations, Thanks for your suggestions during the design phase. About option `email` and `phone`, The implementation of the syntax has been merged into the mainline，seems that we need to get a confirmed design again . and then re-implement these confirmed design. So,  Please feel free to let me know if have any good ideas . Let's using issue #190 for recording. 
2, About dataman , what are some good ways to avoid using dataman in module meta?  Please refer to #190  for requirements. Thanks.

boshengchen:
Address some comments as below:
1, Changed user data transport protocol from dataman to thrift. 
2, Removed user property email and phone. 
3, Added new user property lock status, this option is referenced from mysql.

src/interface/meta.thrift
```diff
@@ -89,6 +101,21 @@ struct EdgeItem {
     4: common.Schema        schema,
 }
 
+
+
+struct UserItem {
+    1: string account,
+    2: string first_name,
```
dangleptr: Do we need first_name, last_name?   
Why not put password into this structure?

src/interface/meta.thrift
```diff
@@ -89,6 +101,21 @@ struct EdgeItem {
...
+    3: string last_name,
+    4: bool   is_lock,
+}
+
+struct RoleItem {
+    1: string account,
```
dangleptr: Why not use userID?

src/interface/meta.thrift
```diff
@@ -300,6 +327,72 @@ struct HBReq {
...
 }
 
+struct CreateUserReq {
+    1: UserItem user,
+    2: string encoded_pwd,
+    3: bool missing_ok,
+}
+
+struct DropUserReq {
+    1: string account,
```
dangleptr: If you don't use "userId" in all requests, why we need it?

src/interface/meta.thrift
```diff
@@ -89,6 +101,21 @@ struct EdgeItem {
...
+struct UserItem {
+    1: string account,
+    2: string first_name,
+    3: string last_name,
+    4: bool   is_lock,
+}
+
+struct RoleItem {
+    1: string account,
+    2: string space,
```
dangleptr: why not spaceId?

src/interface/meta.thrift
```diff
@@ -43,10 +46,19 @@ enum AlterSchemaOp {
 } (cpp.enum_strict)
 
 
+enum RoleType {
+    GOD    = 0x01,
```
dangleptr: Add some comments, what's the meaning for different roles?

src/interface/meta.thrift
```diff
@@ -300,6 +327,72 @@ struct HBReq {
...
+
+struct ListUsersResp {
+    1: ErrorCode code,
+    // Valid if ret equals E_LEADER_CHANGED.
+    2: common.HostAddr  leader,    
+    3: list<UserItem> users,
+}
+
+struct ListRolesReq {
+    1: string space,
```
dangleptr: ditto

src/interface/meta.thrift
boshengchen: Thanks for your review @dangleptr , here are the answers to each question:
1: About first_name and last_name, Discussed with  @sherman-the-tank in issue #190 , Need to add more properties to a user, such as email address, first name, last name, etc.
2:About password is not contained in the structure UserItem，the reason is functional difference is between password and other user attributes. password is only used when connecting or changing password, so password is not normal user attributes. Also, fast read and write do not require serialization or deserialization of structures.

src/interface/meta.thrift
boshengchen: Good suggest!

boshengchen:
About SpaceID and UserID exists multiple places, I've actually thought about it，Because these two structures are used for user management often. So  can use these names directly . I think convert
 from name to id is superfluous in here. What do you think?
UserID and SpaceID are actually useful, they will be used for privilege management. I will use UserID and SpaceID to implement some caches in meta server or meta client.

src/interface/meta.thrift
dangleptr: > 1: About first_name and last_name, Discussed with @sherman-the-tank in issue #190 , Need to add more properties to a user, such as email address, first name, last name, etc.

We'd better just saved the information our system used.  Just supposed that, when you create a account,  the system ask you to fill your personal information,  is it reasonable?

> 2:About password is not contained in the structure UserItem，the reason is functional difference is between password and other user attributes. password is only used when connecting or changing password, so password is not normal user attributes. Also, fast read and write do not require serialization or deserialization of structures.

Make sense.

dangleptr:
> About SpaceID and UserID exists multiple places, I've actually thought about it，Because these two structures are used for user management often. So can use these names directly . I think convert
> from name to id is superfluous in here. What do you think?
> UserID and SpaceID are actually useful, they will be used for privilege management. I will use UserID and SpaceID to implement some caches in meta server or meta client.

It make sense for userId,  but not for spaceID.  We'd better use spaceId to join different entries instead of spaceName.

dangleptr:
By the way,  please refer #483 .  Put the user related processors under a separate dir.

boshengchen:
> It make sense for userId, but not for spaceID. We'd better use spaceId to join different entries instead of spaceName.
That makes sense, let me improve the space name and RoleItem related.

boshengchen:
1,Refactor user processors code structure. 
2,Remove user properties first name and last name. 
3, Add new user properties of resource limit. 
4, Changed related thrift structure from `'name'` to `'id'`, 
5,Add some comments for role.

boshengchen:
Ready to review. Please review again. Thanks.

src/graph/PermissionManager.h
```diff
@@ -0,0 +1,47 @@
...
+ *
+ * This source code is licensed under Apache 2.0 License,
+ * attached with Common Clause Condition 1.0, found in the LICENSES directory.
+ */
+
+
+#ifndef GRAPH_PERMISSIONMANAGER_H
+#define GRAPH_PERMISSIONMANAGER_H
+
+// Operation and permission define:
```
dangleptr: Well done

src/meta/processors/BaseProcessor.inl
```diff
@@ -162,6 +162,18 @@ Status BaseProcessor<RESP>::spaceExist(GraphSpaceID spaceId) {
     return Status::SpaceNotFound();
 }
 
+template<typename RESP>
+Status BaseProcessor<RESP>::userExist(UserID spaceId) {
```
dangleptr: Could we unify the userExist with getUserId ?

src/meta/processors/partsMan/DropSpaceProcessor.cpp
```diff
@@ -41,6 +41,16 @@ void DropSpaceProcessor::process(const cpp2::DropSpaceReq& req) {
     deleteKeys.emplace_back(MetaServiceUtils::indexSpaceKey(req.get_space_name()));
     deleteKeys.emplace_back(MetaServiceUtils::spaceKey(spaceId));
 
+    // delete related role data.
```
dangleptr: Now we just delete the parts information when deleting space.  We could delete the useless user and schema when compaction.

src/meta/processors/usersMan/AuthenticationProcessor.cpp
```diff
@@ -0,0 +1,271 @@
...
+    LOG(INFO) << "Create User " << user.get_account() << ", userId " << userId;
+    data.emplace_back(MetaServiceUtils::userKey(userId),
+                      MetaServiceUtils::userVal(req.get_encoded_pwd(), user));
+    resp_.set_code(cpp2::ErrorCode::SUCCEEDED);
+    resp_.set_id(to(userId, EntryType::USER));
+    doPut(std::move(data));
+}
+
+
+void AlterUserProcessor::process(const cpp2::AlterUserReq& req) {
```
dangleptr: It seems the alter is to change the whole value. 
I want to know what's the language for alter user?

src/meta/processors/usersMan/AuthenticationProcessor.cpp
```diff
@@ -0,0 +1,271 @@
...
+    folly::SharedMutex::WriteHolder wHolder(LockUtils::userLock());
+    auto roleKey = MetaServiceUtils::roleKey(roleItem.get_space_id(), roleItem.get_user_id());
+    resp_.set_id(to(roleItem.get_user_id(), EntryType::USER));
+    resp_.set_code(cpp2::ErrorCode::SUCCEEDED);
+    doRemove(std::move(roleKey));
+}
+
+
+void ChangePasswordProcessor::process(const cpp2::ChangePasswordReq& req) {
+    folly::SharedMutex::WriteHolder wHolder(LockUtils::userLock());
```
dangleptr: Change pwd is a special sentence?

src/meta/processors/usersMan/AuthenticationProcessor.cpp
```diff
@@ -0,0 +1,271 @@
...
+void CheckPasswordProcessor::process(const cpp2::CheckPasswordReq& req) {
+    folly::SharedMutex::ReadHolder rHolder(LockUtils::userLock());
+    auto userRet = getUserId(req.get_account());
+    if (!userRet.ok()) {
+        resp_.set_code(to(userRet.status()));
+        onFinished();
+        return;
+    }
+
+    if (!checkPassword(userRet.value(), req.get_encoded_pwd())) {
```
dangleptr: I want to know when to check the password?

src/meta/processors/BaseProcessor.inl
boshengchen: `userExist` is judged by UserID , But  `getUserId` is  judged by userName. And they query for different tables userIndex and userData. They are same with spaceExist and getSpaceId. I think it's best not to unify them. WDYT?

src/meta/processors/partsMan/DropSpaceProcessor.cpp
boshengchen: Yeah, that makes sense.  Let me modify it later.

src/meta/processors/usersMan/AuthenticationProcessor.cpp
boshengchen: No, Only change properties that require alter.

**Alter user sentences:**
```
ALTER USER account WITH with_user_opt_list

with_user_opt_item
    : KW_ACCOUNT KW_LOCK
    | KW_ACCOUNT KW_UNLOCK
    | KW_MAX_QUERIES_PER_HOUR INTEGER
    | KW_MAX_UPDATES_PER_HOUR INTEGER
    | KW_MAX_CONNECTIONS_PER_HOUR INTEGER
    | KW_MAX_USER_CONNECTIONS INTEGER
    ;

with_user_opt_list
    : with_user_opt_item
    | with_user_opt_list COMMA with_user_opt_item
    ;
```

src/meta/processors/usersMan/AuthenticationProcessor.cpp
boshengchen: Yes, Change password is a separate sentence. No need to verify the password if it is a `GOD` role. else the password must be validated. And `GOD` can change anyone's password. Other roles can only change their own password.

```
change_password_sentence
    : CHANGE PASSWORD account TO new_password
    | CHANGE PASSWORD account FROM old_password TO new_password
    ;
```

src/meta/processors/usersMan/AuthenticationProcessor.cpp
boshengchen: This is a reserved interface，Used for client or terminal connections.

boshengchen:
To improve the drop space , don't delete useless role data.

src/meta/processors/BaseProcessor.inl
dangleptr: Fair enough.

dangleptr:
Jenkins go

boshengchen:
1, Rebased 
2, Resolved compilation errors with depends on  `raftex_obj raftex_thrift_obj wal_obj time_obj`.
Please review again . Thanks.

boshengchen:
Rebased and resolved some conflicts.

src/meta/processors/usersMan/AuthenticationProcessor.cpp
```diff
@@ -0,0 +1,271 @@
...
+        onFinished();
+        return;
+    }
+    std::vector<std::string> keys;
+    keys.emplace_back(MetaServiceUtils::indexUserKey(req.get_account()));
+    keys.emplace_back(MetaServiceUtils::userKey(ret.value()));
+
+    // Collect related roles by user.
+    std::unique_ptr<kvstore::KVIterator> iter;
+    std::string prefix = "__roles__";
```
laura-ding: I think the prefix should be used kRolesTable although it define in cpp file

src/meta/test/CMakeLists.txt
```diff
@@ -171,3 +171,33 @@ nebula_link_libraries(
...
+        AuthProcessorTest.cpp
+        $<TARGET_OBJECTS:meta_service_handler>
+        $<TARGET_OBJECTS:kvstore_obj>
+        $<TARGET_OBJECTS:meta_client>
+        $<TARGET_OBJECTS:meta_thrift_obj>
+        $<TARGET_OBJECTS:common_thrift_obj>
+        $<TARGET_OBJECTS:raftex_obj>
+        $<TARGET_OBJECTS:raftex_thrift_obj>
+        $<TARGET_OBJECTS:wal_obj>
+	$<TARGET_OBJECTS:thrift_obj>
```
laura-ding: Alignment ：185L and 189L

src/meta/test/TestUtils.h
```diff
@@ -184,6 +185,37 @@ class TestUtils {
 
         return sc;
     }
+
+    static StatusOr<UserID> createUser(kvstore::KVStore* kv,
```
laura-ding: Strange alignment

src/meta/processors/usersMan/AuthenticationProcessor.cpp
boshengchen: Good ! We are think alike, I don't feel comfortable using it here either. Let me try.

src/meta/test/CMakeLists.txt
boshengchen: Sharp and bright eyes

boshengchen:
Address laura-ding's comment

src/meta/test/CMakeLists.txt
```diff
@@ -182,11 +182,11 @@ add_executable(
         $<TARGET_OBJECTS:raftex_obj>
         $<TARGET_OBJECTS:raftex_thrift_obj>
         $<TARGET_OBJECTS:wal_obj>
-	$<TARGET_OBJECTS:thrift_obj>
+	    $<TARGET_OBJECTS:thrift_obj>
```
laura-ding: I see it's still not aligned

boshengchen: My bad, resolved.  Thanks for your careful review @laura-ding

dangleptr:
Jenkins go

src/meta/test/CMakeLists.txt
```diff
@@ -171,3 +171,33 @@ nebula_link_libraries(
     gtest
 )
 nebula_add_test(meta_http_test)
+add_executable(
```
laura-ding: image
There are no aligned with 4 Spaces

boshengchen: A stupid mistake，corrected.

boshengchen:
Jenkins go

dangleptr:
Jenkins go

PR #402, opened 2019-05-20
move the KeyUtils from storage to common/base named NebulaKeyUtils

zhangguoqing:


dangleptr:
Jenkins go

dangleptr:
Jenkins go

src/common/base/test/NebulaKeyUtilsTest.cpp
```diff
@@ -0,0 +1,44 @@
+/* Copyright (c) 2018 vesoft inc. All rights reserved.
```
darionyaphet: 2019 ?

zhangguoqing: The 2018 is OK, because the file is copied from src/storage/test/KeyUtilsTest.cpp which was written in 2018 year by someone. This may be that git diff did not track this, and mistakenly thought it was the newly created file in this PR by me.  :-)

Issue #178, opened 2019-02-26
Convert SimplePart to Raft-based Partition


sherman-the-tank:
So that we could benefits from multiple copies of the data

PR #405, opened 2019-05-20
Integrate raft into storage

sherman-the-tank:
All UTs pass. At the moment, all parts still running with single copy. Will enable multiple copies in the future PR

Implemented #178


sherman-the-tank:
This PR replaced PR344

All comments in 344 have been addressed

sherman-the-tank:
jenkins go

src/kvstore/RocksEngine.cpp
```diff
@@ -19,15 +19,90 @@ using fs::FileType;
...
                          std::shared_ptr<rocksdb::CompactionFilterFactory> cfFactory)
         : KVEngine(spaceId)
-        , dataPath_(dataPath) {
-    LOG(INFO) << "open rocksdb on " << dataPath;
-    if (FileUtils::fileType(dataPath.c_str()) == FileType::NOTEXIST) {
-        FileUtils::makeDir(dataPath);
+        , dataPath_(folly::stringPrintf("%s/nebula/%d", dataPath.c_str(), spaceId)) {
+    auto path = folly::stringPrintf("%s/data", dataPath_.c_str());
+    LOG(INFO) << "open rocksdb on " << path;
+    if (FileUtils::fileType(path.c_str()) == FileType::NOTEXIST) {
```
dangleptr: We don't need it.  Rocksdb could create the dir if not exist.

src/kvstore/RocksEngine.cpp
```diff
@@ -182,6 +276,29 @@ ResultCode RocksEngine::removeRange(const std::string& start,
...
+ResultCode RocksEngine::removePrefix(const std::string& prefix) {
+    rocksdb::Slice pre(prefix.data(), prefix.size());
+    rocksdb::ReadOptions readOptions;
+    rocksdb::WriteOptions writeOptions;
+    writeOptions.disableWAL = FLAGS_rocksdb_disable_wal;
+    std::unique_ptr<rocksdb::Iterator> iter(db_->NewIterator(readOptions));
+    iter->Seek(pre);
+    while (iter->Valid()) {
+        if (iter->key().starts_with(pre)) {
+            auto status = db_->Delete(writeOptions, iter->key());
```
dangleptr: Use batch deletes here.

src/kvstore/Common.h
```diff
@@ -8,19 +8,21 @@
...
-    ERR_IO_ERROR         = -7,
+    SUCCEEDED               = 0,
+    ERR_SPACE_NOT_FOUND     = -1,
+    ERR_PART_NOT_FOUND      = -2,
+    ERR_KEY_NOT_FOUND       = -3,
+    ERR_CONSENSUS_ERROR     = -4,
+    ERR_LEADER_CHANGED      = -5,
+    ERR_INVALID_ARGUMENT    = -6,
+    ERR_IO_ERROR            = -7,
+    ERR_UNKNOWN             = -8,
```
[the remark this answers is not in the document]
sherman-the-tank: Usually unknown error goes the last. This ResultCode should only be used inside the storage, so if we need to add more codes, we can always push back the ```ERR_UNKNOWN```

src/kvstore/Common.h
```diff
@@ -8,19 +8,21 @@
...
 };
 
 #define KV_DATA_PATH_FORMAT(path, spaceId) \
```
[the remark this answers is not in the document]
sherman-the-tank: This is the old code. Let me check, if we do not use it anywhere, I can remove it

src/kvstore/Common.h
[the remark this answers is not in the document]
dangleptr: Then give a large number to unknown

src/kvstore/RocksEngine.cpp
sherman-the-tank: Good point 👍 

Will do

src/daemons/MetaDaemon.cpp
```diff
@@ -106,16 +110,25 @@ int main(int argc, char *argv[]) {
     // The meta server has only one space, one part.
     partMan->addPart(0, 0, std::move(peersRet.value()));
 
+    // Generic thread pool
+    auto workers = std::make_shared<nebula::thread::GenericThreadPool>();
```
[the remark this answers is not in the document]
sherman-the-tank: Good point 👍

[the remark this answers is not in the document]
sherman-the-tank: After poking around, unfortunately, there seems no way to access the thrift worker thread. In addition, the thrift::ThreadManager does not support timer

src/daemons/StorageDaemon.cpp
```diff
@@ -108,19 +140,25 @@ int main(int argc, char *argv[]) {
          return EXIT_FAILURE;
     }
 
-    auto ioThreadPool = std::make_shared<folly::IOThreadPoolExecutor>(FLAGS_io_handlers);
+    // Generic thread pool
+    auto workers = std::make_shared<nebula::thread::GenericThreadPool>();
```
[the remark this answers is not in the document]
sherman-the-tank: See comments above

src/kvstore/RocksEngine.cpp
sherman-the-tank: Rocksdb will not create the path. It only creates the necessary files

sherman-the-tank:
jenkins go

src/kvstore/RocksEngine.cpp
dangleptr: Not exactly!   Rocksdb will create the path if not exist,   you could test it.  
But if the parent path not exist,  rocksdb will create path failed.

src/daemons/MetaDaemon.cpp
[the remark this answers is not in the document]
dangleptr: Fine.  Reuse IOThreadPool is enough.

src/kvstore/Common.h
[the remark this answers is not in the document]
dangleptr: You should fix it.  We don't want the error code always different.

src/meta/test/ProcessorTest.cpp
```diff
@@ -41,8 +41,12 @@ using nebula::cpp2::SupportedType;
 using apache::thrift::FragileConstructor::FRAGILE;
 
 TEST(ProcessorTest, AddHostsTest) {
+    auto workers = std::make_shared<thread::GenericThreadPool>();
```
dangleptr: You could move the workers and ioPool related code into TestUtils::initKV

src/kvstore/test/LogEncoderTest.cpp
```diff
@@ -0,0 +1,129 @@
+/* Copyright (c) 2018 vesoft inc. All rights reserved.
```
dangleptr: Copyright 2019

src/kvstore/Common.h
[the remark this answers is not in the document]
sherman-the-tank: Good point 👍  Will do

src/kvstore/RocksEngine.cpp
sherman-the-tank: That's exactly the reason here. We need to create the parent path

src/meta/test/ProcessorTest.cpp
sherman-the-tank: Aha, good idea 👍

src/kvstore/test/LogEncoderTest.cpp
sherman-the-tank: Nice catch 👍

src/daemons/StorageDaemon.cpp
```diff
@@ -44,6 +48,33 @@ static void signalHandler(int sig);
 static Status setupSignalHandler();
 
 
+std::unique_ptr<nebula::kvstore::KVStore> getStoreInstance(
```
laura-ding: Uniform function args alignment：
void funtion(type arg1，
&ensp;&ensp;&ensp;&ensp;&ensp;&ensp;&ensp;&ensp;&ensp;&ensp;type arg2，
&ensp;&ensp;&ensp;&ensp;&ensp;&ensp;&ensp;&ensp;&ensp;&ensp;type arg3)

src/kvstore/LogEncoder.cpp
```diff
@@ -0,0 +1,146 @@
+/* Copyright (c) 2018 vesoft inc. All rights reserved.
```
laura-ding: The new file should use 2019

src/kvstore/LogEncoder.cpp
```diff
@@ -0,0 +1,146 @@
...
+#include "kvstore/LogEncoder.h"
+
+namespace nebula {
+namespace kvstore {
+
+std::string encodeSingleValue(LogType type, folly::StringPiece val) {
+    std::string encoded;
+    encoded.reserve(val.size() + sizeof(int64_t) + 1 + sizeof(uint32_t));
+    // Timstamp (8 bytes)
+    int64_t ts = time::TimeUtils::nowInMSeconds();
```
laura-ding: Define a macro or constant to represent the size of these data type？

src/kvstore/LogEncoder.h
```diff
@@ -0,0 +1,37 @@
+/* Copyright (c) 2018 vesoft inc. All rights reserved.
```
laura-ding: ditto

src/kvstore/LogEncoder.cpp
sherman-the-tank: Good catch! 👍

src/kvstore/Part.cpp
```diff
@@ -5,34 +5,210 @@
...
+
+
+std::string Part::compareAndSet(const std::string& log) {
+    UNUSED(log);
+    LOG(FATAL) << "To be implemented";
+}
+
+
+bool Part::commitLogs(std::unique_ptr<LogIterator> iter) {
+    auto batch = engine_->startBatchWrite();
```
dangleptr: Record the latest committedLogId

src/kvstore/RocksEngine.cpp
```diff
@@ -182,6 +276,35 @@ ResultCode RocksEngine::removeRange(const std::string& start,
 }
 
 
+ResultCode RocksEngine::removePrefix(const std::string& prefix) {
+    rocksdb::Slice pre(prefix.data(), prefix.size());
```
dangleptr: We should use deleteRange.

src/common/network/NetworkUtils.cpp
```diff
@@ -83,19 +83,29 @@ StatusOr<std::unordered_map<std::string, std::string>> NetworkUtils::listDeviceA
     return dev2ipv4s;
 }
 
+
 bool NetworkUtils::getDynamicPortRange(uint16_t& low, uint16_t& high) {
```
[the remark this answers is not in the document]
sherman-the-tank: Not quite understand why we need StatusOr here

src/daemons/StorageDaemon.cpp
sherman-the-tank: We allow both format
```
void func(Type1 arg1,
          Type2 arg2,
          Type3 arg3);
```
or
```
void func(
    Type1 arg1,
    Type2 arg2,
    Type3 arg3);
````
It depends on the length of the argument specification

Comment (sherman-the-tank) [reply src/daemons/StorageDaemon.cpp]: In this case, the length of the argument is pretty long, so we have to employ the second format

File: src/kvstore/RocksEngine.cpp (hunk 14 above)
Comment (sherman-the-tank) [reply src/kvstore/RocksEngine.cpp]: I'm not convinced yet. Especially in the rocksdb version we are using, DeleteRange is an experimental feature, which we should not use

After we upgrade the rocksdb and benchmark to prove DeleteRange is faster, we can certainly switch to it

File: src/kvstore/Part.cpp (hunk 13 above)
Comment (sherman-the-tank) [reply src/kvstore/Part.cpp]: Good catch 👍  This has not been implemented. I'll add it to the ```commit()```

File: src/kvstore/Part.cpp [hunk 16]
```diff
@@ -5,34 +5,235 @@
...
+    }
 }
 
-void SimplePart::asyncRemove(const std::string& key, KVCallback cb) {
-    CHECK_NOTNULL(engine_);
-    auto ret = engine_->remove(key);
-    cb(ret, HostAddr(0, 0));
+
+wal::BufferFlusher* getBufferFlusher() {
+    static wal::BufferFlusher flusher;
```
dangleptr: Only one flusher maybe a bottleneck.  Let's open another issue to make the number configurable.

src/kvstore/RocksEngine.cpp
dangleptr: As your code,  dataPath_ is the parent dir.   Fine,  open another issue to track it if you don't want to change in this pr.

src/kvstore/RocksEngine.cpp
dangleptr: Fine, open another issue to track it.

src/kvstore/Part.cpp
sherman-the-tank: We could make the number of flushing threads in the BufferFlusher configurable

Sure. let's do it in a separate PR

PR #388, opened 2019-05-16
[Query] Support Description about Graph Space

darionyaphet:
Reopen  [Support Description about Graph Space](https://github.com/vesoft-inc/nebula/pull/365)

```
nebula> describe space nba
=================================================
| ID | Name | Partition number | Replica Factor |
=================================================
|  1 |  nba |                1 |              1 |
-------------------------------------------------
```

PR #408, opened 2019-05-21
Allow to create graph space without options

### LINKED ISSUE #350: CREATE SPACE should have default options
e.g. `CREATE SPACE myspace` shall succeed.

dutor:
As title.

This PR will close #350


src/parser/AdminSentences.h
```diff
@@ -223,6 +223,9 @@ class CreateSpaceSentence final : public Sentence {
     }
 
     std::vector<SpaceOptItem*> getOpts() {
+        if (spaceOpts_ == nullptr) {
+            return {};
+        }
```
steppenwolfyuetong: when create graph space without options, CreateSpaceSentence::toString function need to  deal with

src/graph/test/SchemaTest.cpp
```diff
@@ -319,6 +319,20 @@ TEST_F(SchemaTest, metaCommunication) {
...
+        ASSERT_EQ(cpp2::ErrorCode::SUCCEEDED, code);
+    }
+    sleep(FLAGS_load_data_interval_secs + 1);
+    {
+        // TODO To verify the space options are applied correctly.
+        cpp2::ExecutionResponse resp;
+        std::string query = "USE space_with_default_options";
+        auto code = client->execute(query, resp);
+        ASSERT_EQ(cpp2::ErrorCode::SUCCEEDED, code);
+    }
```
steppenwolfyuetong: Here use `describe space_with_default_options` to handle more obvious

src/parser/AdminSentences.h
dutor: roger that!

src/graph/test/SchemaTest.cpp
dutor: Yes, we will do, once #388 is merged.

src/interface/meta.thrift
```diff
@@ -88,8 +88,8 @@ struct ExecResp {
 // Graph space related operations.
 struct CreateSpaceReq {
     1: string space_name,
-    2: i32 parts_num,
-    3: i32 replica_factor,
+    2: optional i32 parts_num,
```
dangleptr: Default field is enough in this case.

dutor: I'm not sure whether it's OK if a non-optional field is not set, is that allowed?

Anyway, the present implementation of meta client set this field blindly. There is no such thing _default_ or _optional_.

Maybe this is a special case where there are values could be used as a default to represent NULL or not set. But we must be careful about cases there aren't such values.

The thrift proto on this struct in upstream has been modified, so this change should be discarded.

dutor:
Jenkins, go!

dutor:
Jenkins go

dutor:
Jenkins go

dutor:
Jenkins go

Issue #67, opened 2019-01-09
[storage] Support filter for query request


dangleptr:
In some scenery, Query engine could push filtering logic to the specific partition in the storage service to improve the performance

PR #409, opened 2019-05-21
Implement filter in storage

dangleptr:
This is the basic version for filter in storage layer.

With this pr,  the filter could work at least. 

Subtask of #67

### DIFF
--- a/src/common/base/Base.h
+++ b/src/common/base/Base.h
@@ -137,6 +137,7 @@ struct PartMeta {
 
 using PartsMap  = std::unordered_map<GraphSpaceID, std::unordered_map<PartitionID, PartMeta>>;
 
+using VariantType = boost::variant<int64_t, double, bool, std::string>;
 // Useful type traits
 
 // Tell if `T' is copy-constructible
--- a/src/common/base/Base.h
+++ b/src/common/base/Base.h
@@ -138,6 +138,23 @@ struct PartMeta {
 using PartsMap  = std::unordered_map<GraphSpaceID, std::unordered_map<PartitionID, PartMeta>>;
 
 using VariantType = boost::variant<int64_t, double, bool, std::string>;
+
+#ifndef VAR_INT64
+#define VAR_INT64 0
+#endif
+
+#ifndef VAR_DOUBLE
+#define VAR_DOUBLE 1
+#endif
+
+#ifndef VAR_BOOL
+#define VAR_BOOL 2
+#endif
+
+#ifndef VAR_STR
+#define VAR_STR 3
+#endif
+
 // Useful type traits
 
 // Tell if `T' is copy-constructible
--- a/src/dataman/RowReader.h
+++ b/src/dataman/RowReader.h
@@ -10,9 +10,11 @@
 #include "base/Base.h"
 #include <gtest/gtest_prod.h>
 #include "gen-cpp2/graph_types.h"
+#include "interface/gen-cpp2/common_types.h"
 #include "dataman/DataCommon.h"
 #include "meta/SchemaProviderIf.h"
 #include "meta/SchemaManager.h"
+#include "base/ErrorOr.h"
 
 namespace nebula {
 
@@ -87,6 +89,64 @@ public:
         folly::StringPiece row,
         std::shared_ptr<const meta::SchemaProviderIf> schema);
 
+    static ErrorOr<ResultType, VariantType> getProp(const RowReader* reader,
+                                                    const std::string& prop) {
+        auto& vType = reader->getSchema()->getFieldType(prop);
+        switch (vType.type) {
+            case nebula::cpp2::SupportedType::BOOL: {
+                bool v;
+                auto ret = reader->getBool(prop, v);
+                if (ret != ResultType::SUCCEEDED) {
+                    return ret;
+                }
+                return v;
+            }
+            case nebula::cpp2::SupportedType::INT: {
+                int64_t v;
+                auto ret = reader->getInt(prop, v);
+                if (ret != ResultType::SUCCEEDED) {
+                    return ret;
+                }
+                return v;
+            }
+            case nebula::cpp2::SupportedType::VID: {
+                VertexID v;
+                auto ret = reader->getVid(prop, v);
+                if (ret != ResultType::SUCCEEDED) {
+                    return ret;
+                }
+                return v;
+            }
+            case nebula::cpp2::SupportedType::FLOAT: {
+                float v;
+                auto ret = reader->getFloat(prop, v);
+                if (ret != ResultType::SUCCEEDED) {
+                    return ret;
+                }
+                return static_cast<double>(v);
+            }
+            case nebula::cpp2::SupportedType::DOUBLE: {
+                double v;
+                auto ret = reader->getDouble(prop, v);
+                if (ret != ResultType::SUCCEEDED) {
+                    return ret;
+                }
+                return v;
+            }
+            case nebula::cpp2::SupportedType::STRING: {
+                folly::StringPiece v;
+                auto ret = reader->getString(prop, v);
+                if (ret != ResultType::SUCCEEDED) {
+                    return ret;
+                }
+                return v.toString();
+            }
+            default:
+                LOG(FATAL) << "Unknown type: " << static_cast<int32_t>(vType.type);
+                return "";
+        }
+    }
+
     virtual ~RowReader() = default;
 
     SchemaVer schemaVer() const noexcept;
@@ -122,6 +182,10 @@ public:
 
     ResultType getTimestamp(const folly::StringPiece name, int64_t& v) const noexcept;
     ResultType getTimestamp(int64_t index, int64_t& v) const noexcept;
+
+    const meta::SchemaProviderIf* getSchema() const {
+        return schema_.get();
+    }
 
     // TODO getPath(const std::string& name) const noexcept;
     // TODO getPath(int64_t index) const noexcept;
--- a/src/storage/CommonUtils.h
+++ b/src/storage/CommonUtils.h
@@ -13,17 +13,11 @@
 namespace nebula {
 namespace storage {
 
-class CommonUtils {
-public:
-    static std::string buildTagKeyInFilter(const std::string& tag, const std::string& prop) {
-        return folly::stringPrintf("%s_%s", tag.c_str(), prop.c_str());
-    }
-};
+using TagProp = std::pair<std::string, std::string>;
 
 struct FilterContext {
-    // key: tagName + propName -> propValue
-    // TODO(heng): figure out a better way.
-    std::unordered_map<std::string, VariantType> tagFilters_;
+    // key: <tagName, propName> -> propValue
+    std::unordered_map<TagProp, VariantType> tagFilters_;
 };
 
 class PropContext {
@@ -53,14 +47,14 @@ public:
                 && filtered_;
     }
 
-    const std::string& tagFilterKey() const {
+    const std::string& tagOrEdgeName() const {
         CHECK(filtered_);
-        return tagFilterKey_;
+        return tagOrEdgeName_;
     }
 
-    void setTagFilterKey(const std::string& tagName, const std::string& propName) {
+    void setTagOrEdgeName(const std::string& tagName) {
         filtered_ = true;
-        tagFilterKey_ = CommonUtils::buildTagKeyInFilter(tagName, propName);
+        tagOrEdgeName_ = tagName;
     }
 
     bool filtered() const {
@@ -78,9 +72,9 @@ public:
     bool    returned_ = false;
 
 private:
-    // filter Key inside tagFilters of FilterContext.  Only valid when filtered_ is true.
-    // Make it private, because it should be accessed through tagFilterKey() and setTagFilterKey().
-    std::string tagFilterKey_;
+    // If the prop is from tag, it is tagName otherwise it is edge name.
+    // It is only valid when filtered_ is true.
+    std::string tagOrEdgeName_;
     // The prop comes from filter.
     bool    filtered_ = false;
 };
@@ -92,12 +86,13 @@ struct TagContext {
 
     TagID tagId_ = 0;
     std::vector<PropContext> props_;
+    std::unordered_map<std::string, int32_t> propNameIndex_;
 
     PropContext* findProp(const std::string& propName) {
-        for (auto& prop : props_) {
-            if (prop.prop_.name == propName) {
-                return &prop;
-            }
+        auto it = propNameIndex_.find(propName);
+        if (it != propNameIndex_.end()) {
+            CHECK(it->second >= 0 && (size_t)it->second < props_.size());
+            return &props_[it->second];
         }
         return nullptr;
     }
@@ -110,8 +105,9 @@ struct TagContext {
         pc.type_ = type;
         pc.prop_.owner = cpp2::PropOwner::SOURCE;
         pc.retIndex_ = props_.size();
-        pc.setTagFilterKey(tagName, propName);
+        pc.setTagOrEdgeName(tagName);
         props_.emplace_back(std::move(pc));
+        propNameIndex_.emplace(propName, props_.size() - 1);
     }
 };
 
--- a/src/storage/CommonUtils.h
+++ b/src/storage/CommonUtils.h
@@ -1,0 +1,130 @@
+/* Copyright (c) 2019 vesoft inc. All rights reserved.
+ *
+ * This source code is licensed under Apache 2.0 License,
+ * attached with Common Clause Condition 1.0, found in the LICENSES directory.
+ */
+
+#ifndef STORAGE_COMMON_H_
+#define STORAGE_COMMON_H_
+
+#include "base/Base.h"
+#include "filter/Expressions.h"
+
+namespace nebula {
+namespace storage {
+
+class CommonUtils {
+public:
+    static std::string buildTagKeyInFilter(const std::string& tag, const std::string& prop) {
+        return folly::stringPrintf("%s_%s", tag.c_str(), prop.c_str());
+    }
+};
+
+struct FilterContext {
+    // key: tagName + propName -> propValue
+    // TODO(heng): figure out a better way.
+    std::unordered_map<std::string, VariantType> tagFilters_;
+};
+
+class PropContext {
+public:
+    enum PropInKeyType {
+        NONE = 0x00,
+        SRC  = 0x01,
+        DST  = 0x02,
+        TYPE = 0x03,
+        RANK = 0x04,
+    };
+
+    PropContext() = default;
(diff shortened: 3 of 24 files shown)


dangleptr:
Jenkins go

src/storage/CommonUtils.h
```diff
@@ -0,0 +1,133 @@
...
+        props_.reserve(8);
+    }
+
+    EdgeType edgeType_ = 0;
+    std::vector<PropContext> props_;
+};
+
+
+
+
```
zlcook: extra lines

src/storage/CommonUtils.h
```diff
@@ -0,0 +1,133 @@
...
+    bool    filtered_ = false;
+};
+
+struct TagContext {
+    TagContext() {
+        props_.reserve(8);
+    }
+
+    TagID tagId_ = 0;
+    std::vector<PropContext> props_;
```
zhangguoqing: How about ```std::unordered_map<PropContext> props_;``` to do findProp?

src/storage/CommonUtils.h
```diff
@@ -0,0 +1,133 @@
...
+
+    bool filtered() const {
+        return filtered_;
+    }
+
+    cpp2::PropDef prop_;
+    nebula::cpp2::ValueType type_;
+    PropInKeyType pikType_ = PropInKeyType::NONE;
+    boost::variant<int64_t, double> sum_ = 0L;
+    int32_t count_    = 0;
```
zhangguoqing: There may be extra spaces.

src/storage/test/QueryBoundTest.cpp
```diff
@@ -26,7 +26,7 @@ void mockData(kvstore::KVStore* kv) {
                 auto key = KeyUtils::vertexKey(partId, vertexId, tagId, 0);
```
zhangguoqing: KeyUtils -> NebulaKeyUtils, you should probably do rebase.

src/storage/test/QueryBoundTest.cpp
```diff
@@ -208,9 +210,131 @@ TEST(QueryBoundTest, inBoundSimpleTest) {
...
+    std::unique_ptr<kvstore::KVStore> kv(TestUtils::initKV(rootPath.path()));
+    auto schemaMan = TestUtils::mockSchemaMan();;
+    mockData(kv.get());
+
+    LOG(INFO) << "Build filter...";
+    auto* tag = new std::string("3001");
+    auto* prop = new std::string("tag_3001_col_0");
+    auto* srcExp = new SourcePropertyExpression(tag, prop);
+    auto* priExp = new PrimaryExpression(20 + 3001L);
+    auto relExp = std::make_unique<RelationalExpression>(srcExp,
```
zhangguoqing: Could you add an example of a compound filter (an expression that compares multiple properties)?

src/storage/CommonUtils.h
```diff
@@ -0,0 +1,133 @@
...
+
+class CommonUtils {
+public:
+    static std::string buildTagKeyInFilter(const std::string& tag, const std::string& prop) {
+        return folly::stringPrintf("%s_%s", tag.c_str(), prop.c_str());
+    }
+};
+
+struct FilterContext {
+    // key: tagName + propName -> propValue
```
whitewum: give a name to tagName+propName ?

src/storage/CommonUtils.h
whitewum: I think the order matters.

whitewum:
@dangleptr is this pr still on working ? or ready ?

dangleptr:
> @dangleptr is this pr still on working ? or ready ?

It is ready.  

I will rebase it and address the comments inside later.

src/common/base/Base.h
```diff
@@ -136,6 +136,7 @@ struct PartMeta {
 
 using PartsMap  = std::unordered_map<GraphSpaceID, std::unordered_map<PartitionID, PartMeta>>;
 
+using VariantType = boost::variant<int64_t, double, bool, std::string>;
```
darionyaphet: `VariantType` was defined duplicate in `src/common/filter/Expressions.h`

src/dataman/RowReader.h
```diff
@@ -87,6 +89,64 @@ class RowReader {
...
+                }
+                return v;
+            }
+            case nebula::cpp2::SupportedType::FLOAT: {
+                float v;
+                auto ret = reader->getFloat(prop, v);
+                if (ret != ResultType::SUCCEEDED) {
+                    return ret;
+                }
+                return static_cast<double>(v);
```
darionyaphet: why don't convert to `float`

src/storage/CommonUtils.h
dangleptr: Good catch

src/dataman/RowReader.h
dangleptr: Because we could use double to represent float.   And we only support 4 types in VariantType

src/storage/CommonUtils.h
dangleptr: The name for what?

src/storage/test/QueryBoundTest.cpp
dangleptr: In TagAndEdgeFilterTest,   it is a compound filter

src/storage/CommonUtils.h
dangleptr: Currently, findProp is used only in checkExp which is invoked once in one request.

src/dataman/RowReader.h
```diff
@@ -87,6 +89,64 @@ class RowReader {
...
+                folly::StringPiece v;
+                auto ret = reader->getString(prop, v);
+                if (ret != ResultType::SUCCEEDED) {
+                    return ret;
+                }
+                return v.toString();
+            }
+            default:
+                LOG(FATAL) << "Unknown type: " << static_cast<int32_t>(vType.type);
+                return "";
```
darionyaphet: It seems could return `ResultType` ?

src/storage/test/QueryBoundTest.cpp
```diff
@@ -217,7 +218,73 @@ TEST(QueryBoundTest, inBoundSimpleTest) {
...
     LOG(INFO) << "Check the results...";
-    checkResponse(resp, false);
+    checkResponse(resp, 30, 2, 20001, 5, false);
+}
+
+TEST(QueryBoundTest, FilterTest_OnlyEdgeFilter) {
+    fs::TempDir rootPath("/tmp/QueryBoundTest.XXXXXX");
+    LOG(INFO) << "Prepare meta...";
+    std::unique_ptr<kvstore::KVStore> kv(TestUtils::initKV(rootPath.path()));
+    auto schemaMan = TestUtils::mockSchemaMan();;
```
whitewum: double ;;

src/storage/test/AdHocSchemaManager.cpp
```diff
@@ -166,8 +166,11 @@ StatusOr<GraphSpaceID> AdHocSchemaManager::toGraphSpaceID(folly::StringPiece spa
 // This interface is disabled
 StatusOr<TagID> AdHocSchemaManager::toTagID(GraphSpaceID space, folly::StringPiece tagName) {
     UNUSED(space);
-    UNUSED(tagName);
-    LOG(FATAL) << "Unimplement";
+    try {
```
whitewum: please check the comment

src/storage/test/QueryBoundTest.cpp
```diff
@@ -272,6 +339,78 @@ TEST(QueryBoundTest,  GenBucketsTest) {
...
+                                                    BoundType::OUT_BOUND);
+    auto f = processor->getFuture();
+    processor->process(req);
+    auto resp = std::move(f).get();
+
+    LOG(INFO) << "Check the results...";
+    EXPECT_EQ(3, resp.result.failed_codes.size());
+    EXPECT_TRUE(nebula::storage::cpp2::ErrorCode::E_INVALID_FILTER
+                    == resp.result.failed_codes[0].code);
+}
```
whitewum: I think only (some) relational & logical expressions are test, what about others?
Besides, the data writen in and filter out are hard to check. I hope you check the result carefully.

src/storage/test/QueryBoundTest.cpp
```diff
@@ -272,6 +339,78 @@ TEST(QueryBoundTest,  GenBucketsTest) {
...
+                                          RelationalExpression::Operator::GE,
+                                          priExp);
+    auto* edgeProp = new std::string("col_0");
+    auto* alias = new std::string("e101");
+    auto* edgeExp = new EdgePropertyExpression(alias, edgeProp);
+    auto* priExp2 = new PrimaryExpression(10007L);
+    auto* right = new RelationalExpression(edgeExp,
+                                           RelationalExpression::Operator::GE,
+                                           priExp2);
+    auto logExp = std::make_unique<LogicalExpression>(left, LogicalExpression::AND, right);
```
whitewum: e.g.,  (a < 0) && (b < 0 || c != e)

whitewum:
Does it support any short-circuit?

src/storage/test/QueryBoundTest.cpp
```diff
@@ -272,6 +339,78 @@ TEST(QueryBoundTest,  GenBucketsTest) {
...
+    auto* priExp = new PrimaryExpression(20 + 3001L);
+    auto* left = new RelationalExpression(srcExp,
+                                          RelationalExpression::Operator::GE,
+                                          priExp);
+    auto* edgeProp = new std::string("col_0");
+    auto* alias = new std::string("e101");
+    auto* edgeExp = new EdgePropertyExpression(alias, edgeProp);
+    auto* priExp2 = new PrimaryExpression(10007L);
+    auto* right = new RelationalExpression(edgeExp,
+                                           RelationalExpression::Operator::GE,
```
whitewum: use LE

src/storage/CommonUtils.h
```diff
@@ -0,0 +1,130 @@
...
```
wadeliuyi: is there any splitter between tagName and propName? else “a” + “bcd” is same with “ab”+”cd”

src/storage/test/QueryBoundTest.cpp
dangleptr: Good catch

src/storage/test/AdHocSchemaManager.cpp
dangleptr: Good catch.  The comment should be fixed.  
AdHocSchemaManager is only used in UTs,  so LOG(FATAL) is ok here.

src/storage/CommonUtils.h
dangleptr: Good catch.

src/storage/test/QueryBoundTest.cpp
dangleptr: The UTs here is to test edgePropFilter and tagPropFilter.  Only the two basic filters is handled in storage layer.  All other filters have been tested in the UTs of Expression module.

src/storage/test/QueryBoundTest.cpp
dangleptr: Your example is to test expressions logic. 
On storage layer,  the only filters being handled are SrcPropFilter and EdgePropFilter.  Of course, maybe we could add UT for EdgeRankFilter

src/storage/test/QueryBoundTest.cpp
dangleptr: If LE,  all edges in UT will pass the filter.

src/storage/QueryBaseProcessor.inl
```diff
@@ -182,20 +97,215 @@ cpp2::ErrorCode QueryBaseProcessor<REQ, RESP>::checkAndBuildContexts(const REQ&
...
+        if (!checkExp(exp_.get())) {
+            return cpp2::ErrorCode::E_INVALID_FILTER;
+        }
+        expCtx_ = std::make_unique<ExpressionContext>();
+        exp_->setContext(expCtx_.get());
+        auto& getters = expCtx_->getters();
+        getters.getDstTagProp = [] (const std::string& alias,
+                                    const std::string& prop) -> VariantType {
+            LOG(FATAL) << "Unsupport get dst tag " << alias << " prop " << prop;
+            return false;
```
wadeliuyi: can return bool for VariantType?

src/storage/CommonUtils.h
```diff
@@ -0,0 +1,130 @@
...
+#include "base/Base.h"
+#include "filter/Expressions.h"
+
+namespace nebula {
+namespace storage {
+
+class CommonUtils {
+public:
+    static std::string buildTagKeyInFilter(const std::string& tag, const std::string& prop) {
+        return folly::stringPrintf("%s_%s", tag.c_str(), prop.c_str());
```
wadeliuyi: I test that we can use “_” in create tag, so “a”+”b_c” is same with “a_b”+”c”

src/storage/CommonUtils.h
```diff
@@ -0,0 +1,130 @@
...
+    void pushFilterProp(const std::string& tagName,
+                        const std::string& propName,
+                        const nebula::cpp2::ValueType& type) {
+        PropContext pc;
+        pc.prop_.name = propName;
+        pc.type_ = type;
+        pc.prop_.owner = cpp2::PropOwner::SOURCE;
+        pc.retIndex_ = props_.size();
+        pc.setTagFilterKey(tagName, propName);
+        props_.emplace_back(std::move(pc));
```
wadeliuyi: why not emplace_back(name, type,...)? then no need construct pc object in stack memory?

src/storage/QueryBaseProcessor.inl
```diff
@@ -182,20 +97,215 @@ cpp2::ErrorCode QueryBaseProcessor<REQ, RESP>::checkAndBuildContexts(const REQ&
...
+                return false;
+            }
+            // TODO(heng): Now we have to scan the whole array to find related tagId,
+            // maybe we could find a better way to solve it.
+            for (auto& tc : tagContexts_) {
+                if (tc.tagId_ == tagId) {
+                    auto* prop = tc.findProp(propName);
+                    if (prop == nullptr) {
+                        tc.pushFilterProp(tagName, propName, field->getType());
+                    } else if (!prop->filtered()) {
```
wadeliuyi: the function pushFilterProp will call setTagFilterKey and the filtered is true, so maybe here no need check?

src/storage/QueryBaseProcessor.inl
```diff
@@ -245,6 +356,34 @@ kvstore::ResultCode QueryBaseProcessor<REQ, RESP>::collectEdgeProps(
...
+                    return value(std::move(res));
+                };
+                getters.getEdgeRank = [&] () -> VariantType {
+                    return rank;
+                };
+                getters.getSrcTagProp = [&, this] (const std::string& tag,
+                                                   const std::string& prop) -> VariantType {
+                    auto tagKey = CommonUtils::buildTagKeyInFilter(tag, prop);
+                    auto it = fcontext->tagFilters_.find(tagKey);
+                    CHECK(it != fcontext->tagFilters_.end());
```
wadeliuyi: why we can sure that iter must not end?

src/storage/QueryBaseProcessor.inl
```diff
@@ -182,20 +97,215 @@ cpp2::ErrorCode QueryBaseProcessor<REQ, RESP>::checkAndBuildContexts(const REQ&
...
+                VLOG(1) << "Skip the bad value for prop " << name;
+                continue;
+            }
+            auto&& v = value(std::move(res));
+            if (prop.fromTagFilter()) {
+                fcontext->tagFilters_.emplace(prop.tagFilterKey(), v);
+            }
+            if (prop.returned_) {
+                switch (v.which()) {
+                    case 0:
```
wadeliuyi: maybe use type micro to replace 0,1,2,3 is more readable.

src/storage/QueryBaseProcessor.inl
```diff
@@ -263,7 +402,10 @@ QueryBaseProcessor<REQ, RESP>::asyncProcessBucket(Bucket bucket) {
         std::vector<OneVertexResp> codes;
         codes.reserve(b.vertices_.size());
         for (auto& pv : b.vertices_) {
-            codes.emplace_back(pv.first, pv.second, processVertex(pv.first, pv.second));
+            FilterContext filterContext;
```
wadeliuyi: why not move parameter FilterContext to processVertex?

src/storage/QueryBaseProcessor.inl
dangleptr: Because we have checkExp at first.

src/storage/QueryBaseProcessor.inl
dangleptr: Good suggestion

src/storage/QueryBaseProcessor.inl
dangleptr: Yep.

src/storage/CommonUtils.h
dangleptr: Good catch.

src/storage/CommonUtils.h
dangleptr: Because we have no related ctor inside PropContext....

src/storage/QueryBaseProcessor.inl
dangleptr: Very good suggestion

src/storage/CommonUtils.h
```diff
@@ -13,17 +13,11 @@
...
-    }
-};
+using TagProp = std::pair<std::string, std::string>;
 
 struct FilterContext {
-    // key: tagName + propName -> propValue
-    // TODO(heng): figure out a better way.
-    std::unordered_map<std::string, VariantType> tagFilters_;
+    // key: <tagName, propName> -> propValue
+    std::unordered_map<TagProp, VariantType> tagFilters_;
```
whitewum: great

src/storage/QueryBaseProcessor.inl
```diff
@@ -274,20 +274,20 @@ void QueryBaseProcessor<REQ, RESP>::collectProps(RowReader* reader,
             }
             auto&& v = value(std::move(res));
             if (prop.fromTagFilter()) {
-                fcontext->tagFilters_.emplace(prop.tagFilterKey(), v);
+                fcontext->tagFilters_.emplace(std::make_pair(prop.tagOrEdgeName(), name), v);
             }
             if (prop.returned_) {
                 switch (v.which()) {
-                    case 0:
+                    case VAR_INT64:
```
whitewum: It's OK if temperly used.

src/storage/CommonUtils.h
wadeliuyi: why define a ctor in PropContext?

PR #413, opened 2019-05-22
[hbasestore] HBase-powered KVStore implementation

### LINKED ISSUE #179: HBase-powered KVStore implementation
Besides NebulaStore (raft-based) KVStore, we want to implement a HBaseStore to use HBase as the storage. That serves two purposes

* Such architecture can be used in latency-insensitive scenario
* Can be a code demo showing how to integrate Nebula with 3rd-party storages

zhangguoqing:
close #179


zhangguoqing:
TODO:
1. use the NebulaKeyUtils #402 to make the storage pass by modifying CMakeLists.txt only.
2. use a test runner to set up a inside hbase testbed.

dangleptr:
> TODO:
> 
> 1. use the NebulaKeyUtils #402 to make the storage pass by modifying CMakeLists.txt only.
> 2. use a test runner to set up a inside hbase testbed.

Excellent!  
How do you run your UTs currently?

zhangguoqing:
Currently, a standalone HBase environment with local file mode was started on my development machine. And all test cases are pass. :)

dangleptr:
> Currently, a standalone HBase environment with local file mode was started on my development machine. And all kvstore test cases are pass. :)

That's cool.  Do you want to implement the runner you mentioned?

zhangguoqing:
@sherman-the-tank said before that he would implement this runner, and it seems that he is very experienced in this case. What I did might not meet his expectations, awkward ^_^|||.

dangleptr:
> awkward

Ha Ha...  That's fine!  
@sherman-the-tank  is too busy recently,  if you want to have a try,  please go ahead.  
We could discuss how to do it.

zhangguoqing:
Thx. But now, I want to take one new case with graph and parser modules. That will help me understand Nebula in a more comprehensive and holistic manner. ;-)

src/interface/hbase.thrift
```diff
@@ -0,0 +1,622 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+// NOTE: The "required" and "optional" keywords for the service methods are purely for documentation
+
+namespace java org.apache.hadoop.hbase.thrift2.generated
```
darionyaphet: Thanks for your great work 🎉 

The IDL under `interface` was used to define the protocol to communication between components.

It seems better to move the file to another place and take a discussion for more details.

dangleptr:
> Thx. But now, I want to take one new case with graph and parser modules. That will help me understand Nebula in a more comprehensive and holistic manner. ;-)

That's fine.  Let's disable the UTs firstly before hbase runner merged in.   Let me take care of it.

src/interface/hbase.thrift
zhangguoqing: You are right, but this hbase.thrift file just copy from [1], do we really want to modify it?  I think that we don't make any changes to it is the best choice. :-)

[1] https://github.com/apache/hbase/blob/rel/1.4.9/hbase-thrift/src/main/resources/org/apache/hadoop/hbase/thrift2/hbase.thrift

src/kvstore/HBaseEngine.cpp
```diff
@@ -0,0 +1,353 @@
...
+
+
+void HBaseEngine::removePart(PartitionID partId) {
+    UNUSED(partId);
+    LOG(INFO) << "HBaseEngine does not support the \"removePart\" method.";
+}
+
+
+std::vector<PartitionID> HBaseEngine::allParts() {
+    std::vector<PartitionID> parts;
```
darionyaphet: Here is no `Partition` in `HBase` and don't support `addPart` and `removePart`.
Why return `0` as result ?

src/kvstore/HBaseEngine.cpp
```diff
@@ -0,0 +1,353 @@
...
+    parts.emplace_back(0);
+    return parts;
+}
+
+
+ResultCode HBaseEngine::ingest(const std::vector<std::string>& files) {
+    UNUSED(files);
+    LOG(INFO) << "HBaseEngine does not support the \"ingest\" method.";
+    return ResultCode::ERR_UNKNOWN;
```
darionyaphet: Would append a new `ErrorCode` which is named `ERR_NOT_SUPPORT` ?

dangleptr:
After skim the pr,  i think we need to refactor the kvstore module. 
Maybe the structure looks like following:
```
kvstore
 |--------- kvstore.h
 |----------inhouse / ...   
 |----------hbase   / ...
```

src/interface/hbase.thrift
zhangguoqing: Yes, agree! I have considered this issue before, and I finally put it here for a while, to see if anyone has a better directory location suggestion? :-)

src/kvstore/HBaseEngine.cpp
zhangguoqing: good idea. 👍

src/kvstore/KVStore.h
```diff
@@ -13,11 +13,20 @@
...
+#include "meta/SchemaManager.h"
 
 namespace nebula {
 namespace kvstore {
 
 struct KVOptions {
+    /**
+     * HBase thrift server address.
+     * */
+    HostAddr hbaseServer_;
```
darionyaphet: move it into `HBaseStore.h` 
PS the HBase relation code moved to a new directory would be better ?

zhangguoqing: Yep, my basic principle for implementing this PR is to try not to change the existing directory structure. I am worried that it is not reasonable. Now, I will adopt your new proposal. thx.

src/kvstore/test/HBaseEngineTest.cpp
```diff
@@ -0,0 +1,312 @@
...
+            edgeWriter << iInt;
+        }
+        for (int32_t iString = 10; iString < 20; iString++) {
+            edgeWriter << folly::stringPrintf("string_col_%d", iString);
+        }
+        auto edgeValue = edgeWriter.encode();
+        edgeData.emplace_back(edgeKey, edgeValue);
+    }
+
+    auto hbase_engine = std::make_unique<HBaseEngine>(
```
darionyaphet: CamelCase

src/kvstore/test/HBaseEngineTest.cpp
```diff
@@ -0,0 +1,312 @@
...
+        for (int32_t iInt = 0; iInt < 10; iInt++) {
+            edgeWriter << iInt;
+        }
+        for (int32_t iString = 10; iString < 20; iString++) {
+            edgeWriter << folly::stringPrintf("string_col_%d", iString);
+        }
+        auto edgeValue = edgeWriter.encode();
+        edgeData.emplace_back(edgeKey, edgeValue);
+    }
+    auto hbase_engine = std::make_unique<HBaseEngine>(
```
darionyaphet: ditto

src/kvstore/test/HBaseEngineTest.cpp
```diff
@@ -0,0 +1,312 @@
...
+        for (int32_t iInt = 0; iInt < 5; iInt++) {
+            edgeWriter << iInt;
+        }
+        for (int32_t iString = 5; iString < 10; iString++) {
+            edgeWriter << folly::stringPrintf("string_col_%d", iString);
+        }
+        auto edgeValue = edgeWriter.encode();
+        edgeData.emplace_back(edgeKey, edgeValue);
+    }
+    auto hbase_engine = std::make_unique<HBaseEngine>(
```
darionyaphet: ditto

src/kvstore/test/HBaseEngineTest.cpp
zhangguoqing: Sorry, my mistakes.

src/kvstore/HBaseEngine.cpp
```diff
@@ -0,0 +1,353 @@
...
+    CHECK_NOTNULL(schemaMan_);
+    LOG(INFO) << "Connect to the HBase thrift server "
+              << network::NetworkUtils::intToIPv4(hbaseServer.first)
+              << ":" << hbaseServer.second;
+    client_ = std::make_unique<HBaseClient>(hbaseServer);
+    CHECK_NOTNULL(client_.get());
+    tableName_ = kHBaseTableNamePrefix + folly::to<std::string>(spaceId_);
+    partsNum_ = this->allParts().size();
+    // TODO(zhangguoqing) check/create the hbase table named tableName_.
+    // Now, the version 1.4.9 of HBase thrift2 does not support APIs for table operations.
```
darionyaphet: why `HBase 1.4.9` ? The latest one is `2.1.4`

src/kvstore/HBaseEngine.cpp
```diff
@@ -0,0 +1,353 @@
...
+    endRowKey.reserve(kMaxRowKeyLength);
+    startRowKey.append(prefix);
+    endRowKey.append(prefix);
+    for (size_t n = 0; n < kMaxRowKeyLength - prefix.size(); n++) {
+        startRowKey.append(reinterpret_cast<const char*>(&kFillMin), sizeof(uint8_t));
+        endRowKey.append(reinterpret_cast<const char*>(&kFillMax), sizeof(uint8_t));
+    }
+    ResultCode code = this->range(startRowKey, endRowKey, storageIter);
+    if (code == ResultCode::ERR_IO_ERROR) {
+        LOG(ERROR) << "Prefix Failed: the HBase I/O error.";
```
darionyaphet: add `prefix` in log info

src/kvstore/HBaseEngine.cpp
```diff
@@ -0,0 +1,353 @@
...
+        : KVEngine(spaceId)
+        , schemaMan_(schemaMan) {
+    CHECK_NOTNULL(schemaMan_);
+    LOG(INFO) << "Connect to the HBase thrift server "
+              << network::NetworkUtils::intToIPv4(hbaseServer.first)
+              << ":" << hbaseServer.second;
+    client_ = std::make_unique<HBaseClient>(hbaseServer);
+    CHECK_NOTNULL(client_.get());
+    tableName_ = kHBaseTableNamePrefix + folly::to<std::string>(spaceId_);
+    partsNum_ = this->allParts().size();
```
darionyaphet: `partsNum_` is not needed

src/interface/hbase.thrift
darionyaphet: IMO, it should be closer to the place of use.

src/kvstore/HBaseEngine.cpp
zhangguoqing: In my opinion, if we really want to use this feature in a real production environment, we should be most compatible with the widely used version rather than the latest version 2.1.4. As far as I know, hbase version 1.4.9 is widely used, and this version of thrift2 is also forward compatible with rel/1.4.0, and which is not substantially different from rel/2.1.4 [1].  The APIs for table operation is support after the branch-2.2 which is not released now.

```
[1] diff hbase.thrift149 hbase.thrift214
78c78, 80
< DELETE_COLUMNS = 1
---
> DELETE_COLUMNS = 1,
> DELETE_FAMILY = 2,
> DELETE_FAMILY_VERSION = 3
210a213,218
> enum TReadType {
> DEFAULT = 1,
> STREAM = 2,
> PREAD = 3
> }
>
229c237,238
< 14: optional bool small
---
> 14: optional TReadType readType
> 15: optional i32 limit
```

darionyaphet: Understood. According to your statement, we should create table from shell before using it .

src/kvstore/HBaseEngine.cpp
zhangguoqing: Yes, this is just for overriding interfaces [1] and [2] of KVEngine. If we remove it, that two interfaces may not looks pretty anymore.  And, in fact, the HBaseStore should has **ONE** big partition(partsNum_ = 1). What do you think about it? 

[1] std::vector<PartitionID> allParts() override;   https://github.com/vesoft-inc/nebula/pull/413/files#diff-cc6e1be77ba890ee429b6461fa20a766L70
[2] int32_t totalPartsNum() override;  https://github.com/vesoft-inc/nebula/pull/413/files#diff-cc6e1be77ba890ee429b6461fa20a766L75

src/kvstore/HBaseEngine.cpp
zhangguoqing: OK.

src/kvstore/HBaseEngine.cpp
zhangguoqing: Yes, for unit tests, these operations[1] will be done by test runner.

[1] https://github.com/vesoft-inc/nebula/pull/413/files#diff-9d431e623cca5bcbd11ca24b91c18bb2R13

src/kvstore/HBaseEngine.cpp
zhangguoqing: There is a **ONE** big partition.   FYI. https://github.com/vesoft-inc/nebula/issues/179#issuecomment-482502357
Another reason is  just for overriding interface of KVEngine.

zhangguoqing:
> After skim the pr, i think we need to refactor the kvstore module.
> Maybe the structure looks like following:
> 
> ```
> kvstore
>  |--------- kvstore.h
>  |----------inhouse / ...   
>  |----------hbase   / ...
> ```

Now, I will just extract hbase as a separate module.   FYI. #344

src/interface/hbase.thrift
zhangguoqing: OK, it will be put into src/kvstore/hbase.

src/kvstore/hbase/HBaseStore.h
```diff
@@ -0,0 +1,128 @@
...
+namespace nebula {
+namespace kvstore {
+
+struct GraphSpaceHBase {
+    Engine engine_;
+    std::set<PartitionID> parts_;
+};
+
+
+class HBaseStore : public KVStore, public Handler {
```
dangleptr: HBaseStore should not derived from Handler.

src/kvstore/CMakeLists.txt
```diff
@@ -3,10 +3,12 @@ add_library(
     Part.cpp
     RocksEngine.cpp
     PartManager.cpp
+    KVStore.cpp
     NebulaStore.cpp
     RocksEngineConfig.cpp
 )
-add_dependencies(kvstore_obj base_obj meta_client)
+add_dependencies(kvstore_obj base_obj meta_client hbasestore_obj)
```
dangleptr: kvstore_obj should NOT depend on hbasestore_obj. 
They are on the same level.

src/kvstore/test/CMakeLists.txt
```diff
@@ -2,6 +2,9 @@ add_executable(
     part_test
     PartTest.cpp
     $<TARGET_OBJECTS:kvstore_obj>
+    $<TARGET_OBJECTS:hbasestore_obj>
```
dangleptr: UTs under kvstore should NOT depends on hbaseXXX

src/kvstore/HBaseEngine.cpp
zhangguoqing: In the end, I still agree with your suggestion.

src/kvstore/hbase/HBaseStore.h
zhangguoqing: Done!

src/kvstore/CMakeLists.txt
zhangguoqing: The hbasestore will be as a plugin.

src/kvstore/test/CMakeLists.txt
zhangguoqing: Should Open a new issue to implement HBaseStore in plugin form.

src/kvstore/plugins/hbase/HBaseStore.h
```diff
@@ -0,0 +1,114 @@
...
+
+private:
+    void addSpace(GraphSpaceID spaceId);
+
+    void removeSpace(GraphSpaceID spaceId);
+
+    Engine newEngine(GraphSpaceID spaceId);
+
+private:
+    std::unordered_map<GraphSpaceID, Engine> kvs_;
```
dangleptr: Why we need it in hbase store?

src/kvstore/plugins/hbase/HBaseStore.h
```diff
@@ -0,0 +1,114 @@
...
+
+    folly::RWSpinLock lock_;
+
+    std::unique_ptr<meta::SchemaManager> schemaMan_{nullptr};
+
+    std::unique_ptr<PartManager> partMan_{nullptr};
```
dangleptr: ditto

src/kvstore/test/CMakeLists.txt
```diff
@@ -2,6 +2,7 @@ add_executable(
     part_test
     PartTest.cpp
     $<TARGET_OBJECTS:kvstore_obj>
+    $<TARGET_OBJECTS:dataman_obj>
```
dangleptr: Do we still need dataman_obj after hbase_obj removed ?

src/kvstore/plugins/hbase/HBaseStore.h
zhangguoqing: Each GraphSpace corresponds to an HBaseEngine, which is associated with a table(operated by an HBaseClient). On each storage node, we use **kv_** to store this mapping relationship.

src/kvstore/plugins/hbase/HBaseStore.h
zhangguoqing: The **schemaMan_** is used to obtain the scheam of Tag and EdgeType. When data is obtained or stored from HBase, scheam is needed to encode and decode the data.

The **partMan_** is used to obtain the partsMap, which can be used to know which GraphSpace is included in the Nebula cluster, and then initialize the GraphSpace to create the corresponding HBaseEngine.

src/kvstore/test/CMakeLists.txt
zhangguoqing: According to the design #179 , when using HBase to store data,  need to perform encode and decode operations, so it needs to use the **dataman** module. We can't remove it here.

src/kvstore/plugins/hbase/HBaseStore.h
dangleptr: I don't think we really need it.   SpaceId equals table and we could pass tableName to HBaseClient interfaces.

src/kvstore/test/CMakeLists.txt
zhangguoqing: Yes, you are right, I will fix it soon.

src/kvstore/plugins/hbase/HBaseStore.h
dangleptr: Ditto.  I don't think we really need partMan_

dangleptr:
By the way, you should disable all UTs before runner merged in.

dangleptr:
Close #179

src/interface/.gitignore
```diff
@@ -1,3 +0,0 @@
-gen-*
```
darionyaphet: why remove it？

zhangguoqing: Its contents has been moved to the .gitignore file in the project root directory.

src/kvstore/plugins/hbase/HBaseStore.h
zhangguoqing: done!

src/kvstore/plugins/hbase/HBaseStore.h
zhangguoqing: done!

dangleptr:
The pr looks better now.  
Jenkins go

dangleptr:
Jenkins go

dangleptr:
Jenkins go

dangleptr:
Jenkins go

dangleptr:
Jenkins go

dangleptr:
Jenkins go

dangleptr:
Jenkins go

PR #415, opened 2019-05-22
Add init of GraphService and check it

### LINKED ISSUE #557: Fix failed UT data_test
http://ci.vesoft.com/job/Nebula-Graph-UnitTest/1223/console

Please recheck it @laura-ding

laura-ding:
close #557

### DIFF
--- a/src/common/network/NetworkUtils.cpp
+++ b/src/common/network/NetworkUtils.cpp
@@ -240,9 +240,6 @@ StatusOr<HostAddr> NetworkUtils::toHostAddr(folly::StringPiece ipPort) {
 }
 
 StatusOr<std::vector<HostAddr>> NetworkUtils::toHosts(const std::string& peersStr) {
-    if (peersStr.empty()) {
-        return Status::Error("Input string(ip:port) empty");
-    }
     std::vector<HostAddr> hosts;
     std::vector<std::string> peers;
     folly::split(",", peersStr, peers, true);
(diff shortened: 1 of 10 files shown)


src/common/network/NetworkUtils.cpp
```diff
@@ -230,6 +230,9 @@ StatusOr<HostAddr> NetworkUtils::toHostAddr(folly::StringPiece ipPort) {
 }
 
 StatusOr<std::vector<HostAddr>> NetworkUtils::toHosts(const std::string& peersStr) {
+    if (peersStr.empty()) {
```
dangleptr: I think we should NOT forbid empty peersStr.  
Because in some cases,  we actually pass in empty string.

src/graph/ExecutionEngine.cpp
```diff
@@ -25,9 +25,10 @@ ExecutionEngine::~ExecutionEngine() {
 
 Status ExecutionEngine::init(std::shared_ptr<folly::IOThreadPoolExecutor> ioExecutor) {
     auto addrs = network::NetworkUtils::toHosts(FLAGS_meta_server_addrs);
-    if (!addrs.ok()) {
+    if (!addrs.ok() || addrs.value().empty()) {
```
dangleptr: If FLAGS_meta_server_addrs is not empty,  in which case "toHosts" return empty array?

src/graph/test/TestEnv.cpp
```diff
@@ -28,7 +28,11 @@ void TestEnv::SetUp() {
     using ThriftServer = apache::thrift::ThriftServer;
     server_ = std::make_unique<ThriftServer>();
     server_->getIOThreadPool()->setNumThreads(1);
```
whitewum: what about multiple threads tests?

laura-ding:
the pr is so long time, close it

PR #419, opened 2019-05-22
Added WallClock class

sherman-the-tank:
To provide much faster xxxNowInSec() and xxxNowInMilliSec() methods

This class will replace TimeUtils class in a separate PR

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -59,6 +59,11 @@ if(NOT CMAKE_BUILD_TYPE)
     set(CMAKE_BUILD_TYPE "Debug")
 endif(NOT CMAKE_BUILD_TYPE)
 
+if(NOT ${CMAKE_BUILD_TYPE} EQUAL "Debug")
+    message(STATUS "Set macro _FORTIFY_SOURCE=2")
+    add_definitions(-D_FORTIFY_SOURCE=2)
+endif()
+
 message(STATUS "CMAKE_BUILD_TYPE: ${CMAKE_BUILD_TYPE}, "
                "options: Debug, Release, RelWithDebInfo, MinSizeRel")
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -59,13 +59,13 @@ if(NOT CMAKE_BUILD_TYPE)
     set(CMAKE_BUILD_TYPE "Debug")
 endif(NOT CMAKE_BUILD_TYPE)
 
-if(NOT ${CMAKE_BUILD_TYPE} EQUAL "Debug")
+if(NOT ${CMAKE_BUILD_TYPE} STREQUAL "Debug")
     message(STATUS "Set macro _FORTIFY_SOURCE=2")
     add_definitions(-D_FORTIFY_SOURCE=2)
 endif()
 
-message(STATUS "CMAKE_BUILD_TYPE: ${CMAKE_BUILD_TYPE}, "
-               "options: Debug, Release, RelWithDebInfo, MinSizeRel")
+message(STATUS "CMAKE_BUILD_TYPE = ${CMAKE_BUILD_TYPE} "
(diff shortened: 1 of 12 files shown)


sherman-the-tank:
jenkins go

src/common/time/WallClock.h
```diff
@@ -0,0 +1,39 @@
...
+ *
+ * The *fast* versions are way faster than the *slow* versions, but not as
+ * precise as the *slow* versions. Choose wisely
+ *
+ */
+class WallClock final {
+public:
+    WallClock() = delete;
+
+    static int64_t slowNowInSec();
```
dangleptr: Do we need slowXXX any more?

Why not change TimeUtils Implementation directly?  Why use another interface?

sherman-the-tank: The name ```TimeUtils``` is too general. It's better to make names more specific and meaningful. What we need here is to get the wall clock time

```slowXXX()``` is more accurate than ```fastXXX()```, but slower, so we keep them for scenarios that require accurate time

Comment (dangleptr) [reply src/common/time/WallClock.h]: > slowXXX() is more accurate than fastXXX(), but slower, so we keep them for scenarios that require accurate time

Sounds reasonable.  

> The name TimeUtils is too general. It's better to make names more specific and meaningful. What we need here is to get the wall clock time

We'd better use another name to avoid copyright infringement

**Comment by dutor** [comment]
Here is the benchmarking:

```
/home/dutor/Wdir/nebula/src/common/time/test/WallClockBenchmark.cpprelative  time/iter  iters/s
gettimeofday_get_msec                                       24.40ns   40.99M
systemclock_get_time_t                           129.21%    18.88ns   52.96M
wallclock_get_msec_slow                          129.91%    18.78ns   53.25M
wallclock_get_msec_fast                          172.47%    14.15ns   70.69M
----------------------------------------------------------------------------
gettimeofday_get_sec                                        17.21ns   58.10M
wallclock_get_sec_slow                            95.37%    18.05ns   55.41M
wallclock_get_sec_fast                           129.28%    13.31ns   75.12M
```

```
/home/dutor/Wdir/nebula/src/common/time/test/DurationBenchmark.cpprelative  time/iter  iters/s
stead_clock_timer                                           44.45ns   22.50M
duration_timer                                   170.50%    26.07ns   38.36M
```

The CPU module is:
```
Intel(R) Xeon(R) CPU E5-2690 v2 @ 3.00GHz
```

File: src/common/time/WallClock.h (hunk 1 above)
Comment (sherman-the-tank) [reply src/common/time/WallClock.h]: WallClock is not copyrighted. It's been used in a lot of places

File: CMakeLists.txt [hunk 2]
```diff
@@ -59,6 +59,11 @@ if(NOT CMAKE_BUILD_TYPE)
     set(CMAKE_BUILD_TYPE "Debug")
 endif(NOT CMAKE_BUILD_TYPE)
 
+if(NOT ${CMAKE_BUILD_TYPE} EQUAL "Debug")
```
dutor: It seems that `STREQUAL` must be used instead here.

sherman-the-tank: Good catch 👍

whitewum:
> Excellent! 👍
> 
> I want to know the usage scenarios of inaccurate methods such as fastNowInSec?

In some case, time accuracy is not so important, just a timestamp is enough.

src/common/time/detail/TscHelper.cpp
```diff
@@ -0,0 +1,78 @@
...
+        });
+    t.detach();
+}
+
+}  // Anonymous namespace
+
+
+volatile uint64_t readTsc() {
+#ifdef DURATION_USE_RDTSCP
+    uint32_t eax, ecx, edx;
```
laura-ding: ecx is not use

dutor:
Jenkins go!

PR #423, opened 2019-05-24
Support desc keyword

darionyaphet:


src/parser/parser.yy
```diff
@@ -596,12 +596,18 @@ describe_tag_sentence
     : KW_DESCRIBE KW_TAG name_label {
         $$ = new DescribeTagSentence($3);
     }
+    | KW_DESC KW_TAG LABEL {
```
dutor: `name_label`

src/parser/parser.yy
```diff
@@ -596,12 +596,18 @@ describe_tag_sentence
...
+        $$ = new DescribeTagSentence($3);
+    }
     ;
 
 describe_edge_sentence
     : KW_DESCRIBE KW_EDGE name_label {
         $$ = new DescribeEdgeSentence($3);
     }
+    | KW_DESC KW_EDGE LABEL {
```
dutor: ditto.

src/parser/parser.yy
```diff
@@ -941,6 +947,9 @@ describe_space_sentence
     : KW_DESCRIBE KW_SPACE LABEL {
         $$ = new DescribeSpaceSentence($3);
     }
+    | KW_DESC KW_SPACE LABEL {
```
dutor: ditto.

src/parser/parser.yy
darionyaphet: fixed ~

steppenwolfyuetong:
This modification is not perfect. I think it is more efficient to use synonyms in` scanner.lex`  files.

dutor:
> This modification is not perfect. I think it is more efficient to use synonyms in` scanner.lex` files.

Good point, but this is a special one. Because `DESC` is not just a synonym to `DESCRIBE`, but also a standalone keyword in the comming `ORDERED BY` statement.

I haven't digged up into the impl of MySQL or other RDBMS. If you have some other smarter ideas, please enlighten me.

steppenwolfyuetong:
> Good point, but this is a special one. Because `DESC` is not just a synonym to `DESCRIBE`, but also a standalone keyword in the comming `ORDERED BY` statement

@dutor,    👍，Yeah，`DESC` also uses in  `ORDER BY` statement.  But here `DESC` is equivalent to `DESCRIBE`.
 
so I think  it could be better using as follow:
```
decribe_command:
       KW_DESC
    |  KW_DESCRIBE
    ;

```
use decribe_command

darionyaphet:
Using `synonym` to handle `KW_DESC` and `KW_DESCRIBE` seems not a good idea.

`DESC` is another keyword.  They are not equivalent.

How to deal with `ORDER BY  DESCRIBE` ?

darionyaphet:
In test case `CREATE SPACE default_space` have modify replica_factor from `3` to `1`,  fixed

darionyaphet:
jenkins go

PR #425, opened 2019-05-24
Use active hosts when creating space

dangleptr:
subtask of #283 

Add a flag hosts_whitelist_enabled to control whether we filter host in white list or not after received hb.


dangleptr:
Jenkins go

src/meta/processors/CreateSpaceProcessor.cpp
```diff
@@ -22,20 +23,25 @@ void CreateSpaceProcessor::process(const cpp2::CreateSpaceReq& req) {
...
-    auto hosts = ret.value();
     auto spaceName = properties.get_space_name();
     auto partitionNum = properties.get_partition_num();
     auto replicaFactor = properties.get_replica_factor();
     VLOG(3) << "Create space " << spaceName << ", id " << spaceId;
-
+    if ((int32_t)hosts.size() < replicaFactor) {
+        LOG(ERROR) << "Not enough hosts existed for replica "
+                   << replicaFactor << ", hosts num " << hosts.size();
+        resp_.set_code(cpp2::ErrorCode::E_UNSUPPORTED);
```
darionyaphet: `Unsupport` is the meaning that this operation does not implement.

`Illegal_Argument` will be better ?

src/meta/processors/HBProcessor.cpp
```diff
@@ -11,10 +11,28 @@ DEFINE_int32(expired_hosts_check_interval_sec, 20,
...
 
 namespace nebula {
 namespace meta {
 
+void HBProcessor::process(const cpp2::HBReq& req) {
+    HostAddr host(req.host.ip, req.host.port);
+    if (FLAGS_hosts_whitelist_enabled
+            && hostExist(MetaServiceUtils::hostKey(host.first, host.second))
+                == Status::HostNotFound()) {
+        LOG(INFO) << "Reject unregistered host " << host << "!";
```
darionyaphet: LOG(ERROR)?

src/meta/processors/CreateSpaceProcessor.cpp
dangleptr: Thrift will return its own failure if related method does not implement.  So i think UNSUPPORTED is ok in this case.

src/meta/processors/HBProcessor.cpp
dangleptr: Actually, it is not an error.

dangleptr:
Let's merge it firstly to avoid pending tasks in the next.

dangleptr:
@darionyaphet @critical27

dangleptr:
Jenkins go

sherman-the-tank:
> Let's merge it firstly to avoid pending tasks in the next.

This SHOULD NOT be the reason to get PR approved! PLEASE remember you are responsible for your vote!

We need to make sure all code committed are high quality, not temporary. Especially NOT just to avoid merge conflict.

dangleptr:
> > Let's merge it firstly to avoid pending tasks in the next.
> 
> This SHOULD NOT be the reason to get PR approved! PLEASE remember you are responsible for your vote!
> 
> We need to make sure all code committed are high quality, not temporary. Especially NOT just to avoid merge conflict.

Do you have any concerns about this pr?

dangleptr:
> This SHOULD NOT be the reason to get PR approved! PLEASE remember you are responsible for your vote!

This pr have been reviewed,  don't you see the comments by @darionyaphet ?
And @critical27 has been worked around the work recently.  Why you say this pr is not reviewed?
I have left enough time for all your guys to review each pr each time.  

You could have no suggestions or comments, and you could vote -1 if you have any other thoughts, but please don't say the pr is NOT reviewed just because you haven't reviewed it.  @sherman-the-tank

PR #445, opened 2019-05-27
Optimize GQLParser by using raw buffer instead of istream

dutor:
Now, there are still two instances of `std::istream`(`yyin`) and `std::ostream`(`yyout`) left in the `yyFlexLexer` class, i.e. the base class of `GraphScanner`, which is provided by *Flex*, which we cannot optimize them away directly.

For a workaround, we could make the whole parser reentrant, even on the failure cases, so that the parser could be used as a static, thread-local instance.


dutor:
Jenkins, please run tests.

src/parser/GQLParser.h
```diff
@@ -16,6 +16,21 @@ namespace nebula {
...
+                end_ = nullptr;
+                return 0;
+            }
+            int left = end_ - pos_;
+            auto n = maxSize > left ? left : maxSize;
+            ::memcpy(buf, pos_, n);
+            pos_ += n;
+            return n;   // Number of bytes we actually filled in `buf'
+        };
+        scanner_.setReadBuffer(std::move(readBuffer));
```
boshengchen: Great work. let me benefit greatly by this PR. I have a little question ：
Here, we using memcpy, copy the memory from `pos_` to `but`.  How to destroy both?
Is it the below destructor? 
```
yyFlexLexer::~yyFlexLexer()
{
	delete [] yy_state_buf;
	yyfree(yy_start_stack  );
	yy_delete_buffer( YY_CURRENT_BUFFER );
	yyfree(yy_buffer_stack  );
}
```

src/parser/GQLParser.h
```diff
@@ -24,11 +39,18 @@ class GQLParser {
...
+    StatusOr<std::unique_ptr<SequentialSentences>> parse(std::string query) {
+        // Since GraphScanner needs a writable buffer, we have to copy the query string
+        buffer_ = std::move(query);
+        pos_ = &buffer_[0];
+        end_ = pos_ + buffer_.size();
+
         auto ok = parser_.parse() == 0;
         if (!ok) {
+            pos_ = nullptr;
+            end_ = nullptr;
```
boshengchen: :+1:

src/parser/GQLParser.h
dutor: Yeah. Flex itself maintains a mutable buffer, of which we don't bother to take care of the lifetime.

But you reminded me to find an existing leak that happens on the failure case. Thanks for that.

dutor: Addressed, please have a check.

dutor:
> Unit testing failed.

Due to failure in raft, jenkins please run tests again!

dutor:
Jenkins, go!

PR #446, opened 2019-05-28
Move the value's serialization from Graph into Storage

darionyaphet:
* Move the value's serialization from `graph layer` into `storage layer`.

* Add field name

They are used to judge index operation.


src/graph/InsertEdgeExecutor.cpp
```diff
@@ -47,51 +48,59 @@ void InsertEdgeExecutor::execute() {
...
         }
+
         {
             auto &out = edges[index++];
             out.key.set_src(src);
             out.key.set_dst(dst);
             out.key.set_ranking(rank);
             out.key.set_edge_type(edgeType_);
-            out.props = writer.encode();
+            for (auto property : properties_) {
```
dutor: Serialize the property names for every row of values?

I seriously don't agree with that.

darionyaphet: In `Storage Layer`, I wish to check the fields is indexed or not .

dangleptr:
After talked with @sherman-the-tank,  we'd better keep the serialization at query engine.  
There are several reasons:
1.  Performance,   thrift encoding/decoding is much slower than dataman.
2. In our case, we only need to extract some column value from encoded value.  And dataman is designed for it,  it support read one column without decode the whole value. 

@darionyaphet

darionyaphet:
`dataman` support read columns to avoid deserialize the whole binary value. This is a good point. :)

As a result, should just support fields in storage layer to make sure I can pick out the fields .

PR #449, opened 2019-05-28
Fixed the build

sherman-the-tank:
Some linux releases have turned on _FORTIFY_SOURCE=2, which broke our build. This PR tries to fix it


sherman-the-tank:
jenkins go

dutor:
As we have discussed, we should add the _FORTIFY_SOURCE explicitly to avoid such inconsistency between platforms.

sherman-the-tank:
Fix #474

Issue #440, opened 2019-05-27
Support displaying online hosts in "SHOW HOSTS" command


dangleptr:
subtask of #283 

After #425 merged-in,  we create space on real active hosts.  So we expect to show active hosts in "SHOW HOSTS" command, maybe as follows:

```
=============================
|          Ip |  Port | Status |
=============================
| 192.168.8.5 | 65500 | online |
-----------------------
| 192.168.8.1 | 65500 | offline |
-----------------------

```

critical27:
on the fly

whitewum:
@dangleptr  add commands to the nGQL doc pls.

whitewum:
@jude-zhu @Amber1990Zhang can U help to check this command, and add to nGQL doc?

Amber1990Zhang:
> @jude-zhu @Amber1990Zhang can U help to check this command, and add to nGQL doc?

Got it.

PR #450, opened 2019-05-28
Support displaying online hosts in "SHOW HOSTS" command and save hosts info in rocskdb

critical27:
as title #440


critical27:
jenkins go

dangleptr:
You could run UTs locally with ctest, and the failure could be check on http://ci.vesoft.com/job/Nebula-Graph-UnitTest/

dangleptr:
Close #440

critical27:
jenkins go

dangleptr:
Please rebased on master

src/graph/test/SchemaTest.cpp
```diff
@@ -34,14 +35,16 @@ TEST_F(SchemaTest, metaCommunication) {
...
+                network::NetworkUtils::toHosts("127.0.0.1:1000, 127.0.0.1:1100").value());
     }
     {
         cpp2::ExecutionResponse resp;
         std::string query = "SHOW HOSTS";
         client->execute(query, resp);
-        std::vector<uniform_tuple_t<std::string, 2>> expected{
-            {"127.0.0.1", "1000"},
-            {"127.0.0.1", "1100"},
+        std::vector<uniform_tuple_t<std::string, 3>> expected{
```
dangleptr: We'd better test one case with host offline.

src/meta/processors/BaseProcessor.inl
```diff
@@ -126,6 +127,44 @@ StatusOr<std::vector<nebula::cpp2::HostAddr>> BaseProcessor<RESP>::allHosts() {
     return hosts;
 }
 
+template<typename RESP>
+StatusOr<std::vector<cpp2::HostItem>> BaseProcessor<RESP>::allHostsWithStatus() {
+    std::vector<cpp2::HostItem> hostItems;
+    const auto& prefix = MetaServiceUtils::hostPrefix();
+    std::unique_ptr<kvstore::KVIterator> iter;
+    auto ret = kvstore_->prefix(kDefaultSpaceId_, kDefaultPartId_, prefix, &iter);
+    if (ret != kvstore::ResultCode::SUCCEEDED) {
```
dangleptr: If whitelist disabled,  no hosts existed in rocksdb is normal.

src/graph/test/SchemaTest.cpp
critical27: roger that

src/meta/processors/BaseProcessor.inl
critical27: nice catch

critical27:
jenkins go

src/graph/test/SchemaTest.cpp
critical27: SchemaTest has other ut which need to use active hosts. I add ut in meta/test/ProcessorTest.

critical27:
jenkins go

src/meta/processors/BaseProcessor.h
```diff
@@ -170,6 +170,11 @@ class BaseProcessor {
      * */
     StatusOr<std::vector<nebula::cpp2::HostAddr>> allHosts();
 
+    /**
+     * Get all hosts with online/offline status.
+     * */
+    StatusOr<std::vector<cpp2::HostItem>> allHostsWithStatus();
```
dangleptr: We'd better put the method in ListHostProcessor if no other processors use it.

critical27:
jenkins go

src/interface/meta.thrift
```diff
@@ -89,6 +89,11 @@ struct EdgeItem {
     4: common.Schema        schema,
 }
 
+struct HostItem {
+    1: common.HostAddr      hostAddr,
+    2: string               status,
```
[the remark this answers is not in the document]
critical27: yeah

steppenwolfyuetong:
Excellent!

critical27:
jenkins go

critical27:
jenkins go

critical27:
jenkins go

critical27:
jenkins go

src/meta/client/MetaClient.cpp
```diff
@@ -438,12 +454,12 @@ folly::Future<StatusOr<bool>> MetaClient::addHosts(const std::vector<HostAddr>&
...
 
-folly::Future<StatusOr<std::vector<HostAddr>>> MetaClient::listHosts() {
+folly::Future<StatusOr<std::vector<HostStatus>>> MetaClient::listHosts() {
     cpp2::ListHostsReq req;
     return getResponse(std::move(req), [] (auto client, auto request) {
                 return client->future_listHosts(request);
             }, [this] (cpp2::ListHostsResp&& resp) -> decltype(auto) {
-                return this->to(resp.hosts);
+                return this->toHostStatus(resp.hosts);
             });
```
dangleptr: Be careful here! 
Now we need to list hosts with status,  but as current implementation,  the activeHostManager only valid on leader.  So we need to send the request to leader or we save the status in rocksdb

critical27: If we save the status in rocksdb, we need to get host info from rocksdb when instantiate ActiveHostsMan, am I right?

critical27:
According to @sherman-the-tank suggestion of specification, remove ActiveHostsManHolder. Also,  I modify host info to be stored in rocksdb instead of memory to support multi replica acco.

critical27:
jenkins go

critical27:
jenkins go

dutor:
Jenkins, go!

src/meta/ActiveHostsMan.cpp
```diff
@@ -0,0 +1,109 @@
...
+    while (it != hostsMap_.end()) {
+        if ((now - it->second.lastHBTimeInSec_) > expirationInSeconds_ && it->second.isOnline_) {
+            LOG(INFO) << it->first << " expired! last hb time " << it->second.lastHBTimeInSec_;
+            data.emplace_back(MetaServiceUtils::hostKey(it->first.first, it->first.second),
+                              MetaServiceUtils::hostValOffline());
+            it->second.isOnline_ = false;
+        }
+        it++;
+    }
+
```
dangleptr: The lock scope should be finished here.

src/meta/ActiveHostsMan.cpp
```diff
@@ -0,0 +1,109 @@
+/* Copyright (c) 2018 vesoft inc. All rights reserved.
```
dangleptr: 2019

src/meta/ActiveHostsMan.cpp
```diff
@@ -0,0 +1,109 @@
...
+#include "meta/processors/BaseProcessor.h"
+
+namespace nebula {
+namespace meta {
+
+ActiveHostsMan::ActiveHostsMan(int32_t intervalSeconds, int32_t expiredSeconds,
+                               kvstore::KVStore* kv)
+    : intervalSeconds_(intervalSeconds)
+    , expirationInSeconds_(expiredSeconds) {
+    if (kv != nullptr) {
```
dangleptr: Do you allow to pass in an empty kv?

src/meta/ActiveHostsMan.cpp
```diff
@@ -0,0 +1,109 @@
...
+        kvstore_ = kv;
+    }
+
+    init();
```
dangleptr: Please don't call init in ctor.  Because ctor has no returned value if init failed.

src/meta/ActiveHostsMan.cpp
```diff
@@ -0,0 +1,109 @@
...
+                              &ActiveHostsMan::cleanExpiredHosts,
+                              this);
+}
+
+void ActiveHostsMan::updateHostInfo(const HostAddr& hostAddr, const HostInfo& info) {
+    folly::RWSpinLock::ReadHolder rh(&lock_);
+    auto it = hostsMap_.find(hostAddr);
+    if (it == hostsMap_.end()) {
+        folly::RWSpinLock::UpgradedHolder uh(&lock_);
+        hostsMap_.emplace(std::move(hostAddr), std::move(info));
```
dangleptr: If it is a new heartbeat, we should store it in kvstore as online.

src/meta/ActiveHostsMan.h
```diff
@@ -32,97 +33,48 @@ struct HostInfo {
...
-    ActiveHostsManHolder() = delete;
-    ~ActiveHostsManHolder() = delete;
-
-    static ActiveHostsMan* hostsMan() {
-       static auto hostsMan
-           = std::make_unique<ActiveHostsMan>(FLAGS_expired_hosts_check_interval_sec,
-                                              FLAGS_expired_threshold_sec);
-       return hostsMan.get();
-    }
+    bool isOnline_ = false;
```
dangleptr: Do we need this field?  If exists in ActiveHostMan, it means the host is online.

src/meta/ActiveHostsMan.cpp
critical27: some ut which only use getActiveHosts pass in an empty kv

src/meta/ActiveHostsMan.cpp
critical27: yeah, good catch, maybe I will rename it like "initHostMap" to support raft later

src/meta/ActiveHostsMan.cpp
critical27: I add it in HBProcessor

src/meta/ActiveHostsMan.cpp
critical27: good catch

src/meta/ActiveHostsMan.h
critical27: Maybe it is not mandatory, but necessary, so we can get active hosts from memory instead of get in kvstore.

critical27:
Jenkins, go!

critical27:
Jenkins go

src/meta/ActiveHostsMan.h
```diff
@@ -34,95 +37,53 @@ struct HostInfo {
...
-        CHECK_GE(expiredSeconds, intervalSeconds)
-            << "expiredSeconds " << expiredSeconds
-            << " should >= intervalSeconds " << intervalSeconds;
-        CHECK(checkThread_.start());
-        checkThread_.addTimerTask(intervalSeconds * 1000,
-                                  intervalSeconds * 1000,
-                                  &ActiveHostsMan::cleanExpiredHosts,
-                                  this);
+    static ActiveHostsMan* instance() {
+        static ActiveHostsMan activeHostsMan(FLAGS_expired_hosts_check_interval_sec,
```
dangleptr: We'd better use a unique_ptr to hold the instance,  so it could be released when the process shut down.

src/meta/ActiveHostsMan.h
```diff
@@ -34,95 +37,53 @@ struct HostInfo {
...
+                                             FLAGS_expired_threshold_sec);
+        return &activeHostsMan;
     }
 
     ~ActiveHostsMan() {
         checkThread_.stop();
         checkThread_.wait();
+        delete hbProcessor_;
```
dangleptr: Why?

src/meta/ActiveHostsMan.h
```diff
@@ -34,95 +37,53 @@ struct HostInfo {
...
-        folly::RWSpinLock::ReadHolder rh(&lock_);
-        auto it = hostsMap_.find(hostAddr);
-        if (it == hostsMap_.end()) {
-            folly::RWSpinLock::UpgradedHolder uh(&lock_);
-            hostsMap_.emplace(std::move(hostAddr), std::move(info));
-        } else {
-            it->second.lastHBTimeInSec_ = info.lastHBTimeInSec_;
-        }
-    }
+    void init(kvstore::KVStore* kv);
```
dangleptr: Do we need the kv ptr passed in?

critical27: HBProcessor need a kv ptr. Maybe it make more sense to pass a kv ptr than a  HBProcessor?

critical27:
Jenkins go

src/daemons/MetaDaemon.cpp
```diff
@@ -115,6 +116,8 @@ int main(int argc, char *argv[]) {
             nebula::kvstore::KVStore::instance(std::move(options)));
 
     auto handler = std::make_shared<nebula::meta::MetaServiceHandler>(kvstore.get());
+    auto activeHostsMan = nebula::meta::ActiveHostsMan::instance(kvstore.get());
+    activeHostsMan->loadHostMap();
```
dangleptr: We'd better move "activeHostsMan->loadHostMap()" inside instance(),   and use std::call_once to to init the host map at the first time.

src/meta/test/ActiveHostsManTest.cpp
```diff
@@ -6,18 +6,22 @@
...
 
 namespace nebula {
 namespace meta {
 
 TEST(ActiveHostsManTest, NormalTest) {
+    fs::TempDir rootPath("/tmp/ActiveHostsManTest.XXXXXX");
+    std::unique_ptr<kvstore::KVStore> kv(TestUtils::initKV(rootPath.path()));
     auto now = time::TimeUtils::nowInSeconds();
-    ActiveHostsMan ahMan(1, 1);
+    ActiveHostsMan ahMan(1, 1, kv.get());
```
dangleptr: We don't the kv instance in the UT

src/meta/ActiveHostsMan.cpp
```diff
@@ -0,0 +1,117 @@
...
+                                    UpdateCallback cb) {
+    folly::RWSpinLock::ReadHolder rh(&lock_);
+    auto it = hostsMap_.find(hostAddr);
+    if (it == hostsMap_.end()) {
+        {
+            folly::RWSpinLock::UpgradedHolder uh(&lock_);
+            hostsMap_.emplace(std::move(hostAddr), std::move(info));
+        }
+        if (cb != nullptr) {
+            cb(hostAddr.first, hostAddr.second);
```
dangleptr: hostAddr has been moved into hostsMap,  you can't use it anymore.

critical27: good catch!

src/daemons/MetaDaemon.cpp
critical27: But loadHostMap use other non-static members, maybe call_once won't work.

critical27:
Addressed comments and rebased on master.

critical27:
Jenkins go

critical27:
Jenkins go

critical27:
Jenkins go!

src/meta/processors/BaseProcessor.h
```diff
@@ -23,6 +23,9 @@ namespace meta {
 
 using nebula::network::NetworkUtils;
 
+const PartitionID kDefaultPartId_ = 0;
```
dangleptr: Put it some where more common,  use "static const" and remove the underscore in name.

critical27:
Jenkins go

src/meta/processors/BaseProcessor.h
critical27: Some processors only depends on this header file, I will rename it anyway.

PR #454, opened 2019-05-28
Remove -march=native when compiling RocksDB

dutor:
`-march=native` would make the compiler more radical to use newer instruction set, which may introduce incompatibilities.


dutor:
Please hold on this PR, I just found that there are the same problem in other libraries, e.g. zstd.

dutor:
It seems that zstd has some kind of dynamic dispatching mechanism to detect BMI instruction set at runtime. It's probably OK to run on architectures which didn't support such instructions, even though I haven't verify this.

Now let this patch merged in firstly.

dutor:
jenkins, go

PR #455, opened 2019-05-28
Support auto completion of keywords in console

### LINKED ISSUE #448: Is it possible to add automatic completion of the function variables?
nebula> create space shakespeare(partition_num = 10, replica_factor = 1)
// it's boring to keep type the same thing and time-consuming, will it be possible if I type 'par' here the following thing would be added automatically?

dutor:
For now, we only support simple completions based on the keywords. In future, we may add support to sub-command, or we could even combine readline and our lexer and parser.

This will close #448


src/console/CliManager.cpp
```diff
@@ -240,5 +243,313 @@ void CliManager::loadHistory() {
...
+// will be set, thus we will do completion with its sub keywords.
+static bool isStartOfStatement(std::string &primaryKeyword);
+
+// Given the prefix and a collection of keywords, retrieve the longest common prefix
+// e.g. given `u' as the prefix and [USE, USER, USERS] as the collection, will return `USE'
+static auto longestCommonPrefix(std::string prefix,
+                                const std::vector<std::string>& words);
+
+// Callback by realine if an auto completion is triggered
+static char** completer(const char *text, int start, int end);
```
darionyaphet: why add `start` and `end` ?

share/resources/completion.json
```diff
@@ -0,0 +1,85 @@
...
+            "sub_keywords": [
+                "VERTEX", "EDGE", "VALUES", "NO", "OVERWRITE"
+            ]
+        },
+        "SHOW" : {
+            "sub_keywords": [
+                "HOSTS", "TAGS", "EDGES", "SPACES", "USERS", "ROLES", "USER"
+            ]
+        },
+        "DESCRIBE" : {
```
darionyaphet: `DESC` is miss ?

src/console/CliManager.cpp
dutor: This is a callback function, so they are required while we don't use them.

share/resources/completion.json
dutor: It was not available when this PR initiated.

Let me add it in.

src/console/CliManager.cpp
```diff
@@ -240,5 +243,313 @@ void CliManager::loadHistory() {
...
+        typeNames.emplace_back(tname.asString());
+    }
+
+    return Status::OK();
+}
+
+
+void CliManager::initAutoCompletion() {
+    // The completion function
+    ::rl_attempted_completion_function = completer;
```
dangleptr: Is there no default auto-complete module in readline?  This is a very common feature.

dutor: Yes, the default one is to complete filenames.

dutor: If you mean whether readline has builtin support to all these functions here(Given a collection of keywords, readline does all completion works for you). I'm afraid the answer seems no. I had been following up the official documents.

dangleptr: 。。。。The default module is useless if not allow users to pass custom keywords.

dutor: Yes it is. Actually, you could infer from the `completer` function, that if we don't have to support sub-keywords completion, all we need is to retrieve the longest common prefix of all completion candidates.

Issue #437, opened 2019-05-26
does nGQL support comments , such as # or // or /* */, and line break, such as 


whitewum:
for a very very large query, it's common to add some comments, and breaks it into multiple lines.
Think about a 50 lines query.

steppenwolfyuetong:
For comment，
There are column level comment and table level comment.
There are three comment formats in mysql, e.g. `#, /* */, -- `

dutor:
As discussed offline, I will assign this feature request to @laura-ding to accomplish.

laura-ding:
Has supported,  pr is 457

whitewum:
great! like it!

PR #457, opened 2019-05-29
Added support to comments in nGQL 

laura-ding:
This PR partially implements issue #437


dutor:
Please adjust the commit message according to the title of this PR while merging in.

Issue #456, opened 2019-05-28
A compound sentences bug: "console> use spaceid; show tags; "


whitewum:
nebula>  use myspace_test; show tags;     //ERROR
Please choose a graph space with `USE spaceName' firstly 

nebula>  use myspace_test;                       // use space first;                                                                                                         Execution succeeded (Time spent: 158/925 us)                                                                                                                             nebula>  use myspace_test; show tags;      // then this compound sentence is ok                                                                                                                              ==============                                                                                                                                                           |       Name |

laura-ding:
I will fix it

laura-ding:
When multi sentences input, the code first to do prepare() , if the handler of prepare() dependence previous command result，prepare() will fail. so the prepare() cannot handle depends on the Previous command result. So move the handler to execute()

PR #475, opened 2019-05-30
fix issue #456

laura-ding:


dutor:
There are some jobs, that are performed in `prepare`, that must be performed before `execute`, you could investigate what `PipeExecutor::prepair` does.

laura-ding:
Close first, fix later

PR #479, opened 2019-05-30
fix build

laura-ding:


laura-ding:
Thanks for @steppenwolfyuetong

laura-ding:
jenkins go

Issue #172, opened 2019-02-26
Advanced data modification


sherman-the-tank:
More complicate data modification, such as

* INSERT NOT EXISTS
* UPDATE ... WHERE (conditional modification)
* DELETE ... WHERE (conditional deletion)

zhangguoqing:
@zlcook's #171 is implementing the **DELETE** statement, so I am implementing the **UPDATE** statement.


#### [Update a vertex](https://github.com/vesoft-inc/nebula/blob/master/docs/nGQL.md#update-a-vertex) 
The following statement updates a vertex

<span style="color:blue">**UPDATE VERTEX**</span>  <vertex\_id> <span style="color:blue">**SET**</span> \<update\_decl\> [<span style="color:blue">**WHERE**</span> <conditions\>] [<span style="color:blue">**YIELD**</span> <field\_list>]

<update\_decl> ::= <update\_form1> | <update\_form2>
<update\_form1> ::= <prop\_name> = <expression\> {,<prop\_name> = <expression\>}+
<update\_form2> ::= (<prop\_list>) = (<value\_list>) | (<prop\_list>) = <var\>

#### [Update an edge](https://github.com/vesoft-inc/nebula/blob/master/docs/nGQL.md#update-an-edge)
The following statement updates an edge

<span style="color:blue">**UPDATE EDGE**</span>  <vertex\_id> -> <vertex\_id> [@<weight\>] <span style="color:blue">**OF**</span> <edge\_type> <span style="color:blue">**SET**</span> <update\_decl> [<span style="color:blue">**WHERE**</span> <conditions\>] [<span style="color:blue">**YIELD**</span> <field\_list>]

dangleptr:
> More complicate data modification, such as
> 
> * INSERT NOT EXISTS
> * UPDATE ... WHERE (conditional modification)
> * DELETE ... WHERE (conditional deletion)

Currently,  we don't support where clause.   For update, we only support update specified edge or vertex

PR #481, opened 2019-05-30
Implement delete edges

zlcook:
Issue #172 
Delete Sentence: 
* `DELETE EDGE edge_type(<vertex_id>-><vertex_id>[@\<weight>] [, <vertex_id>-><vertex_id>[@\<weight>]]*)

### DIFF
--- a/src/graph/test/DeleteEdgeTest.cpp
+++ b/src/graph/test/DeleteEdgeTest.cpp
@@ -1,0 +1,290 @@
+/* Copyright (c) 2019 vesoft inc. All rights reserved.
+ *
+ * This source code is licensed under Apache 2.0 License,
+ * attached with Common Clause Condition 1.0, found in the LICENSES directory.
+ */
+
+#include "base/Base.h"
+#include "graph/test/TestEnv.h"
+#include "graph/test/TestBase.h"
+#include "meta/test/TestUtils.h"
+
+DECLARE_int32(load_data_interval_secs);
+
+namespace nebula {
+namespace graph {
+
+class DeleteEdgeTest : public TestBase {
+protected:
+    void SetUp() override {
+        TestBase::SetUp();
+        // ...
+    }
+
+    void TearDown() override {
+        // ...
+        TestBase::TearDown();
+    }
+
+    static void SetUpTestCase() {
+        client_ = gEnv->getClient();
+
+        ASSERT_NE(nullptr, client_);
+
+        ASSERT_TRUE(prepareSchema());
+    }
+
+    static void TearDownTestCase() {
+        ASSERT_TRUE(removeData());
+        client_.reset();
+    }
+
+protected:
+    static AssertionResult prepareSchema();
+
+    static AssertionResult removeData();
+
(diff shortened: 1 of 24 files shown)


whitewum:
should this pr rebase to master? @zlcook

src/graph/test/DeleteEdgeTest.cpp
```diff
@@ -0,0 +1,223 @@
...
+    {
+        cpp2::ExecutionResponse resp;
+        std::string cmd = "INSERT VERTEX person(name, age) VALUES 1001:(\"Zhangsan\", 22),"
+                          "1002:(\"Lisi\", 23), 1003:(\"Jack\", 18), 1004:(\"Rose\", 19)";
+        auto code = client_->execute(cmd, resp);
+        ASSERT_EQ(cpp2::ErrorCode::SUCCEEDED, code);
+    }
+    {
+        cpp2::ExecutionResponse resp;
+        std::string cmd = "INSERT EDGE friend(intimacy) VALUES 1001->1002@15:(90),"
```
whitewum: add two (different types or different ranks) edgess with the same src and the same dst. Then delete one . 

and try to double delete the same edge.

delete a not exist edge. 

check: both the in and out keys are not exists in storage ( not partially deleted)

zlcook: Ok

zlcook:
> I think current storage client can't guarantee that both in-key and out-key can be deleted. Since they are in different shard on different machine. Would you add a comment to address this?

You are right, I will add a comment.

> BTW. What about the feature to delete a vertex (with all the versions)?
 
* Basic idea:
graph layer need send twice requests to complete it.  
1. First request, delete the vertex and it's out-edges( they located in same partition), then construct `EdgeKeys` of in-edges (located in different partitions ) according to out-edges and return to graph layer.
2. Second request, graph layer delete the in-edges from the first request.

In order to ensure the in-edges can eventually be by removed, we can record the `vertex` needed to be deleted  in `meta server`.  If the `second request` failed, we can continue to delete in-edges by compaction.

whitewum:
> > I think current storage client can't guarantee that both in-key and out-key can be deleted. Since they are in different shard on different machine. Would you add a comment to address this?
> 
> You are right, I will add a comment.
> 
> > BTW. What about the feature to delete a vertex (with all the versions)?
> 
> * Basic idea:
>   graph layer need send twice requests to complete it.
> 
> 1. First request, delete the vertex and it's out-edges( they located in same partition), then construct `EdgeKeys` of in-edges (located in different partitions ) according to out-edges and return to graph layer.
> 2. Second request, graph layer delete the in-edges coming from the first request.
> 
> In order to ensure the in-edges can eventually be removed, we can record the `vertex` needed to be deleted in `meta server`. If the `second request` failed, we can continue to delete in-edges by compaction.

@dutor @dangleptr what do you think about delete a vertex with all the edges.  In GB, it allows dangling edges.

dangleptr:
Ready to review now?

PR #483, opened 2019-05-31
Refactor meta processors code structure

dangleptr:
There are too many processors in meta.  And more processors are on the way.  

Move related processors into some sub dir under processors.


dangleptr:
Jenkins go

PR #484, opened 2019-05-31
Some minor fixes

dutor:
One thing to remember: calling `std::move` on local objects(when `return`) and temporary objects(when passed as arguments) will make the copy elision optimization impossible.


steppenwolfyuetong:
By the way,  you are a great god in C++. 👍

PR #395, opened 2019-05-17
To add a web interface to indicate the status of the three daemons

steppenwolfyuetong:
#Issue #237 . To add a web interface to indicate the status of the three daemons

use as follow:

http://127.0.0.1:XXXX/status
http://127.0.0.1:XXXX
When viewing the status of a submodule, use
http://127.0.0.1:XXXX/status/XXXX/XXXX

Issue #406, opened 2019-05-21
To fix or improve the dependencies between different modules


dutor:
The only explicit dependencies we have to specify is the parts regarding the auto generated source files, i.e. *.yy, *.lex, *.thrift, etc. All others shall be resolved automatically by the building system.

Besides, it's the very linking phase of some executable or shared object that requires(depends on) other modules(libraries, objects), rather than the compiling phase of individual objects. 

Besides, generated header files of the thrift files are now missing in the OUTPUT list, I think it might be the root cause of how we reach the current sistuation.

laura-ding:
we should offer a place to save dependence packages.

sherman-the-tank:
Done

PR #488, opened 2019-05-31
Clean up dependency on proxygen libs

dangleptr:
I found the problem when i write a UT with no dependency of proxygen lib.  
It is introduced by #210 and #395  

Please pay more attention to your dependency.  @steppenwolfyuetong @darionyaphet

### DIFF
--- a/src/graph/CMakeLists.txt
+++ b/src/graph/CMakeLists.txt
@@ -2,7 +2,6 @@ add_library(
     graph_obj OBJECT
     GraphFlags.cpp
     GraphService.cpp
-    GraphHttpHandler.cpp
     ClientSession.cpp
     SessionManager.cpp
     ExecutionEngine.cpp
@@ -47,5 +46,9 @@ add_dependencies(
     stats_obj
 )
 
+add_library(
+    graph_http_obj OBJECT
+    GraphHttpHandler.cpp
+)
 
 add_subdirectory(test)
--- a/src/graph/CMakeLists.txt
+++ b/src/graph/CMakeLists.txt
@@ -39,7 +39,6 @@ add_dependencies(
     base_obj
     parser_obj
     graph_thrift_obj
-    ws_obj
     meta_thrift_obj
     meta_client
     schema_obj
--- a/src/meta/CMakeLists.txt
+++ b/src/meta/CMakeLists.txt
@@ -44,7 +44,6 @@ add_dependencies(
     common_thrift_obj
     kvstore_obj
     thread_obj
-    ws_obj
     stats_obj
 )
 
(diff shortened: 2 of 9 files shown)


dangleptr:
Jenkins go

src/storage/test/StorageHttpHandlerTest.cpp
```diff
@@ -43,32 +43,6 @@ class StorageHttpHandlerTestEnv : public ::testing::Environment {
...
-    auto mClient
-        = std::make_unique<meta::MetaClient>(threadPool, std::move(addrsRet.value()), true);
-    mClient->init();
-    std::string dataPath = folly::stringPrintf("%s/data", rootPath.path());
-    auto sc = TestUtils::mockStorageServer(mClient.get(),
-                                           dataPath.c_str(),
-                                           localIp,
-                                           localDataPort);
-
     {
```
steppenwolfyuetong: Can explain that the mockserver is not started, how can the latter test run?

src/graph/CMakeLists.txt
```diff
@@ -47,5 +46,9 @@ add_dependencies(
     stats_obj
 )
 
+add_library(
+    graph_http_obj OBJECT
+    GraphHttpHandler.cpp
+)
```
steppenwolfyuetong: Is there no dependency here?

dangleptr: Just depend on base_obj.  Good catch

src/storage/test/StorageHttpHandlerTest.cpp
dangleptr: We just need the http server.

src/graph/CMakeLists.txt
dutor: Here is adding a library, OBJECT library, no dependencies.

src/meta/CMakeLists.txt
```diff
@@ -49,6 +48,13 @@ add_dependencies(
     stats_obj
 )
 
+add_library(
+    meta_http_handler OBJECT
+    MetaHttpHandler.cpp
+)
+add_dependencies(
+    meta_http_handler
+    base_obj)
```
steppenwolfyuetong: right bracket format

dutor:
Executables and shared libraries depend on other libraries, while static or OBJECT libraries depend on other headers, not libraries.

Now `add_dependencies` brings us too many chaos, we have to address this recently.

Please refer to this issue #406 .

dutor:
Besides, by move `third-party` out from our repo, `add_dependencies` is no longer needed anymore.

dangleptr:
> Executables and shared libraries depend on other libraries, while static or OBJECT libraries depend on other headers, not libraries.
> 
> Now `add_dependencies` brings us too many chaos, we have to address this recently.
> 
> Please refer to this issue #406 .

Make sense.  OBJECT should just depend on the headers.  
"add_dependencies" only work for Executables

dangleptr:
> Besides, by moving `third-party` out from our repo, `add_dependencies` is no longer needed anymore.

Very good.

PR #490, opened 2019-06-03
Fix build

laura-ding:


laura-ding:
jenkins go

dangleptr:
Which one build failed?

laura-ding:
@dangleptr 
image

darionyaphet:
I think I've had similar problems.

Could you describe it in detail ?

dangleptr:
@laura-ding Got it.  In multi-process make,  the dependency order should be ensured.

laura-ding:
test

PR #493, opened 2019-06-04
Add some benchmarks

dangleptr:
1.  Benchmark for std::transform VS range for
2.  Benchmark for std::for_each  VS range for
3.  Benchmark for MultiVersion VS SingleVersion in rocksdb


dangleptr:
Jenkins go

src/common/base/test/RangeVsTransformBenchmark.cpp
```diff
@@ -0,0 +1,130 @@
...
+BENCHMARK(Test1_RangeTestStr) {
+    std::vector<int32_t> from;
+    std::vector<std::string> to;
+    BENCHMARK_SUSPEND {
+        from.resize(1000, 0);
+        to.resize(1000);
+    }
+    int32_t index = 0;
+    for (auto& i : from) {
+        to[index++] = std::to_string(i);
```
dutor: You are testing `std::to_string`, not `std::transform` vs. range for.

src/common/base/test/RangeVsTransformBenchmark.cpp
```diff
@@ -0,0 +1,130 @@
...
+
+-O3
+[heng.chen@hp-server nebula]$ src/common/base/test/_build/range_vs_transform__bm
+============================================================================
+/data/heng/workspace/nebula/src/common/base/test/RangeVsTransformBenchmark.cpprelative  time/iter  iters/s
+============================================================================
+Test1_RangeTestStr                                          84.19us   11.88K
+Test1_TransformStr                               100.01%    84.18us   11.88K
+----------------------------------------------------------------------------
+Test2_RangeTestInt                                         222.70ns    4.49M
```
dutor: Please expain why `-O3` is better than `-O2` for this case.

src/common/base/test/CMakeLists.txt
```diff
@@ -110,3 +110,11 @@ nebula_link_libraries(
     gtest_main
 )
 nebula_add_test(nebulakey_utils_test)
+
+add_compile_options(-O3)
```
dutor: Please use `target_compile_options` instead.

src/common/base/test/RangeVsTransformBenchmark.cpp
```diff
@@ -0,0 +1,130 @@
...
+int main(int argc, char** argv) {
+    folly::init(&argc, &argv, true);
+    folly::runBenchmarks();
+    return 0;
+}
+
+
+/*
+-O2
+[heng.chen@hp-server nebula]$ src/common/base/test/_build/range_vs_transform__bm
```
dutor: Better no to expose such kind of privacy.

Better post the CPU model you are testing on.

src/common/base/test/RangeVsTransformBenchmark.cpp
```diff
@@ -0,0 +1,130 @@
...
+============================================================================
+/data/heng/workspace/nebula/src/common/base/test/RangeVsTransformBenchmark.cpprelative  time/iter  iters/s
```
dutor: ditto. 

Besides, try to align the headers according to the testing results

src/common/base/test/RangeVsTransformBenchmark.cpp
dangleptr: int to std::string is a common case with std::transform.   We'd better to figure out the accurate perf under this case.

src/common/base/test/RangeVsTransformBenchmark.cpp
dangleptr: Let me try it.

dangleptr: Generate the asm for the snippet
```
int from[100000];
int to[100000];

int f() {
    int index = 0;
    for (auto& i : from) {
        to[index++] = i;
    }
    return 0;
}
```
With -O2
```
.LFB1898:
	.cfi_startproc
	xorl	%eax, %eax
	.p2align 4,,10
	.p2align 3
.L2:
	movl	from(%rax), %edx
	addq	$4, %rax
	movl	%edx, to-4(%rax)
	cmpq	$400000, %rax
	jne	.L2
	xorl	%eax, %eax
	ret
	.cfi_endproc
```
With -O3
```
_Z1fv:
.LFB1898:
	.cfi_startproc
	subq	$8, %rsp
	.cfi_def_cfa_offset 16
	movl	$400000, %edx
	movl	$from, %esi
	movl	$to, %edi
	call	memcpy
	xorl	%eax, %eax
	addq	$8, %rsp
	.cfi_def_cfa_offset 8
	ret
	.cfi_endproc
```

Notice that,  with O3,  there is no loop,  just call memcpy directly,  very smart!

src/common/base/test/RangeVsTransformBenchmark.cpp
dutor: Actually, I had digged into the implementations of follybenchmark, trying to make it able to use relative paths, but ended up with no solution.

src/common/base/test/RangeVsTransformBenchmark.cpp
dutor: Excellent.

src/common/base/test/RangeVsTransformBenchmark.cpp
dutor: Fair enough.

src/common/base/test/CMakeLists.txt
```diff
@@ -110,3 +110,11 @@ nebula_link_libraries(
     gtest_main
 )
 nebula_add_test(nebulakey_utils_test)
+
+add_executable(range_vs_transform__bm RangeVsTransformBenchmark.cpp)
+nebula_link_libraries(
+    range_vs_transform__bm
```
darionyaphet: one more underscore

src/kvstore/test/MultiVersionBenchmark.cpp
```diff
@@ -0,0 +1,111 @@
...
+        if (FLAGS_do_compact) {
+            rocksdb::CompactRangeOptions croptions;
+            db->CompactRange(croptions, nullptr, nullptr);
+        }
+    }
+    auto start = genKey(0, 0);
+    rocksdb::ReadOptions roptions;
+    rocksdb::Iterator* iter = db->NewIterator(roptions);
+    if (iter) {
+        iter->Seek(rocksdb::Slice(start));
```
darionyaphet: could  std::move here ?

src/kvstore/test/MultiVersionBenchmark.cpp
```diff
@@ -0,0 +1,111 @@
...
+            for (int v = 0; v < FLAGS_versions; v++) {
+                std::string key;
+                if (withVersion) {
+                    key = genKey(i, v);
+                } else {
+                    key = genKey(i, 0);
+                }
+                auto val = folly::stringPrintf("val_%d_%d", i, v);
+                db->Put(woptions,
+                        rocksdb::Slice(key.data(), key.size()),
```
darionyaphet: should std::move key and val  ?

src/common/base/test/CMakeLists.txt
```diff
@@ -110,3 +110,11 @@ nebula_link_libraries(
...
+    follybenchmark
+    boost_regex
```
darionyaphet: will use `boost_regex` ?

src/kvstore/test/MultiVersionBenchmark.cpp
dangleptr: For slice,  move equals copy

src/kvstore/test/MultiVersionBenchmark.cpp
dangleptr: ditto

src/common/base/test/CMakeLists.txt
dangleptr: because it is depended by folly benchmark

src/common/base/test/CMakeLists.txt
dangleptr: Good catch

dangleptr:
Jenkins go

src/kvstore/test/MultiVersionBenchmark.cpp
dutor: > For slice, move equals copy

FYI. This also appies to StringPiece.

PR #494, opened 2019-06-04
Implement compaction filter for storage

### LINKED ISSUE #286: Implement compaction filter for storage
* Data should be removed after related schema has been deleted.
 * TTL and max-version should be supported.

dangleptr:
Close #286


dangleptr:
Jenkins go

src/storage/CompactionFilter.h
```diff
@@ -8,40 +8,100 @@
...
     }
 
-    const char* Name() const override {
-        return "NebulaCompactionFilter";
+    bool indexValid(const folly::StringPiece&) const {
+        return true;
     }
+
+private:
+    mutable std::string lastKeyWithNoVerison_;
```
dutor: Is the filter called from a single thread? If not, `lastKeyWithNoVersion` shoud be protected.

src/storage/CompactionFilter.h
```diff
@@ -8,40 +8,100 @@
...
+    }
+
+    bool ttlValid(GraphSpaceID, const folly::StringPiece&) const {
+        return true;
+    }
+
+    bool filterVersions(const folly::StringPiece& key) const {
+        folly::StringPiece keyWithNoVersion = NebulaKeyUtils::keyWithNoVersion(key);
+        if (keyWithNoVersion == lastKeyWithNoVerison_) {
+            // TODO(heng): we could support max-versions configuration in schema if needed.
```
zlcook: `max-versions` configuration means that we can save multiple keys with different versions, Right ?

src/storage/CompactionFilter.h
dangleptr: Yes,  sub-compaction is closed by default.  Of course, we could check the config again in code.

src/storage/CompactionFilter.h
dangleptr: Yes,  in some cases, we need to track the history versions of a vertex or edge.  The feature will be implemented in the future.   We call it time-series graph db.  HAHA

src/kvstore/CompactionFilter.h
```diff
@@ -0,0 +1,98 @@
...
+    }
+
+    std::unique_ptr<rocksdb::CompactionFilter>
+    CreateCompactionFilter(const rocksdb::CompactionFilter::Context& context) override {
+        auto now = time::TimeUtils::nowInSeconds();
+        if (context.is_full_compaction) {
+            LOG(INFO) << "Do full compaction!";
+            lastRunCustomFilterTimeSec_ = now;
+            return std::unique_ptr<rocksdb::CompactionFilter>(
+                            new KVCompactionFilter(spaceId_, createKVFilter()));
```
critical27: Do we have to consider default full compaction without filter?

dangleptr: One important purpose for full compaction is to remove useless data.  So we should use custom filter.

src/storage/CompactionFilter.h
```diff
@@ -8,40 +8,100 @@
...
+        } else {
+            VLOG(3) << "Skip the system key inside, key " << key;
+        }
+        return false;
+    }
+
+    bool schemaValid(GraphSpaceID spaceId, const folly::StringPiece& key) const {
+        if (NebulaKeyUtils::isVertex(key)) {
+            auto tagId = NebulaKeyUtils::getTagId(key);
+            if (schemaMan_->getNewestTagSchemaVer(spaceId, tagId) < 0) {
```
steppenwolfyuetong: When there are multiple versions of the data, do we only save the latest data? Are the rest deleted in the compaction?

src/kvstore/CompactionFilter.h
```diff
@@ -0,0 +1,98 @@
...
+        : spaceId_(spaceId)
+        , kvFilter_(std::move(kvFilter)) {
+    }
+
+    bool Filter(int,
+                const rocksdb::Slice& key,
+                const rocksdb::Slice& val,
+                std::string*,
+                bool*) const override {
+        return kvFilter_->filter(spaceId_,
```
steppenwolfyuetong: Could remove some of the parameters that are not used?

dangleptr: It is rocksdb's interface.

src/storage/CompactionFilter.h
dangleptr: yes

src/storage/CompactionFilter.h
```diff
@@ -8,40 +8,100 @@
...
+                VLOG(3) << "Space " << spaceId << ", EdgeType " << edgeType << " invalid";
+                return false;
+            }
+        }
+        return true;
+    }
+
+    bool ttlValid(GraphSpaceID, const folly::StringPiece&) const {
+        return true;
+    }
```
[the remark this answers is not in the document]
dangleptr: Good point.

src/storage/CompactionFilter.h
whitewum: awesome

dangleptr:
> Everything looks fine to me. Maybe we can try to use FilterV2 to reduce some IO potentially later.

Good point.  Let's check the filterV2 usage.

PR #495, opened 2019-06-04
Fix http2 port name

laura-ding:


etc/nebula-graphd.conf.default
```diff
@@ -41,4 +41,4 @@
 # HTTP service port
 --ws_http_port=13000
 # HTTP2 service port
---ws_h2_portt=13002
```
dutor: Sorry about this mistake.

laura-ding:
408 pr  include this fix

PR #502, opened 2019-06-10
User privilege manager in graph layer.

boshengchen:
This one depends on PR #400 and PR #480 .
Please hold on this one .


src/graph/UserAccessControl.cpp
```diff
@@ -0,0 +1,194 @@
...
+#define    ACL_CREATESPACE        (1<<22)
+#define    ACL_DROPSPACE          (1<<23)
+#define    ACL_DESCRIBESPACE      (1<<24)
+#define    ACL_YIELD              (1<<25)
+#define    ACL_CREATEUSER         (1<<26)
+#define    ACL_DROPUSER           (1<<27)
+#define    ACL_ALTERUSER          (1<<28)
+#define    ACL_GRANT              (1<<29)
+#define    ACL_REVOKE             (1<<30)
+#define    ACL_CHANGEPASSWORD     (1<<31)
```
dutor: I would recommend to use `enum`s.

Besides, you have used  up all of the 32 bits, while there are definitely new statements in future. So, You should use a 64bits unsigned integer to represent ACLs instead. BTW. `1` ought to be `1UL`.

src/graph/UserAccessControl.cpp
```diff
@@ -0,0 +1,194 @@
...
+
+
+int32_t UserAccessControl::getACLMode(RoleType type) {
+    //  Operation and permission define reference UserAccessControl.h
+    int32_t acl = 0;
+    switch (type) {
+        case RoleType::GOD:
+            //  God has all privileges
+            acl |= (ACL_ADDHOSTS
+                   |ACL_REMOVEHOSTS
```
dutor: There ought to be spaces before and after an operator.

src/meta/client/MetaClient.cpp
```diff
@@ -643,6 +715,126 @@ MetaClient::removeRange(std::string segment, std::string start, std::string end)
...
+    req.set_encoded_pwd(std::move(password));
+    req.set_missing_ok(missingOk);
+    return getResponse(std::move(req), [] (auto client, auto request) {
+        return client->future_createUser(request);
+    }, [] (cpp2::ExecResp&& resp) -> UserID {
+        return resp.get_id().get_user_id();
+    }, true);
+}
+
+folly::Future<StatusOr<bool>>
```
dutor: `StatusOr<bool>` is **identical** to `Status`

src/meta/client/MetaClient.cpp
```diff
@@ -643,6 +715,126 @@ MetaClient::removeRange(std::string segment, std::string start, std::string end)
...
+        return std::move(resp).get_user_item();
+    });
+}
+
+folly::Future<StatusOr<std::unordered_map<UserID, cpp2::UserItem>>>
+MetaClient::listUsers() {
+    cpp2::ListUsersReq req;
+    return getResponse(std::move(req), [] (auto client, auto request) {
+        return client->future_listUsers(request);
+    }, [] (cpp2::ListUsersResp&& resp) -> decltype(auto) {
```
dutor: Please do a research on the difference between `auto` versus `decltype(auto)`, don't use the latter one unless it's necessary.

src/graph/UserAccessControl.cpp
boshengchen: `enum` is better. That makes sense 👍 
32 bits is really not enough if a new operation is added，I will change it to 64 bits.

bright-starry-sky:
Close this one . Will track this with #1873 .

Issue #473, opened 2019-05-29
Implement config manager 


dangleptr:
Just like SchemaManager/PartManager,  we need a ConfigManager to manage all configs for current cluster. 

Please upload your design document here.  @critical27

sherman-the-tank:
Here are some requirements

1. This meant to replace any gflag that might be modified during the runtime. Going forward, the rule is using gflags for any start-up configuration and those configurations should never be changed during the runtime; using CM for any configuration that could be changed during the runtime

1. All configuration in CM should be persisted on the Meta Server and can be restored during the daemon start-up time

1. CM should be as easy as gflags regarding the reading interface

1. The configuration in CM should always affect all hosts in the cluster. *Whether to support only modifying the configuration on one host is a open question

critical27:
Let us look into config key pattern: 
table_prefix + global_flag/spaceId + len(module) + module + len(name) + name + config_type.

With global_flag or spaceId, we can save configuration of different spaces, and "module" is something like "storage" or "graph" to distinguish configurations of same name.

As to config value, I think config_type and value will be enough.

critical27:
Besides, @dangleptr suggests that maybe we can support a level of hosts to support set configuration for one host later. The priority will be host > space > global.

Also, maybe we need a local_mode in config manager to support use configurations of local file for debugging.

dangleptr:
> Besides, @dangleptr suggests that maybe we can support a level of hosts to support set configuration for one host later. The priority will be host > space > global.
> 
> Also, maybe we need a local_mode in config manager to support use configurations of local file for debugging.

If you have host level configurations on meta server,  local_mode is needless.

dangleptr:
> Let us look into config key pattern:
> table_prefix + global_flag/spaceId + len(module) + module + len(name) + name + config_type.
> 
> With global_flag or spaceId, we can save configuration of different spaces, and "module" is something like "storage" or "graph" to distinguish configurations of same name.
> 
> As to config value, I think config_type and value will be enough.

Now, i am thinking about how about saving the default value on meta service

sherman-the-tank:
Since the ConfManager will replace most of the gflags, which are used on the critical path, we need to make sure the READ operation on the ConfManager is fast enough

critical27:
> Since the ConfManager will replace most of the gflags, which are used on the critical path, we need to make sure the READ operation on the ConfManager is fast enough

[...]

PR #504, opened 2019-06-10
Add config manager to store configurations that could be changed during the runtime

critical27:
GflagsManager is used to maintain all gflags in our program. It works like this:
1. When we start-up, we will parse all gflags passed in, and try to register on meta server.
2. We can list all gflags, get some specified gflag, or update some **mutable** glfags in console.
3. GlfagsManager will poll gflags from meta server periodically, and update those gflags that value is updated. So for mutable gflags, we can change them dynamically. (e.g. `load_data_interval_secs`)

### How to use config manager
There are configs for graph/meta/storage in our system. We can set config of one specified module, or we can set the config of same name to same value in different module. See example belows.

### **in console**
* We can update those **mutable** variables like this:

Set and get a specified module's config.
```
UPDATE VARIABLES graph:load_data_interval_secs=10
GET VARIABLES graph:load_data_interval_secs
```
it will show like this:
```
==============================================================
| module |                    name |  type |    mode | value |
==============================================================
|  GRAPH | load_data_interval_secs | INT64 | MUTABLE |    10 |
--------------------------------------------------------------
```

Set and get config of same name to all module
```
UPDATE VARIABLES graph:load_data_interval_secs=10
GET VARIABLES graph:load_data_interval_secs
```
it will show like this:
```
===============================================================
|  module |                    name |  type |    mode | value |
===============================================================
|   GRAPH | load_data_interval_secs | INT64 | MUTABLE |    10 |
---------------------------------------------------------------
|    META | load_data_interval_secs | INT64 | MUTABLE |    10 |
---------------------------------------------------------------
| STORAGE | load_data_interval_secs | INT64 | MUTABLE |    10 |
---------------------------------------------------------------
```

* For **immutable** variables, we can only get. 

Or we can list configs in all module or single module
```
SHOW VARIABLES
SHOW VARIABLES graph
```

#### **in code**
For those immutable gflags (used to initialize or never changes), GflagsManager will not effect them, and the code use them just as the way before (FLAGS_xxx);

For those mutable gflags, GflagsManager will update its value once get new value from meta server.
But if we want to use mutable gflags, we may have conflicts, in [gflags](https://github.com/gflags/gflags/blob/master/src/gflags.h.in) it says:
`These programmatic ways to access flags are thread-safe, but direct access is only thread-compatible.`
So as for mutalbe gflags, we need to get its value by calling GetCommandLineOption, see the example below:
```
     std::string flag;
     gflags::GetCommandLineOption("load_config_interval_secs", &flag);
     size_t delayMS = stoi(flag) * 1000 + folly::Random::rand32(900);
```
GetCommandLineOption only gives you a value in string, so you may need to transform to the type you want.

### Some issues for now
1. For those mutable gflgas, we need to specify its name in hard code, any better ideas?
2. When console print result, type of different rows must be the same, so I transform all values to string for now. (when we list all variables, it contains value of different type)
3. For now all uint32 and int32 glags are converted to int64 for now. (The reason is we can only parse int64, double, bool, string in console)


critical27:
This pr is still in process, refer to #473.

Done: 
Config Manager interface.

Todo:
1. Config key to be decided according to @sherman-the-tank's comments in #473.
2. Reload config key when daemon start-up.
3. Support get/set in parser.

critical27:
Jenkins go

src/interface/meta.thrift
```diff
@@ -416,6 +416,46 @@ struct CheckPasswordReq {
...
+enum ConfigType {
+    INT64   = 0x00,
+    DOUBLE  = 0x01,
+    BOOL    = 0x02,
+    STRING  = 0x03,
+} (cpp.enum_strict)
+
+struct ConfigItem {
+    1: string               space,
+    2: string               module,
```
dangleptr: Use enum for module

src/interface/meta.thrift
```diff
@@ -416,6 +416,46 @@ struct CheckPasswordReq {
...
 
+enum ConfigType {
+    INT64   = 0x00,
+    DOUBLE  = 0x01,
+    BOOL    = 0x02,
+    STRING  = 0x03,
+} (cpp.enum_strict)
+
+struct ConfigItem {
+    1: string               space,
```
dangleptr: Why not spaceId ?

src/meta/ConfigManager.h
```diff
@@ -0,0 +1,200 @@
...
+class ConfigManager {
+    FRIEND_TEST(ConfigManTest, MetaConfigManTest);
+    FRIEND_TEST(ConfigManTest, MemConfigManagerTest);
+
+public:
+    static std::unique_ptr<ConfigManager> instance(MetaClient* client = nullptr);
+
+    virtual ~ConfigManager();
+
+    virtual folly::Future<StatusOr<bool>> setConfig(folly::StringPiece space,
```
dangleptr: Why use Future?

src/meta/ConfigManager.h
```diff
@@ -0,0 +1,200 @@
...
+                                                    folly::StringPiece module,
+                                                    folly::StringPiece name,
+                                                    const VariantType &value,
+                                                    bool checkRegistered = true);
+
+    virtual folly::Future<StatusOr<ConfigItem>> getConfigAsInt64(folly::StringPiece space,
```
dangleptr: Ditto

src/meta/ConfigManager.h
```diff
@@ -0,0 +1,200 @@
...
+
+
+    virtual folly::Future<StatusOr<bool>> registerConfig(const std::string& space,
+                                                         const std::string& module,
+                                                         const std::string& name,
+                                                         const VariantTypeEnum& type,
+                                                         const VariantType& defaultValue);
+
+protected:
+    virtual folly::Future<StatusOr<bool>> setConfigAsInt64(folly::StringPiece space,
```
dangleptr: Why make the methods to be protected?

src/meta/ConfigManager.h
```diff
@@ -0,0 +1,200 @@
...
+                                                          bool value) = 0;
+
+    virtual folly::Future<StatusOr<bool>> setConfigAsString(folly::StringPiece space,
+                                                            folly::StringPiece module,
+                                                            folly::StringPiece name,
+                                                            folly::StringPiece value) = 0;
+
+    explicit ConfigManager(MetaClient *metaClient);
+
+    MetaClient *metaClient_{nullptr};
```
dangleptr: Do we really need it in ConfigManager interface?

src/meta/client/MetaClient.h
```diff
@@ -260,6 +313,10 @@ class MetaClient {
     MetaChangedListener*  listener_{nullptr};
     bool                  sendHeartBeat_ = false;
     std::atomic_bool      ready_{false};
+    MetaConfigMap         metaConfigMap_;
+    folly::RWSpinLock     configCacheLock_;
+    thread::GenericWorker loadCfgThread_;
```
dangleptr: Could we reuse current background thread?

src/interface/meta.thrift
critical27: Sure

src/interface/meta.thrift
critical27: At first I encrypt with spaceId, but later I found I need to get space name when I get or list config. There is only a map of <name, spaceId> in MetaClient, no need to add another one? Or do you have any better ideas?

src/meta/ConfigManager.h
critical27: We only use the setConfig method in ConfigManger for console now,  which will call the Int64 and other method.

src/graph/ConfigExecutor.cpp
```diff
@@ -0,0 +1,318 @@
...
+    std::string module;
+    if (configItem_ != nullptr) {
+        if (configItem_->getSpace() != nullptr) {
+            space = *configItem_->getSpace();
+        }
+        if (configItem_->getModule() != nullptr) {
+            module = *configItem_->getModule();
+        }
+    }
+
```
wadeliuyi: maybe need return directly when space and module is empty string

src/meta/client/MetaClient.h
critical27: Of course, but I think maybe we need to different load interval?

src/graph/ConfigExecutor.cpp
```diff
@@ -0,0 +1,318 @@
...
+            onError_(std::move(resp).status());
+            return;
+        }
+
+        std::vector<std::string> header;
+        header.push_back("space");
+        header.push_back("module");
+        header.push_back("name");
+        header.push_back("type");
+        header.push_back("value");
```
wadeliuyi: use initialize list maybe more performance like std::vector<std::string> header({"a", "b", "c"})

src/graph/ConfigExecutor.cpp
critical27: When space and module is empty, it will show all configs just like "show variables" in mysql.

src/graph/ConfigExecutor.cpp
critical27: Yeah, you're right, roger that!

src/graph/ConfigExecutor.cpp
wadeliuyi: so can there generate a default response directly when parameters are empty?

src/meta/ConfigManager.h
critical27: These interface is all for console which goes to meta server, I will add some interface for code and get/set from cache.

src/graph/ConfigExecutor.cpp
```diff
@@ -0,0 +1,318 @@
...
+        resp_ = std::make_unique<cpp2::ExecutionResponse>();
+        resp_->set_column_names(std::move(header));
+
+        auto item = std::move(resp).value();
```
wadeliuyi: what is the use about "std::move(resp)" here?

boshengchen:
I have a question, how to avoid reset the special configurations at run time?  Such as configuration `data path`  or `security manager` , etc.  These parameters cannot be reset during cluster running.

src/meta/ConfigManager.cpp
```diff
@@ -0,0 +1,279 @@
...
+    CHECK_NOTNULL(metaClient_);
+}
+
+ConfigManager::~ConfigManager() {
+    if (nullptr != metaClient_) {
+        metaClient_ = nullptr;
+    }
+}
+
+std::unique_ptr<ConfigManager> ConfigManager::instance(MetaClient* client) {
```
wadeliuyi: named instance like a singleton module.

critical27:
> I have a question, how to avoid reset the special configurations at run time? Such as configuration `data path` or `security manager` , etc. These parameters cannot be reset during cluster running.

Yeah, just as pr's title, Config Manager is only for store configurations that could be changed during the runtime. For those configurations can not be changed, we still use gflags.

src/meta/client/MetaClient.cpp
```diff
@@ -957,5 +966,211 @@ folly::Future<StatusOr<bool>> MetaClient::heartbeat() {
...
+StatusOr<std::vector<ConfigItem>>
+MetaClient::listConfigsFromCache(const std::string& space, const std::string& module) {
+    if (!configReady_) {
+        return Status::Error("Config cache not ready!");
+    }
+    std::vector<ConfigItem> ret;
+    {
+        folly::RWSpinLock::ReadHolder holder(configCacheLock_);
+        if (!space.empty()) {
+            for (const auto& entry : metaConfigMap_[space]) {
```
wadeliuyi: why can use operator[] here?

src/graph/ConfigExecutor.cpp
critical27: For now, there is no default response. Only if no configs was set before indeed, it get empty result.

src/meta/client/MetaClient.cpp
critical27: If we specify a space, only return configuration of this space. Whether read configs from cache or kvstore is still to be decided.

src/meta/processors/configMan/RegConfigProcessor.cpp
```diff
@@ -0,0 +1,42 @@
...
+
+#include "meta/processors/configMan/RegConfigProcessor.h"
+
+namespace nebula {
+namespace meta {
+
+void RegConfigProcessor::process(const cpp2::RegConfigReq& req) {
+    std::vector<kvstore::KV> data;
+    {
+        folly::SharedMutex::ReadHolder rHolder(LockUtils::configLock());
```
dangleptr: You should use write lock when registering gflags

src/meta/ClientBasedGflagsManager.cpp
```diff
@@ -0,0 +1,142 @@
...
+}
+
+template<typename ValueType>
+folly::Future<StatusOr<bool>> ClientBasedGflagsManager::set(const cpp2::ConfigModule& module,
+                                                            const std::string& name,
+                                                            const cpp2::ConfigType& type,
+                                                            const ValueType& value) {
+    std::string valueStr;
+    valueStr.append(reinterpret_cast<const char*>(&value), sizeof(value));
+    return metaClient_->setConfig(module, name, type, valueStr);
```
dangleptr: std::move

src/meta/ClientBasedGflagsManager.cpp
```diff
@@ -0,0 +1,142 @@
...
+}
+
+template<>
```
dangleptr: Do we need the template keyword?

src/meta/ClientBasedGflagsManager.cpp
```diff
@@ -0,0 +1,142 @@
...
+    return metaClient_->listConfigs(module);
+}
+
+folly::Future<StatusOr<bool>>
+ClientBasedGflagsManager::registerConfig(const cpp2::ConfigModule& module, const std::string& name,
+                                         const cpp2::ConfigType& type, const cpp2::ConfigMode& mode,
+                                         const std::string& value) {
+    auto item = toThriftConfigItem(module, name, type, mode, value);
+    std::vector<cpp2::ConfigItem> items;
+    items.emplace_back(item);
```
dangleptr: move

src/meta/ClientBasedGflagsManager.cpp
```diff
@@ -0,0 +1,142 @@
...
+}
+
+void ClientBasedGflagsManager::regCfgThreadFunc() {
+    auto status = metaClient_->regConfig(gflagsDeclared_).get();
+    if (status.ok()) {
+        LOG(INFO) << "Register flags ok " << gflagsDeclared_.size();
+        return;
+    }
+
+    std::string name = "load_config_interval_secs";
```
dangleptr: Why the snip code about "load_config_interval_secs" flag

src/meta/GflagsManager.h
```diff
@@ -0,0 +1,127 @@
...
+    void updateConfigCache(const std::vector<cpp2::ConfigItem>& items);
+
+    void updateGflagsValue(const ConfigItem& item);
+
+    template<typename ValueType>
+    std::string gflagsValueToThriftValue(const gflags::CommandLineFlagInfo& flag);
+
+    cpp2::ConfigModule                  module_{cpp2::ConfigModule::UNKNOWN};
+    std::vector<cpp2::ConfigItem>       gflagsDeclared_;
+    thread::GenericWorker               bgThread_;
```
dangleptr: Do we need the bgThread in base class?

src/meta/ClientBasedGflagsManager.cpp
critical27: Good catch!

src/meta/ClientBasedGflagsManager.cpp
critical27: It's template specialization for std::string

src/meta/ClientBasedGflagsManager.cpp
critical27: Good catch!

src/meta/ClientBasedGflagsManager.cpp
critical27: I will modify the init process.

src/meta/GflagsManager.h
critical27: Yeah, I will move it to ClientBasedGflagsManager

critical27:
Jenkins go

src/meta/ClientBasedGflagsManager.cpp
```diff
@@ -0,0 +1,95 @@
...
+namespace nebula {
+namespace meta {
+
+ClientBasedGflagsManager::ClientBasedGflagsManager(MetaClient *client) {
+    metaClient_ = client;
+    CHECK_NOTNULL(metaClient_);
+}
+
+ClientBasedGflagsManager::~ClientBasedGflagsManager() {
+    metaClient_ = nullptr;
```
darionyaphet: I'm not sure using reset will better ?

src/meta/KVBasedGflagsManager.cpp
```diff
@@ -0,0 +1,141 @@
...
+    UNUSED(module); UNUSED(name);
+    LOG(FATAL) << "Unimplement!";
+    return Status::OK();
+}
+
+folly::Future<StatusOr<std::vector<cpp2::ConfigItem>>>
+KVBasedGflagsManager::listConfigs(const cpp2::ConfigModule& module) {
+    UNUSED(module);
+    LOG(FATAL) << "Unimplement!";
+    return Status::OK();
```
darionyaphet: Although LOG(FATAL) will break down the programming, it's seems using `Status::NotSupported()` to replace `Status::OK()` would be reasonable. just personal understanding. :)

src/meta/ClientBasedGflagsManager.cpp
critical27: It's a raw pointer, not a smart one.

src/meta/KVBasedGflagsManager.cpp
critical27: Sure, sounds reasonable

src/graph/ConfigExecutor.cpp
```diff
@@ -0,0 +1,292 @@
...
+        }
+        if (configItem_->getName() != nullptr) {
+            name = *configItem_->getName();
+        }
+        if (configItem_->getValue() != nullptr) {
+            value = configItem_->getValue()->eval();
+        }
+    }
+
+    if (module == meta::cpp2::ConfigModule::UNKNOWN) {
```
wadeliuyi: move this check after line  99.

src/meta/client/MetaClient.cpp
```diff
@@ -995,5 +1003,181 @@ folly::Future<StatusOr<int64_t>> MetaClient::balance() {
...
+        for (auto& item : items) {
+            std::pair<cpp2::ConfigModule, std::string> key = {item.module_, item.name_};
+            metaConfigMap.emplace(std::move(key), std::move(item));
+        }
+        {
+            // For any configurations that is in meta, update in cache to replace previous value
+            folly::RWSpinLock::WriteHolder holder(configCacheLock_);
+            for (const auto& entry : metaConfigMap) {
+                auto& key = entry.first;
+                if (metaConfigMap_.find(key) == metaConfigMap_.end() ||
```
wadeliuyi: the if will traversing 3 times the map if key is exist in metaConfigMap_.

src/meta/processors/configMan/ListConfigsProcessor.cpp
```diff
@@ -0,0 +1,41 @@
...
+
+    std::vector<cpp2::ConfigItem> items;
+    while (iter->valid()) {
+        auto key = iter->key();
+        auto value = iter->val();
+        auto item = MetaServiceUtils::parseConfigValue(value);
+        auto configName = MetaServiceUtils::parseConfigKey(key);
+        item.set_module(configName.first);
+        item.set_name(configName.second);
+        items.emplace_back(item);
```
wadeliuyi: move(item)

src/meta/processors/configMan/GetConfigProcessor.cpp
```diff
@@ -0,0 +1,53 @@
...
+    std::string configKey = MetaServiceUtils::configKey(module, name);
+    auto ret = doGet(configKey);
+    if (!ret.ok()) {
+        return;
+    }
+
+    cpp2::ConfigItem item = MetaServiceUtils::parseConfigValue(ret.value());
+    item.set_module(module);
+    item.set_name(name);
+    items.emplace_back(std::move(item));
```
wadeliuyi: move(item)

src/meta/processors/configMan/RegConfigProcessor.cpp
```diff
@@ -0,0 +1,42 @@
...
+            auto mode = item.get_mode();
+            auto value = item.get_value();
+
+            std::string configKey = MetaServiceUtils::configKey(module, name);
+            // ignore config which has been registered before
+            if (doGet(configKey).ok()) {
+                continue;
+            }
+            std::string configValue = MetaServiceUtils::configValue(type, mode, value);
+            data.emplace_back(configKey, configValue);
```
wadeliuyi: move key and value

src/meta/processors/configMan/SetConfigProcessor.cpp
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    cpp2::ConfigItem item = MetaServiceUtils::parseConfigValue(ret.value());
+    if (item.get_mode() == cpp2::ConfigMode::IMMUTABLE) {
+        resp_.set_code(cpp2::ErrorCode::E_CONFIG_IMMUTABLE);
+        onFinished();
+        return;
+    }
+    std::string configValue = MetaServiceUtils::configValue(type, mode, value);
+    data.emplace_back(configKey, configValue);
```
wadeliuyi: move

src/meta/processors/configMan/SetConfigProcessor.cpp
```diff
@@ -0,0 +1,59 @@
...
+}
+
+void SetConfigProcessor::setOneConfig(const cpp2::ConfigModule& module, const std::string& name,
+                                      const cpp2::ConfigType& type, const cpp2::ConfigMode& mode,
+                                      const std::string& value, std::vector<kvstore::KV>& data) {
+    std::string configKey = MetaServiceUtils::configKey(module, name);
+    auto ret = doGet(std::move(configKey));
+    if (!ret.ok()) {
+        resp_.set_code(cpp2::ErrorCode::E_NOT_FOUND);
+        onFinished();
```
wadeliuyi: onFinished will delete this, so here can not call onFinished, because caller will call setOneConfig many times.

src/graph/ExecutionContext.cpp
```diff
@@ -15,6 +15,10 @@ ExecutionContext::~ExecutionContext() {
         sm_ = nullptr;
     }
 
+    if (nullptr != gflagsManager_) {
+        gflagsManager_ = nullptr;
```
wadeliuyi: maybe here can set nullptr directly.

src/meta/KVBasedGflagsManager.cpp
```diff
@@ -0,0 +1,141 @@
...
+        LOG(ERROR) << "Should not reach here";
+    }
+}
+
+Status KVBasedGflagsManager::registerGflags() {
+    // based on kv, so we will retry until register config successfully
+    bool succeeded = false;
+    while (!succeeded) {
+        std::vector<kvstore::KV> data;
+        folly::SharedMutex::WriteHolder wHolder(LockUtils::configLock());
```
wadeliuyi: why need write lock?

src/meta/KVBasedGflagsManager.cpp
```diff
@@ -0,0 +1,141 @@
...
+                        auto addr = value(std::move(leader));
+                        // HostAddr(0, 0) means no leader exists
+                        if (addr != HostAddr(0, 0)) {
+                            succeeded = true;
+                        }
+                    }
+                }
+                baton.post();
+            });
+            baton.wait();
```
wadeliuyi: it is dangerous to call wait after current thread hold a write lock.

src/meta/processors/configMan/ListConfigsProcessor.cpp
critical27: good catch!

src/meta/client/MetaClient.cpp
critical27: fixed

src/meta/processors/configMan/SetConfigProcessor.cpp
critical27: I think almost every handler will call onFinished, if a new request come, it will create a new instance to handle it.

src/meta/processors/configMan/RegConfigProcessor.cpp
critical27: good catch!

src/meta/KVBasedGflagsManager.cpp
critical27: Maybe you need to refer to PR #622.

src/meta/processors/configMan/SetConfigProcessor.cpp
wadeliuyi: SetConfigProcessor::process will call 3 times setOneConfig, if the first one call onFinished and delete this, the left two call will crash.

src/graph/ExecutionContext.cpp
critical27: I just keep the same style as above - -.

src/meta/KVBasedGflagsManager.cpp
critical27: It seems we want make sure that we register succeed, and it it called only once when we init.

src/meta/KVBasedGflagsManager.cpp
wadeliuyi: like you say this function just call once from init in main, and this will block the main thread, then the whole process just one thread, so no need lock?

src/meta/processors/configMan/SetConfigProcessor.cpp
critical27: I see, good catch.

src/meta/KVBasedGflagsManager.cpp
critical27: Yeah, just in case.

whitewum:
Great Job

PR #510, opened 2019-06-11
Use FLAGS_local_ip to assign local ip in StorageDaemon

CPWstatic:


CPWstatic:
jenkins go

src/daemons/StorageDaemon.cpp
```diff
@@ -22,6 +22,7 @@ DEFINE_int32(port, 44500, "Storage daemon listening port");
 DEFINE_bool(reuse_port, true, "Whether to turn on the SO_REUSEPORT option");
 DEFINE_string(data_path, "", "Root data path, multi paths should be split by comma."
                              "For rocksdb engine, one path one instance.");
+DEFINE_string(local_ip, "", "Local ip speicified for NetworkUtils::getLocalIP");
```
dutor: As a normal user, I cannot figure out what this config is exactly used for. 

What about _IP address which is used to identify this server, combined with the listen port_

Issue #515, opened 2019-06-12
Balancer


dangleptr:
* Balance will generate balance plan according to current active hosts and parts allocation
 * For the plan, we hope after moving the least parts , it will reach a reasonable state.
 * Only one balance plan could be invoked at the same time.
 * Each balance plan has one id,  and we could show the status by "balance id" command and after FO, we could go on the balance plan by type "balance id"
 * Each balance plan contains many balance tasks,  the task represents the minimum movement unit.
 * We save the whole balancePlan state in kvstore to do failover.
 * Each balance task contains serval steps. And it should be executed step by step.  
 * One task failed will result in the whole balance plan failed.  
 * Currently, we hope tasks for the same part could be invoked serially

This is subtask of #182

sherman-the-tank:
I wonder if we want to introduce the **RESOURCE GROUP**/**HOST GROUP** concept. We can talk about it offline

dangleptr:
let's close it now.  The balancer on meta has been finished.

PR #516, opened 2019-06-12
Balance skeleton on meta. 

dangleptr:
Subtask of #515 

The skeleton for balance.   AdminClient have been implemented in #553


dangleptr:
Jenkins go

whitewum:
do u mean re-balance?

dangleptr:
Jenkins go

dangleptr:
> do u mean re-balance?

Yes.

dangleptr:
Jenkins go

src/interface/meta.thrift
```diff
@@ -311,6 +312,19 @@ struct HBReq {
...
+    1: optional common.GraphSpaceID space_id,
+    // Specify the balance id to check the status of the related balance plan
+    2: optional i64 id,
+}
+
+struct BalanceResp {
+    1: ErrorCode        code,
+    2: i64              id,
+    // Valid if code equals E_LEADER_CHANGED.
+    3: common.HostAddr  leader,
```
spacewalkman: If it's valid conditionally, why not make it `optional`

src/meta/processors/admin/BalancePlan.h
```diff
@@ -0,0 +1,73 @@
...
+class BalancePlan {
+    friend class Balancer;
+public:
+    enum class Status : uint8_t {
+        START              = 0x01,
+        IN_PROGRESS        = 0x02,
+        FINISHED           = 0x03,
+    };
+
+    explicit BalancePlan(BalanceID id, kvstore::KVStore* kv)
```
spacewalkman: no need to specify `explicit`

src/meta/processors/admin/BalanceTask.cpp
```diff
@@ -0,0 +1,224 @@
...
+    }
+    switch (status_) {
+        case Status::START: {
+            LOG(INFO) << taskIdStr_ << "Start to move part!";
+            status_ = Status::CHANGE_LEADER;
+            ret_ = Result::IN_PROGRESS;
+            startTimeMs_ = time::TimeUtils::nowInSeconds();
+        }
+        case Status::CHANGE_LEADER: {
+            LOG(INFO) << taskIdStr_ << "Ask the src to give up the leadership.";
```
spacewalkman: Log the current leader `src`  could be helpful to diagnostic

src/meta/processors/admin/BalanceTask.cpp
```diff
@@ -0,0 +1,224 @@
...
+                    ret_ = Result::FAILED;
+                } else {
+                    status_ = Status::ADD_PART_ON_DST;
+                }
+                invoke();
+            });
+            break;
+        }
+        case Status::ADD_PART_ON_DST: {
+            LOG(INFO) << taskIdStr_ << "Open the part as leaner on dst.";
```
spacewalkman: ditto, put which `part` on which `dst` in log

src/meta/processors/admin/BalanceTask.cpp
```diff
@@ -0,0 +1,224 @@
...
+                    ret_ = Result::FAILED;
+                } else {
+                    status_ = Status::ADD_LEARNER;
+                }
+                invoke();
+            });
+            break;
+        }
+        case Status::ADD_LEARNER: {
+            LOG(INFO) << taskIdStr_ << "Add learner dst.";
```
spacewalkman: ditto

src/meta/processors/admin/BalanceTask.cpp
```diff
@@ -0,0 +1,224 @@
...
+                    ret_ = Result::FAILED;
+                } else {
+                    status_ = Status::REMOVE_PART_ON_SRC;
+                }
+                invoke();
+            });
+            break;
+        }
+        case Status::REMOVE_PART_ON_SRC: {
+            LOG(INFO) << taskIdStr_ << "Close part on src host.";
```
spacewalkman: ditto

src/meta/processors/admin/BalanceTask.cpp
```diff
@@ -0,0 +1,224 @@
...
+                } else {
+                    ret_ = Result::SUCCECCED;
+                    status_ = Status::END;
+                }
+                invoke();
+            });
+            break;
+        }
+        case Status::END: {
+            LOG(INFO) << taskIdStr_ <<  "Part has been moved successfully!";
```
spacewalkman: ditto

src/meta/processors/admin/BalanceTask.cpp
```diff
@@ -0,0 +1,224 @@
...
+        folly::Baton<true, std::atomic> baton;
+        bool ret = false;
+        kv_->asyncMultiPut(kDefaultSpaceId,
+                           kDefaultPartId,
+                           std::move(data),
+                           [this, &ret, &baton] (kvstore::ResultCode code) {
+            if (kvstore::ResultCode::SUCCEEDED == code) {
+                ret = true;
+            } else {
+                LOG(INFO) << taskIdStr_ << "Can't persist task!";
```
spacewalkman: IMO,  this is at LEAST a WARN level log and in ill-format 
```c++
LOG(INFO) << "Can't persist task:"<<taskIdStr_;
```

src/meta/processors/admin/BalanceTask.cpp
```diff
@@ -0,0 +1,224 @@
...
+                    ret_ = Result::FAILED;
+                } else {
+                    status_ = Status::MEMBER_CHANGE;
+                }
+                invoke();
+            });
+            break;
+        }
+        case Status::MEMBER_CHANGE: {
+            LOG(INFO) << taskIdStr_ << "Send member change request to the leader"
```
spacewalkman: ditto

src/meta/processors/admin/BalanceTask.cpp
```diff
@@ -0,0 +1,224 @@
...
+                    ret_ = Result::FAILED;
+                } else {
+                    status_ = Status::UPDATE_PART_META;
+                }
+                invoke();
+            });
+            break;
+        }
+        case Status::UPDATE_PART_META: {
+            LOG(INFO) << taskIdStr_ << "Update meta for part.";
```
spacewalkman: ditto

src/meta/processors/admin/AdminClient.h
```diff
@@ -0,0 +1,70 @@
...
+
+#include "base/Base.h"
+#include <folly/executors/IOThreadPoolExecutor.h>
+#include "base/Status.h"
+#include "thrift/ThriftClientManager.h"
+
+namespace nebula {
+namespace meta {
+
+class FaultInjector {
```
spacewalkman: What's the point of `Injector` part in the name?  To inject(introduce) a fault?  Kind of misleading, maybe we should come up with a better name?   FaultRecovery/FaultRejector/SafeGuardian etc, WDYT? Correct me if i misunderstood

src/meta/processors/admin/Balancer.h
```diff
@@ -0,0 +1,96 @@
...
+1. Balance will generate balance plan according to current active hosts and parts allocation
+2. For the plan, we hope after moving the least parts , it will reach a reasonable state.
+3. Only one balance plan could be invoked at the same time.
+4. Each balance plan has one id, and we could show the status by "balance id" command
+   and after FO, we could go on the balance plan by type "balance id"
+5. Each balance plan contains many balance tasks, the task represents the minimum movement unit.
+6. We save the whole balancePlan state in kvstore to do failover.
+7. Each balance task contains serval steps. And it should be executed step by step.
+8. One task failed will result in the whole balance plan failed.
+9. urrently, we hope tasks for the same part could be invoked serially
```
spacewalkman: typo? Currently?

src/meta/processors/admin/Balancer.h
```diff
@@ -0,0 +1,96 @@
...
+#include "meta/processors/admin/BalancePlan.h"
+
+namespace nebula {
+namespace meta {
+/**
+1. Balance will generate balance plan according to current active hosts and parts allocation
+2. For the plan, we hope after moving the least parts , it will reach a reasonable state.
+3. Only one balance plan could be invoked at the same time.
+4. Each balance plan has one id, and we could show the status by "balance id" command
+   and after FO, we could go on the balance plan by type "balance id"
```
spacewalkman: `go on` ---> `resume`  be better?

src/meta/processors/admin/Balancer.cpp
```diff
@@ -0,0 +1,264 @@
...
+    for (auto it = hostParts.begin(); it != hostParts.end(); it++) {
+        hosts.emplace_back(it->first, it->second.size());
+    }
+    std::sort(hosts.begin(), hosts.end(), [](const auto& l, const auto& r) {
+        return l.second < r.second;
+    });
+    return hosts;
+}
+
+StatusOr<HostAddr> Balancer::pickupHost(
```
spacewalkman: May be we should give the function with a more meaningful name: `hostWithMinimalParts`, WDYT?

src/meta/processors/admin/Balancer.cpp
```diff
@@ -0,0 +1,264 @@
...
+    balanceParts(newHostParts, totalParts, tasks);
+    return tasks;
+}
+
+void Balancer::balanceParts(std::unordered_map<HostAddr, std::vector<PartitionID>>& newHostParts,
+                            int32_t totalParts,
+                            std::vector<BalanceTask>& tasks) {
+    float avgLoad = static_cast<float>(totalParts)/newHostParts.size();
+    LOG(INFO) << "The expect avg load is " << avgLoad;
+    int32_t minLoad = avgLoad * 0.9;
```
spacewalkman: Does this criteria is fixed or tunable?  It would be better to make it tunable(a config item)? The smaller the gap, the bigger the rebalance load

src/meta/processors/admin/Balancer.h
```diff
@@ -0,0 +1,96 @@
...
+#include "network/NetworkUtils.h"
+#include "meta/processors/admin/AdminClient.h"
+#include "meta/processors/admin/BalanceTask.h"
+#include "meta/processors/admin/BalancePlan.h"
+
+namespace nebula {
+namespace meta {
+/**
+1. Balance will generate balance plan according to current active hosts and parts allocation
+2. For the plan, we hope after moving the least parts , it will reach a reasonable state.
```
spacewalkman: Could be better if we put the rational of `Why we want to to move the LEAST instead of the MAXIMAL parts` here?
Intuitionally，BALANCE means to move data from the MORE side to the LESS side. Why do it counter-intuitively?
IMO, we should doc the following(at least):

- What we are balancing? 
- What's the criteria of  balance/imbalance?
- How can we admin it?
- How can we tune it?

src/meta/processors/admin/BalanceProcessor.cpp
```diff
@@ -0,0 +1,49 @@
...
+
+#include "meta/processors/admin/BalanceProcessor.h"
+#include "meta/processors/admin/Balancer.h"
+
+namespace nebula {
+namespace meta {
+
+void BalanceProcessor::process(const cpp2::BalanceReq& req) {
+    if (req.get_space_id() != nullptr) {
+        LOG(ERROR) << "Unsupport balance for specal space " << *req.get_space_id();
```
critical27: typo

src/meta/processors/admin/BalancePlan.h
```diff
@@ -0,0 +1,73 @@
...
+#include "meta/processors/admin/BalanceTask.h"
+
+namespace nebula {
+namespace meta {
+
+class BalancePlan {
+    friend class Balancer;
+public:
+    enum class Status : uint8_t {
+        START              = 0x01,
```
critical27: Maybe NOT_STATED is better? It confuses me when I read code first time.

src/meta/processors/admin/Balancer.h
critical27: Maybe @dangleptr means "moving the minimum parts in quanity" here.

src/meta/processors/admin/Balancer.cpp
```diff
@@ -0,0 +1,264 @@
...
+    CHECK_GT(hosts.size(), 1);
+    auto maxPartsHost = hosts.back();
+    auto minPartsHost = hosts.front();
+    while (maxPartsHost.second > maxLoad || minPartsHost.second < minLoad) {
+        CHECK_GE(maxPartsHost.second, avgLoad);
+        CHECK_GE(avgLoad, minPartsHost.second);
+        auto& partsFrom = newHostParts[maxPartsHost.first];
+        auto& partsTo = newHostParts[minPartsHost.first];
+        std::vector<PartitionID> diff;
+        std::set_difference(partsFrom.begin(), partsFrom.end(), partsTo.begin(), partsTo.end(),
```
critical27: This balance algorithm works fine when we have only one replica. Just out of curiosity, what if we support multi-replica later, I think set_difference won't always get the minimum move, do you have better ideas?

src/meta/processors/admin/BalancePlan.h
critical27: Maybe NOT_STARTED will be better? It confuses me when I review at the first time.

src/meta/processors/admin/BalanceProcessor.cpp
critical27: typo "specal"

src/meta/processors/admin/BalanceProcessor.cpp
```diff
@@ -0,0 +1,49 @@
...
+        resp_.set_code(cpp2::ErrorCode::E_UNSUPPORTED);
+        onFinished();
+        return;
+    }
+    if (req.get_id() != nullptr) {
+        LOG(ERROR) << "Unsupport show status for special balance plan, id=" << *req.get_id();
```
spacewalkman: `special `? Do you mean `specific`?

src/meta/processors/admin/Balancer.h
dangleptr: Yes.  The principle is we reach balanced state with the least part movements

src/meta/processors/admin/Balancer.h
dangleptr: Good point.

src/meta/processors/admin/BalancePlan.h
critical27: Sounds ok, WAITING also make sense.

src/meta/processors/admin/AdminClient.h
dangleptr: Fault Injector is an component to inject some faults in the UTs.  
It could help us to mock the some corner cases.

src/meta/processors/admin/Balancer.cpp
dangleptr: I have updated the minLoad equals floor(avgLoad) and maxLoad equals ceil(avgLoad) now.

src/meta/processors/admin/Balancer.cpp
whitewum: IMO pickupHost could be better to be a interface function. what if another pickup method (by working load?).

dangleptr: HA...  I think both of them are OK now.  
As @spacewalkman mentioned above,  hostWithMinimalParts could be more direct.  Maybe we could change it with another name when the algorithm changed.

src/meta/processors/admin/BalanceProcessor.h
```diff
@@ -0,0 +1,34 @@
...
+#include "meta/processors/BaseProcessor.h"
+#include "meta/ActiveHostsMan.h"
+#include "time/TimeUtils.h"
+
+namespace nebula {
+namespace meta {
+
+class BalanceProcessor : public BaseProcessor<cpp2::BalanceResp> {
+public:
+    static BalanceProcessor* instance(kvstore::KVStore* kvstore) {
```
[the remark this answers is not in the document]
dangleptr: The delete will be invoked in onFinished method inside BaseProcessor.  
It is used when process is one async method.

src/meta/test/BalancerTest.cpp
```diff
@@ -0,0 +1,548 @@
...
+    }
+    {
+        std::vector<Status> sts{Status::Error("transLeader failed!"),
+                                Status::OK(),
+                                Status::OK(),
+                                Status::OK(),
+                                Status::OK(),
+                                Status::OK(),
+                                Status::OK()};
+        std::unique_ptr<FaultInjector> injector(new TestFaultInjector(std::move(sts)));
```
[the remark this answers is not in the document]
dangleptr: We can't use make_unique in this case.  Because the unique_ptr is for FaultInjector, but it takes pointer from TestFaultInjector.  There is an implicit convert inside it.

src/meta/processors/admin/Balancer.h
```diff
@@ -0,0 +1,119 @@
...
+#include "network/NetworkUtils.h"
+#include "meta/processors/admin/AdminClient.h"
+#include "meta/processors/admin/BalanceTask.h"
+#include "meta/processors/admin/BalancePlan.h"
+
+namespace nebula {
+namespace meta {
+/**
+There are two interfaces public:
+ * Balance:  it will construct a balance plan and invoked it. If last balance plan is not succeeded, it will
```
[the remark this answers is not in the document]
dangleptr: Good suggestion.  Let me think about it.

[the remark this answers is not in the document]
whitewum: BTW, another option is to let DBA generate a balance plan manually  by some tools ( let's suppose the pickup algorithm doesn't meet DBA's requirement) , add give the plan to the meta service to do the job.

Make a system smart enough to deal all the situations at the beginning stage is every hard ( and even not possible). It needs more operation data/logs to make a model to train the system. Human experience/plan may be more efficient.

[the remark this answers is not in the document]
dangleptr: Sounds good.  I will add TODO comments here.  Let's address them in the future.

src/meta/test/BalancerTest.cpp
```diff
@@ -0,0 +1,548 @@
...
+        hostParts.emplace(HostAddr(2, 0), std::vector<PartitionID>{1, 2, 3, 4});
+        hostParts.emplace(HostAddr(3, 0), std::vector<PartitionID>{});
+        int32_t totalParts = 12;
+        std::vector<BalanceTask> tasks;
+        VLOG(1) << "=== original map ====";
+        dump(hostParts, tasks);
+        balancer->balanceParts(0, 0, hostParts, totalParts, tasks);
+        VLOG(1) << "=== new map ====";
+        dump(hostParts, tasks);
+        for (auto it = hostParts.begin(); it != hostParts.end(); it++) {
```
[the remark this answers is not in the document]
dangleptr: For balance,  we care more about numbers.  
For example,  in this case,   we don't care the new host takes {1, 2, 3},  {2, 3, 4}, {1, 3, 4} 
All of them are correct.

src/meta/test/BalancerTest.cpp
```diff
@@ -0,0 +1,548 @@
...
+        hostParts.emplace(HostAddr(7, 0), std::vector<PartitionID>{});
+        int32_t totalParts = 27;
+        std::vector<BalanceTask> tasks;
+        VLOG(1) << "=== original map ====";
+        dump(hostParts, tasks);
+        balancer->balanceParts(0, 0, hostParts, totalParts, tasks);
+        VLOG(1) << "=== new map ====";
+        dump(hostParts, tasks);
+        for (auto it = hostParts.begin(); it != hostParts.end(); it++) {
+            EXPECT_GE(4, it->second.size());
```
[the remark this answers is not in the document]
dangleptr: yes, we can't fix the result because the algorithm maybe changed.  But we should ensure the numbers is balanced.

src/meta/processors/admin/BalancePlan.cpp
```diff
@@ -0,0 +1,198 @@
...
+const std::string kBalancePlanTable = "__b_plan__"; // NOLINT
+
+void BalancePlan::dispatchTasks() {
+    // Key -> spaceID + partID,  Val -> List of task index in tasks_;
+    std::unordered_map<std::pair<GraphSpaceID, PartitionID>, std::vector<int32_t>> partTasks;
+    int32_t index = 0;
+    for (auto& task : tasks_) {
+        partTasks[std::make_pair(task.spaceId_, task.partId_)].emplace_back(index++);
+    }
+    buckets_.resize(std::min(tasks_.size(), (size_t)FLAGS_task_concurrency));
```
critical27: Maybe use partTasks.size() instead of tasks_.size() will be better?

src/meta/processors/admin/BalanceTask.cpp
```diff
@@ -0,0 +1,234 @@
...
+            status_ = Status::CHANGE_LEADER;
+            ret_ = Result::IN_PROGRESS;
+            startTimeMs_ = time::TimeUtils::nowInSeconds();
+        }
+        case Status::CHANGE_LEADER: {
+            LOG(INFO) << taskIdStr_ << "Ask the src to give up the leadership.";
+            SAVE_STATE();
+            client_->transLeader(spaceId_, partId_, src_, dst_).thenValue([this](auto&& resp) {
+                if (!resp.ok()) {
+                    ret_ = Result::FAILED;
```
critical27: How about sleep a while if failed or retry only a few times, otherwise it would invoke all the times?

src/meta/processors/admin/BalancePlan.cpp
dangleptr: Good point!

src/meta/processors/admin/BalanceTask.cpp
dangleptr: Good suggestion!  I will wrap retry logic inside AdminClient.

dangleptr:
Jenkins go

PR #518, opened 2019-06-12
Fixed number of worker threads

dutor:
As title.

### DIFF
--- a/etc/nebula-graphd.conf.default
+++ b/etc/nebula-graphd.conf.default
@@ -36,8 +36,10 @@
 --session_idle_timeout_secs=60000
 # The number of threads to accept incoming connections
 --num_accept_threads=1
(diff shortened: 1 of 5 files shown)


dangleptr:
Good one.  We need the configuration for storage/meta too.  Do you want to take care of it.  @dutor

dutor:
> Good one. We need the configuration for storage/meta too. Do you want to take care of it. @dutor

I had a local patch trying to unify the startup code in the three daemons. I will address these issue in that PR.

dangleptr:
> > Good one. We need the configuration for storage/meta too. Do you want to take care of it. @dutor
> 
> I had a local patch trying to unify the startup code in the three daemons. I will address these issue in that PR.

Very nice.  You are the man!

PR #522, opened 2019-06-13
Fixed issues on linking against linunwind and removed maybe-non-existing option of Git command

laura-ding:


laura-ding:
jenkins go

PR #523, opened 2019-06-13
Use basic type instead of int32/int64

zlcook:
As title


laura-ding:
jenkins go

dangleptr:
Jenkins go

PR #524, opened 2019-06-13
fix metad crashed when reboot

critical27:
Fix meta daemons crashed because of invalid operation when leader has not been elected.

### DIFF
--- a/src/meta/ActiveHostsMan.cpp
+++ b/src/meta/ActiveHostsMan.cpp
@@ -16,7 +16,7 @@ ActiveHostsMan::ActiveHostsMan(int32_t intervalSeconds, int32_t expiredSeconds,
     : intervalSeconds_(intervalSeconds)
     , expirationInSeconds_(expiredSeconds) {
     if (kv != nullptr) {
-        kvstore_ = kv;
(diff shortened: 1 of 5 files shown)


src/meta/ActiveHostsMan.cpp
```diff
@@ -47,11 +47,14 @@ void ActiveHostsMan::updateHostInfo(const HostAddr& hostAddr, const HostInfo& in
     }
     if (kvstore_ != nullptr && !data.empty()) {
         folly::SharedMutex::WriteHolder wHolder(LockUtils::spaceLock());
+        bool ret = true;
         kvstore_->asyncMultiPut(kDefaultSpaceId, kDefaultPartId, std::move(data),
-                                [] (kvstore::ResultCode code) {
-            CHECK_EQ(code, kvstore::ResultCode::SUCCEEDED);
+                                [&ret] (kvstore::ResultCode code) {
+            ret = (code == kvstore::ResultCode::SUCCEEDED);
```
dangleptr: The callback is invoked in another thread.  The "ret" maybe invalid there.

critical27:
Jenkins go

PR #482, opened 2019-05-31
Support insert multi tags and edges

laura-ding:
1、Support insert multi tags
2、Support insert multi edges

PR #525, opened 2019-06-14
Check if properties and values match when inserting vertices and inse…

steppenwolfyuetong:
Check if properties and values match when inserting vertices and inserting edges


steppenwolfyuetong:
@dutor @dangleptr @laura-ding ,  please review, thx.

steppenwolfyuetong:
`PR #482(Support insert multi tags and edges )` has already contained this PR, all close this PR.

PR #537, opened 2019-06-22
Implementation of Order by

CPWstatic:


CPWstatic:
jenkins go

CPWstatic:
jenkins go

src/graph/OrderByExecutor.cpp
```diff
@@ -0,0 +1,205 @@
+/* Copyright (c) 2018 vesoft inc. All rights reserved.
```
laura-ding: use 2019

src/graph/OrderByExecutor.h
```diff
@@ -0,0 +1,43 @@
+/* Copyright (c) 2018 vesoft inc. All rights reserved.
```
laura-ding: ditto

src/parser/parser.yy
```diff
@@ -74,6 +74,8 @@ class GraphScanner;
     nebula::SchemaPropItem                 *create_schema_prop_item;
     nebula::SchemaPropList                 *alter_schema_prop_list;
     nebula::SchemaPropItem                 *alter_schema_prop_item;
+	nebula::OrderFactor                    *order_factor;
```
laura-ding: alignment

src/parser/parser.yy
```diff
@@ -701,10 +707,47 @@ drop_edge_sentence
     }
     ;
 
+order_factor
```
laura-ding: Add  UTs of parser

src/parser/scanner.lex
```diff
@@ -96,6 +96,10 @@ BY                          ([Bb][Yy])
 IN                          ([Ii][Nn])
 TTL_DURATION                ([Tt][Tt][Ll][_][Dd][Uu][Rr][Aa][Tt][Ii][Oo][Nn])
 TTL_COL                     ([Tt][Tt][Ll][_][Cc][Oo][Ll])
+ORDER                       ([Oo][Rr][Dd][Ee][Rr])
```
laura-ding: Add  UTs of scanner

src/graph/OrderByExecutor.cpp
```diff
@@ -0,0 +1,205 @@
...
+        }
+        case Type::datetime:
+        {
+            if (!(lhs.value_.datetime == rhs.value_.datetime)) {
+                return lhs.value_.datetime < rhs.value_.datetime;
+            }
+        }
+        default:
+        {
+            return true;
```
laura-ding: should be `return false`?

src/graph/test/OrderByTest.cpp
```diff
@@ -0,0 +1,224 @@
+/* Copyright (c) 2018 vesoft inc. All rights reserved.
```
laura-ding: ditto

src/graph/test/TraverseTestBase.h
```diff
@@ -0,0 +1,787 @@
+/* Copyright (c) 2018 vesoft inc. All rights reserved.
```
laura-ding: ditto

src/graph/OrderByExecutor.h
```diff
@@ -0,0 +1,43 @@
...
+    void setupResponse(cpp2::ExecutionResponse &resp) override;
+
+private:
+    std::unique_ptr<InterimResult> setupInterimResult();
+
+private:
+    OrderBySentence                                            *sentence_{nullptr};
+    std::unique_ptr<InterimResult>                              inputs_;
+    std::vector<cpp2::RowValue>                                 rows_;
+    std::vector<std::pair<int64_t, OrderFactor::OrderType>>     sortFactors_;
```
laura-ding: Explain uint64_t is the field index of schema

src/graph/OrderByExecutor.cpp
CPWstatic: you are right

src/graph/OrderByExecutor.cpp
```diff
@@ -0,0 +1,205 @@
...
+bool ColumnValue::operator < (const ColumnValue& rhs) const {
+    DCHECK_EQ(type_, rhs.type_);
+    auto& lhs = *this;
+    switch (lhs.type_) {
+        case Type::bool_val:
+        {
+            if (!(lhs.value_.bool_val == rhs.value_.bool_val)) {
+                return lhs.value_.bool_val < rhs.value_.bool_val;
+            }
+        }
```
wadeliuyi: may be need break after every cace

CPWstatic: you are right

src/graph/OrderByExecutor.h
CPWstatic: check SchemaProvider::getFieldIndex, it returns integer

src/parser/parser.yy
CPWstatic: it looks well in vim, try to fix it

src/parser/scanner.lex
```diff
@@ -184,6 +188,10 @@ IP_OCTET                    ([0-9]|[1-9][0-9]|1[0-9][0-9]|2[0-4][0-9]|25[0-5])
 {TTL_COL}                   { return TokenType::KW_TTL_COL; }
 {TRUE}                      { yylval->boolval = true; return TokenType::BOOL; }
 {FALSE}                     { yylval->boolval = false; return TokenType::BOOL; }
+{ORDER}                     { return TokenType::KW_ORDER; }
+{ASCEND}                    { return TokenType::KW_ASCEND; }
+{ASC}                       { return TokenType::KW_ASC; }
+{DESCEND}                   { return TokenType::KW_DESCEND; }
```
steppenwolfyuetong: No, no , should not exist  "order by xxx  ascend" or "  "order by xxx  descend" in nGQL.  

Only exist as follows "order by xxxx  asc" or "  "order by xxx desc",    the  default is asc in oder by clause.

CPWstatic: ok , i'll delete them

src/parser/test/ParserTest.cpp
```diff
@@ -702,5 +702,45 @@ TEST(Parser, Annotation) {
     }
 }
 
+TEST(Parser, Agg) {
+    {
+        GQLParser parser;
+        std::string query = "ORDER BY $-.id";
+        auto result = parser.parse(query);
+        ASSERT_TRUE(result.ok()) << result.status();
+    }
```
steppenwolfyuetong: Can the `ORDER BY` clause be used alone?

src/parser/parser.yy
```diff
@@ -701,10 +707,41 @@ drop_edge_sentence
...
+        $1->addFactor($3);
+        $$ = $1;
+    }
+    ;
+
+order_by_sentence
+    : KW_ORDER KW_BY order_factors {
+        $$ = new OrderBySentence($3);
+    }
+    ;
```
steppenwolfyuetong: IMO, it should be order_by_clause, not order_by_sentence.

src/graph/OrderByExecutor.cpp
```diff
@@ -0,0 +1,226 @@
...
+            }
+            break;
+        }
+        case Type::date:
+        {
+            if (!(lhs.value_.date == rhs.value_.date)) {
+                return lhs.value_.date < rhs.value_.date;
+            }
+            break;
+        }
```
steppenwolfyuetong: To date, datetime, are you comparing by string or by corresponding value?

src/graph/OrderByExecutor.cpp
```diff
@@ -0,0 +1,226 @@
...
+        case Type::datetime:
+        {
+            if (!(lhs.value_.datetime == rhs.value_.datetime)) {
+                return lhs.value_.datetime < rhs.value_.datetime;
+            }
+            break;
+        }
```
steppenwolfyuetong: ditto

src/parser/parser.yy
CPWstatic: i was discussing this with dutor and sheman before,  it would be better a sentence at present, or it would made the go executor too much bloated. For further reason is that we have no optimization for execution plan for now.  We should do some optimization for clause, and we'd prefer doing some go-clauses in storage layer for performance. But it is a challenge to this optimization, we'll disscuss this and put it on the agenda as soon as possible.

src/graph/OrderByExecutor.cpp
CPWstatic: by value, how do u think? it depends on the implementation of storage actually. Date and Datetime is not implement in storage. For now, i am doing the sort with the union type of ColumnValue, because it is comparable.

src/parser/test/ParserTest.cpp
CPWstatic: i treat _ORDER BY_ as a sentence for now. It would not be a syntax error, but return nothing.

src/parser/parser.yy
steppenwolfyuetong: I think `order by` as a separate sql is weird, it should be an algorithm  or a  clause

src/graph/OrderByExecutor.cpp
steppenwolfyuetong: Although `date` and `datetime` is not implemented,  what I really care about is how to store these two types, whether they are stored as strings or converted to timestamps.

CPWstatic: Do u mean that _how to store in storage_?

src/graph/OrderByExecutor.cpp
```diff
@@ -0,0 +1,226 @@
...
+        return false;
+    };
+
+    if (!sortFactors_.empty()) {
+        std::sort(rows_.begin(), rows_.end(), comparator);
+    }
+
+    if (onResult_) {
+        LOG(INFO) << "on result";
```
dutor: Do we really need to a log here?

src/graph/OrderByExecutor.cpp
```diff
@@ -0,0 +1,226 @@
...
+        }
+        auto pair = std::make_pair(schema->getFieldIndex(field), factor->orderType());
+        sortFactors_.emplace_back(std::move(pair));
+    }
+}
+
+void OrderByExecutor::execute() {
+    FLOG_INFO("Executing Order By: %s", sentence_->toString().c_str());
+    auto comparator = [this] (cpp2::RowValue& lhs, cpp2::RowValue& rhs) {
+        auto lhsColumns = lhs.get_columns();
```
dutor: You are copying the columns. 

IMO. You could eliminate the copying with `const auto &lhsColumns = ...`, provided that `lhs` and `rhs` are passed by const reference.

src/graph/OrderByExecutor.cpp
```diff
@@ -0,0 +1,226 @@
...
+namespace graph {
+namespace cpp2 {
+
+bool ColumnValue::operator < (const ColumnValue& rhs) const {
+    DCHECK_EQ(type_, rhs.type_);
+    auto& lhs = *this;
+    switch (lhs.type_) {
+        case Type::bool_val:
+        {
+            if (!(lhs.value_.bool_val == rhs.value_.bool_val)) {
```
dutor: Any reason to check equalness first?

dutor: As the offline discussion, the equality checking here is not necessary.

src/graph/test/OrderByTest.cpp
```diff
@@ -0,0 +1,231 @@
...
+    }
+}
+
+TEST_F(OrderByTest, WrongFactor) {
+    std::string go = "GO FROM %ld OVER serve YIELD "
+                     "$^[player].name as name, serve.start_year as start, $$[team].name as team";
+    {
+        cpp2::ExecutionResponse resp;
+        auto &player = players_["Boris Diaw"];
+        // Will not do sort if field name not exist in input schema
```
whitewum: a bit weird. a warning to the user is better

src/parser/test/ParserTest.cpp
whitewum: Mark a TODO to explain.

src/graph/OrderByExecutor.cpp
CPWstatic: thank you!!!

src/parser/test/ScannerTest.cpp
```diff
@@ -317,6 +317,13 @@ TEST(Scanner, Basic) {
         CHECK_SEMANTIC_TYPE("TTL_COL", TokenType::KW_TTL_COL),
         CHECK_SEMANTIC_TYPE("ttl_col", TokenType::KW_TTL_COL),
         CHECK_SEMANTIC_TYPE("Ttl_col", TokenType::KW_TTL_COL),
+        CHECK_SEMANTIC_TYPE("ORDER", TokenType::KW_ORDER),
+        CHECK_SEMANTIC_TYPE("Order", TokenType::KW_ORDER),
+        CHECK_SEMANTIC_TYPE("order", TokenType::KW_ORDER),
+        CHECK_SEMANTIC_TYPE("ASC", TokenType::KW_ASC),
+        CHECK_SEMANTIC_TYPE("Asc", TokenType::KW_ASC),
+        CHECK_SEMANTIC_TYPE("asc", TokenType::KW_ASC),
```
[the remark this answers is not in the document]
CPWstatic: desc already exist.

dutor:
There are conflicts with upstream. Please resolve them.

CPWstatic:
> There are conflicts with upstream. Please resolve them.

done

PR #540, opened 2019-06-24
Do prepare after setting onResult_ in PipeExecutor::prepare

CPWstatic:
Do prepare after setting onResult_ in PipeExecutor::prepare, or these cases will not work well:
***GO ... | GO ... | GO ...***
***GO ... | (GO ... | GO ...)***


CPWstatic:
jenkins go

PR #541, opened 2019-06-25
Use folly::Function instead of std::function in kvstore callback

dangleptr:
std::function has two defects:
1.   It must be copy constructible,  so we can't pass object which is NOT in capture list.
2.  It has problems with const-ness method.

For more information, please refer to https://github.com/facebook/folly/blob/master/folly/Function.h

### DIFF
--- a/src/kvstore/NebulaStore.cpp
+++ b/src/kvstore/NebulaStore.cpp
@@ -313,7 +313,7 @@ void NebulaStore::asyncMultiRemove(GraphSpaceID spaceId,
                                    KVCallback cb) {
     folly::RWSpinLock::ReadHolder rh(&lock_);
     CHECK_FOR_WRITE(spaceId, partId, cb);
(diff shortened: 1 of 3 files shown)


dangleptr:
Jenkins go

dangleptr:
Jenkins go

src/kvstore/NebulaStore.cpp
```diff
@@ -313,7 +313,7 @@ void NebulaStore::asyncMultiRemove(GraphSpaceID spaceId,
                                    KVCallback cb) {
     folly::RWSpinLock::ReadHolder rh(&lock_);
     CHECK_FOR_WRITE(spaceId, partId, cb);
-    return partIt->second->asyncMultiRemove(std::move(keys), cb);
+    return partIt->second->asyncMultiRemove(std::move(keys), std::move(cb));
```
boshengchen: 👍

Issue #532, opened 2019-06-19
Merge host info from kvstore and memory


critical27:
For now, ActiveHostsMan will load host info from kvstore when it start-up, so it have to wait until kvstore is ready. We can load host info in background thread and merge result from kvstore and memory to update the online/offline status. In that way, we can get rid of sleep when initializing kvstore.

dangleptr:
Very good.

andychow01:
Very good.

sherman-the-tank:
Why do we persist the active host info in kvstore? If this is the cache of the meta information we discussed earlier, we need to store other information as well.

Please shed more light, or we can talk about this offline

critical27:
@sherman-the-tank 
As #440, we add some host status of online/offline. So we persist the host status in kvstore.

critical27:
In #550, @andychow01 fix this by loading info from kvstore lazily, when kvstore is ready, we can merge info from kvstore and memory. The reason of merging is that the host status in kvstore may be outdated because of leadership change.

PR #544, opened 2019-06-25
merge host info

andychow01:
fix issue 532:
    https://github.com/vesoft-inc/nebula/issues/532

merge host infos from kvstore.


src/meta/ActiveHostsMan.cpp
```diff
@@ -94,8 +99,11 @@ void ActiveHostsMan::loadHostMap() {
         HostInfo info;
         info.lastHBTimeInSec_ = time::TimeUtils::nowInSeconds();
         if (iter->val() == MetaServiceUtils::hostValOnline()) {
```
critical27: I think we need to set status of hosts "which is online in kvstore but not in memory" to offline? Because the info in memory is more updated. After leader really receive heartbeat, it will set info to online.

src/meta/test/ActiveHostsMergeTest.cpp
```diff
@@ -0,0 +1,54 @@
+
```
critical27: Maybe just put the ut in ActiveHostsManTest?

andychow01: yes!

PR #546, opened 2019-06-25
fix unreachable code and imcomplete UT in FileUtils

### LINKED ISSUE #545: Unreachable code and flaky UT in FileUtils
```c++
    if (recursively) {
            if (dEnt->d_type == DT_DIR && !recursively) {	   
                  // unreachable code here
            }
   }
```
And UT does not test the non-recursive part properly, should test a parent with a sub dir and a file under it.

spacewalkman:
close #545


spacewalkman:
```c++
    if (recursively) {
            if (dEnt->d_type == DT_DIR && !recursively) {	   
                  // unreachable code here
            }
   }
```
And UT does not test the non-recursive part properly, should test a parent with a sub dir and a file under it.

src/common/fs/FileUtils.cpp
```diff
@@ -16,62 +16,60 @@ static const int32_t kMaxPathLen = 1024;
 
 namespace detail {
 
-bool removeDir(const char* path, bool recursively) {
-    if (recursively) {
```
dutor: Good catch!

But I notice that `removeDir` behaves differently with the standard `rmdir` API on what's recursive. This might be counter intuitive, at least to me. But let us discuss this later.

src/common/fs/FileUtils.cpp
```diff
@@ -95,7 +93,6 @@ StatusOr<std::string> FileUtils::readLink(const char *path) {
     return std::string(buffer, len);
 }
 
-
```
dutor: @sherman-the-tank encourages this styling. Yet this has not been synced to others.

src/common/fs/FileUtils.cpp
```diff
@@ -488,24 +466,28 @@ void FileUtils::Iterator::fileNext() {
...
+                                    path_.c_str(),
+                                    ::strerror(errno));
             return;
         }
     } else if (type_ == FileType::REGULAR) {
         fstream_ = std::make_unique<std::ifstream>();
         fstream_->open(path_);
         if (!fstream_->is_open()) {
-            status_ = Status::Error("open `%s': %s", path_.c_str(), ::strerror(errno));
+            status_ = Status::Error("open `%s': %s",
```
dutor: Such changes are OK to me. But for your reference, our linting rule on column width is 100c, which you could check on `src/.linters`.

spacewalkman: I undo the change related to the linter, just focus on the logic part and repushed.

PR #549, opened 2019-06-26
Fixed crash on leader change

dutor:
`leader` in `storage:cpp2::ResultCode` is an optional field, and `get_leader()` returns `nullptr` if not set.


dutor:
Jenkins, go!

dutor:
Jenkins go!

PR #550, opened 2019-06-26
merge host info from kvstore

andychow01:
fix bug:
    https://github.com/vesoft-inc/nebula/issues/532


src/meta/ActiveHostsMan.cpp
```diff
@@ -120,6 +120,26 @@ void ActiveHostsMan::cleanExpiredHosts() {
...
+    // merge host info from kvstore
+    if (kvstore_ != nullptr) {
+        const auto& prefix = MetaServiceUtils::hostPrefix();
+        std::unique_ptr<kvstore::KVIterator> iter;
+        auto ret = kvstore_->prefix(kDefaultSpaceId, kDefaultPartId, prefix, &iter);
+        if (ret == kvstore::ResultCode::SUCCEEDED) {
+            while (iter->valid()) {
+                auto host = MetaServiceUtils::parseHostKey(iter->key());
+                if (iter->val() == MetaServiceUtils::hostValOnline()) {
+                    bool found = hostsMap_.find({host.ip, host.port}) == hostsMap_.end();
```
critical27: Maybe you need a lock here

critical27: Also maybe just use if (hostsMap_.find({host.ip, host.port}) == hostsMap_.end()) here, you name it "found", which means "not found in memory", it confuses.

critical27: Here is another case, what if it is offline in kvstore, but online in memory, should we handle it? If we don't, those host will stay offline forever, the result of getActiveHosts and "show hosts" will be different.

src/meta/test/ActiveHostsManTest.cpp
```diff
@@ -37,6 +37,56 @@ TEST(ActiveHostsManTest, NormalTest) {
     }
 }
 
+TEST(ActiveHostsManTest, NormalTest2) {
```
spacewalkman: Should give it more meaningful test name like `MergeHostTest` ? WDYT?

src/meta/ActiveHostsMan.cpp
```diff
@@ -120,6 +120,27 @@ void ActiveHostsMan::cleanExpiredHosts() {
...
+        auto ret = kvstore_->prefix(kDefaultSpaceId, kDefaultPartId, prefix, &iter);
+        if (ret == kvstore::ResultCode::SUCCEEDED) {
+            while (iter->valid()) {
+                auto host = MetaServiceUtils::parseHostKey(iter->key());
+                if (iter->val() == MetaServiceUtils::hostValOnline()) {
+                    folly::RWSpinLock::ReadHolder rh(&lock_);
+                    bool notFound = hostsMap_.find({host.ip, host.port}) == hostsMap_.end();
+                    if (notFound) {
+                        data.emplace_back(MetaServiceUtils::hostKey(host.ip, host.port),
+                                MetaServiceUtils::hostValOffline());
```
dangleptr: Alignment

src/meta/test/ActiveHostsManTest.cpp
```diff
@@ -37,6 +37,56 @@ TEST(ActiveHostsManTest, NormalTest) {
...
+    fs::TempDir rootPath("/tmp/ActiveHostsMergeTest.XXXXXX");
+    std::unique_ptr<kvstore::KVStore> kv(TestUtils::initKV(rootPath.path()));
+    {
+        std::vector<kvstore::KV> data;
+        for (auto i = 0; i < 3; i++) {
+            data.emplace_back(MetaServiceUtils::hostKey(0, i),
+                MetaServiceUtils::hostValOnline());
```
dangleptr: Alignment

src/meta/test/ActiveHostsManTest.cpp
dangleptr: +1

src/meta/test/ActiveHostsManTest.cpp
```diff
@@ -37,6 +37,56 @@ TEST(ActiveHostsManTest, NormalTest) {
...
+        } else {
+            ++offlineNum;
+        }
+        iter->next();
+    }
+    ASSERT_EQ(3, onlineNum);
+    ASSERT_EQ(0, offlineNum);
+
+    ActiveHostsMan ahMan(1, 1, kv.get());
+    sleep(3);
```
dangleptr: sleep(intervalSeconds + 1)

src/meta/test/ActiveHostsManTest.cpp
andychow01: fixed

src/meta/ActiveHostsMan.cpp
andychow01: fixed

src/meta/test/ActiveHostsManTest.cpp
andychow01: fixed

src/meta/test/ActiveHostsManTest.cpp
andychow01: fixed

PR #552, opened 2019-06-28
Support command with RAFT

dangleptr:
Support sending command with RAFT.  
When encountering a command log,   leader will hold the logs after it until the command log committed. 

It is the foundation for the following features include "Transfer leader", "member change" etc. 

Subtask of #515


dangleptr:
Jenkins go

src/kvstore/raftex/RaftPart.h
```diff
@@ -123,6 +128,11 @@ class RaftPart : public std::enable_shared_from_this<RaftPart> {
      ***************************************************************/
     folly::Future<AppendLogResult> casAsync(std::string log);
 
+    /**
+     * Asynchronously send one command.
+     * */
+    folly::Future<AppendLogResult> commandAsync(std::string log);
```
sherman-the-tank: Maybe rename it to ```sendCommandAsync()``` to make it consistent with other ```verbXxxxAsync()``` methods?

src/kvstore/raftex/RaftPart.h
```diff
@@ -325,6 +335,14 @@ class RaftPart : public std::enable_shared_from_this<RaftPart> {
             return singlePromises_.back().getFuture();
         }
 
+        folly::Future<ValueType> getAndRollSharedFuture() {
```
sherman-the-tank: Not sure why you need to use SharedPromise here. I think what you need is a SinglePromise, just like CAS does

dangleptr:
> From what I understand here, it seems COMMAND is very similar to CAS. The only difference is at the commit stage. So I think we can treat COMMAND same as CAS until we are ready to commit

Yes,  COMMAND is similar to CAS, but not the same.  There are two differences:
 * Normal logs after CAS could be committed together.   In opposite,  Normal logs after COMMAND should be hold until the COMMAND committed,  but the logs **before** COMMAND could be committed together.  This is the purpose about changes in LogIterator. 
 * CAS maybe failed.  So we use SinglePromise for it.  But COMMAND not,  so it could shared one promise with the normal logs before it.  This is the purpose about changes in PromiseSet. 

@sherman-the-tank

src/kvstore/raftex/RaftPart.h
```diff
@@ -39,10 +39,23 @@ enum class AppendLogResult {
     E_WAL_FAILURE = -6,
 };
 
+enum class LogType {
+    NORMAL      = 0x00,
+    CAS         = 0x01,
+    /**
+      COMMAND is similar to CAS, but not the same. There are two differences:
```
sherman-the-tank: Well done 👍

src/kvstore/raftex/RaftPart.h
```diff
@@ -326,6 +344,14 @@ class RaftPart : public std::enable_shared_from_this<RaftPart> {
             return singlePromises_.back().getFuture();
         }
 
+        folly::Future<ValueType> getAndRollSharedFuture() {
+            if (rollSharedPromise_) {
+                sharedPromises_.emplace_back();
+            }
+            rollSharedPromise_ = true;
```
critical27: It really confuses me about rollSharedPromise_ at first :joy:.

PR #553, opened 2019-06-28
Implement admin client for balancer

dangleptr:
The PR is based on #516 .

Subtask of #515

### DIFF
--- a/src/meta/processors/admin/AdminClient.cpp
+++ b/src/meta/processors/admin/AdminClient.cpp
@@ -329,6 +329,7 @@ StatusOr<std::vector<HostAddr>> AdminClient::getPeers(GraphSpaceID spaceId, Part
         case kvstore::ResultCode::ERR_KEY_NOT_FOUND:
             return Status::Error("Key Not Found");
         default:
+            LOG(WARNING) << "Get peers failed, error " << static_cast<int32_t>(code);
             break;
     }
     return Status::Error("Get Failed");
--- a/src/meta/processors/admin/AdminClient.cpp
+++ b/src/meta/processors/admin/AdminClient.cpp
@@ -75,7 +75,7 @@ folly::Future<Status> AdminClient::addLearner(GraphSpaceID spaceId, PartitionID 
     }
     folly::Promise<Status> pro;
     auto f = pro.getFuture();
-    getResponse(ret.value(), 0, req, [] (auto client, auto request) {
+    getResponse(ret.value(), 0, std::move(req), [] (auto client, auto request) {
         return client->future_addLearner(request);
     }, 0, std::move(pro), FLAGS_max_retry_times_admin_op);
     return f;
@@ -95,7 +95,7 @@ folly::Future<Status> AdminClient::waitingForCatchUpData(GraphSpaceID spaceId,
     }
     folly::Promise<Status> pro;
     auto f = pro.getFuture();
-    getResponse(ret.value(), 0, req, [] (auto client, auto request) {
+    getResponse(ret.value(), 0, std::move(req), [] (auto client, auto request) {
         return client->future_waitingForCatchUpData(request);
     }, 0, std::move(pro), FLAGS_max_retry_times_admin_op);
     return f;
@@ -114,7 +114,7 @@ folly::Future<Status> AdminClient::memberChange(GraphSpaceID spaceId, PartitionI
     }
     folly::Promise<Status> pro;
     auto f = pro.getFuture();
-    getResponse(ret.value(), 0, req, [] (auto client, auto request) {
+    getResponse(ret.value(), 0, std::move(req), [] (auto client, auto request) {
         return client->future_memberChange(request);
     }, 0, std::move(pro), FLAGS_max_retry_times_admin_op);
     return f;
@@ -203,7 +203,7 @@ folly::Future<Status> AdminClient::getResponse(
                                                          t.exception().what().c_str())));
             return;
         }
-        auto&& resp = t.value();
+        auto&& resp = std::move(t).value();
         p.setValue(respGen(std::move(resp)));
     });
     return f;
--- a/src/meta/processors/admin/AdminClient.cpp
+++ b/src/meta/processors/admin/AdminClient.cpp
@@ -5,90 +5,333 @@
  */
 
 #include "meta/processors/admin/AdminClient.h"
+#include "meta/MetaServiceUtils.h"
+#include "meta/processors/Common.h"
+
+DEFINE_int32(max_retry_times_admin_op, 3, "max retry times for admin request!");
 
 namespace nebula {
 namespace meta {
 
 folly::Future<Status> AdminClient::transLeader(GraphSpaceID spaceId,
-                                                      PartitionID partId,
-                                                      const HostAddr& leader,
-                                                      const HostAddr& dst) {
-    UNUSED(spaceId);
-    UNUSED(partId);
-    UNUSED(leader);
-    UNUSED(dst);
+                                               PartitionID partId,
+                                               const HostAddr& leader,
+                                               const HostAddr& dst) {
     if (injector_) {
         return injector_->transLeader();
     }
-    return Status::OK();
+    storage::cpp2::TransLeaderReq req;
+    req.set_space_id(spaceId);
+    req.set_part_id(partId);
+    req.set_new_leader(to(dst));
+    return getResponse(leader, std::move(req), [] (auto client, auto request) {
+               return client->future_transLeader(request);
+           }, [this] (auto&& resp) -> Status {
+               switch (resp.get_code()) {
+                   case storage::cpp2::ErrorCode::SUCCEEDED:
+                   case storage::cpp2::ErrorCode::E_LEADER_CHANGED: {
+                       return Status::OK();
+                   }
+                   default:
+                       return Status::Error("Unknown code %d",
+                                            static_cast<int32_t>(resp.get_code()));
+               }
+           });
 }
 
 folly::Future<Status> AdminClient::addPart(GraphSpaceID spaceId,
-                                                  PartitionID partId,
-                                                  const HostAddr& host,
-                                                  bool asLearner) {
-    UNUSED(spaceId);
-    UNUSED(partId);
-    UNUSED(host);
-    UNUSED(asLearner);
+                                           PartitionID partId,
+                                           const HostAddr& host,
+                                           bool asLearner) {
     if (injector_) {
         return injector_->addPart();
     }
-    return Status::OK();
+    storage::cpp2::AddPartReq req;
+    req.set_space_id(spaceId);
+    req.set_part_id(partId);
+    req.set_as_learner(asLearner);
+    return getResponse(host, std::move(req), [] (auto client, auto request) {
+               return client->future_addPart(request);
+           }, [this] (auto&& resp) -> Status {
+               if (resp.get_code() == storage::cpp2::ErrorCode::SUCCEEDED) {
+                   return Status::OK();
+               } else {
+                   return Status::Error("Add part failed! code=%d",
+                                        static_cast<int32_t>(resp.get_code()));
+               }
+           });
 }
 
 folly::Future<Status> AdminClient::addLearner(GraphSpaceID spaceId, PartitionID partId) {
-    UNUSED(spaceId);
-    UNUSED(partId);
     if (injector_) {
         return injector_->addLearner();
     }
-    return Status::OK();
+    storage::cpp2::AddLearnerReq req;
+    req.set_space_id(spaceId);
+    req.set_part_id(partId);
+    auto ret = getPeers(spaceId, partId);
+    if (!ret.ok()) {
+        return ret.status();
+    }
+    folly::Promise<Status> pro;
+    auto f = pro.getFuture();
+    getResponse(ret.value(), 0, req, [] (auto client, auto request) {
+        return client->future_addLearner(request);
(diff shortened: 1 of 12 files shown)


src/meta/processors/admin/BalancePlan.cpp
```diff
@@ -0,0 +1,198 @@
...
+        partTasks[std::make_pair(task.spaceId_, task.partId_)].emplace_back(index++);
+    }
+    buckets_.resize(std::min(tasks_.size(), (size_t)FLAGS_task_concurrency));
+    for (auto it = partTasks.begin(); it != partTasks.end(); it++) {
+        size_t minNum = tasks_.size();
+        int32_t i = 0, minIndex = 0;
+        for (auto& bucket : buckets_) {
+            if (bucket.size() < minNum) {
+                minNum = bucket.size();
+                minIndex = i;
```
wadeliuyi: if no break after set minIndex=i, the minIndex is the largest one that match condition bucket.size() < minNum

src/meta/processors/admin/BalancePlan.cpp
```diff
@@ -0,0 +1,198 @@
...
+            };  // onError
+        }  // for (auto j = 0; j < buckets_[i].size(); j++)
+    }  // for (auto i = 0; i < buckets_.size(); i++)
+
+    for (auto& bucket : buckets_) {
+        if (!bucket.empty()) {
+            tasks_[bucket[0]].invoke();
+        }
+    }
+    saveInStore(true);
```
wadeliuyi: saveInStore(true) will be called in onFilish, this is the secend called, why?

src/meta/processors/admin/BalancePlan.h
```diff
@@ -0,0 +1,95 @@
...
+         * */
+        FAILED             = 0x04,
+    };
+
+    BalancePlan(BalanceID id, kvstore::KVStore* kv, AdminClient* client)
+        : id_(id)
+        , kv_(kv)
+        , client_(client) {}
+
+    void addTask(BalanceTask task) {
```
wadeliuyi: why parameter is not BalanceTask &&task?

src/meta/processors/admin/Balancer.cpp
```diff
@@ -0,0 +1,316 @@
...
+            auto spaceId = MetaServiceUtils::spaceId(iter->key());
+            spaces.push_back(spaceId);
+            iter->next();
+        }
+    }
+    plan_ = std::make_unique<BalancePlan>(time::TimeUtils::nowInSeconds(), kv_, client_.get());
+    for (auto spaceId : spaces) {
+        auto tasks = genTasks(spaceId);
+        for (auto& task : tasks) {
+            plan_->addTask(std::move(task));
```
wadeliuyi: why not define a function “addTasks(vector<task>&&)” in plan?

src/meta/processors/admin/Balancer.cpp
```diff
@@ -0,0 +1,316 @@
...
+    int32_t minLoad = std::floor(avgLoad);
+    int32_t maxLoad = std::ceil(avgLoad);
+    auto hosts = sortedHostsByParts(newHostParts);
+    CHECK_GT(hosts.size(), 1);
+    auto maxPartsHost = hosts.back();
+    auto minPartsHost = hosts.front();
+    auto lastDelta = maxPartsHost.second - minPartsHost.second + 1;
+    while (maxPartsHost.second > maxLoad
+            || minPartsHost.second < minLoad
+            || maxPartsHost.second - minPartsHost.second < lastDelta) {
```
wadeliuyi: why is less than lastDelta? If current delta is 0, this function will process the loop, but will do nothing because of “partsFrom.size() <= partsTo.size() + 1”

src/meta/processors/admin/BalancePlan.cpp
dangleptr: Because we need to update the balancePlan state

src/meta/processors/admin/BalancePlan.h
dangleptr: Because I want to accept const reference and right reference both.

src/meta/processors/admin/Balancer.cpp
dangleptr: A HA.  There are two reasons:
1.  AddTask is designed firstly.
2.  AddTask is more convenient in UTs

src/meta/processors/admin/Balancer.cpp
dangleptr: Because i want the delta be more small after each iteration.  And if noActions happened, the loop will be broken.

src/meta/processors/admin/BalancePlan.cpp
dangleptr: minNum is updated each time.

dangleptr:
Jenkins go

src/meta/test/AdminClientTest.cpp
```diff
@@ -0,0 +1,263 @@
...
+        });
+        baton.wait();
+        auto peersRet = client->getPeers(0, 1);
+        CHECK(peersRet.ok());
+        auto hosts = std::move(peersRet).value();
+        ASSERT_EQ(3, hosts.size());
+        ASSERT_EQ(HostAddr(localIp, sc2->port_), hosts[0]);
+        ASSERT_EQ(HostAddr(localIp, sc1->port_), hosts[1]);
+        ASSERT_EQ(HostAddr(1, 1), hosts[2]);
+    }
```
whitewum: let's try more than 3 hosts.

src/meta/test/AdminClientTest.cpp
```diff
@@ -0,0 +1,263 @@
...
+    {
+        LOG(INFO) << "Test removePart...";
+        folly::Baton<true, std::atomic> baton;
+        client->removePart(0, 0, {localIp, sc->port_}).then([&baton](auto&& st) {
+            CHECK(st.ok());
+            baton.post();
+        });
+        baton.wait();
+    }
```
whitewum: what about add part again?
or let's loop the test a few times

src/meta/processors/admin/AdminClient.cpp
```diff
@@ -5,90 +5,333 @@
...
+        auto resp = std::move(t).value();
+        switch (resp.get_code()) {
+            case storage::cpp2::ErrorCode::SUCCEEDED: {
+                p.setValue(Status::OK());
+                return;
+            }
+            case storage::cpp2::ErrorCode::E_LEADER_CHANGED: {
+                if (retry < retryLimit) {
+                    HostAddr leader(resp.get_leader().get_ip(), resp.get_leader().get_port());
+                    int32_t leaderIndex = 0;
```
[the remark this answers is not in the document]
dangleptr: The code here is to find the index of leader in hosts.

src/meta/processors/admin/AdminClient.cpp
```diff
@@ -5,90 +5,333 @@
...
+                    return;
+                }
+                p.setValue(Status::Error("Unknown code %d", static_cast<int32_t>(resp.get_code())));
+                return;
+            }
+        }
+    });
+}
+
+nebula::cpp2::HostAddr AdminClient::to(const HostAddr& addr) {
```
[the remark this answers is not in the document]
dangleptr: How about toThriftHost ?

src/meta/processors/admin/AdminClient.cpp
```diff
@@ -5,90 +5,333 @@
...
+    nebula::cpp2::HostAddr thriftAddr;
+    thriftAddr.set_ip(addr.first);
+    thriftAddr.set_port(addr.second);
+    return thriftAddr;
+}
+
+StatusOr<std::vector<HostAddr>> AdminClient::getPeers(GraphSpaceID spaceId, PartitionID partId) {
+    CHECK_NOTNULL(kv_);
+    auto partKey = MetaServiceUtils::partKey(spaceId, partId);
```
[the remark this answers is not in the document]
dangleptr: Compiler will optimize it.  The partKey will reuse the returned string's allocation.

src/meta/processors/admin/AdminClient.cpp
```diff
@@ -5,90 +5,333 @@
...
+            std::vector<HostAddr> hosts;
+            hosts.resize(partHosts.size());
+            std::transform(partHosts.begin(), partHosts.end(), hosts.begin(), [](const auto& h) {
+                return HostAddr(h.get_ip(), h.get_port());
+            });
+            return hosts;
+        }
+        case kvstore::ResultCode::ERR_KEY_NOT_FOUND:
+            return Status::Error("Key Not Found");
+        default:
```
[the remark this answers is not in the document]
dangleptr: Good point.

src/meta/processors/admin/BalanceTask.cpp
```diff
@@ -38,7 +38,7 @@ void BalanceTask::invoke() {
         case Status::CHANGE_LEADER: {
             LOG(INFO) << taskIdStr_ << "Ask the src to give up the leadership.";
             SAVE_STATE();
-            client_->transLeader(spaceId_, partId_, src_, dst_).thenValue([this](auto&& resp) {
+            client_->transLeader(spaceId_, partId_, src_).thenValue([this](auto&& resp) {
```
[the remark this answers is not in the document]
dangleptr: Good catch.  We should use MS here.

src/meta/test/AdminClientTest.cpp
dangleptr: AddPart again will be tested on server side.  The UT is just to test client interface.

src/meta/test/BalanceIntegrationTest.cpp
```diff
@@ -0,0 +1,35 @@
...
+#include <folly/synchronization/Baton.h>
+#include "meta/processors/admin/Balancer.h"
+#include "meta/test/TestUtils.h"
+#include "storage/test/TestUtils.h"
+#include "fs/TempDir.h"
+
+namespace nebula {
+namespace meta {
+
+TEST(BalanceIntegrationTest, SimpleTest) {
```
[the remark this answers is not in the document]
dangleptr: Yes.  The BalanceIntegrationTest will go on when server-side is ready.

src/meta/test/AdminClientTest.cpp
dangleptr: Yes. We have more situations to test. Let's open another issue to track it.

dangleptr:
Jenkins go

src/meta/processors/admin/AdminClient.cpp
```diff
@@ -5,90 +5,334 @@
...
+                                           bool asLearner) {
     if (injector_) {
         return injector_->addPart();
     }
-    return Status::OK();
+    storage::cpp2::AddPartReq req;
+    req.set_space_id(spaceId);
+    req.set_part_id(partId);
+    req.set_as_learner(asLearner);
+    return getResponse(host, std::move(req), [] (auto client, auto request) {
```
wadeliuyi: I can not understand why need move(req) here, the req’s memory is in caller stack, and the parameter req of getResponse passed by value?

src/meta/processors/admin/AdminClient.cpp
```diff
@@ -5,90 +5,334 @@
...
+    storage::cpp2::CatchUpDataReq req;
+    req.set_space_id(spaceId);
+    req.set_part_id(partId);
+    auto ret = getPeers(spaceId, partId);
+    if (!ret.ok()) {
+        return ret.status();
+    }
+    folly::Promise<Status> pro;
+    auto f = pro.getFuture();
+    getResponse(ret.value(), 0, req, [] (auto client, auto request) {
```
wadeliuyi: and here why not move(req)?

src/meta/processors/admin/AdminClient.cpp
dangleptr: Maybe req will be changed in the future.  So use move here.

src/meta/processors/admin/AdminClient.cpp
dangleptr: A HA.  Good catch!  We should use move.

src/meta/processors/admin/AdminClient.cpp
```diff
@@ -5,90 +5,334 @@
...
+                if (retry < retryLimit) {
+                    HostAddr leader(resp.get_leader().get_ip(), resp.get_leader().get_port());
+                    int32_t leaderIndex = 0;
+                    for (auto& h : hosts) {
+                        if (h == leader) {
+                            break;
+                        }
+                        leaderIndex++;
+                    }
+                    LOG(INFO) << "Return leder change from " << hosts[index]
```
whitewum: leader

src/meta/processors/admin/AdminClient.cpp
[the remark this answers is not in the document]
whitewum: sounds better

src/meta/test/AdminClientTest.cpp
```diff
@@ -0,0 +1,263 @@
...
+}
+
+TEST(AdminClientTest, RetryTest) {
```
whitewum: Great!

src/meta/test/AdminClientTest.cpp
whitewum: Yep. It's simple for the client,  but a valued corned case.

dangleptr:
> Please consider some corner cases: 1.operations failed at the server side, or 2. client crashed and resume to do some job (what should client do).

Yes,  all interfaces here reentrant functions.  When we implement them on server-side, we should keep the principle

PR #554, opened 2019-06-28
Implement update vertex and edge

zhangguoqing:
Now, we support updating one vertex or edge with constraints， all operations about read-modify-write are atomic.  
The filters and yields expressions in update sentence will be pushed down to storage layer. 

Partial-Issue:  #172


src/graph/UpdateEdgeExecutor.cpp
```diff
@@ -0,0 +1,225 @@
...
+                                           edgeTypeName_->c_str(), propName);
+                    break;
+                }
+                updateItem.field = propName;
+                updateItem.value = Expression::encode(item->value());
+            } else {
+                status = Status::Error("Invalid update item: " + item->toString());
+                break;
+            }
+            updateItems_.emplace_back(updateItem);
```
wadeliuyi: maybe need call std::move and move this line after 91

wadeliuyi: why do not use std::move?

zhangguoqing: thx, done!

src/parser/test/ParserTest.cpp
```diff
@@ -594,28 +607,36 @@ TEST(Parser, InsertEdge) {
 TEST(Parser, UpdateEdge) {
     {
         GQLParser parser;
-        std::string query = "UPDATE EDGE 12345 -> 54321 SET amount=3.14,time=1537408527";
+        std::string query = "UPDATE EDGE 12345 -> 54321 OF transfer "
```
darionyaphet: UPDATE EDGE `edge name` will better ?

src/graph/UpdateEdgeExecutor.cpp
```diff
@@ -0,0 +1,574 @@
...
+    yieldExpCtx_ = std::make_unique<ExpressionContext>();
+    do {
+        status = checkIfGraphSpaceChosen();
+        if (!status.ok()) {
+            break;
+        }
+        insertable_ = sentence_->getInsertable();
+        status = sentence_->getSrcId()->prepare();
+        if (!status.ok()) {
+            return status;
```
wadeliuyi: break

src/graph/UpdateEdgeExecutor.cpp
```diff
@@ -0,0 +1,574 @@
...
+        }
+        auto src = sentence_->getSrcId()->eval();
+        if (!Expression::isInt(src)) {
+            return Status::Error("SRC Vertex ID should be of type integer");
+        }
+        edge_.set_src(Expression::asInt(src));
+
+        status = sentence_->getSrcId()->prepare();
+        if (!status.ok()) {
+            return status;
```
wadeliuyi: ditto

src/graph/UpdateEdgeExecutor.cpp
```diff
@@ -0,0 +1,574 @@
...
+            if (!status.ok()) {
+                status = Status::Error("Invalid update item value expression: "
+                                       + item->toString());
+                break;
+            }
+            updateItem.value = Expression::encode(item->value());
+            updateItems_.emplace_back(std::move(updateItem));
+        }
+    }
+    return status;
```
wadeliuyi: if updateList == nullptr, maybe should return some error?

src/graph/UpdateEdgeExecutor.cpp
```diff
@@ -0,0 +1,574 @@
...
+    return Status::OK();
+}
+
+
+Status UpdateEdgeExecutor::prepareNeededProps() {
+    auto status = Status::OK();
+    do {
+        if (updateItems_.empty()) {
+            status = Status::Error("There must be some correct update items");
+            break;
```
wadeliuyi: maybe should move the check to each function of prepare, if check failed, the process will not do the next step.

src/graph/UpdateEdgeExecutor.cpp
```diff
@@ -0,0 +1,574 @@
...
+        }
+        for (auto *col : yields_) {
+            col->expr()->setContext(yieldExpCtx_.get());
+            status = col->expr()->prepare();
+            if (!status.ok()) {
+                break;
+            }
+            col->expr()->setContext(expCtx_.get());
+            status = col->expr()->prepare();
```
wadeliuyi: no check the status?

src/graph/UpdateEdgeExecutor.cpp
```diff
@@ -0,0 +1,574 @@
...
+    auto future = ectx()->storage()->getVertexProps(spaceId,
+                                                    std::move(vertices),
+                                                    std::move(returns));
+    auto *runner = ectx()->rctx()->runner();
+    auto cb = [this, &baton] (auto &&result) {
+        auto completeness = result.completeness();
+        if (completeness != 100) {
+            auto msg = "Internal error when fetch dst vertex props.";
+            LOG(ERROR) << msg;
+            onError_(Status::Error(msg));
```
wadeliuyi: miss baton.post?

src/graph/UpdateEdgeExecutor.cpp
```diff
@@ -0,0 +1,574 @@
...
+            this->vertexHolder_ = std::make_unique<VertexHolder>();
+        }
+        for (auto& resp : result.responses()) {
+            this->vertexHolder_->add(resp);
+        }
+        baton.post();
+    };
+    auto error = [this] (auto &&e) {
+        LOG(ERROR) << "Exception caught: " << e.what();
+        DCHECK(onError_);
```
wadeliuyi: ditto

src/graph/UpdateEdgeExecutor.cpp
```diff
@@ -0,0 +1,574 @@
...
+        onError_(Status::Error("Internal error: getVertexProps!"));
```
wadeliuyi: maybe here can not call onError_, because onError_ will free the point "this", so if this thread wake up from baton.wait, the memory of "this" has been free.

src/graph/UpdateVertexExecutor.cpp
```diff
@@ -0,0 +1,301 @@
...
+    }
+    return Status::OK();
+}
+
+
+Status UpdateVertexExecutor::prepareNeededProps() {
+    auto status = Status::OK();
+    do {
+        if (updateItems_.empty()) {
+            status = Status::Error("There must be some correct update items!");
```
wadeliuyi: maybe move the check to each prepare function is better.

src/graph/UpdateVertexExecutor.cpp
```diff
@@ -0,0 +1,301 @@
...
+            onError_(Status::Error("Internal Error"));
+            return;
+        }
+        this->finishExecution(std::move(resp));
+    };
+    auto error = [this] (auto &&e) {
+        LOG(ERROR) << "Exception caught: " << e.what();
+        DCHECK(onError_);
+        onError_(Status::Error("Internal error"));
+        return;
```
wadeliuyi: no need return.

src/storage/client/StorageClient.h
```diff
@@ -127,6 +127,25 @@ class StorageClient {
         std::vector<storage::cpp2::PropDef> returnCols,
         folly::EventBase* evb = nullptr);
 
+    folly::SemiFuture<StorageRpcResponse<storage::cpp2::UpdateResponse>> updateVertex(
+        GraphSpaceID space,
+        std::vector<VertexID> vertices,
```
wadeliuyi: use &&, then there is no call rvalue construction function.

src/storage/client/StorageClient.h
```diff
@@ -127,6 +127,25 @@ class StorageClient {
...
+        std::string filter,
+        std::vector<storage::cpp2::UpdateItem> updateItems,
+        std::vector<storage::cpp2::PropDef> returnCols,
+        bool insertable,
+        folly::EventBase* evb = nullptr);
+
+    folly::SemiFuture<StorageRpcResponse<storage::cpp2::UpdateResponse>> updateEdge(
+        GraphSpaceID space,
+        std::vector<storage::cpp2::EdgeKey> edges,
```
wadeliuyi: ditto

src/graph/UpdateEdgeExecutor.cpp
```diff
@@ -0,0 +1,574 @@
...
+            // Very bad, it should delete the upsert positive edge!!!
+            DCHECK(onError_);
+            onError_(Status::Error("Internal error about insert reversely edge."));
+            return;
+        }
+    };
+    auto error = [this] (auto &&e) {
+        LOG(ERROR) << "Exception caught: " << e.what();
+        DCHECK(onError_);
+        onError_(Status::Error("Internal error: addEdges"));
```
wadeliuyi: call onError_ in other thread will free "this" pointer, and current thread will get a UAF error.

src/graph/UpdateEdgeExecutor.cpp
```diff
@@ -0,0 +1,574 @@
...
+}
+
+
+void UpdateEdgeExecutor::fetchDstVertexProps() {
+    auto spaceId = ectx()->rctx()->session()->space();
+    auto statusOr = getDstProps();
+    if (!statusOr.ok()) {
+        auto msg = "Internal error about getDstProps.";
+        LOG(ERROR) << msg;
+        onError_(Status::Error(msg));
```
wadeliuyi: take care of onError_

src/graph/UpdateVertexExecutor.cpp
```diff
@@ -0,0 +1,301 @@
...
+    yieldExpCtx_ = std::make_unique<ExpressionContext>();
+    do {
+        status = checkIfGraphSpaceChosen();
+        if (!status.ok()) {
+            break;
+        }
+        insertable_ = sentence_->getInsertable();
+        status = sentence_->getVid()->prepare();
+        if (!status.ok()) {
+            return status;
```
wadeliuyi: break

src/storage/UpdateEdgeProcessor.cpp
```diff
@@ -0,0 +1,382 @@
...
+    {
+        folly::Baton<true, std::atomic> baton;
+        this->kvstore_->asyncMultiPut(spaceId_, partId, std::move(data),
+                [&baton, &ret, partId, this] (kvstore::ResultCode code) {
+            VLOG(3) << "spaceId: " << this->spaceId_ << ", partId: "
+                    << partId << ", code:" << static_cast<int32_t>(code);
+            ret = code;
+            baton.post();
+        });
+        baton.wait();
```
wadeliuyi: maybe here can not wait directly, because asyncMultiPut will run raft, and it will take a long time, the wait will let a worker thread sleep, it will cut down the throughput of system

src/storage/UpdateEdgeProcessor.cpp
```diff
@@ -0,0 +1,382 @@
...
+        return it->second;
+    };
+    getters.getEdgeProp = [&, this] (const std::string &prop) -> VariantType {
+        auto res = RowReader::getPropByName(this->filterReader_.get(), prop);
+        CHECK(ok(res));
+        auto val = value(std::move(res));
+        VLOG(1) << "Hit edgeProp for edge prop " << prop << ", value " << val;
+        return val;
+    };
+    if (this->exp_ != nullptr) {
```
wadeliuyi: maybe filter_ is more readable.

src/storage/UpdateEdgeProcessor.cpp
```diff
@@ -0,0 +1,382 @@
...
+                    break;
+                }
+                record[prop.retIndex_] = value(std::move(res));
+                index++;
+            }
+        }
+        for (auto& tc : tagContexts_) {
+            for (auto& prop : tc.props_) {
+                if (prop.returned_) {
+                    auto res = RowReader::getPropByIndex(reader.get(), index);
```
wadeliuyi: maybe here can not reuse index if break from line 157.

src/graph/UpdateEdgeExecutor.cpp
zhangguoqing: done!

src/graph/UpdateEdgeExecutor.cpp
zhangguoqing: done!

src/graph/UpdateEdgeExecutor.cpp
zhangguoqing: The updateList always isn't nullptr. 
`auto items = sentence_->updateList()->items();`

src/graph/UpdateEdgeExecutor.cpp
zhangguoqing: done!

src/graph/UpdateEdgeExecutor.cpp
zhangguoqing: done, check again！

src/graph/UpdateEdgeExecutor.cpp
zhangguoqing: done!

src/graph/UpdateEdgeExecutor.cpp
zhangguoqing: done!

src/graph/UpdateEdgeExecutor.cpp
zhangguoqing: fixed!

src/graph/UpdateVertexExecutor.cpp
zhangguoqing: done!

src/graph/UpdateVertexExecutor.cpp
zhangguoqing: done!

src/graph/UpdateEdgeExecutor.cpp
zhangguoqing: done!

src/graph/UpdateEdgeExecutor.cpp
zhangguoqing: done！

src/graph/UpdateVertexExecutor.cpp
zhangguoqing: done！

src/storage/UpdateEdgeProcessor.cpp
zhangguoqing: return kvstore::ResultCode::ERR_FALSE_UPDATER;

src/storage/UpdateEdgeProcessor.cpp
zhangguoqing: filterReader_  rename to  filter_?

src/storage/UpdateEdgeProcessor.cpp
zhangguoqing: If disable the baton.wait(), the test case will failure when checking the kvstore, because the updated data has not written to the kvstore.

src/storage/UpdateEdgeProcessor.cpp
wadeliuyi: sorry, I mean exp_ rename to filter_

src/storage/UpdateEdgeProcessor.cpp
wadeliuyi: here can get a future and add a lambda to worker like function "RaftPart::replicateLogs":
get_future(xxx).then(worker, []{check the kvstore});

src/graph/UpdateVertexExecutor.cpp
wadeliuyi: delete return status

src/storage/UpdateVertexProcessor.cpp
```diff
@@ -0,0 +1,364 @@
...
+                       props,
+                       fcontext,
+                       collector);
+    return ret;
+}
+
+
+kvstore::ResultCode UpdateVertexProcessor::processVertex(PartitionID partId,
+                                                         VertexID vId) {
+    tagReader_.clear();
```
wadeliuyi: this function is too long, maybe should split it.

src/storage/UpdateEdgeProcessor.cpp
zhangguoqing: The `exp_` is the member of QueryBaseProcessor rather than UpdateEdgeProcessor. I try not to modify other code.

src/graph/UpdateEdgeExecutor.cpp
```diff
@@ -0,0 +1,616 @@
...
+    };
+    std::move(future).via(runner).thenValue(cb).thenError(error);
+}
+
+
+VariantType UpdateEdgeExecutor::VertexHolder::get(const VertexID id,
+                                                  const int64_t index) const {
+    auto iter = data_.find(id);
+
+    // TODO(dutor) We need a type to represent NULL or non-existing prop
```
wadeliuyi: why is todo dutor?

src/storage/UpdateVertexProcessor.cpp
```diff
@@ -0,0 +1,366 @@
...
+                            auto log = kvstore::encodeMultiValues(kvstore::OP_MULTI_PUT, data);
+                            return log;
+                        }
+                        return std::string("");
+                    },
+                    [&] (kvstore::ResultCode code) {
+                        updateRet = code;
+                        baton.post();
+                    });
+            baton.wait();
```
dangleptr: Don't wait here for the response,  it will hurt the performance

src/interface/storage.thrift
```diff
@@ -203,6 +205,44 @@ struct CatchUpDataReq {
...
+    1: bool upsert,
+    2: common.VertexID vertex_id,       // Only valid when update vertex
+    3: EdgeKey edge_key,                // Only valid when update edge
+    4: binary data,
+}
+
+struct UpdateResponse {
+    1: required ResponseCommon result,
+    2: optional common.Schema schema,   // return column related props
+    3: optional list<UpdateRespData> return_data,
```
dangleptr: Why return a list of the response?

Do we support update batch of vertex in one sentence?

src/interface/storage.thrift
```diff
@@ -203,6 +205,44 @@ struct CatchUpDataReq {
...
+struct UpdateItem {
+    1: required binary name,    // Tag or EdgeType schema name
+    2: required binary field,   // property name of scheam
+    3: required binary value,   // expression which is encoded
+}
+
+struct UpdateVertexRequest {
+    1: common.GraphSpaceID space_id,
+    2: map<common.PartitionID, list<common.VertexID>>(cpp.template = "std::unordered_map") parts,
+    3: common.EdgeType edge_type,
```
dangleptr: What's the purpose about this edge_type?

src/interface/storage.thrift
```diff
@@ -203,6 +205,44 @@ struct CatchUpDataReq {
...
+
+struct UpdateItem {
+    1: required binary name,    // Tag or EdgeType schema name
+    2: required binary field,   // property name of scheam
+    3: required binary value,   // expression which is encoded
+}
+
+struct UpdateVertexRequest {
+    1: common.GraphSpaceID space_id,
+    2: map<common.PartitionID, list<common.VertexID>>(cpp.template = "std::unordered_map") parts,
```
dangleptr: I am not sure how to support update batch of vertex in one update sentence

src/graph/UpdateEdgeExecutor.cpp
zhangguoqing: fixed!

src/interface/storage.thrift
zhangguoqing: yep, fixed!

src/interface/storage.thrift
zhangguoqing: Want to call the QueryBaseProcessor<REQ, RESP>::checkAndBuildContexts.

src/interface/storage.thrift
zhangguoqing: It already changed to a single update!

src/storage/UpdateVertexProcessor.cpp
zhangguoqing: OK, recheck it now please!

src/storage/QueryBaseProcessor.inl
```diff
@@ -207,8 +210,7 @@ bool QueryBaseProcessor<REQ, RESP>::checkExp(const Expression* exp) {
         case Expression::kEdgeType: {
             return true;
         }
-        case Expression::kAliasProp:
-        case Expression::kEdgeProp: {
```
dangleptr: Why remove kEdgeProp?

src/storage/UpdateEdgeProcessor.cpp
```diff
@@ -0,0 +1,373 @@
...
+        return ret;
+    }
+    // Only use the latest version.
+    if (iter && iter->valid()) {
+        key_ = iter->key().toString();
+        edgeReader_ = RowReader::getEdgePropReader(schemaMan_,
+                                                   iter->val(),
+                                                   spaceId_,
+                                                   edgeKey.edge_type);
+        filterReader_ = RowReader::getEdgePropReader(schemaMan_,
```
dangleptr: Why we have two readers?

src/storage/UpdateEdgeProcessor.cpp
```diff
@@ -0,0 +1,373 @@
...
+    if (retCode != cpp2::ErrorCode::SUCCEEDED) {
+        LOG(ERROR) << "Failure build contexts!";
+        this->pushResultCode(retCode, partId);
+        this->onFinished();
+        return;
+    }
+    updateItems_ = std::move(req.get_update_items());
+
+    CHECK_NOTNULL(kvstore_);
+    auto edgeKey = req.get_edge_key();
```
dangleptr: std::move(req).get_edge_key()

src/storage/UpdateEdgeProcessor.cpp
```diff
@@ -0,0 +1,373 @@
...
+         cpp2::UpdateResponse>::checkAndBuildContexts(req);
+    if (retCode != cpp2::ErrorCode::SUCCEEDED) {
+        return retCode;
+    }
+
+    // build context of the update items
+    if (this->expCtx_ == nullptr) {
+        this->expCtx_ = std::make_unique<ExpressionContext>();
+    }
+    for (auto& item : req.get_update_items()) {
```
dangleptr: What's purpose about this code?   I can't understand it

src/storage/UpdateEdgeProcessor.cpp
```diff
@@ -0,0 +1,373 @@
...
+    VLOG(3) << "Process spaceId: " << this->spaceId_
+            << ", partId:  " << partId
+            << ", src: " << edgeKey.get_src()
+            << ", edge_type: " << edgeKey.get_edge_type()
+            << ", dst: " << edgeKey.get_dst()
+            << ", ranking: " << edgeKey.get_ranking();
+    kvstore::ResultCode ret = kvstore::ResultCode::SUCCEEDED;
+    this->kvstore_->asyncAtomicOp(spaceId_, partId,
+        [&, this] () -> std::string {
+            ret = this->processEdge(partId, edgeKey);
```
dangleptr: Use another name for "processEdge/processData",   i don't know what are you doing inside. 

And for update,   the code structure should look like 

```
auto edgeValue = getEdge()
if (checkFilter()) {
      updateAndWriteBackEdge()
      return true;
} else {
      return false;
}
```

src/storage/QueryBaseProcessor.inl
zhangguoqing: There is already no EdgePropertyExpression yet.  

FYI. [PR669](https://github.com/vesoft-inc/nebula/pull/669)

src/storage/UpdateEdgeProcessor.cpp
zhangguoqing: The edgeReader_ is used to transfer to updater which is std::move. The filterReader_ is used to get the value of Expression.

src/storage/UpdateEdgeProcessor.cpp
zhangguoqing: There maybe some expression in the value of update items. Here is the context of getting the relevant expression.

src/storage/UpdateEdgeProcessor.cpp
zhangguoqing: ok, i will update it.

src/storage/UpdateEdgeProcessor.cpp
zhangguoqing: ok, I will update it as you suggested.

src/interface/storage.thrift
```diff
@@ -203,6 +205,40 @@ struct CatchUpDataReq {
     3: common.HostAddr     target,
 }
 
+struct UpdateResponse {
+    1: required ResponseCommon result,
+    2: optional common.Schema schema,   // return column related props schema
+    3: optional binary data,            // return column related props value
+    4: optional bool upsert = false,    // it's true when need to be inserted by UPSERT
```
dangleptr: What' the purpose about this field in response

src/interface/storage.thrift
```diff
@@ -203,6 +205,40 @@ struct CatchUpDataReq {
...
+
+struct UpdateVertexRequest {
+    1: common.GraphSpaceID space_id,
+    2: common.VertexID vertex_id,
+    3: common.PartitionID part_id,
+    4: binary filter,
+    5: list<UpdateItem> update_items,
+    6: list<PropDef> return_columns,
+    7: bool insertable,
+    8: common.EdgeType edge_type,
```
dangleptr: Why we need edge_type in UpdateVertex request?

src/interface/storage.thrift
```diff
@@ -203,6 +205,40 @@ struct CatchUpDataReq {
...
+
+struct UpdateEdgeRequest {
+    1: common.GraphSpaceID space_id,
+    2: EdgeKey edge_key,
+    3: common.PartitionID part_id,
+    4: binary filter,
+    5: list<UpdateItem> update_items,
+    6: list<PropDef> return_columns,
+    7: bool insertable,
+    8: common.EdgeType edge_type,
```
dangleptr: We already have edge_key,  why we need edge_type here.

src/storage/UpdateEdgeProcessor.cpp
dangleptr: Great

src/parser/parser.yy
```diff
@@ -46,6 +46,7 @@ class GraphScanner;
     nebula::VertexIDList                   *vid_list;
     nebula::OverClause                     *over_clause;
     nebula::WhereClause                    *where_clause;
+    nebula::WhereClause                    *when_clause;
```
dangleptr: WhenClause

src/interface/storage.thrift
dangleptr: Don't add an extra field just due to some internal call method.   You could Implement your own checkAndBuild.

src/storage/QueryBaseProcessor.inl
dangleptr: Good

src/parser/test/ParserTest.cpp
```diff
@@ -626,28 +639,36 @@ TEST(Parser, InsertEdge) {
...
+                            "WHEN transfer.amount > 3.14 && $$.person.name == \"dutor\"";
         auto result = parser.parse(query);
         ASSERT_TRUE(result.ok()) << result.status();
     }
     {
         GQLParser parser;
-        std::string query = "UPDATE EDGE 12345 -> 54321 SET amount=3.14,time=1537408527 "
-                            "WHERE amount > 3.14 YIELD amount,time";
+        std::string query = "UPDATE EDGE 12345 -> 54321 OF transfer "
+                            "SET amount = 3.14 + $^.job.salary, time = 1537408527 "
```
dangleptr: Very good feature.

src/storage/UpdateEdgeProcessor.cpp
dangleptr: Could we unify them

src/storage/UpdateEdgeProcessor.cpp
```diff
@@ -0,0 +1,373 @@
...
+                                       tc.tagId_,
+                                       tc.props_,
+                                       &fcontext,
+                                       &collector);
+        if (ret != kvstore::ResultCode::SUCCEEDED) {
+            return ret;
+        }
+    }
+
+    // the Where clause for filting
```
dangleptr: The filter is not used for where clause?

src/storage/UpdateEdgeProcessor.cpp
```diff
@@ -0,0 +1,373 @@
...
+                if (prop.prop_.name == field && prop.returned_) {
+                    retIndex = prop.retIndex_;
+                    break;
+                }
+            }
+        }
+        switch (expValue.which()) {
+            case VAR_INT64: {
+                auto v = boost::get<int64_t>(expValue);
+                updater_->setInt(field, v);
```
dangleptr: What if the filed is not int type?  how to deal with it

src/storage/UpdateEdgeProcessor.cpp
```diff
@@ -0,0 +1,373 @@
...
+
+    // the Set clause for updating
+    SchemaVer ver = edgeReader_->schemaVer();
+    const auto constSchema = this->schemaMan_->getEdgeSchema(
+            spaceId_, this->edgeContext_.edgeType_, ver);
+    auto schema = std::const_pointer_cast<meta::SchemaProviderIf>(constSchema);
+    // TODO(zhangguoqing) When the edge is to be inserted, it can't instantiate
+    // the class RowUpdater() with edgeReader_ parameter, otherwise the updater_
+    // doing encoded will report the error: Failed to encode updated data.
+    if (resp_.get_upsert()) {
```
dangleptr: I can't understand this code.   Why we check field in response?   And i think we don't need to support upsert in this processor,  we could do it in another one.

src/interface/storage.thrift
```diff
@@ -203,6 +205,40 @@ struct CatchUpDataReq {
...
 
+struct UpdateResponse {
+    1: required ResponseCommon result,
+    2: optional common.Schema schema,   // return column related props schema
+    3: optional binary data,            // return column related props value
+    4: optional bool upsert = false,    // it's true when need to be inserted by UPSERT
+}
+
+struct UpdateItem {
+    1: required binary name,    // TagName or EdgeType
```
dangleptr: I am not sure how to represent "update vertex id  tag1.prop1 = tag2.prop2 + tag3.prop3" with your UpdateItem

src/graph/UpdateEdgeExecutor.cpp
```diff
@@ -0,0 +1,426 @@
...
+    DCHECK(sentence_ != nullptr);
+    Status status = Status::OK();
+    expCtx_ = std::make_unique<ExpressionContext>();
+    do {
+        status = checkIfGraphSpaceChosen();
+        if (!status.ok()) {
+            break;
+        }
+        insertable_ = sentence_->getInsertable();
+        status = sentence_->getSrcId()->prepare();
```
dangleptr: Why we need an expression for srcId?

src/graph/UpdateEdgeExecutor.cpp
```diff
@@ -0,0 +1,426 @@
...
+            break;
+        }
+        auto src = sentence_->getSrcId()->eval();
+        if (src.ok() && !Expression::isInt(src.value())) {
+            status = Status::Error("SRC Vertex ID should be of type integer");
+            break;
+        }
+        edge_.set_src(Expression::asInt(src.value()));
+
+        status = sentence_->getDstId()->prepare();
```
dangleptr: Ditto

src/graph/UpdateEdgeExecutor.cpp
```diff
@@ -0,0 +1,426 @@
...
+        yields_ = clause->columns();
+    }
+    return Status::OK();
+}
+
+
+Status UpdateEdgeExecutor::prepareNeededProps() {
+    auto status = Status::OK();
+    do {
+        if (filter_ != nullptr) {
```
dangleptr: Why we need filter again in QueryEngine,   you have push it to storage service

src/graph/UpdateEdgeExecutor.cpp
```diff
@@ -0,0 +1,426 @@
...
+            status = filter_->prepare();
+            if (!status.ok()) {
+                status = Status::Error("Invalid when clause: " + filter_->toString());
+                break;
+            }
+        }
+        if (yields_.empty()) {
+            break;
+        }
+        for (auto *col : yields_) {
```
dangleptr: Ditto.  The yield expressions should be calculated on storage service.  We don't need this code

src/graph/UpdateEdgeExecutor.cpp
```diff
@@ -0,0 +1,426 @@
...
+            col->expr()->setContext(expCtx_.get());
+            status = col->expr()->prepare();
+            if (!status.ok()) {
+                break;
+            }
+        }
+    } while (false);
+
+    if (expCtx_->hasDstTagProp()) {
```
dangleptr: What's this?

src/graph/UpdateEdgeExecutor.cpp
```diff
@@ -0,0 +1,426 @@
...
+            }
+            auto res = RowReader::getPropByIndex(reader.get(), index);
+            if (ok(res)) {
+                return value(std::move(res));
+            }
+            return Status::Error(folly::sformat("{}.{} was not exist", tagName, prop));
+        };
+        std::vector<VariantType> record;
+        record.reserve(yields_.size());
+        for (auto *column : yields_) {
```
dangleptr: We don't need to execute the expression again.

src/graph/UpdateEdgeExecutor.cpp
```diff
@@ -0,0 +1,426 @@
...
+    for (auto *col : yields_) {
+        if (col->alias() == nullptr) {
+            columnNames.emplace_back(col->expr()->toString());
+        } else {
+            columnNames.emplace_back(*col->alias());
+        }
+    }
+    resp_->set_column_names(std::move(columnNames));
+
+    std::vector<cpp2::RowValue> rows;
```
dangleptr: Why we need a vector here?

src/graph/UpdateEdgeExecutor.h
```diff
@@ -0,0 +1,95 @@
...
+    void insertReverselyEdge();
+
+    // All required data have arrived, finish the execution.
+    void finishExecution(storage::cpp2::UpdateResponse &&rpcResp);
+
+    /**
+     * A container to hold the mapping from vertex id to its properties, used for lookups
+     * during the final evaluation process.
+     */
+    class VertexHolder final {
```
dangleptr: Why we need VertexHolder?  There is only one vertex or edge

src/interface/storage.thrift
zhangguoqing: In storage, both UpdateVertexProcessor and UpdateEdgeProcessor call the function QueryBaseProcessor::checkAndBuildContexts first, in their own checkAndBuildContexts functions.  The QueryBaseProcessor::checkAndBuildContexts  should use the req.edge_type.

FYI. https://github.com/vesoft-inc/nebula/blob/master/src/storage/QueryBaseProcessor.inl#L39

src/interface/storage.thrift
zhangguoqing: ditto!

src/interface/storage.thrift
dangleptr: Don't design your interface as implementation internal.

src/interface/storage.thrift
zhangguoqing: If insert the edge in UPSERT sentence, the response should be notified, then send the request for inserting the reversely edge.

src/parser/parser.yy
zhangguoqing: Do we want to achieve WhenCluse? I think it is the same as where.

src/storage/UpdateEdgeProcessor.cpp
zhangguoqing: I have no idea to unify them, do you have any advice?

src/storage/UpdateEdgeProcessor.cpp
zhangguoqing: I think, they are almost always the same, and it will very redundant when separated.

src/storage/UpdateEdgeProcessor.cpp
zhangguoqing: The UpdateEdgeExecutor::prepareSet has checked the field in it's schema in graph layer.

src/storage/UpdateEdgeProcessor.cpp
zhangguoqing: Here is  checkFilter(), rather than traversing filter.

src/interface/storage.thrift
zhangguoqing: yep, done!

src/storage/UpdateEdgeProcessor.cpp
zhangguoqing: It has been refactored.

src/storage/UpdateEdgeProcessor.cpp
zhangguoqing: done!

src/interface/storage.thrift
zhangguoqing: It has been refactored which remove edge_type away.

src/storage/UpdateEdgeProcessor.cpp
zhangguoqing: It has put into the storage layer.

src/storage/UpdateEdgeProcessor.cpp
zhangguoqing: It has been refactored.

src/graph/UpdateEdgeExecutor.cpp
zhangguoqing: I have make it  to the INTEGER.

src/graph/UpdateEdgeExecutor.cpp
zhangguoqing: yep, done!

src/graph/UpdateEdgeExecutor.cpp
zhangguoqing: You are right, execute  all of them at the storage service now!

src/graph/UpdateEdgeExecutor.h
zhangguoqing: Wrong code, it has been removed!

src/graph/UpdateEdgeExecutor.cpp
zhangguoqing: It always a vector when return.

src/interface/storage.thrift
zhangguoqing: name = tag1,  prop = prop1, value = Expression::encode(tag2.prop2 + tag3.prop3)

src/graph/UpdateEdgeExecutor.cpp
```diff
@@ -0,0 +1,239 @@
...
+Status UpdateEdgeExecutor::prepareYield() {
+    auto *clause = sentence_->yieldClause();
+    if (clause != nullptr) {
+        yields_ = clause->columns();
+    }
+    return Status::OK();
+}
+
+
+std::vector<std::string> UpdateEdgeExecutor::getReturnProps() {
```
dangleptr: rename it to be getReturnColumns

src/graph/UpdateEdgeExecutor.cpp
```diff
@@ -0,0 +1,239 @@
...
+    std::vector<std::string> returnColumns;
+    for (auto *col : yields_) {
+        auto column = Expression::encode(col->expr());
+        returnColumns.emplace_back(column);
```
dangleptr: std::move(column)

src/graph/UpdateEdgeExecutor.cpp
zhangguoqing: Check if the DestPropertyExpression is in all expressions.  It has been put into the storage service.

src/parser/parser.yy
dangleptr: To avoid misunderstanding,  we'd better use WhenClause,  of course it's implementation could be the same  with WhereClause.

src/parser/test/ParserTest.cpp
zhangguoqing: I can't understand your idea, can you elaborate on it? thx.

src/storage/client/StorageClient.h
zhangguoqing: Now it only one vertex.

src/storage/client/StorageClient.h
zhangguoqing: Now it only one an EdgeKey.

src/graph/UpdateEdgeExecutor.cpp
```diff
@@ -0,0 +1,239 @@
...
+    }
+    return std::move(returnColumns);
```
dangleptr: We don't need "move" here.

src/parser/MutateSentences.h
```diff
@@ -329,75 +351,142 @@ class UpdateList final {
...
+        kind_ = Kind::kUpdateVertex;
+    }
+
     void setInsertable(bool insertable) {
         insertable_ = insertable;
     }
 
-    void setVid(Expression *vid) {
-        vid_.reset(vid);
+    const bool getInsertable() const {
```
dangleptr: We don't need "const" specifier for "bool"

src/storage/UpdateVertexProcessor.cpp
```diff
@@ -0,0 +1,360 @@
...
+               }
+               default: {
+                   LOG(FATAL) << "Unknown VariantType: " << v.which();
+                   return;
+               }
+           }
+           respScheam.columns.emplace_back(std::move(column));
+        }
+        resp_.set_schema(std::move(respScheam));
+        resp_.set_data(std::move(writer.encode()));
```
dangleptr: we don't need move here

src/storage/UpdateVertexProcessor.cpp
```diff
@@ -0,0 +1,360 @@
...
+        return ret;
+    }
+
+    if (iter && iter->valid()) {
+        auto reader = RowReader::getTagPropReader(this->schemaMan_,
+                                                  iter->val(),
+                                                  this->spaceId_,
+                                                  tagId);
+        const auto constSchema = reader->getSchema();
+        for (auto prop : props) {
```
dangleptr: auto&

src/storage/UpdateVertexProcessor.cpp
```diff
@@ -0,0 +1,360 @@
...
+            auto index =  constSchema->getFieldIndex(prop.prop_.name);
+            auto res = RowReader::getPropByIndex(reader.get(), index);
```
dangleptr: You could use RowReader::getPropByName directly.

src/storage/UpdateVertexProcessor.cpp
```diff
@@ -0,0 +1,360 @@
...
+                return kvstore::ResultCode::ERR_UNKNOWN;
+            }
+            auto&& v = value(std::move(res));
+            tagFilters_.emplace(std::make_pair(tagId, prop.prop_.name), v);
+        }
+        if (updateTagIds_.find(tagId) != updateTagIds_.end()) {
+            auto schema = std::const_pointer_cast<meta::SchemaProviderIf>(constSchema);
+            auto updater =
+                std::unique_ptr<RowUpdater>(new RowUpdater(std::move(reader), schema));
+            pendingUpdater_[tagId] = std::make_unique<KeyUpdaterPair>();
```
dangleptr: pendingUpdater_[tagId] will do search tagId inside pendingUpdater_. 

You should use auto it = updateTagIds_.find(tagId),  then use the it directly here.

src/storage/UpdateVertexProcessor.cpp
```diff
@@ -0,0 +1,360 @@
...
+            auto res = RowReader::getPropByIndex(reader.get(), index);
+            if (!ok(res)) {
+                VLOG(1) << "Skip the bad value for tag: " << tagId
+                        << ", prop " << prop.prop_.name;
+                return kvstore::ResultCode::ERR_UNKNOWN;
+            }
+            auto&& v = value(std::move(res));
+            tagFilters_.emplace(std::make_pair(tagId, prop.prop_.name), v);
+        }
+        if (updateTagIds_.find(tagId) != updateTagIds_.end()) {
```
dangleptr: What's the updateTagIds_ used for?

src/storage/client/StorageClient.inl
```diff
@@ -158,6 +157,49 @@ folly::SemiFuture<StorageRpcResponse<Response>> StorageClient::collectResponse(
     return context->promise.getSemiFuture();
 }
 
+
+template<class Request, class RemoteFunc, class Response>
+folly::Future<StatusOr<Response>> StorageClient::getResponse(
```
dangleptr: I noticed the function has been supported in #868

src/storage/UpdateEdgeProcessor.cpp
```diff
@@ -0,0 +1,398 @@
...
+    auto edgeKey = req.get_edge_key();
+    this->edgeContext_.edgeType_ = edgeKey.get_edge_type();
+    auto retCode = checkAndBuildContexts(req);
+    if (retCode != cpp2::ErrorCode::SUCCEEDED) {
+        LOG(ERROR) << "Failure build contexts!";
+        this->pushResultCode(retCode, partId);
+        this->onFinished();
+        return;
+    }
+    updateItems_ = std::move(req.get_update_items());
```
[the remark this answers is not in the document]
zhangguoqing: done!

src/graph/UpdateEdgeExecutor.cpp
zhangguoqing: done!

src/graph/UpdateEdgeExecutor.cpp
zhangguoqing: done!

src/storage/client/StorageClient.inl
zhangguoqing: He copy mine, I will rebase master. :)

src/graph/UpdateEdgeExecutor.cpp
zhangguoqing: done!

src/parser/MutateSentences.h
zhangguoqing: done!

src/storage/UpdateVertexProcessor.cpp
zhangguoqing: done!

src/storage/UpdateVertexProcessor.cpp
zhangguoqing: yep!

src/storage/UpdateVertexProcessor.cpp
zhangguoqing: done!

src/parser/parser.yy
zhangguoqing: OK, how about  adding the line `using WhenClause = WhereClause;` in src/parser/Clauses.h ?

src/storage/UpdateVertexProcessor.cpp
```diff
@@ -0,0 +1,366 @@
...
+                    v = folly::to<bool>(false);
+                    break;
+                default:
+                    LOG(FATAL) << "Unimplemented";
+                    return kvstore::ResultCode::ERR_UNKNOWN;
+            }
+            tagFilters_.emplace(std::make_pair(tagId, propName), v);
+        }
+        auto schema = std::const_pointer_cast<meta::SchemaProviderIf>(constSchema);
+        auto updater = std::unique_ptr<RowUpdater>(new RowUpdater(schema));
```
critical27: Why not use a RowWriter here? Is this because keep a same interface?

zhangguoqing: Because there are maybe some fields to be updated in the `updateAndWriteBack()` function.  RowUpdate also can to be encoded to string.

src/interface/storage.thrift
dangleptr: make sense

src/parser/Clauses.h
```diff
@@ -151,6 +151,7 @@ class WhereClause final {
     std::unique_ptr<Expression>                 filter_;
 };
 
+using WhenClause = WhereClause;
```
dangleptr: Well done.

src/parser/scanner.lex
```diff
@@ -27,7 +27,6 @@ static constexpr size_t MAX_STRING = 4096;
 GO                          ([Gg][Oo])
 AS                          ([Aa][Ss])
 TO                          ([Tt][Oo])
-OR                          ([Oo][Rr])
```
dangleptr: Why remove it?

src/storage/UpdateVertexProcessor.cpp
```diff
@@ -0,0 +1,366 @@
...
+            auto&& v = value(std::move(res));
+            tagFilters_.emplace(std::make_pair(tagId, prop.prop_.name), v);
+        }
+        if (updateTagIds_.find(tagId) != updateTagIds_.end()) {
+            auto schema = std::const_pointer_cast<meta::SchemaProviderIf>(constSchema);
+            auto updater =
+                std::unique_ptr<RowUpdater>(new RowUpdater(std::move(reader), schema));
+            pendingUpdater_[tagId] = std::make_unique<KeyUpdaterPair>();
+            auto it = pendingUpdater_.find(tagId);
+            it->second->key = iter->key().toString();
```
dangleptr: write the code in this way. 
```
auto& tagUpdater = pendingUpdater_[tagId];
tagUpdater->key = iter->key().toString();
tagUpdater->updater = std::move(updater);
```


BTW.  pendingUpdater_ is confused,  'tagUpdaters_' maybe a better name.

src/storage/UpdateVertexProcessor.cpp
```diff
@@ -0,0 +1,366 @@
...
+                    break;
+                default:
+                    LOG(FATAL) << "Unimplemented";
+                    return kvstore::ResultCode::ERR_UNKNOWN;
+            }
+            tagFilters_.emplace(std::make_pair(tagId, propName), v);
+        }
+        auto schema = std::const_pointer_cast<meta::SchemaProviderIf>(constSchema);
+        auto updater = std::unique_ptr<RowUpdater>(new RowUpdater(schema));
+        pendingUpdater_[tagId] = std::make_unique<KeyUpdaterPair>();
```
dangleptr: ditto.

src/storage/UpdateVertexProcessor.cpp
critical27: I guess, for each tagId  need to update or upsert, we get corresponding schema and updater, if it pass filter, then we can update it.

src/parser/scanner.lex
zhangguoqing: Previously `UPDATE OR INSERT`, but now changed to `UPSERT`.  So `OR` does not have to exist.

src/storage/UpdateVertexProcessor.cpp
zhangguoqing: yep!

src/storage/UpdateVertexProcessor.cpp
zhangguoqing: Because we may update multi-tags in `Set` Clause. The `updateTagIds_ ` is used to record all `Tag IDs` at `checkAndBuildContexts` for update items. Here generate `updater` should only for the TagID that is recorded in `updateTagIds_`.

src/storage/UpdateVertexProcessor.cpp
zhangguoqing: @critical27 thx, you are all right.

src/storage/UpdateVertexProcessor.cpp
zhangguoqing: done!

dangleptr:
Jenkins go

dangleptr:
Jenkins go

dangleptr:
Jenkins go

zhangguoqing:
Jenkins go

dangleptr:
Jenkins go

src/graph/UpdateEdgeExecutor.cpp
```diff
@@ -0,0 +1,258 @@
...
+            status = Status::Error("SRC Vertex ID should be of type integer");
+            break;
+        }
+        edge_.set_src(Expression::asInt(src.value()));
+
+        status = sentence_->getDstId()->prepare();
+        if (!status.ok()) {
+            break;
+        }
+        auto dst = sentence_->getDstId()->eval();
```
wadeliuyi: what about dst.ok() is false?

src/graph/UpdateEdgeExecutor.cpp
```diff
@@ -0,0 +1,258 @@
...
+        status = checkIfGraphSpaceChosen();
+        if (!status.ok()) {
+            break;
+        }
+        insertable_ = sentence_->getInsertable();
+        status = sentence_->getSrcId()->prepare();
+        if (!status.ok()) {
+            break;
+        }
+        auto src = sentence_->getSrcId()->eval();
```
wadeliuyi: what about src.ok() is false?

src/graph/UpdateEdgeExecutor.cpp
```diff
@@ -0,0 +1,258 @@
...
+        }
+    }
+    resp_->set_column_names(std::move(columnNames));
+
+    std::vector<cpp2::RowValue> rows;
+    if (rpcResp.__isset.schema && rpcResp.__isset.data) {
+        std::shared_ptr<ResultSchemaProvider> schema;
+        if (rpcResp.get_schema() != nullptr) {
+            schema = std::make_shared<ResultSchemaProvider>(rpcResp.schema);
+        }
```
wadeliuyi: what about rpcResp.get_schema() == nullptr?

src/graph/UpdateEdgeExecutor.cpp
```diff
@@ -0,0 +1,258 @@
...
+                    case VAR_BOOL:
+                        row[index].set_bool_val(boost::get<bool>(column));
+                        break;
+                    case VAR_STR:
+                        row[index].set_str(boost::get<std::string>(column));
+                        break;
+                    default:
+                        LOG(FATAL) << "Unknown VariantType: " << column.which();
+                }
+            }
```
wadeliuyi: if ok(res) is false, maybe you need insert a default value, else the row is not match the yields_ fields.

src/graph/UpdateEdgeExecutor.cpp
```diff
@@ -0,0 +1,258 @@
...
+    reverselyEdge.key.set_src(edge_.dst);
+    reverselyEdge.key.set_dst(edge_.src);
+    reverselyEdge.key.set_ranking(edge_.ranking);
+    reverselyEdge.key.set_edge_type(-edge_.edge_type);
+    reverselyEdge.props = "";
+    edges.emplace_back(reverselyEdge);
+    auto space = ectx()->rctx()->session()->space();
+    auto future = ectx()->storage()->addEdges(space, std::move(edges), false);
+    auto *runner = ectx()->rctx()->runner();
+    auto cb = [this] (auto &&resp) {
```
wadeliuyi: can not capture this, because after you call insertReverselyEdge, you will call finishExecution, and finishExecution will delete this. so here is a problem USE AFTER FREE

src/graph/UpdateEdgeExecutor.cpp
```diff
@@ -0,0 +1,258 @@
...
+            DCHECK(onError_);
+            onError_(std::move(resp).status());
+            return;
+        }
+        auto rpcResp = std::move(resp).value();
+        if (insertable_ && rpcResp.get_upsert()) {
+            // TODO(zhangguoqing) Making the reverse edge of insertion is transactional
+            this->insertReverselyEdge();
+        }
+        this->finishExecution(std::move(rpcResp));
```
wadeliuyi: can not call finishExecution after insertReverselyEdge.

src/graph/UpdateVertexExecutor.cpp
```diff
@@ -0,0 +1,206 @@
...
+                    case VAR_BOOL:
+                        row[index].set_bool_val(boost::get<bool>(column));
+                        break;
+                    case VAR_STR:
+                        row[index].set_str(boost::get<std::string>(column));
+                        break;
+                    default:
+                        LOG(FATAL) << "Unknown VariantType: " << column.which();
+                }
+            }
```
wadeliuyi: ditto

src/storage/UpdateEdgeProcessor.cpp
```diff
@@ -0,0 +1,376 @@
...
+        VLOG(1) << "Hit edgeProp for prop: " << prop << ", value: " << it->second;
+        return it->second;
+    };
+
+    if (this->exp_ != nullptr) {
+        auto filterResult = this->exp_->eval();
+        if (filterResult.ok() && !Expression::asBool(filterResult.value())) {
+            VLOG(1) << "Filter skips the update";
+            return false;
+        }
```
wadeliuyi: what about filterResult.ok() is false?

src/storage/UpdateVertexProcessor.cpp
```diff
@@ -0,0 +1,344 @@
...
+            auto updater =
+                std::unique_ptr<RowUpdater>(new RowUpdater(std::move(reader), schema));
+            tagUpdaters_[tagId] = std::make_unique<KeyUpdaterPair>();
+            auto& tagUpdater = tagUpdaters_[tagId];
+            tagUpdater->key = iter->key().toString();
+            tagUpdater->updater = std::move(updater);
+        }
+    } else if (insertable_ && updateTagIds_.find(tagId) != updateTagIds_.end()) {
+        resp_.set_upsert(true);
+        const auto constSchema = this->schemaMan_->getTagSchema(this->spaceId_, tagId);
```
wadeliuyi: becareful about constSchema, maybe it is a nullptr

src/storage/UpdateVertexProcessor.cpp
```diff
@@ -0,0 +1,344 @@
...
+
+    // build context of the update items
+    for (auto& item : req.get_update_items()) {
+        std::string name = item.get_name();
+        std::string prop = item.get_prop();
+        auto* sourcePropExp = new SourcePropertyExpression(&name, &prop);
+        sourcePropExp->setContext(this->expCtx_.get());
+        auto status = sourcePropExp->prepare();
+        if (!status.ok() || !this->checkExp(sourcePropExp)) {
+            return cpp2::ErrorCode::E_INVALID_UPDATER;
```
wadeliuyi: before return, maybe need delete sourcePropExp, else memory leak.

src/storage/UpdateEdgeProcessor.cpp
```diff
@@ -0,0 +1,376 @@
...
+    }
+    // build context of the update items
+    for (auto& item : req.get_update_items()) {
+        std::string edgeName = item.get_name();
+        std::string prop = item.get_prop();
+        auto* edgePropExp = new AliasPropertyExpression(new std::string(""), &edgeName, &prop);
+        edgePropExp->setContext(this->expCtx_.get());
+        auto status = edgePropExp->prepare();
+        if (!status.ok() || !this->checkExp(edgePropExp)) {
+            return cpp2::ErrorCode::E_INVALID_UPDATER;
```
wadeliuyi: where will delete the edgePropExp ptr?

src/graph/UpdateEdgeExecutor.cpp
zhangguoqing: Please note the **dst.ok()** in the If-Clause. :)
```
 50         auto dst = sentence_->getDstId()->eval();
 51         if (dst.ok() && !Expression::isInt(dst.value())) {
 52             status = Status::Error("DST Vertex ID should be of type integer");
 53             break;
 54         }
```

zhangguoqing:
Jenkins go

src/graph/UpdateEdgeExecutor.cpp
wadeliuyi: maybe is if (!dst.ok() || !Expression::isInt(dst.value()))

src/storage/UpdateVertexProcessor.cpp
zhangguoqing: We can't delete it, because all members of Expression are smart ptr.
FYI.  https://github.com/vesoft-inc/nebula/blob/master/src/storage/QueryBaseProcessor.inl#L249

src/parser/scanner.lex
dangleptr: Make sense

src/storage/UpdateEdgeProcessor.cpp
zhangguoqing: ditto!

src/storage/UpdateEdgeProcessor.cpp
zhangguoqing: OK, how about ```if (!filterResult.ok() || !Expression::asBool(filterResult.value())) {}```

src/graph/UpdateEdgeExecutor.cpp
zhangguoqing: done, no check it.

src/graph/UpdateEdgeExecutor.cpp
zhangguoqing: thx. it's good.

src/graph/UpdateEdgeExecutor.cpp
zhangguoqing: done!

src/graph/UpdateEdgeExecutor.cpp
zhangguoqing: yep, if ok(res) is false, it will do `onError_`.

src/storage/UpdateVertexProcessor.cpp
zhangguoqing: done!

zhangguoqing:
Jenkins go

src/graph/UpdateEdgeExecutor.cpp
zhangguoqing: done!

src/graph/UpdateEdgeExecutor.cpp
zhangguoqing: done!

dangleptr:
Jenkins go

dangleptr:
Jenkins go

PR #555, opened 2019-06-28
Faster logging

sherman-the-tank:
This PR overwrites most of most of the macros from glog/logging.h to avoid any string operation if the FLAGS_minloglevel > severity


sherman-the-tank:
jenkins go

src/daemons/MetaDaemon.cpp
```diff
@@ -16,6 +16,7 @@
 #include "kvstore/NebulaStore.h"
 #include "meta/ActiveHostsMan.h"
 
+using nebula::operator<<;
```
dangleptr: Great!  That's a good idea!

src/common/base/test/LoggingBenchmark.cpp
```diff
@@ -0,0 +1,93 @@
...
+Benchmark number is taken from WSL running on i7-8650
+
+============================================================================
+/home/sherman/Workspace/nebula/src/common/base/test/LoggingBenchmark.cpprelative  time/iter  iters/s
+============================================================================
+glog_output_logs                                             3.13us  319.07K
+optimized_output_logs                            100.10%     3.13us  319.39K
+----------------------------------------------------------------------------
+glog_skip_logs                                               1.76us  567.45K
+optimized_skip_logs                                 inf%     0.00fs  Infinity
```
dangleptr: Great

src/common/base/test/LoggingBenchmark.cpp
```diff
@@ -0,0 +1,93 @@
...
+    folly::runBenchmarks();
+    return 0;
+}
+
+
+/*
+Benchmark number is taken from WSL running on i7-8650
+
+============================================================================
+/home/sherman/Workspace/nebula/src/common/base/test/LoggingBenchmark.cpprelative  time/iter  iters/s
```
dangleptr: Remove the relative path.

src/common/base/Logging.h
```diff
@@ -0,0 +1,222 @@
...
+    (FLAGS_minloglevel > google::GLOG_FATAL)
+#endif
+
+
+/*****************************************
+ *
+ * Normal loggings
+ *
+ ****************************************/
+#define _LOG_INFO \
```
dangleptr: Where did you override VLOG?

src/common/base/test/LoggingBenchmark.cpp
sherman-the-tank: Oops, forgot doing it.

Good catch 👍

src/daemons/MetaDaemon.cpp
dutor: I must add some explanation here.

The reason why we need to `using` this stream operator(function) explicitly, is that `HostAddr` is an alias type to `std::pair`, whose actual type is declared in namespace `std` but not namespace `nebula`. In other words, if `HostAddr` is an actual user defined type, instead of an alias, `using` is not necessary here, since the ADL rule(Argument Dependent Lookup) should have covered this.

src/common/base/Logging.h
```diff
@@ -0,0 +1,222 @@
...
+    _MINLOGLEVEL_GT_INFO \
+        ? (void) 0 \
+        : google::LogMessageVoidify() & COMPACT_GOOGLE_LOG_INFO.stream()
```
dutor: Damn, what a trick!

dutor:
> Unit testing failed.

This failure was due to the known issue that changes on schema infos haven't been synced across different modules.

To address this problem, we need to adjust the waiting time in test cases, and also to check the presence of related schema before using, instead of crash on CHECK.

PR #566, opened 2019-07-02
add a clang-format config.

### LINKED ISSUE #738: We should use the clang-format to format the code.
We should use the [clang-format](https://clang.llvm.org/docs/ClangFormatStyleOptions.html) to format the code.

monadbobo:
Nebula code is to use the Google C++ Style,
so added clang-format config to automatically format the code.

close #738


monadbobo:
close #738

PR #568, opened 2019-07-02
Fix bugs in RAFT

dangleptr:
The pr fix four bugs: 
1.  When commit logs, we hope it is under protection by raftLock_.  
2.  Write WALs should be protected by raftLock_.
3.  Term maybe changed when committed,  we should handle it. 
4.  When write wal failed, reset replicatingLogs_ to be false.

### DIFF
--- a/src/kvstore/raftex/RaftPart.cpp
+++ b/src/kvstore/raftex/RaftPart.cpp
@@ -401,14 +401,7 @@ void RaftPart::appendLogsInternal(AppendLogsIterator iter) {
     LogID prevLogId = 0;
     TermID prevLogTerm = 0;
     LogID committed = 0;
-    {
-        std::lock_guard<std::mutex> g(raftLock_);
-        currTerm = term_;
-        prevLogId = lastLogId_;
-        prevLogTerm = lastLogTerm_;
-        committed = committedLogId_;
-    }
-
+    LogID lastId = 0;
     if (iter.valid()) {
         VLOG(2) << idStr_ << "Ready to append logs from id "
                 << iter.logId() << " (Current term is "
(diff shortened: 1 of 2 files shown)


dangleptr:
Jenkins go

whitewum:
how you find this bug? can you reproduce it before the fix.

dangleptr:
> how you find this bug? can you reproduce it before the fix.

Sorry,  it is hard to reproduce it with current code structure.  I will implement test framework for raft in the future.  Currently, we have to pay more attention to code review.

whitewum:
> > how you find this bug? can you reproduce it before the fix.
> 
> Sorry, it is hard to reproduce it with current code structure. I will implement test framework for raft in the future. Currently, we have to pay more attention to code review.

for a distributed storage system,  code review is really far from enough. There are so many failover situations and states transitions.
I DO think the test framework is critical.
 So do our financial clients concern.

dangleptr:
> > > how you find this bug? can you reproduce it before the fix.
> > 
> > 
> > Sorry, it is hard to reproduce it with current code structure. I will implement test framework for raft in the future. Currently, we have to pay more attention to code review.
> 
> for a distributed storage system, code review is really far from enough. There are so many failover situations and states transitions.
> I DO think the test framework is critical.
> So do our financial clients concern.

Agreed.

dangleptr:
Jenkins go

critical27:
Maybe we also need to handle term's problem in processElectionResponses. It is similar to append logs, when we send start a election, the term in AskForVoteReq should be same with currentTerm. WDYT?

dangleptr:
> Maybe we also need to handle term's problem in processElectionResponses. It is similar to append logs, when we send start an election, the term in AskForVoteReq should be same with currentTerm. WDYT?

Good point.  We should handle it.

src/kvstore/raftex/RaftPart.cpp
```diff
@@ -572,63 +591,83 @@ void RaftPart::processAppendLogResponses(
...
-        }
-
-        // Step 3: Commit the batch
-        if (commitLogs(std::move(walIt))) {
-            std::lock_guard<std::mutex> g(raftLock_);
-
-            committedLogId_ = lastLogId;
+            auto walIt = wal_->iterator(committedId + 1, lastLogId);
+            // Step 3: Commit the batch
+            if (commitLogs(std::move(walIt))) {
```
sherman-the-tank: Good catch again 👍 

This absolutely is a potential problem. But I think we might want to take different solution to fix it. The reason I left ```commitLogs()``` out of the lock is because committing logs could take very long time. Putting it in the lock would block the normal write.

What about making WAL thread safe?

sherman-the-tank: Discussed with @dangleptr offline, he will remove the ```raftLock``` in the log append logic, so that appending logs will not be blocked by ```commitLogs()``` function call

wadeliuyi: I have a question, if we remove the raftLock, how can we synchronize the multi thread to commit log?

PR #569, opened 2019-07-02
Fixed bug on overflown number literals in scanner

### LINKED ISSUE #534: tooooooooooo loooooooongggggggggggggggg id/version number over flow
what if the id 
insert vertex xxx 100000000000000000000000000000000000000000()
the result is over flow:

nebula> go from 10100000000000000000000000000000000000000000000000000000 over like                                                                    =======================                                                                                                                               |                  id |                                                                                                                               =======================                                                                                                                               | 9223372036854775807 |                                                                                                                               -----------------------                                                                                                                               Got 1 rows (Time spent: 9218/9974 us)          

I don't think over flow is a good idea.

dutor:
As title.

This PR will close #534


src/parser/scanner.lex
```diff
@@ -249,20 +249,64 @@ IP_OCTET                    ([0-9]|[1-9][0-9]|1[0-9][0-9]|2[0-4][0-9]|25[0-5])
...
+                                    while (i < yyleng && yytext[i] == '0') {
+                                        i++;
+                                    }
+                                    if (yyleng - i > 16) {
+                                        yyterminate();
+                                    }
+                                }
                                 int64_t val = 0;
                                 sscanf(yytext, "%lx", &val);
                                 yylval->intval = val;
```
steppenwolfyuetong: The maximum value of int64 is 9,223,372,036,854,775,807,
Hexadecimal is 0X7FFF FFFF FFFF FFFF FFFF FFFF FFFF FFFF
when value is in [ 0X8000 0000 0000 0000 0000 0000 0000 0000, 0XFFFF FFFF FFFF FFFF FFFF FFFF FFFF FFFF] ， the values still overflow

src/parser/scanner.lex
```diff
@@ -249,20 +249,64 @@ IP_OCTET                    ([0-9]|[1-9][0-9]|1[0-9][0-9]|2[0-4][0-9]|25[0-5])
...
                             }
 0[Xx]{HEX}+                 {
+                                if (yyleng > 18) {
+                                    auto i = 2;
+                                    while (i < yyleng && yytext[i] == '0') {
+                                        i++;
+                                    }
+                                    if (yyleng - i > 16) {
+                                        yyterminate();
+                                    }
```
steppenwolfyuetong: The maximum value of int64 is 9,223,372,036,854,775,807,
Hexadecimal is 0X7FFF FFFF FFFF FFFF FFFF FFFF FFFF FFFF
when value is in [ 0X8000 0000 0000 0000 0000 0000 0000 0000, 0XFFFF FFFF FFFF FFFF FFFF FFFF FFFF FFFF] ， the value is still overflow.

So, here add esle statement

src/parser/scanner.lex
```diff
@@ -249,20 +249,64 @@ IP_OCTET                    ([0-9]|[1-9][0-9]|1[0-9][0-9]|2[0-4][0-9]|25[0-5])
...
+                                }
                                 int64_t val = 0;
                                 sscanf(yytext, "%lx", &val);
```
steppenwolfyuetong: Should we judge the return value of sscanf?

src/parser/scanner.lex
dutor: Hex is a way to specify integers as raw bytes, there is no such thing overflow in terms of literals.

src/parser/scanner.lex
dutor: Yes, we should have done that, even though the regex and checking above already guarantee the correctness.

steppenwolfyuetong: 👍

src/parser/scanner.lex
steppenwolfyuetong: if （yyleng - i == 16 && yytext[i] > 7 ),   Does the hexadecimal  value not overflow?
the value in decimal is great than 9,223,372,036,854,775,807.

src/parser/test/ScannerTest.cpp
```diff
@@ -350,6 +373,17 @@ TEST(Scanner, Basic) {
         CHECK_SEMANTIC_VALUE(".123", TokenType::DOUBLE, 0.123),
         CHECK_SEMANTIC_VALUE("123.456", TokenType::DOUBLE, 123.456),
 
+        CHECK_SEMANTIC_VALUE("0xFFFFFFFFFFFFFFFF", TokenType::INTEGER, 0xFFFFFFFFFFFFFFFFL),
```
steppenwolfyuetong: when use int64, the value "0xFFFFFFFFFFFFFFFF"  has already has overflowed.

src/parser/scanner.lex
steppenwolfyuetong: Yeah， you are right,   [-9,223,372,036,854,775,808     9,223,372,036,854,775,807] is [0, 0XFFFF FFFF FFFF FFFF]
👍

monadbobo: you can replace  magic number "18" with  std::numeric_limits<int64_t>::digits10 .

dutor: @monadbobo But here we are dealing with hexdecimal, instead of base 10.

PR #570, opened 2019-07-02
fix crash problem when webservice start failed

wadeliuyi:
Webservice hold a global thread point named wsThread_, and need call this thread’s join function if the service start failed, or else the process will crash.


darionyaphet:
When `nebula::WebService::stop()` be called, I find it will crash. Did you have a similar problem?

wadeliuyi:
> When `nebula::WebService::stop()` be called, I find it will crash. Did you have a similar problem?

where do you call function "nebula::WebService::stop()"? 
You can see the diff which I modified, and it not crash when I test the case: 
1: start and work normal.
2: stop normal.
3: start but some port has been used.

src/daemons/GraphDaemon.cpp
```diff
@@ -133,6 +135,7 @@ int main(int argc, char *argv[]) {
     status = setupSignalHandler();
     if (!status.ok()) {
```
darionyaphet: `setupSignalHandler()` will always return `Status::OK()` 

Do you sure `nebula::WebService::stop()` will be called ?

darionyaphet:
> where do you call function "nebula::WebService::stop()"?

`gServer->serve()` will block and I add `nebula::WebService::stop()` before main() returning .

wadeliuyi:
> > where do you call function "nebula::WebService::stop()"?
> 
> `gServer->serve()` will block and I add `nebula::WebService::stop()` before main() returning .

yes, process will crash if you call “nebula::WebService::stop()” before main() return.

because we call “nebula::WebService::stop()” in signalHandler from other thread, this will set _M_id in wsThread_, follow is one implement in stdc++:
void thread::join()
 {
    if (_M_id != id())
      __e = __gthread_join(_M_id._M_thread, 0);
    if (__e)
      __throw_system_error(__e);
    _M_id = id();
  }
 so if you call jion() again from main thread, the _M_id is not same with main thread, then will throw a exception, and I write a test sample, it crash too:
bool stop = false;

void signalHandler(int sig) {
    switch (sig) {
        case SIGINT:
        case SIGTERM:
            t.join();
            stop = true;
            break;
        default:
            printf("other signal\n");
    }
}

void setupSignalHandler() {
    ::signal(SIGPIPE, SIG_IGN);
    ::signal(SIGINT, signalHandler);
    ::signal(SIGTERM, signalHandler);
}

int main(void)
{
    setupSignalHandler();
    auto a = []() {
    std::vector<std::string> t({"a", "b", "c"});
    std::string a;
    a.reserve(128);

    printf("a legnth is %d\n", a.length());};

    t = std::thread(a);
    while (!stop)sleep(1);
    printf("main stopped\n");
    t.join();
    return 0;
}

src/daemons/GraphDaemon.cpp
```diff
@@ -112,6 +101,19 @@ int main(int argc, char *argv[]) {
...
+
+    LOG(INFO) << "Starting Graph HTTP Service";
+    // http://127.0.0.1:XXXX/status is equivalent to http://127.0.0.1:XXXX
+    nebula::WebService::registerHandler("/status", [] {
+        return new nebula::graph::GraphHttpHandler();
+    });
+    status = nebula::WebService::start();
+    if (!status.ok()) {
+        LOG(ERROR) << "Failed to start web service: " << status;
+        nebula::WebService::stop();
```
dutor: IMHO, This not a proper way to address this problem. `start` and `stop` ought to be paired only if the `start` succeeded. Otherwise, it's counter intuitive.

I think you should consider to join and destroy the internal thread if `start` fails.

wadeliuyi: thanks very much, good point.

wadeliuyi:
ready for test

wadeliuyi:
jenkins go!

PR #572, opened 2019-07-03
To deal with illegal characters in parser

### LINKED ISSUE #562: [nGQL] Chinese punctune ；make service crash
It's better to check the string.
at least not make service crash.

dutor:
As title.

This will close #562


src/parser/test/ScannerTest.cpp
```diff
@@ -374,6 +374,9 @@ TEST(Scanner, Basic) {
         CHECK_SEMANTIC_VALUE("\"\\\\\\\110 \"", TokenType::STRING, "\\H "),
         CHECK_SEMANTIC_VALUE("\"\\\\\\\\110 \"", TokenType::STRING, "\\\\110 "),
         CHECK_SEMANTIC_VALUE("\"\\\\\\\\\110 \"", TokenType::STRING, "\\\\H "),
+
+
+        CHECK_SEMANTIC_VALUE("\"己所不欲，勿施于人\"", TokenType::STRING, "己所不欲，勿施于人"),
```
whitewum: before the fix, does this sentence make service crash?
My point is can you re-produce the problem. I'm not sure the bug is due to non-asiic code

dutor: It's another story. You could refer to discussions in #562 for details.

src/parser/scanner.lex
```diff
@@ -317,18 +319,40 @@ IP_OCTET                    ([0-9]|[1-9][0-9]|1[0-9][0-9]|2[0-4][0-9]|25[0-5])
...
                             }
 "#".*                       // Skip the annotation
 "//".*                      // Skip the annotation
 "--".*                      // Skip the annotation
 "/*"                        { BEGIN(COMMENT); }
 <COMMENT>"*/"               { BEGIN(INITIAL); }
 <COMMENT>([^*]|\n)+|.       // Skip the annotation
-.                           { printf("error %c\n", *yytext); yyterminate(); }
+.                           {
+                                /**
```
whitewum: lesson learned

PR #575, opened 2019-07-03
To detect TTY in nebula console

dutor:
When `stdin` is not detached to a TTY, command history will not be saved by default, and no prompt will be printed.


src/console/CliManager.cpp
```diff
@@ -16,6 +16,7 @@
 
 DECLARE_string(u);
 DECLARE_string(p);
+DEFINE_bool(enable_history, false, "Whether to force saving the command history");
```
steppenwolfyuetong: Great God, I have a question.
This ".nebula_history" file is used to save history commands. Is there a file size limit? When the file is large, does it take a long time to start initialization (loadHistory)?

How did you think about this? Thx.

dutor: This is good question. We will take good care of this, in another PR.

dutor: This is an option to **force** the history output, even no tty attached. With this option set to false, the history saving is automatical.

monadbobo: OK, I get it.

PR #582, opened 2019-07-04
To link executables with -no-pie

dutor:
Some compilers are built with `-enable-default-pie`. That is no problem generally, but glog is buggy to deal with this when retrieve the backtrace. So we disable pie from our own side.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -47,7 +47,7 @@ endif()
 
 set(CMAKE_CXX_STANDARD 14)
 
-set(CMAKE_EXE_LINKER_FLAGS "-static-libstdc++")
+set(CMAKE_EXE_LINKER_FLAGS "-static-libstdc++ -no-pie")
 
 # CMake macro to specify number of jobs to build third-party
 if (NOT THIRD_PARTY_JOBS)


monadbobo:
Could you describe the problem caused by glog in more detail?

dutor:
> Could you describe the problem caused by glog in more detail?

Actually, glog introduced a bug that will cause the symbol resolve fail for PIE, which is exactly the version we are currently using, which has been fixed the the latest release.

steppenwolfyuetong:
> > Could you describe the problem caused by glog in more detail?
> 
> Actually, glog introduced a bug that will cause the symbol resolve fail for PIE, which is exactly the version we are currently using, which has been fixed the the latest release.

@dutor   I have a question, why not use the latest version of glog ?

dutor:
> > > Could you describe the problem caused by glog in more detail?
> > 
> > 
> > Actually, glog introduced a bug that will cause the symbol resolve fail for PIE, which is exactly the version we are currently using, which has been fixed the the latest release.
> 
> @dutor I have a question, why not use the latest version of glog ?

First, it makes little sense to us to enable PIE, by default. Second, we should try our best to reduce mandatory requirements on our dependencies, especially when we have better options to deal with the problems.

FYI. we are going to move the third-party out from our repo.

steppenwolfyuetong:
> > > > Could you describe the problem caused by glog in more detail?
> > > 
> > > 
> > > Actually, glog introduced a bug that will cause the symbol resolve fail for PIE, which is exactly the version we are currently using, which has been fixed the the latest release.
> > 
> > 
> > @dutor I have a question, why not use the latest version of glog ?
> 
> First, it makes little sense to us to enable PIE, by default. Second, we should try our best to reduce mandatory requirements on our dependencies, especially when we have better options to deal with the problems.
> 
> FYI. we are going to move the third-party out from our repo.

f1634089a5014f20919a619ed3eb2e2e

monadbobo:
LGTM!

Issue #583, opened 2019-07-04
storage service crashed when stopping service


monadbobo:
I compile the nebula with addresssanitizer.

When I started, the query engine service crashed due to some exceptions. At this point, I stopped the storage service and then caused the storage service to crash. The following was the output of addresssanitizer.

> ==17116==ERROR: AddressSanitizer: heap-use-after-free on address 0x60400011f060 at pc 0x000001805b39 bp 0x7fcbfc2af400 sp 0x7fcbfc2af3f0
READ of size 8 at 0x60400011f060 thread T8
    #0 0x1805b38 in nebula::meta::MetaClient::heartbeat() src/meta/client/MetaClient.cpp:957
    #1 0x17f0850 in nebula::meta::MetaClient::heartBeatThreadFunc() src/meta/client/MetaClient.cpp:67
    #2 0x1926f9d in void std::__invoke_impl<void, void (nebula::meta::MetaClient::*&)(), nebula::meta::MetaClient*&>(std::__invoke_memfun_deref, void (nebula::meta::MetaClient::*&)(), nebula::meta::MetaClient*&) (/data/pagefault/tools/nebula/bin/nebula-storaged+0x1926f9d)
    #3 0x1910f61 in std::__invoke_result<void (nebula::meta::MetaClient::*&)(), nebula::meta::MetaClient*&>::type std::__invoke<void (nebula::meta::MetaClient::*&)(), nebula::meta::MetaClient*&>(void (nebula::meta::MetaClient::*&)(), nebula::meta::MetaClient*&) /usr/include/c++/8/bits/invoke.h:95
    #4 0x18e9e60 in void std::_Bind<void (nebula::meta::MetaClient::*(nebula::meta::MetaClient*))()>::__call<void, , 0ul>(std::tuple<>&&, std::_Index_tuple<0ul>) (/data/pagefault/tools/nebula/bin/nebula-storaged+0x18e9e60)
    #5 0x18bb3fd in void std::_Bind<void (nebula::meta::MetaClient::*(nebula::meta::MetaClient*))()>::operator()<, void>() (/data/pagefault/tools/nebula/bin/nebula-storaged+0x18bb3fd)
    #6 0x189dcbf in std::_Function_handler<void (), std::_Bind<void (nebula::meta::MetaClient::*(nebula::meta::MetaClient*))()> >::_M_invoke(std::_Any_data const&) /usr/include/c++/8/bits/std_function.h:297
[...]

PR #584, opened 2019-07-04
Fixed a bug where the storage service might crash (#583)

monadbobo:
When the storage service deamon is started, the nebula::meta::MetaClient will use the MetaServerBasedPartManager object(MetaClient::registerListener), but the MetaServerBasedPartManager object will be released before nebula::meta::MetaClient(through KVStore object),so we need to manually release nebula::meta::MetaClient here before release KVStore object.

### DIFF
--- a/src/daemons/StorageDaemon.cpp
+++ b/src/daemons/StorageDaemon.cpp
@@ -200,10 +200,12 @@ int main(int argc, char *argv[]) {
         gServer->setIOThreadPool(ioThreadPool);
         gServer->serve();  // Will wait until the server shuts down
     } catch (const std::exception& e) {
+        metaClient.reset();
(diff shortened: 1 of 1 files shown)


monadbobo:
jenkins go!

dutor:
Recently, I have found that `src/storage/test/StorageClientTest.cpp` has the similar issue.

dangleptr:
A better way is unregister listener in MetaServerBasedPartManager dtor.  MetaClient should NOT depend on one component in kvstore.

monadbobo:
> A better way is unregister listener in MetaServerBasedPartManager dtor. MetaClient should NOT depend on one component in kvstore.

the SchemaManager also used the MetaClient, so I don't think it should use the  unique_ptr to represent MetaClient.

PR #586, opened 2019-07-04
Fix init bug in nebula store

### LINKED ISSUE #585: Two different storage engine may contain same part when NebulaStore init.
there is a problem that we config two data path like a and b, then there are two engines engine_a and engine_b, and engine_a has two parts 1 and 2, engine_b has 3. when call addPart with partId 1, then minIndex is engine_b, and the result cause that part 1 in two engine

dangleptr:
Fix #585   #562 

Close #585


dangleptr:
Jenkins go

wadeliuyi:
LGTM

PR #588, opened 2019-07-05
To be able to build under GCC 9.1

dutor:
As title.


dutor:
Due to the build time exceeded 30min.

Jenkins, go!

PR #595, opened 2019-07-08
Fixed a bug where the storage service might crash (#583)

monadbobo:
summary:

  When the storage service deamon is started, the nebula::meta::MetaClient
  will use the MetaServerBasedPartManager object(MetaClient::registerListener),
  but the MetaServerBasedPartManager object will be released before
  nebula::meta::MetaClient(through KVStore object).

  Added the MetaClient::unRegisterListener method to stop the background thread
  and reset the listener_ object, And the unRegisterListener is called
  when the MetaServerBasedPartManager object was released.


critical27:
I am a little confused. Since PartManager `has a` MetaClient, MetaClient `has a` listener, and PartManager object will be released before MetaClient, why don't we just set the MetaClient in PartManager to nullptr?

monadbobo:
> I am a little confused. Since PartManager `has a` MetaClient, MetaClient `has a` listener, and PartManager object will be released before MetaClient, why don't we just set the MetaClient in PartManager to nullptr?
> The resource in MetaClient will be released in ~MetaClient. I think it's ok.

That's because two threads will be started in metaclient, and MetaServerBasedPartManager will be used in these two threads.

src/meta/client/MetaClient.h
```diff
@@ -68,10 +68,18 @@ class MetaClient {
...
 
+    void deInit();
+
     void registerListener(MetaChangedListener* listener) {
+        CHECK(listener_ == nullptr);
         listener_ = listener;
     }
 
+    void unRegisterListener() {
+        deInit();
```
critical27: I understood now. Maybe you should not call deInit here? Because other's still use the bgThread in MetaClient, e.g. SchemaManager?

monadbobo: Here we need to stop the background thread first, otherwise it will have race conditions in listener_.

dangleptr: We'd better protect listener_ in MetaClient instead of stop bgThread directly.   Because we don't want meta client is controlled by kvstore.

monadbobo: According to my understanding.
1 metaclient itself is dependent on the listener_ object.
2 If listener_ is released, both two threads of metaclient will not work, then there is no need for metaclient at this time.

dangleptr: Not exactly.  Listener_ is not necessary for meta client.

src/meta/client/MetaClient.h
```diff
@@ -69,9 +69,16 @@ class MetaClient {
     void init();
 
     void registerListener(MetaChangedListener* listener) {
+        CHECK(listener_ == nullptr);
```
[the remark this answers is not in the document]
monadbobo: > Totally, the pr lgtm. Does the crash problem fixed after unRegisterListener?

Yes, this PR will fix the crash.

PR #597, opened 2019-07-09
Implementation of Distinct

CPWstatic:
GO ... YIELD DISTINCT ...


src/graph/GoExecutor.cpp
```diff
@@ -75,6 +79,13 @@ void GoExecutor::execute() {
         onEmptyInputs();
         return;
     }
+    if (distinct_) {
+        std::unordered_set<VertexID> uniqID;
+        for (auto id : starts_) {
+            uniqID.emplace(id);
+        }
+        starts_ = std::vector<VertexID>(uniqID.begin(), uniqID.end());
```
whitewum: what if distinct & order by?

src/graph/test/GoTest.cpp
```diff
@@ -154,5 +154,24 @@ TEST_F(GoTest, DISABLED_OneStepInOutBound) {
     }
 }
 
+TEST_F(GoTest, Distinct) {
+    {
+        cpp2::ExecutionResponse resp;
+        auto &player = players_["Boris Diaw"];
+        auto *fmt = "GO FROM %ld OVER like "
+                    "| GO FROM $-.id OVER like | GO FROM $-.id OVER serve "
+                    "YIELD DISTINCT serve._dst, $$[team].name";
```
whitewum: try the return is null

src/parser/parser.yy
```diff
@@ -95,6 +95,7 @@ class GraphScanner;
 %token KW_ROLES KW_BY
 %token KW_TTL_DURATION KW_TTL_COL
 %token KW_ORDER KW_ASC
+%token KW_DISTINCT
```
whitewum: better to add a case in parser ut.

src/graph/GoExecutor.cpp
```diff
@@ -530,7 +554,15 @@ std::unique_ptr<InterimResult> GoExecutor::setupInterimResult(RpcResponse &&rpcR
                     LOG(FATAL) << "Unknown VariantType: " << column.which();
             }
         }
-        rsWriter->addRow(writer);
+        std::string encode = writer.encode();
+        if (distinct_) {
+            if (uniqResult->count(encode) == 0) {
+                uniqResult->emplace(encode);
```
whitewum: uniqResult // only emplace the first encode. right?

CPWstatic: Yes.

src/graph/GoExecutor.cpp
CPWstatic: Order by is standalone statement, they will not affect each other.

src/parser/test/ParserTest.cpp
```diff
@@ -940,4 +940,22 @@ TEST(Parser, ReentrantRecoveryFromFailure) {
...
+        std::string query = "GO FROM 1 over friend "
+                            "YIELD DISTINCT friend.name as name, friend.age as age";
+        auto result = parser.parse(query);
+        ASSERT_TRUE(result.ok()) << result.status();
+    }
+    {
+        GQLParser parser;
+        std::string query = "GO FROM 1 OVER like "
+                            "| GO FROM $-.id OVER like | GO FROM $-.id OVER serve "
+                            "YIELD DISTINCT serve._dst, $$.team.name";
```
whitewum: syntax error: yield col1, distinct col2

CPWstatic: That is what we want, and I add this test case in parser.

src/graph/GoExecutor.cpp
dutor: You could have just called `emplace` for once, since it will tell you whether the value exists.

whitewum:
please remember to update nGQL.md

src/graph/GoExecutor.cpp
```diff
@@ -530,7 +554,16 @@ std::unique_ptr<InterimResult> GoExecutor::setupInterimResult(RpcResponse &&rpcR
                     LOG(FATAL) << "Unknown VariantType: " << column.which();
             }
         }
-        rsWriter->addRow(writer);
+        // TODO Consider float/double, and need to reduce mem copy.
```
whitewum: right

src/graph/GoExecutor.cpp
```diff
@@ -530,7 +554,16 @@ std::unique_ptr<InterimResult> GoExecutor::setupInterimResult(RpcResponse &&rpcR
...
+        std::string encode = writer.encode();
+        if (distinct_) {
+            auto ret = uniqResult->emplace(encode);
+            if (ret.second) {
```
whitewum: I wonder which is more efficient?

Issue #600, opened 2019-07-09
Column type checks are ignored when create or alter tag | edge.


boshengchen:
**Describe the bug**
In source `parser.yy`, the column_spec_list clause is used to create or alter tag | edge, column types require a validation check.

**How To Reproduce**
  1. CREATE TAG person(name, age, married bool)
  2. ALTER TAG person ADD (col1), CHANGE (married), DROP (age int)
  3. CREATE EDGE man(name, age)

**Expected behavior**
SyntaxError

sherman-the-tank:
Can someone take care of this?

laura-ding:
Has be solved, so close it.

laura-ding:
Has solved, so close it.

PR #601, opened 2019-07-09
fix bug for column data type check

boshengchen:
Related to issuer #600 

The following statements expect a syntax error.  but has not been caught before
1. CREATE TAG person(name, age, married bool)
2. ALTER TAG person ADD (col1), CHANGE (married), DROP (age int)
3. CREATE EDGE man(name, age)

### DIFF
--- a/src/parser/parser.yy
+++ b/src/parser/parser.yy
@@ -39,6 +39,7 @@ class GraphScanner;
     nebula::SequentialSentences            *sentences;
     nebula::ColumnSpecification            *colspec;
     nebula::ColumnSpecificationList        *colspeclist;
+    nebula::ColumnNameList                 *colsnamelist;
     nebula::ColumnType                      type;
     nebula::StepClause                     *step_clause;
     nebula::FromClause                     *from_clause;
@@ -157,6 +158,7 @@ class GraphScanner;
 
 %type <colspec> column_spec
 %type <colspeclist> column_spec_list
+%type <colsnamelist> column_name_list
 
 %type <with_user_opt_list> with_user_opt_list
 %type <with_user_opt_item> with_user_opt_item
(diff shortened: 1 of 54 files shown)


boshengchen:
Jenkins go

src/parser/parser.yy
```diff
@@ -587,12 +587,14 @@ create_tag_sentence
         $$ = new CreateTagSentence($3, new ColumnSpecificationList(), $6);
     }
     | KW_CREATE KW_TAG name_label L_PAREN column_spec_list R_PAREN create_schema_prop_list {
+        CHECK_COLUMN_SPACES(false, $5);
```
dutor: How about to add another structure, e.g. `ColumnNameList`, for `DROP`?

src/parser/test/ParserTest.cpp
```diff
@@ -307,6 +307,77 @@ TEST(Parser, EdgeOperation) {
         ASSERT_TRUE(result.ok()) << result.status();
     }
 }
+// Column space format test, expected SyntaxError
+TEST(Parser, ColumnSpacesTest) {
+    {
+        GQLParser parser;
+        std::string query = "CREATE TAG person(name, age, married bool)";
+        auto result = parser.parse(query);
+        ASSERT_FALSE(result.ok()) << result.status();
```
dutor: Be careful about `StatusOr`. It holds either a _value_ or _Status_, but never both.

The same to all other places.

src/parser/parser.yy
boshengchen: Good idea ! It's better to add a new structure for `DROP`.

boshengchen:
Address dutor's comment. add a new structure ColumnNameList.

boshengchen:
Jenkins go

PR #603, opened 2019-07-10
Fixed dangling pointer in Balancer

dutor:
As title.

### DIFF
--- a/src/meta/processors/admin/Balancer.cpp
+++ b/src/meta/processors/admin/Balancer.cpp
@@ -63,8 +63,9 @@ bool Balancer::recovery() {
         plan_ = std::make_unique<BalancePlan>(corruptedPlans[0], kv_, client_.get());
         plan_->onFinished_ = [this] () {
             bool expected = true;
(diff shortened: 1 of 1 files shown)


dangleptr:
As original implementation,  after plan_.reset(),  we did call any members in plan_ in onFinished_.  

The program crashed?

dutor:
> As original implementation, after plan_.reset(), we did not access any members in plan_ in onFinished_.
> 
> The program crashed?

After `plan_` got released, `onFinished_` also no longer persists, as well as all data it has captured, `this` in this case. So `running_` cannot be reached any more, even though it's still alive.

dangleptr:
> > As original implementation, after plan_.reset(), we did not access any members in plan_ in onFinished_.
> > The program crashed?
> 
> After `plan_` got released, `onFinished_` also no longer persists, as well as all data it has captured, `this` in this case. So `running_` cannot be reached any more, even though it's still alive.

Fair enough!

dutor:
Thanks to @wadeliuyi for discovering the race condition problem.

PR #604, opened 2019-07-10
Remove the raftlock_ when appending logs into raft queue

dangleptr:
To improve the efficient,  we don't need to acquire the raftLock_ in appendLogs when replicating logs.   All we need to do is just putting the log in the queue and return the future.

All checks could be done when we really need to replicate the logs.


src/kvstore/raftex/RaftPart.cpp
```diff
@@ -382,22 +371,34 @@ folly::Future<AppendLogResult> RaftPart::appendLogAsync(ClusterID source,
...
         }
-        if (replicatingLogs_) {
-            VLOG(2) << idStr_
-                    << "Another AppendLogs request is ongoing,"
-                       " just return";
-            return retFuture;
-        } else {
+
+        bool expected = false;
+        if (replicatingLogs_.compare_exchange_strong(expected, true)) {
```
wadeliuyi: replicatingLogs_ is protected by logsLock_, maybe no need CAS?

src/kvstore/raftex/RaftPart.cpp
```diff
@@ -1244,45 +1234,44 @@ folly::Future<AppendLogResult> RaftPart::sendHeartbeat() {
...
         if (!logs_.empty()) {
             LOG(WARNING) << idStr_
                          << "There is logs in the buffer,"
                             " stop sending the heartbeat";
             return AppendLogResult::SUCCEEDED;
         }
 
-        if (replicatingLogs_) {
+        bool expected = false;
+        if (!replicatingLogs_.compare_exchange_strong(expected, true)) {
```
wadeliuyi: ditto

src/kvstore/raftex/RaftPart.cpp
dangleptr: replicatingLogs_ has no relationship with logsLock_.   We can't assume it will be accessed under protection by logsLock_,  actually we accessed it in many places directly without logsLock_

dangleptr:
Jenkins go

dangleptr:
Jenkins go

src/kvstore/raftex/RaftPart.cpp
```diff
@@ -382,22 +371,34 @@ folly::Future<AppendLogResult> RaftPart::appendLogAsync(ClusterID source,
...
         }
     }
 
+    {
+        std::lock_guard<std::mutex> g(raftLock_);
+        auto res = canAppendLogs();
+        if (res != AppendLogResult::SUCCEEDED) {
+            LOG(ERROR) << idStr_
+                       << "Cannot append logs, clean the buffer";
+            sendingPromise_.setValue(std::move(res));
```
wadeliuyi: maybe no need move?

dangleptr: Good point.  Actually, in this case move equals copy.  It is harmless.

src/kvstore/raftex/RaftPart.cpp
```diff
@@ -651,49 +641,48 @@ void RaftPart::processAppendLogResponses(
...
-                            });
-                        logs_.clear();
-                    } else {
-                        replicatingLogs_ = false;
-                        VLOG(2) << idStr_ << "No more log to be replicated";
-                    }
-                }
+                firstLogId = lastLogId_ + 1;
             } else {
                 LOG(FATAL) << idStr_ << "Failed to commit logs";
```
wadeliuyi: if commitLogs() failed, why not return?

src/kvstore/raftex/RaftPart.cpp
```diff
@@ -651,49 +641,48 @@ void RaftPart::processAppendLogResponses(
...
+        // Continue to process the original AppendLogsIterator if necessary
+        iter.resume();
+        if (iter.empty()) {
+            std::lock_guard<std::mutex> lck(logsLock_);
+            if (logs_.size() > 0) {
+                // continue to replicate the logs
+                sendingPromise_ = std::move(cachingPromise_);
+                cachingPromise_.reset();
+                iter = AppendLogsIterator(
+                    firstLogId,
```
wadeliuyi: if commitLogs() failed, firstLogId is 0, is there any problem?

src/kvstore/raftex/RaftPart.cpp
dangleptr: If commitLogs failed, there must be some really horrible things happened.  To protect the data,  we must stop the server.

wadeliuyi: LOG(FATAL) will stop storage server?

dangleptr:
Jenkins go

PR #606, opened 2019-07-10
Make heartbeat to be a special empty log

dangleptr:
This pr is to fix the problem that when follower becomes leader,  it will not commit the log with previous term.  As the paper said,  the logic is correct.  But the problem is that as current heart beat implementation, we will not commit anything when received response from followers.  It is incorrect. 

So this pr will use a special empty log to represent heartbeat,   the log will be written into WAL but skipped when committing.   It has no harm to performance because we don't send heartbeat when replicating logs.   And it will force the uncommitted logs to commit when the new leader send heartbeat with it's current term.


dangleptr:
Jenkins go

critical27:
LGTM. It really simplifies code.

PR #611, opened 2019-07-11
Issue#424 to make `Expression::eval` to be able to do error handling

monadbobo:
Summary:

1. When the expression is evaluated, some wrong judgments are added.
   When the expression was failed, an error will be replied to the user.
2. When the user's destination edge for inserting an vertical does not exist,
   we need to skip this dst edge when the user query the prop of dst edge( Issue#580).

### DIFF
--- a/src/graph/GoExecutor.cpp
+++ b/src/graph/GoExecutor.cpp
@@ -10,6 +10,7 @@
 #include "dataman/RowSetReader.h"
 #include "dataman/ResultSchemaProvider.h"
 
+
 namespace nebula {
 namespace graph {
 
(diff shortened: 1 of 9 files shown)


src/graph/GoExecutor.cpp
```diff
@@ -616,8 +628,12 @@ void GoExecutor::setupResponseBody(RpcResponse &rpcResp, cpp2::ExecutionResponse
         rows.emplace_back();
         rows.back().set_columns(std::move(row));
     };
-    processFinalResult(rpcResp, cb);
+
+    if (!processFinalResult(rpcResp, cb)) {
+      return false;
```
wadeliuyi: two spaces?

wadeliuyi: here is two spaces;

src/graph/GoExecutor.cpp
```diff
@@ -657,21 +673,35 @@ void GoExecutor::processFinalResult(RpcResponse &rpcResp, Callback cb) const {
             RowSetReader rsReader(eschema, vdata.edge_data);
             auto iter = rsReader.begin();
             while (iter) {
```
wadeliuyi: I do not understand why not judge the condition “iter != rsRead.end()” in the “while”?

PR #613, opened 2019-07-11
wait for metaserer ready(loop)

andychow01:
MetaClient instance wait meta serer ready(loop), then do other things like start threadpool.....

### DIFF
--- a/src/meta/client/MetaClient.cpp
+++ b/src/meta/client/MetaClient.cpp
@@ -47,13 +47,13 @@ bool MetaClient::waitForMetadReady() {
             auto ret = heartbeat().get();
             if (!ret.ok()) {
                 LOG(ERROR) << "Heartbeat failed, status:" << ret.status();
-                std::this_thread::sleep_for(std::chrono::milliseconds(100));
+                std::this_thread::sleep_for(std::chrono::milliseconds(1000));
                 continue;
             }
         }  // end if
         loadDataThreadFunc();
         if (!ready_) {
-            std::this_thread::sleep_for(std::chrono::milliseconds(100));
+            std::this_thread::sleep_for(std::chrono::milliseconds(1000));
         }
     }  // end while
 
--- a/src/meta/client/MetaClient.cpp
+++ b/src/meta/client/MetaClient.cpp
@@ -17,9 +17,11 @@ namespace meta {
 
 MetaClient::MetaClient(std::shared_ptr<folly::IOThreadPoolExecutor> ioThreadPool,
                        std::vector<HostAddr> addrs,
+                       HostAddr localHost,
                        bool sendHeartBeat)
     : ioThreadPool_(ioThreadPool)
     , addrs_(std::move(addrs))
+    , localHost_(localHost)
     , sendHeartBeat_(sendHeartBeat) {
     CHECK(ioThreadPool_ != nullptr) << "IOThreadPool is required";
     CHECK(!addrs_.empty())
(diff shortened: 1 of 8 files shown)


src/meta/client/MetaClient.cpp
```diff
@@ -39,8 +39,24 @@ MetaClient::~MetaClient() {
...
 
 
-void MetaClient::init() {
+void MetaClient::waitForMetadReady() {
     loadDataThreadFunc();
+
+    if (!ready_ && sendHeartBeat_) {
+        while (true) {
+            if (listener_ == nullptr) {
+                VLOG(1) << "Can't send heartbeat due to listener_ is nullptr!";
```
wadeliuyi: Function “heartbeat” call “auto localHost = listener_->getLocalHost();”, so need continue here

src/meta/client/MetaClient.cpp
```diff
@@ -39,8 +39,24 @@ MetaClient::~MetaClient() {
...
+            }
+            auto ret = heartbeat().get();
+            if (!ret.ok()) {
+                LOG(ERROR) << "Heartbeat failed, status:" << ret.status();
+                ::sleep(1);
```
wadeliuyi: maybe can use std::this_thread::sleep_for to sleep short

src/meta/client/MetaClient.cpp
andychow01: fixed! Thanks! But I think exit(EXIT_FAILURE) will be more reasonable!

src/meta/client/MetaClient.cpp
andychow01: fixed! Thanks!

src/meta/client/MetaClient.cpp
```diff
@@ -39,8 +39,25 @@ MetaClient::~MetaClient() {
 }
 
 
-void MetaClient::init() {
+void MetaClient::waitForMetadReady() {
     loadDataThreadFunc();
```
dangleptr: Let's figure out what's ready for Meta client. 
Currently,  meta client sends HB to meta server on storage layer,  but it does not on query engine side. 
On storage side,  meta client ready means that server accept the HB and local cache is filled. 
On graph side,  meta client ready means that local cache is filled. 

So we should handle both situations above.

src/meta/client/MetaClient.cpp
wadeliuyi: > fixed! Thanks! But I think exit(EXIT_FAILURE) will be more reasonable!

so that mean metaclient is strong depend listener?

src/meta/client/MetaClient.cpp
andychow01: Modified! Please review!

src/meta/client/MetaClient.cpp
```diff
@@ -39,8 +41,19 @@ MetaClient::~MetaClient() {
...
+    while (!ready_) {
+        if (sendHeartBeat_) {
+            auto ret = heartbeat().get();
+            if (!ret.ok()) {
+                LOG(ERROR) << "Heartbeat failed, status:" << ret.status();
+                std::this_thread::sleep_for(std::chrono::milliseconds(100));
+                continue;
+            }
+        }  // end if
+        loadDataThreadFunc();
```
critical27: Maybe you need to sleep a while here? If it a MetaClient of graph, and fail to load data, it will try to call listSpaces all the time?

andychow01: Now I have added a sleep. Thanks!

src/meta/client/MetaClient.cpp
andychow01: Now, the logic has changed. Metaclient don't strong depend on listener!

src/meta/client/MetaClient.cpp
```diff
@@ -39,8 +41,22 @@ MetaClient::~MetaClient() {
 }
 
 
-void MetaClient::init() {
-    loadDataThreadFunc();
+bool MetaClient::waitForMetadReady() {
+    while (!ready_) {
```
wadeliuyi: is there any method to stop this loop if the ready_ alway is false?

src/meta/test/MetaClientTest.cpp
```diff
@@ -469,7 +475,7 @@ TEST(MetaClientTest, HeartbeatTest) {
         }
     }
     sleep(FLAGS_heartbeat_interval_secs + 1);
-    ASSERT_EQ(1, ActiveHostsMan::instance()->getActiveHosts().size());
```
dangleptr: Why the changes?

Issue #580, opened 2019-07-04
vertexId not exists, but return props are not null 


jude-zhu:
First, I define the tag and edge schema
```
CREATE TAG student(name string, age int, gender string)
CREATE EDGE like(likeness double)
```

And then insert a vertex and an edge
```
INSERT VERTEX student(name, age, gender) VALUES 201:("Mike", 18, "male");
INSERT EDGE like (likeness) VALUES 201 -> 203:(93.2);
```
the dest. vertex is undefined. However, when I try
```
GO FROM 201 OVER like YIELD $^[student].name AS Student, $$[student].name AS Friend, $$[student].age AS Age, $$[student].gender AS Gender
```
the dest. props are not null

![20190704112851](https://user-images.githubusercontent.com/51590253/60638162-df736f00-9e4f-11e9-8072-766d358a5b1d.jpg)

monadbobo:
When the storage layer performs a vertex lookup(QueryBoundProcessor::processVertex), if the  vertex does not exist, the vertex data should not be set at this time. And the query engine also needs to handle this situation, I will fix this problem.

PR #616, opened 2019-07-11
Issue#424 to make `Expression::eval` to be able to do error handling.

monadbobo:
Summary:

1. When the expression is evaluated, some wrong judgments are added.
   When the expression was failed, an error will be replyed to the user.
2. When the user's destination vertex for inserting an edge does not exist,
   we need to skip this dst vertex when the user query the prop of dst vertex(Issue#580).

### DIFF
--- a/src/common/filter/Expressions.cpp
+++ b/src/common/filter/Expressions.cpp
@@ -144,7 +144,7 @@ std::string InputPropertyExpression::toString() const {
 }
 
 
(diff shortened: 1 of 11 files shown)


monadbobo:
I think the fix for issue #580 needs to provide a null type to be a perfect fix, so I will resubmit a PR to fix Issue#424 separately.

Issue #182, opened 2019-02-26
Support scaling out for storage service. 


sherman-the-tank:
Meta server can orchestra partition movement/re-balance in the cluster

PR #619, opened 2019-07-12
Add preProcessor for each log before it being writen into WAL

dangleptr:
It will be used for command processing. 
Subtask of #182


dangleptr:
Jenkins go

PR #620, opened 2019-07-12
Fix bug about stopping twice for part

dangleptr:
Thanks for @wadeliuyi reporting.  The part will be stopped inside removePartition.


src/kvstore/NebulaStore.cpp
```diff
@@ -274,8 +274,6 @@ void NebulaStore::removePart(GraphSpaceID spaceId, PartitionID partId) {
         if (partIt != spaceIt->second->parts_.end()) {
             auto* e = partIt->second->engine();
             CHECK_NOTNULL(e);
-            // Stop the raft
-            partIt->second->stop();
```
sherman-the-tank: Where is the other place?

PR #622, opened 2019-07-12
Fixed UpgradedHolder-related race condition in ActiveHostsMan

dutor:
`UpgradedHolder` makes a `RWSpinLock` or `ReadHolder` progress to the `UPGRADED` phase. And `UPGRADED` is an intermediate phase to transit(upgrade) from `READ` to `WRITE`.

### DIFF
--- a/src/meta/ActiveHostsMan.cpp
+++ b/src/meta/ActiveHostsMan.cpp
@@ -37,10 +37,17 @@ bool ActiveHostsMan::updateHostInfo(const HostAddr& hostAddr, const HostInfo& in
         using ReadHolder = folly::RWSpinLock::ReadHolder;
         using WriteHolder = folly::RWSpinLock::WriteHolder;
         using UpgradedHolder = folly::RWSpinLock::UpgradedHolder;
+        // Acquire the read lock
         ReadHolder rh(&lock_);
         auto it = hostsMap_.find(hostAddr);
         if (it == hostsMap_.end()) {
-            WriteHolder wh(UpgradedHolder(&lock_));
+            // Step to Upgraded phase to stop new readers
+            UpgradedHolder uh(&lock_);
+            // Unlock the read lock
+            rh.reset();
+            // Wait for existing readers to finish and then acquire the write lock
+            WriteHolder wh(std::move(uh));
+            // Now we are safe to write
             hostsMap_.emplace(hostAddr, std::move(info));
             data.emplace_back(MetaServiceUtils::hostKey(hostAddr.first, hostAddr.second),
                               MetaServiceUtils::hostValOnline());
--- a/src/meta/ActiveHostsMan.cpp
+++ b/src/meta/ActiveHostsMan.cpp
@@ -34,10 +34,13 @@ ActiveHostsMan::ActiveHostsMan(int32_t intervalSeconds, int32_t expiredSeconds,
 bool ActiveHostsMan::updateHostInfo(const HostAddr& hostAddr, const HostInfo& info) {
     std::vector<kvstore::KV> data;
     {
-        folly::RWSpinLock::ReadHolder rh(&lock_);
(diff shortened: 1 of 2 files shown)


wadeliuyi:
Perfect!

src/meta/ActiveHostsMan.cpp
```diff
@@ -34,10 +34,13 @@ ActiveHostsMan::ActiveHostsMan(int32_t intervalSeconds, int32_t expiredSeconds,
...
     {
-        folly::RWSpinLock::ReadHolder rh(&lock_);
+        using ReadHolder = folly::RWSpinLock::ReadHolder;
+        using WriteHolder = folly::RWSpinLock::WriteHolder;
+        using UpgradedHolder = folly::RWSpinLock::UpgradedHolder;
+        ReadHolder rh(&lock_);
         auto it = hostsMap_.find(hostAddr);
         if (it == hostsMap_.end()) {
-            folly::RWSpinLock::UpgradedHolder uh(&lock_);
+            WriteHolder wh(UpgradedHolder(&lock_));
```
boshengchen: Like this.

dutor:
Here is a snippet to demo this issue.
```cpp
    folly::RWSpinLock lock;
    auto reader = std::thread([&] () {
        folly::RWSpinLock::ReadHolder rh(lock);
        fprintf(stderr, "read locked\n");
        sleep(5);
        fprintf(stderr, "read unlocked\n");
    });

    auto writer = std::thread([&]() {
        sleep(1);
        folly::RWSpinLock::ReadHolder rh(lock);
        folly::RWSpinLock::UpgradedHolder uh(lock);
        fprintf(stderr, "upgraded\n");
        //rh.reset();
        //folly::RWSpinLock::WriteHolder wh(std::move(uh));
        //fprintf(stderr, "write locked\n");
        sleep(10);
    });

    reader.join();
    writer.join();
```

dutor:
@monadbobo @wadeliuyi @boshengchen 

Please note that the first commit of this PR is buggy. It has a deadlock problem, because the writer-to-be thread, aka. the current thread, is a reader itself. So it must be release the read lock before acquire the write lock.

src/meta/ActiveHostsMan.cpp
```diff
@@ -34,10 +34,20 @@ ActiveHostsMan::ActiveHostsMan(int32_t intervalSeconds, int32_t expiredSeconds,
...
+        using UpgradedHolder = folly::RWSpinLock::UpgradedHolder;
+        // Acquire the read lock
+        ReadHolder rh(&lock_);
         auto it = hostsMap_.find(hostAddr);
         if (it == hostsMap_.end()) {
-            folly::RWSpinLock::UpgradedHolder uh(&lock_);
+            // Step to Upgraded phase to stop new readers
+            UpgradedHolder uh(&lock_);
+            // Unlock the read lock
+            rh.reset();
```
monadbobo: Is hostsMap_  the  single-writer-many-reader?

critical27: It is multi-writer, update by heartbeat.

dutor: No, it's a multi-writer multi-reader case

dutor:
For simplicity, we decide to use `WriteHolder` instead.

If anyone has questions about the `UpgradedHolder`, please contact @dangleptr or me for details.

dangleptr:
Thanks for taking care of it.  @dutor 
It is my mistake to use UpgradeLock in this situation which is not Single-writer-multi-reader.  

For more information about UpgradeLock concept, please refer https://www.boost.org/doc/libs/1_58_0/doc/html/thread/synchronization.html#thread.synchronization.mutex_concepts.upgrade_lockable

PS.  The implementation in folly is different with boost.  
In folly, upgradeLock and sharedLock are mutually exclusive

PR #627, opened 2019-07-15
 fix crash when thrift service port has been used

wadeliuyi:
1: fix crash when thrift service port has been used
2: fix storage crash when raft service port has been used

### DIFF
--- a/src/daemons/GraphDaemon.cpp
+++ b/src/daemons/GraphDaemon.cpp
@@ -7,6 +7,8 @@
 #include "base/Base.h"
 #include "network/NetworkUtils.h"
 #include <signal.h>
+#include <errno.h>
+#include <string.h>
 #include "base/Status.h"
 #include "fs/FileUtils.h"
 #include "process/ProcessUtils.h"
@@ -24,7 +26,7 @@ using nebula::network::NetworkUtils;
 static std::unique_ptr<apache::thrift::ThriftServer> gServer;
 
 static void signalHandler(int sig);
-static void setupSignalHandler();
+static Status setupSignalHandler();
 static Status setupLogging();
 static void printHelp(const char *prog);
 
@@ -130,26 +132,60 @@ int main(int argc, char *argv[]) {
     gServer->setThreadStackSizeMB(5);
 
     // Setup the signal handlers
-    setupSignalHandler();
+    status = setupSignalHandler();
+    if (!status.ok()) {
+        LOG(ERROR) << status;
+        nebula::WebService::stop();
+        return EXIT_FAILURE;
+    }
 
     FLOG_INFO("Starting nebula-graphd on %s:%d\n", localIP.c_str(), FLAGS_port);
     try {
         gServer->serve();  // Blocking wait until shut down via gServer->stop()
     } catch (const std::exception &e) {
+        nebula::WebService::stop();
         FLOG_ERROR("Exception thrown while starting the RPC server: %s", e.what());
         return EXIT_FAILURE;
     }
 
+    nebula::WebService::stop();
     FLOG_INFO("nebula-graphd on %s:%d has been stopped", localIP.c_str(), FLAGS_port);
 
     return EXIT_SUCCESS;
 }
 
 
-void setupSignalHandler() {
-    ::signal(SIGPIPE, SIG_IGN);
-    ::signal(SIGINT, signalHandler);
-    ::signal(SIGTERM, signalHandler);
+Status installHandler(int signum, sighandler_t handler) {
+    // signal() returns the previous value of the signal handler, or SIG_ERR on error.
+    sighandler_t ret = ::signal(signum, handler);
+    if (ret == SIG_ERR) {
+        return Status::Error("Failed to install handler for %d(%s), error: %s", 
+                                signum, ::strsignal(signum), ::strerror(errno));
+    }
+    else {
+        return Status::OK();
+    }
+}
+
+
+Status setupSignalHandler() {
+    Status status = Status::OK();
+    do {
+        status = installHandler(SIGPIPE, SIG_IGN);
+        if (!status.ok())
+            break;
+
+        status = installHandler(SIGINT, signalHandler);
+        if (!status.ok())
+            break;
+
+        status = installHandler(SIGTERM, signalHandler);
+        if (!status.ok())
+            break;
+
+    } while (false);
+
+    return status;
 }
 
 
@@ -158,7 +194,6 @@ void signalHandler(int sig) {
         case SIGINT:
         case SIGTERM:
             FLOG_INFO("Signal %d(%s) received, stopping this server", sig, ::strsignal(sig));
-            nebula::WebService::stop();
             gServer->stop();
             break;
         default:
--- a/src/daemons/GraphDaemon.cpp
+++ b/src/daemons/GraphDaemon.cpp
@@ -5,6 +5,7 @@
  */
 
 #include "base/Base.h"
+#include "common/base/SignalHandler.h"
 #include "network/NetworkUtils.h"
 #include <signal.h>
 #include <errno.h>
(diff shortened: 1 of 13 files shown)


src/daemons/GraphDaemon.cpp
```diff
@@ -141,21 +143,49 @@ int main(int argc, char *argv[]) {
...
     }
 
+    nebula::WebService::stop();
     FLOG_INFO("nebula-graphd on %s:%d has been stopped", localIP.c_str(), FLAGS_port);
 
     return EXIT_SUCCESS;
 }
 
 
+Status installHandler(int signum, sighandler_t handler) {
```
dutor: Please check a signal handler in src/common/SignalHandler

wadeliuyi: > Please check a signal handler in src/common/SignalHandler

I am sorry that I can not find file named SignalHandler in src/common, Do you mean add a new file named SignalHandler in src/common to erase repeat code of “installHandler”?

wadeliuyi:
Jenkins go.

src/daemons/GraphDaemon.cpp
dutor: src/common/base

wadeliuyi: > src/common/base

good point, thx, fixed

wadeliuyi:
Jenkins go.

src/kvstore/raftex/RaftexService.h
```diff
@@ -49,6 +46,23 @@ class RaftexService : public cpp2::RaftexServiceSvIf {
     void removePartition(std::shared_ptr<RaftPart> part);
 
 private:
+    void initThriftServer(std::shared_ptr<folly::IOThreadPoolExecutor> pool, uint16_t port = 0);
+    bool setup();
+    void serve();
+
+    enum RaftServiceStatus{
```
dangleptr: Do we really need the complicated status here?

src/kvstore/raftex/RaftexService.cpp
```diff
@@ -21,57 +21,97 @@ std::shared_ptr<RaftexService> RaftexService::createService(
...
+bool RaftexService::setup() {
+    LOG(INFO) << "Starting the Raftex Service";
 
-        {
-            std::lock_guard<std::mutex> g(svc->readyMutex_);
-            svc->ready_ = true;
-        }
-        svc->readyCV_.notify_all();
+    try {
+        server_->setup();
```
dangleptr: Is the step make the whole process crashed if port is used?

wadeliuyi: yes, server_->setup() will call “AsyncServerSocket::bindSocket”, and this function will throw an exception
if (fsp::bind(fd, saddr, address.getActualSize()) != 0) {
    if (!isExistingSocket) {
      closeNoInt(fd);
    }
    folly::throwSystemError(
        errno, "failed to bind to async server socket: " + address.describe());
  }

then I can see it in the nebula-storaged.INFO:
E0716 14:20:59.741447  2314 ThriftServer.cpp:440] Got an exception while setting up the server: 98failed to bind to async server socket: [::]:18002: Address already in use

dangleptr: So we could handle it in a more simple way.  Just move server_->setup out from the background thread,  and return error if failed.  In waitUntilReady,  after starting up the event loop in bg thread,  we could pend there to check the loop status before returned.  

PS.  You could take the base/test/ServerContext.h as your reference.

src/kvstore/NebulaStore.cpp
```diff
@@ -62,16 +62,19 @@ NebulaStore::~NebulaStore() {
...
 
-void NebulaStore::init() {
+bool NebulaStore::init() {
     LOG(INFO) << "Start the raft service...";
     workers_ = std::make_shared<thread::GenericThreadPool>();
     workers_->start(FLAGS_num_workers);
     raftService_ = raftex::RaftexService::createService(ioPool_, raftAddr_.second);
-    raftService_->waitUntilReady();
+    if (!raftService_->start()) {
+        LOG(INFO) << "Start the raft service failed";
```
darionyaphet: `LOG(ERROR)` will be better ?

src/daemons/GraphDaemon.cpp
```diff
@@ -130,26 +133,35 @@ int main(int argc, char *argv[]) {
     gServer->setThreadStackSizeMB(5);
 
     // Setup the signal handlers
-    setupSignalHandler();
+    status = setupSignalHandler();
+    if (!status.ok()) {
+        LOG(ERROR) << status;
```
darionyaphet: Please append more log info, just display status message maybe confuse.

src/kvstore/raftex/RaftexService.cpp
```diff
@@ -94,6 +148,10 @@ void RaftexService::stop() {
         LOG(INFO) << "All partitions have stopped";
     }
     server_->stop();
+
+    // reclaim resource
+    waitUntilStop();
```
darionyaphet: Does set `status_ ` to `STATUS_STOPPING`  is needed ?

src/kvstore/NebulaStore.cpp
wadeliuyi: good point, thx, I will fix it.

src/kvstore/raftex/RaftexService.cpp
wadeliuyi: I have think about your ideal deeply with see the source of socket setup, and you are right, I will fix it, thanks very much.

src/kvstore/raftex/RaftexService.cpp
wadeliuyi: yes, I will fix it, thx.

src/daemons/GraphDaemon.cpp
wadeliuyi: the status is return from SignalHandler::installInternal, and it will take enough info to get the error reason. thx.

src/kvstore/raftex/RaftexService.cpp
wadeliuyi: I check the function setup again, the function will init event base by “serveEventBase_ = eventBaseManager_->getEventBase();”, and this is thread local variable, so I can not call set up in main thread.
and I will use a atomic_int to replace the “status_ and cv”, thx.

src/kvstore/raftex/RaftexService.cpp
wadeliuyi: yes, I will fix it, thx.

src/kvstore/raftex/RaftexService.h
wadeliuyi: yes, I will fix it, thx.

wadeliuyi:
Jenkins go.

src/kvstore/raftex/RaftexService.cpp
```diff
@@ -94,11 +125,15 @@ void RaftexService::stop() {
...
```
dangleptr: Why move waitUntilStop inside stop.  
Typically, when we want to stop the service.  What we do is stop and waitUntilStop.  In which way, you could do some other  things between stop and waitUntilStop

wadeliuyi: Because I see that we just use raft service in nebula store, and stop it in nebula store destructor, I want make it more simply, then move waitUntilStop in stop.

dangleptr: Not only in NebulaStore.  Raft UTs also use it.

wadeliuyi: fixed

wadeliuyi:
Jenkins go.

wadeliuyi:
Jenkins go

Issue #485, opened 2019-05-31
User varint to represent data version 


dangleptr:
For most cases,  we don't need multi versions data.  So to save space used, we could use varint to represent data version.

sherman-the-tank:
Do you mean the schema version number in the encoded data? That's already varint-like

dangleptr:
> Do you mean the schema version number in the encoded data? That's already varint-like

NO, i mean the version in vertex or edge

sherman-the-tank:
Got it. You meant the version number in the vertex/edge key

zlcook:
**Basic idea** : 

* The create space sentence: 
```
CREATE SPACE  name_label ( 
  [partition_num= INTEGER], 
  [replica_factor = INTEGER], 
  [time_series = {'Y'| 'N'}]
 )
```
`time_series = Y` means that all `Tag/Edge` in `space` totaly support multi versions data. `time_series = N`  is the opposite side.
 
* The Tag's key design:
```
PartitionID + VertexID + TagID + encodeVarint(TagVersion)
```
 If `space` doesn't support multi versions data, the TagVersion is always 0.

* The Edge's key is similar to Tag

whitewum:
is it possible to change from Y to N ? or N to Y.  1. prevent this change happens.  or 2. use compaction.
is Y or N as a default value?

zlcook:
> is it possible to change from Y to N ? or N to Y. 1. prevent this change happens. or 2. use compaction.
WDYT @dangleptr ? 
> is Y or N as a default value?
The default value is `N`

whitewum:
> > is it possible to change from Y to N ? or N to Y. 1. prevent this change happens. or 2. use compaction.
> 
> WDYT @dangleptr ?
> 
> > is Y or N as a default value?
> 
> The default value is `N`

remember add a case to cover these situations. both 1 and 2.

dangleptr:
Looks good to me. 
We should be careful about changes in NebulaKeyUtils,  we have used some fixed length of vertex or edge in some places.

sherman-the-tank:
@zlcook The design sounds good to me. Let's do it!!

PR #628, opened 2019-07-15
use varint to represent data version in the key

zlcook:
As title, Refer to Issue #485


zlcook:
jenkins go!

src/storage/test/AdHocSchemaManager.cpp
```diff
@@ -9,6 +9,22 @@
...
+}
+
+bool AdHocSchemaManager::isSupportTimeSeries(GraphSpaceID space) {
+    folly::RWSpinLock::ReadHolder rh(seriesLock_);
+    auto it = timeSeries_.find(space);
+    if (it == timeSeries_.end()) {
+        // Not found
+        return false;
+    } else {
+        return timeSeries_[space];
```
wadeliuyi: maybe you can return iter->second directly

zlcook: Good catch :)

src/common/base/NebulaKeyUtils.inl
```diff
@@ -0,0 +1,37 @@
...
+namespace nebula {
+
+template<typename T>
+typename std::enable_if<std::is_integral<T>::value, int32_t>::type
+NebulaKeyUtils::encodeVersion(T v, uint8_t *encode) {
+    size_t len = folly::encodeVarint(v, encode);
+    DCHECK_GT(len, 0UL);
+    uint8_t temp;
+    // Keep the original order
+    for (size_t i = 0; i < len / 2; i++) {
```
critical27: Why do we need to reverse the order both in encode and decode?

zlcook: > Why do we need to reverse the order both in encode and decode?

Sorry for the late reply! 
The reason：
* 300 is greater than 200, however, the numbers(ps: 200, 300) encoded by `folly::encodeVarint` does not guarantee the original order.  In order to maintain the original order, reverse the byte order.  [folly::encodeVarint principle](https://developers.google.com/protocol-buffers/docs/encoding?csw=1#varints)

* The `key` in RocksDb sorted in lexicographic order.

src/common/base/NebulaKeyUtils.inl
```diff
@@ -0,0 +1,37 @@
...
+        encode[len-i-1] = temp;
+    }
+    return len;
+}
+
+template<typename T>
+typename std::enable_if<std::is_integral<T>::value>::type
+NebulaKeyUtils::decodeVersion(const char* data, int32_t len, T& v) {
+    const uint8_t* encode = reinterpret_cast<const uint8_t*>(data);
+    uint8_t copy[10];
```
wadeliuyi: Is it sure that len is less than 10?

zlcook: Yes, the encoded value of the 8-byte largest integer occupies up to 10 bytes.

zlcook:
Jenkins go

dangleptr:
Close it.  We don't change the format of our key in 1.x any more.  Thanks~

PR #630, opened 2019-07-15
Fixed false negative report when variable holds empty result

dutor:
As title.


dutor:
> Could you add a test to non-existent var

Sure.

PR #635, opened 2019-07-15
update query grammar

Amber1990Zhang:
Update query grammar, `[]`is ommitted and a `.` is added, eg:
the origin: $$[student].name
the new one: $$.student.name

Vertex ID can be generated with hash or specified manually.


Amber1990Zhang:
Please review quickly so I can post to the web. @dutor @whitewum @dangleptr

PR #640, opened 2019-07-16
Fix get wrong data when multi tags have same prop name

### LINKED ISSUE #581: Return value not correct when a vertex has same name in multiple tags
First, I create three tags and an edge
```
CREATE TAG course(name string, credits int)
CREATE TAG student(name string, age int, gender string)
CREATE EDGE select(grade int)
```
Then, insert the following data
```
INSERT VERTEX course(name, credits),building(name) VALUES 102:("English", 6, "No11");
INSERT VERTEX student(name, age, gender) VALUES 202:("Jane", 17, "female");
INSERT EDGE select (grade) VALUES 202 -> 102:(3);
```
when query
```
GO FROM 202 OVER select YIELD $^[student].name AS Student, $$[course].name AS Curriculum, select.grade AS Grade, $$[building].name AS location
```

return unexpected Curriculum

laura-ding:
close #581 
Before send to storaged, sort properties by tag, and read them in the order they are sent.


src/graph/GoExecutor.h
```diff
@@ -154,13 +154,17 @@ class GoExecutor final : public TraverseExecutor {
...
     public:
-        VariantType get(VertexID id, const std::string &prop) const;
+        VariantType get(VertexID id, int64_t index) const;
         void add(const storage::cpp2::QueryResponse &resp);
         const auto* schema() const {
             return schema_.get();
         }
 
     private:
+        // The schema include multi vertexes, and multi tags of one vertex
```
zlcook: Typo:  vertexes -> vertices

src/graph/GoExecutor.h
```diff
@@ -154,13 +154,17 @@ class GoExecutor final : public TraverseExecutor {
...
+        // eg: get 3 vertexex, vertex A has tag1.prop1, vertex B has tag2.prop2,
```
zlcook: ditto

src/graph/GoExecutor.h
laura-ding: `vertexes` and `vertices` are all right

laura-ding:
jenkins go

src/graph/test/DataTest.cpp
```diff
@@ -106,6 +106,26 @@ AssertionResult DataTest::prepareSchema() {
                                << " failed, error code "<< static_cast<int32_t>(code);
         }
     }
+    // Test same propName diff tyep in diff tags
```
whitewum: typo

src/graph/test/DataTest.cpp
```diff
@@ -106,6 +106,26 @@ AssertionResult DataTest::prepareSchema() {
...
+    // Test same propName same type in diff tags
+    {
+        cpp2::ExecutionResponse resp;
+        std::string cmd = "CREATE TAG interest(name string)";
+        auto code = client_->execute(cmd, resp);
+        if (cpp2::ErrorCode::SUCCEEDED != code) {
+            return TestError() << "Do cmd:" << cmd
+                               << " failed, error code "<< static_cast<int32_t>(code);
+        }
+    }
```
whitewum: 1. how about use for-loop to create? because I don't see any case to cover: add 10000000 tags/edge_types.
```
for ()

2. create tag xxx+i (name int, name2 string)

2. create tag xxy+i (name string, name2 int)
```

src/graph/test/DataTest.cpp
```diff
@@ -269,6 +289,62 @@ TEST_F(DataTest, InsertVertex) {
...
+        ASSERT_EQ(cpp2::ErrorCode::SUCCEEDED, code);
+        std::vector<std::tuple<std::string, int64_t, std::string, int64_t, int64_t>> expected = {
+            {"Joy", 90, "Petter", 19, 456},
+        };
+        ASSERT_TRUE(verifyResult(resp, expected));
+    }
+    // Test same prop name same type in diff tags
+    {
+        cpp2::ExecutionResponse resp;
+        std::string cmd = "INSERT VERTEX person(name, age),interest(name) "
```
whitewum: for this test case:

insert vertex person(name, age), interest(name)  

/* 1. can I miss any field? e.g. person(name), interest(name) 
if no, please add a case*/

"VALUES hash(\"Bob\"):(\"Bob\", 19, \"basketball\")";

/* 2. can I miss any filed? e.g."VALUES hash(\"Bob\"):(\"Bob\", /*19,*/ \"basketball\")";

if no, please add a case */

PR #646, opened 2019-07-17
wait for metaserer ready

andychow01:
When storage server starts and query engine init, they should wait for meta server ready(in a loop)，then do other things!

### DIFF
--- a/src/graph/ExecutionEngine.cpp
+++ b/src/graph/ExecutionEngine.cpp
@@ -29,7 +29,10 @@ Status ExecutionEngine::init(std::shared_ptr<folly::IOThreadPoolExecutor> ioExec
         return addrs.status();
     }
     metaClient_ = std::make_unique<meta::MetaClient>(ioExecutor, std::move(addrs.value()));
-    metaClient_->init();
+    if (!metaClient_->waitForMetadReady()) {
+        LOG(ERROR) << "ExecutionEngine::init waitForMetadReady failed";
+        return Status::Error("ExecutionEngine::init waitForMetadReady failed");
+    }
 
     schemaManager_ = meta::SchemaManager::create();
     schemaManager_->init(metaClient_.get());
--- a/src/graph/test/TestEnv.cpp
+++ b/src/graph/test/TestEnv.cpp
@@ -39,15 +39,28 @@ void TestEnv::SetUp() {
     auto addrsRet
         = network::NetworkUtils::toHosts(folly::stringPrintf("127.0.0.1:%d", metaServerPort()));
     CHECK(addrsRet.ok()) << addrsRet.status();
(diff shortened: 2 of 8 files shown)


src/graph/ExecutionEngine.cpp
```diff
@@ -29,7 +29,10 @@ Status ExecutionEngine::init(std::shared_ptr<folly::IOThreadPoolExecutor> ioExec
         return addrs.status();
     }
     metaClient_ = std::make_unique<meta::MetaClient>(ioExecutor, std::move(addrs.value()));
-    metaClient_->init();
+    if (!metaClient_->waitForMetadReady()) {
```
wadeliuyi: the function is always return true, is this check really need?

andychow01: Now, because there is no clusterId, so this function aways return true. The next PR which support clusterId will change this code. Thanks!

src/storage/test/StorageClientTest.cpp
```diff
@@ -42,14 +42,24 @@ TEST(StorageClientTest, VerticesInterfacesTest) {
...
+    auto& localHost = hostRet.value();
     auto mClient
-        = std::make_unique<meta::MetaClient>(threadPool, std::move(addrsRet.value()), true);
-    mClient->init();
+        = std::make_unique<meta::MetaClient>(threadPool, std::move(addrs), localHost, true);
+    LOG(INFO) << "Add hosts and create space....";
+    auto r = mClient->addHosts({HostAddr(localIp, localDataPort)}).get();
+    ASSERT_TRUE(r.ok());
+    mClient->waitForMetadReady();
+    while (meta::ActiveHostsMan::instance()->getActiveHosts().size() == 0) {
```
critical27: Do we need this? If MetaClient has send heartbeat successfully, it should have been an active host.

src/graph/test/TestEnv.cpp
```diff
@@ -39,15 +39,28 @@ void TestEnv::SetUp() {
...
+    auto hostRet = nebula::network::NetworkUtils::toHostAddr("127.0.0.1", storagePort);
+    if (!hostRet.ok()) {
+        LOG(ERROR) << "Bad local host addr, status:" << hostRet.status();
+    }
+    auto& localhost = hostRet.value();
+    mClient_ = std::make_unique<meta::MetaClient>(threadPool,
+                                                  std::move(addrsRet.value()),
+                                                  localhost,
+                                                  true);
+    auto r = mClient_->addHosts({localhost}).get();
```
critical27: Why we add a host and remove it later? waitForMetadReady need it for what?

andychow01: If we do not remove it, later it will affect the next test. Some test case request there is no host!

PR #647, opened 2019-07-17
wait for metaserer ready

andychow01:


src/meta/client/MetaClient.cpp
```diff
@@ -39,8 +41,22 @@ MetaClient::~MetaClient() {
 }
 
 
-void MetaClient::init() {
-    loadDataThreadFunc();
+bool MetaClient::waitForMetadReady() {
+    while (!ready_) {
```
dutor: The caller has no way to abort this retry loop. This is not appropriate.

andychow01: Now, the code has changed! Please review! Thanks!

src/meta/client/MetaClient.cpp
```diff
@@ -39,8 +41,27 @@ MetaClient::~MetaClient() {
...
+}
+
+
+bool MetaClient::waitForMetadReady(int count, int retryIntervalSecs) {
+    int tryCount = count;
+    while (!isMetadReady() && ((count == -1) || (tryCount > 0))) {
+        --tryCount;
+        ::sleep(retryIntervalSecs);
+    }  // end while
+
```
wadeliuyi: when count > 0, and ready_ is false after break the loop, we need start bgThread_ too?

src/graph/ExecutionEngine.cpp
```diff
@@ -29,7 +29,11 @@ Status ExecutionEngine::init(std::shared_ptr<folly::IOThreadPoolExecutor> ioExec
         return addrs.status();
     }
     metaClient_ = std::make_unique<meta::MetaClient>(ioExecutor, std::move(addrs.value()));
-    metaClient_->init();
+    // load data try 3 time
+    bool loadDataOk = metaClient_->waitForMetadReady(3);
+    if (!loadDataOk) {
+        LOG(ERROR) << "ExecutionEngine::init loadData by thread error!";
```
wadeliuyi: storaged return failed when load meta data failed, but why here not?

dutor: return on failure.

src/meta/client/MetaClient.cpp
```diff
@@ -39,8 +41,27 @@ MetaClient::~MetaClient() {
...
+            ready_ = false;
+            return ready_;
+        }
+    }  // end if
     loadDataThreadFunc();
+    return ready_;
+}
+
+
+bool MetaClient::waitForMetadReady(int count, int retryIntervalSecs) {
```
dutor: On what condition does this function fail?

src/graph/ExecutionEngine.cpp
darionyaphet: when the data loading failed we should return ?

darionyaphet:
Basically LGTM  🎉  Please look over the inline comments :)

src/meta/client/MetaClient.cpp
andychow01: Next version, we will add clusterId support, when mete client meets wrong clusterId, then will fail.
This version, it will be all success!

src/graph/ExecutionEngine.cpp
andychow01: Now, the code has changed!

andychow01: When load data error, then return error!

andychow01: Now, when load data error, then return error!

PR #649, opened 2019-07-17
Remove requests queue inside Host

dangleptr:
We don't need to hold all requests inside host.  There are two defects:
1.  Catch up data will be slow,  because we have to send request one by one.
2.  More memory  occupied due to the queue.

### DIFF
--- a/src/kvstore/raftex/Host.cpp
+++ b/src/kvstore/raftex/Host.cpp
@@ -27,9 +27,11 @@ Host::Host(const HostAddr& addr, std::shared_ptr<RaftPart> part, bool isLearner)
         , addr_(addr)
         , isLearner_(isLearner)
         , idStr_(folly::stringPrintf(
-            "[Host: %s:%d] ",
+            "%s[Host: %s:%d] ",
+            part_->idStr_.c_str(),
             NetworkUtils::intToIPv4(addr_.first).c_str(),
-            addr_.second)) {
+            addr_.second))
+        , cachingPromise_(folly::SharedPromise<cpp2::AppendLogResponse>()) {
 }
 
 
@@ -44,8 +46,8 @@ void Host::waitForStop() {
 }
 
 
-cpp2::ErrorCode Host::checkStatus(std::lock_guard<std::mutex>& lck) const {
-    UNUSED(lck);
+cpp2::ErrorCode Host::checkStatus() const {
+    CHECK(!lock_.try_lock());
     if (stopped_) {
         VLOG(2) << idStr_ << "The host is stopped, just return";
         return cpp2::ErrorCode::E_HOST_STOPPED;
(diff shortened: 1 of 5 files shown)


dangleptr:
Jenkins go

src/kvstore/raftex/Host.cpp
```diff
@@ -106,18 +107,13 @@ folly::Future<cpp2::AppendLogResponse> Host::appendLogs(
...
-                                    committedLogId,
-                                    lastLogTermSent,
-                                    lastLogIdSent)));
-                return requests_.back().first.getFuture();
+            if (cachingPromise_.size() <= FLAGS_max_outstanding_requests) {
+                pendingReq_ = std::make_tuple(term,
+                                              logId,
+                                              committedLogId,
+                                              lastLogTermSent,
+                                              lastLogIdSent);
```
wadeliuyi: I think the caller’s parameter name “currTerm, lastLogId, committedId, prevLogTerm, prevLogId” is more readable than here.

src/kvstore/raftex/Host.cpp
```diff
@@ -185,8 +182,8 @@ folly::Future<cpp2::AppendLogResponse> Host::appendLogsInternal(
             {
                 std::lock_guard<std::mutex> g(self->lock_);
                 self->promise_.setValue(r);
+                self->cachingPromise_.setValue(r);
```
wadeliuyi: why the cached request no need retry?

src/kvstore/raftex/Host.cpp
dangleptr: prevLogTerm, prevLogId sounds good

src/kvstore/raftex/Host.cpp
dangleptr: Next request will fill the gap.  So we don't need retry here.

src/kvstore/raftex/Host.cpp
```diff
@@ -138,20 +135,21 @@ folly::Future<cpp2::AppendLogResponse> Host::appendLogs(
...
         logIdToSend_ = logId;
-        lastLogTermSent_ = lastLogTermSent;
-        lastLogIdSent_ = lastLogIdSent;
+        lastLogTermSent_ = prevLogTerm;
+        lastLogIdSent_ = prevLogId;
         committedLogId_ = committedLogId;
 
-        promise_ = folly::Promise<cpp2::AppendLogResponse>();
+        promise_ = std::move(cachingPromise_);
+        cachingPromise_ = folly::SharedPromise<cpp2::AppendLogResponse>();
```
critical27: Do we need to reset pendingReq_ here?

dangleptr: Good catch!

PR #652, opened 2019-07-18
Fix the same name tag and edge in same space #598

### LINKED ISSUE #598: Can tag name be identical to an edge type name within the same space
I defined an edge type named _player2_ , then created a tag named _player2_ in the same space successfully.

image

Should the name be unique within one space?

laura-ding:
close #598

### DIFF
--- a/src/graph/ShowExecutor.cpp
+++ b/src/graph/ShowExecutor.cpp
@@ -164,13 +164,19 @@ void ShowExecutor::showTags() {
         std::vector<std::string> header{"Name"};
         resp_->set_column_names(std::move(header));
 
+        std::set<std::string> uniqueTags;
         for (auto &tag : value) {
+            uniqueTags.emplace(tag.get_tag_name());
+        }
+
+        for (auto &item : uniqueTags) {
             std::vector<cpp2::ColumnValue> row;
             row.resize(1);
-            row[0].set_str(tag.get_tag_name());
+            row[0].set_str(item);
             rows.emplace_back();
             rows.back().set_columns(std::move(row));
         }
+
         resp_->set_rows(std::move(rows));
         DCHECK(onFinish_);
         onFinish_();
@@ -204,10 +210,15 @@ void ShowExecutor::showEdges() {
         std::vector<std::string> header{"Name"};
         resp_->set_column_names(std::move(header));
 
+        std::set<std::string> uniqueEdges;
         for (auto &edge : value) {
+            uniqueEdges.emplace(edge.get_edge_name());
+        }
+
+        for (auto &item : uniqueEdges) {
             std::vector<cpp2::ColumnValue> row;
             row.resize(1);
-            row[0].set_str(edge.get_edge_name());
+            row[0].set_str(item);
             rows.emplace_back();
             rows.back().set_columns(std::move(row));
         }
--- a/src/interface/meta.thrift
+++ b/src/interface/meta.thrift
@@ -11,24 +11,25 @@ namespace go nebula.meta
 include "common.thrift"
 
 enum ErrorCode {
-    SUCCEEDED = 0,
+    SUCCEEDED          = 0,
 
     // RPC Failure
-    E_DISCONNECTED = -1,
-    E_FAIL_TO_CONNECT = -2,
-    E_RPC_FAILURE = -3,
+    E_DISCONNECTED     = -1,
+    E_FAIL_TO_CONNECT  = -2,
+    E_RPC_FAILURE      = -3,
 
-    E_LEADER_CHANGED = -11,
+    E_LEADER_CHANGED   = -11,
 
     // Operation Failure
-    E_NO_HOSTS       = -21,
-    E_EXISTED        = -22,
-    E_NOT_FOUND      = -23,
-    E_INVALID_HOST   = -24,
-    E_UNSUPPORTED    = -25,
-    E_NOT_DROP       = -26,
+    E_NO_HOSTS         = -21,
+    E_EXISTED          = -22,
+    E_NOT_FOUND        = -23,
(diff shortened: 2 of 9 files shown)


laura-ding:
jenkins go

src/graph/ShowExecutor.cpp
```diff
@@ -164,13 +164,19 @@ void ShowExecutor::showTags() {
         std::vector<std::string> header{"Name"};
         resp_->set_column_names(std::move(header));
 
+        std::set<std::string> uniqueTags;
```
whitewum: i think the order doesn't matter. Am I right?

src/interface/meta.thrift
```diff
@@ -11,23 +11,24 @@ namespace go nebula.meta
 include "common.thrift"
 
 enum ErrorCode {
-    SUCCEEDED = 0,
+    SUCCEEDED          = 0,
```
whitewum: great

src/interface/meta.thrift
```diff
@@ -11,23 +11,24 @@ namespace go nebula.meta
...
-    E_UNSUPPORTED    = -25,
-    E_NOT_DROP       = -26,
+    E_NO_HOSTS         = -21,
+    E_EXISTED          = -22,
+    E_NOT_FOUND        = -23,
+    E_INVALID_HOST     = -24,
+    E_UNSUPPORTED      = -25,
+    E_NOT_DROP         = -26,
     E_BALANCER_RUNNING = -27,
+    E_CONFLICT         = -28,
```
whitewum: why a new errorcode? how about existed? (what about mysql? table vs view)

src/graph/ShowExecutor.cpp
laura-ding: Users can get more friendly read when showing tags or edges

src/meta/processors/schemaMan/CreateEdgeProcessor.cpp
```diff
@@ -11,6 +11,18 @@ namespace meta {
 
 void CreateEdgeProcessor::process(const cpp2::CreateEdgeReq& req) {
     CHECK_SPACE_ID_AND_RETURN(req.get_space_id());
+    {
+        folly::SharedMutex::ReadHolder rHolder(LockUtils::tagLock());
+        auto conflictRet = getTagId(req.get_space_id(), req.get_edge_name());
+        if (conflictRet.ok()) {
+            LOG(ERROR) << "Create Edge Failed :" << req.get_edge_name() << " has same name tag";
```
dutor: No space before `:`

_Failed to create edge `edge_type`: already exists.

src/meta/processors/schemaMan/CreateTagProcessor.cpp
```diff
@@ -11,17 +11,31 @@ namespace meta {
 
 void CreateTagProcessor::process(const cpp2::CreateTagReq& req) {
     CHECK_SPACE_ID_AND_RETURN(req.get_space_id());
+    {
+        // if there is an edge of the same name
+        folly::SharedMutex::ReadHolder rHolder(LockUtils::edgeLock());
+        auto conflictRet = getEdgeType(req.get_space_id(), req.get_tag_name());
+        if (conflictRet.ok()) {
+            LOG(ERROR) << "Create Tag Failed :" << req.get_tag_name() << " has same name edge";
```
dutor: ditto.

src/meta/processors/schemaMan/CreateTagProcessor.cpp
```diff
@@ -11,17 +11,31 @@ namespace meta {
 
 void CreateTagProcessor::process(const cpp2::CreateTagReq& req) {
     CHECK_SPACE_ID_AND_RETURN(req.get_space_id());
+    {
+        // if there is an edge of the same name
```
dutor: Also add this comment to the edge creation.

src/meta/processors/schemaMan/CreateEdgeProcessor.cpp
```diff
@@ -11,6 +11,20 @@ namespace meta {
 
 void CreateEdgeProcessor::process(const cpp2::CreateEdgeReq& req) {
     CHECK_SPACE_ID_AND_RETURN(req.get_space_id());
+    {
+        // if there is an tag of the same name
+        folly::SharedMutex::ReadHolder rHolder(LockUtils::tagLock());
```
wadeliuyi: In multithread env, we can not separate this check in two lock, because between read.unlock and write.lock, other thread may insert some same record.

src/meta/processors/schemaMan/CreateTagProcessor.cpp
```diff
@@ -11,17 +11,32 @@ namespace meta {
...
+        folly::SharedMutex::ReadHolder rHolder(LockUtils::edgeLock());
```
wadeliuyi: ditto

src/meta/processors/schemaMan/CreateEdgeProcessor.cpp
laura-ding: Good catch! I have to think about it, I think to use the same lock to do it, but it will handle slower in tag and edge. It is hard to happen when thread A creates tag tagName, thread B create edge tagName at the same timestamp.

dutor: @wadeliuyi We might as well leave this potential race condition problem to address in the future.

@laura-ding Better to add a TODO here.

wadeliuyi: but "hard to happen" not mean "do not happen", and create tag and edge is not the key path of our program, so performance is not key point?

PR #654, opened 2019-07-18
Support Ingest SST Files

darionyaphet:
In this PR, you can ingest the SST files after download from `HDFS`.

`INGEST` command can be used in console.


src/meta/MetaHttpIngestHandler.cpp
```diff
@@ -0,0 +1,146 @@
...
+bool MetaHttpIngestHandler::ingestSSTFiles(GraphSpaceID space, const std::string& path) {
+    std::unique_ptr<kvstore::KVIterator> iter;
+    auto prefix = MetaServiceUtils::partPrefix(space);
+    auto ret = kvstore_->prefix(0, 0, prefix, &iter);
+    if (ret != kvstore::ResultCode::SUCCEEDED) {
+        LOG(ERROR) << "Fetch Parts Failed";
+        return false;
+    }
+
+    std::vector<std::string> storageIPs;
```
wadeliuyi: why not use set?

src/kvstore/NebulaStore.cpp
```diff
@@ -387,23 +387,19 @@ void NebulaStore::asyncRemovePrefix(GraphSpaceID spaceId,
...
+                               const std::string& extra) {
     decltype(spaces_)::iterator it;
     folly::RWSpinLock::ReadHolder rh(&lock_);
     RETURN_IF_SPACE_NOT_FOUND(spaceId, it);
     for (auto& engine : it->second->engines_) {
-        auto parts = engine->allParts();
         std::vector<std::string> extras;
+        auto parts = nebula::fs::FileUtils::listAllDirsInDir(extra.c_str());
         for (auto part : parts) {
+            auto path = folly::stringPrintf("%s/%s", extra.c_str(), part.c_str());
```
wadeliuyi: maybe we should separate the different part to different engine, like nebula store init?

src/graph/IngestExecutor.cpp
```diff
@@ -0,0 +1,79 @@
...
+
+void IngestExecutor::execute() {
+    auto *mc = ectx()->getMetaClient();
+    auto  addresses = mc->getAddresses();
+    auto  metaHost = network::NetworkUtils::intToIPv4(addresses[0].first);
+    auto  spaceId = ectx()->rctx()->session()->space();
+    auto *path  = sentence_->path();
+
+    auto func = [metaHost, path, spaceId]() {
+        auto tmp = "/usr/bin/curl -G \"http://%s:%d/%s?path=%s&space=%d\"";
```
wadeliuyi: here use auto is not good.

src/meta/MetaHttpIngestHandler.cpp
darionyaphet: Search operations have logarithmic complexity. It's better

src/meta/MetaHttpIngestHandler.cpp
```diff
@@ -0,0 +1,146 @@
...
+#include "meta/MetaServiceUtils.h"
+#include "meta/MetaHttpIngestHandler.h"
+#include "webservice/Common.h"
+#include "network/NetworkUtils.h"
+#include "process/ProcessUtils.h"
+#include <proxygen/httpserver/RequestHandler.h>
+#include <proxygen/lib/http/ProxygenErrorEnum.h>
+#include <proxygen/httpserver/ResponseBuilder.h>
+
+DEFINE_int32(storage_ingest_http_port, 12000, "Storage daemon's http port");
```
dangleptr: ditto

src/kvstore/NebulaStore.cpp
darionyaphet: Good Catch ! I missed

src/meta/MetaHttpIngestHandler.cpp
```diff
@@ -0,0 +1,146 @@
...
+            auto storageIP = network::NetworkUtils::intToIPv4(host.get_ip());
+            if (std::find(storageIPs.begin(), storageIPs.end(), storageIP) == storageIPs.end()) {
+                storageIPs.push_back(std::move(storageIP));
+            }
+        }
+        iter->next();
+    }
+
+    std::atomic<int> completed(0);
+    std::vector<std::thread> threads;
```
wadeliuyi: I think it is more friendly that use GenericThreadPool to replace threads, if the cluster has more than 1000’s server, this will waste too many cpu resource, and more dangle is that meta server is center of our cluster.

src/storage/StorageHttpIngestHandler.cpp
```diff
@@ -0,0 +1,103 @@
...
+                .sendWithEOM();
+            return;
+        default:
+            break;
+    }
+
+    if (ingestSSTFiles(space_, path_)) {
+        LOG(ERROR) << "SSTFile ingest successfully " << path_;
+        ResponseBuilder(downstream_)
+            .status(200, "SSTFile ingest successfully")
```
wadeliuyi: why not define all of the magic string and code in a common file? because a lot of place use the magic string

darionyaphet: Good Idea. I will move the status code into `webservice/Common.h`  thanks  👍

src/storage/test/CompactionTest.cpp
```diff
@@ -16,7 +16,7 @@ namespace nebula {
 namespace storage {
 
 void mockData(kvstore::KVStore* kv) {
-    for (auto partId = 0; partId < 3; partId++) {
+    for (auto partId = 1; partId <= 3; partId++) {
```
wadeliuyi: convert auto to int

src/meta/MetaHttpIngestHandler.cpp
```diff
@@ -0,0 +1,168 @@
...
+            auto command = folly::stringPrintf("/usr/bin/curl -G \"%s\"", url.c_str());
+            LOG(INFO) << "Command: " << command;
+            auto ingestResult = ProcessUtils::runCommand(command.c_str());
+            return ingestResult.ok() && ingestResult.value() == "SSTFile ingest successfully";
+        };
+        auto future = pool.addTask(dispatcher);
+        futures.push_back(std::move(future));
+    }
+
+    std::atomic<bool> successfully{true};
```
wadeliuyi: here no need atomic, the wait() will synchronize the pool and current thread to access successfully

src/meta/MetaHttpIngestHandler.cpp
```diff
@@ -0,0 +1,168 @@
...
+            auto storageIP = network::NetworkUtils::intToIPv4(host.get_ip());
+            if (std::find(storageIPs.begin(), storageIPs.end(), storageIP) == storageIPs.end()) {
+                storageIPs.insert(std::move(storageIP));
+            }
+        }
+        iter->next();
+    }
+
+    std::vector<folly::SemiFuture<bool>> futures;
+    static nebula::thread::GenericThreadPool pool;
```
wadeliuyi: if we want to stop the meta service, how can we stop the pool?

darionyaphet: when running `GenericThreadPool`'s destructor , `stop()` will be called.

src/meta/MetaHttpDownloadHandler.cpp
```diff
@@ -165,34 +174,50 @@ bool MetaHttpDownloadHandler::dispatchSSTFiles(const std::string& hdfsHost,
         return false;
     }
 
-    std::atomic<int> completed(0);
-    std::vector<std::thread> threads;
+    std::vector<folly::SemiFuture<bool>> futures;
+    static nebula::thread::GenericThreadPool pool;
+    std::call_once(downloadPoolStartFlag, []() {
```
critical27: Maybe you can define downloadPoolStartFlag with `static`, no need to define as global, just a suggestion.

src/kvstore/NebulaStore.cpp
```diff
@@ -387,27 +387,31 @@ void NebulaStore::asyncRemovePrefix(GraphSpaceID spaceId,
...
     RETURN_IF_SPACE_NOT_FOUND(spaceId, it);
     for (auto& engine : it->second->engines_) {
         auto parts = engine->allParts();
         std::vector<std::string> extras;
         for (auto part : parts) {
+            LOG(INFO) << "Loading Part " << part;
+            auto path = folly::stringPrintf("%s/%d", extra.c_str(), part);
+            auto files = nebula::fs::FileUtils::listAllFilesInDir(path.c_str(), true, "*.sst");
+
+            if (files.size() == 0) {
```
critical27: Do we always have sst files in a partition? Maybe some partitions have, some don't, WDYT.

darionyaphet: Good catch  I should check path before list files.

src/meta/MetaHttpDownloadHandler.cpp
darionyaphet: 👍  fixed

src/graph/IngestExecutor.cpp
```diff
@@ -0,0 +1,78 @@
...
+}
+
+void IngestExecutor::execute() {
+    auto *mc = ectx()->getMetaClient();
+    auto  addresses = mc->getAddresses();
+    auto  metaHost = network::NetworkUtils::intToIPv4(addresses[0].first);
+    auto  spaceId = ectx()->rctx()->session()->space();
+
+    auto func = [metaHost, spaceId]() {
+        std::string tmp = "http://%s:%d/%s?space=%d";
```
wadeliuyi: static const char * is better?

src/meta/MetaHttpDownloadHandler.cpp
```diff
@@ -164,34 +172,43 @@ bool MetaHttpDownloadHandler::dispatchSSTFiles(const std::string& hdfsHost,
...
+        auto future = pool_->addTask(dispatcher);
+        futures.push_back(std::move(future));
     }
 
-    for (auto &thread : threads) {
-        thread.join();
-    }
-    return completed == (int32_t)hostPartition.size();
+    bool successfully{true};
+    folly::collectAll(futures).then([&](const std::vector<folly::Try<bool>>& tries) {
```
wadeliuyi: std::move(futures)

src/meta/MetaHttpIngestHandler.cpp
```diff
@@ -0,0 +1,161 @@
...
+                                           FLAGS_ws_storage_http_port, space);
+            auto ingestResult = nebula::http::HttpClient::get(url);
+            return ingestResult.ok() && ingestResult.value() == "SSTFile ingest successfully";
+        };
+        auto future = pool_->addTask(dispatcher);
+        futures.push_back(std::move(future));
+    }
+
+    bool successfully{true};
+    folly::collectAll(futures).then([&](const std::vector<folly::Try<bool>>& tries) {
```
wadeliuyi: ditto

src/meta/MetaHttpIngestHandler.cpp
```diff
@@ -0,0 +1,161 @@
...
+
+void MetaHttpIngestHandler::onError(ProxygenError error) noexcept {
+    LOG(ERROR) << "Web Service MetaHttpIngestHandler got error : "
+               << proxygen::getErrorString(error);
+}
+
+bool MetaHttpIngestHandler::ingestSSTFiles(GraphSpaceID space) {
+    std::unique_ptr<kvstore::KVIterator> iter;
+    auto prefix = MetaServiceUtils::partPrefix(space);
+    auto ret = kvstore_->prefix(0, 0, prefix, &iter);
```
wadeliuyi: do not use 0

src/storage/StorageHttpDownloadHandler.cpp
```diff
@@ -126,45 +135,44 @@ void StorageHttpDownloadHandler::onError(ProxygenError error) noexcept {
...
-    });
 
     for (auto& part : parts) {
-        auto downloader = [hdfsHost, hdfsPort, hdfsPath, localPath, part, this]() {
-            int32_t partInt;
+        auto downloader = [hdfsHost, hdfsPort, hdfsPath, part, this]() {
+            PartitionID partId;
             try {
-                partInt = folly::to<int32_t>(part);
+                partId = folly::to<PartitionID>(part);
```
wadeliuyi: move out of this lambda

src/meta/MetaHttpIngestHandler.cpp
```diff
@@ -0,0 +1,161 @@
...
+    if (ret != kvstore::ResultCode::SUCCEEDED) {
+        LOG(ERROR) << "Fetch Parts Failed";
+        return false;
+    }
+
+    std::set<std::string> storageIPs;
+    while (iter->valid()) {
+        for (auto host : MetaServiceUtils::parsePartVal(iter->val())) {
```
wadeliuyi: auto &host

src/graph/IngestExecutor.cpp
darionyaphet: use `char *` could avoid to call `c_str()`

src/kvstore/NebulaStore.cpp
```diff
@@ -542,6 +550,25 @@ ErrorOr<ResultCode, std::shared_ptr<SpacePartInfo>> NebulaStore::space(GraphSpac
     return it->second;
 }
 
+ErrorOr<ResultCode, std::string> NebulaStore::getDataPath(GraphSpaceID spaceId,
+                                                          PartitionID partId) {
+    folly::RWSpinLock::ReadHolder rh(&lock_);
```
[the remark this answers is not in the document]
darionyaphet: It's seems ok and which I refer to is engine()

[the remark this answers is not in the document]
darionyaphet: And seems I don't need `std::shard_ptr<Part>` just to check if space & part is existed :)

src/storage/test/StorageHttpDownloadHandlerTest.cpp
```diff
@@ -26,46 +26,70 @@ class StorageHttpDownloadHandlerTestEnv : public ::testing::Environment {
...
+        pool_ = std::make_unique<nebula::thread::GenericThreadPool>();
+        pool_->start(1);
+        auto poolPtr = pool_.get();
+
         VLOG(1) << "Starting web service...";
-        WebService::registerHandler("/download", [] {
+        WebService::registerHandler("/download", [poolPtr, this] {
             auto handler =  new storage::StorageHttpDownloadHandler();
-            handler->init(helper.get());
+            handler->init(helper.get(), poolPtr, kv_.get());
```
dangleptr: Why download depends on kvstore?

src/graph/IngestExecutor.cpp
```diff
@@ -0,0 +1,80 @@
...
+
+void IngestExecutor::execute() {
+    auto *mc = ectx()->getMetaClient();
+    auto  addresses = mc->getAddresses();
+    auto  metaHost = network::NetworkUtils::intToIPv4(addresses[0].first);
+    auto  spaceId = ectx()->rctx()->session()->space();
+    auto *path  = sentence_->path();
+
+    auto func = [metaHost, path, spaceId]() {
+        std::string tmp = "/usr/bin/curl -G \"http://%s:%d/%s?path=%s&space=%d\"";
```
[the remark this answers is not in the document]
darionyaphet: Actually, I have update this issue. Maybe you should reflush before review ? 

thanks for you help :)

src/storage/test/StorageHttpDownloadHandlerTest.cpp
darionyaphet: I will get download path from kv store

PR #662, opened 2019-07-19
Implemented type casting

### LINKED ISSUE #637: Type Conversion crash
When an operator is used with operands of different types, graphd crashes.

when adding numbers and character, graphd crashes.

when compare numbers and characters, the result:

dutor:
This pr implements and fixes #637


src/graph/test/YieldTest.cpp
```diff
@@ -40,11 +40,11 @@ TEST_F(YieldTest, Basic) {
     }
     {
         cpp2::ExecutionResponse resp;
-        std::string query = "YIELD 1+1";
+        std::string query = "YIELD 1+1, '1+1', (int)3.14, (string)(1+1), (string)true";
```
whitewum: 1+1 < 3

src/common/filter/Expressions.h
```diff
@@ -209,6 +210,56 @@ class Expression {
...
+        LOG(FATAL) << "unknown type: " << value.which();
+    }
+
+    static int64_t toInt(const VariantType &value) {
+        switch (value.which()) {
+            case 0:
+                return boost::get<int64_t>(value);
+            case 1:
+                return static_cast<int64_t>(boost::get<double>(value));
+            case 3:
```
laura-ding: case 2 and case 3

dutor: Shame on me.

src/common/filter/Expressions.h
```diff
@@ -209,6 +210,56 @@ class Expression {
         return std::abs(left - right) < EPSILON;
     }
 
+    // Procedures used to do type casting
+    static std::string toString(const VariantType &value) {
+        char buf[1024];
+        switch (value.which()) {
+            case 0:
+                return folly::to<std::string>(boost::get<int64_t>(value));
+            case 1:
```
wadeliuyi: why not use micro?

dutor:
Jenkins, go

src/common/filter/Expressions.h
```diff
@@ -209,6 +210,56 @@ class Expression {
...
+        switch (value.which()) {
+            case 0:
+                return static_cast<double>(boost::get<int64_t>(value));
+            case 1:
+                return boost::get<double>(value);
+            case 2:
+                return boost::get<bool>(value) ? 1.0 : 0.0;
+            case 3:
+                // TODO(dutor) error handling
+                return folly::to<double>(boost::get<std::string>(value));
```
CPWstatic: I think folly::tryTo would be better here, or exception would occur if an unparseable string is provided.

src/common/filter/Expressions.h
```diff
@@ -209,6 +210,56 @@ class Expression {
...
+    static int64_t toInt(const VariantType &value) {
+        switch (value.which()) {
+            case 0:
+                return boost::get<int64_t>(value);
+            case 1:
+                return static_cast<int64_t>(boost::get<double>(value));
+            case 2:
+                return boost::get<bool>(value) ? 1.0 : 0.0;
+            case 3:
+                return boost::get<int64_t>(value);
```
CPWstatic: We should do type casting here, right?

src/common/filter/Expressions.h
dutor: I will address such problems after #669 is merged.

src/common/filter/Expressions.h
dutor: Yes, this was a big mistake.

dutor:
@CPWstatic I had resolved the conflicts. But the error handling work mentioned above is not addressed. I am going to do it in the next PR, since there are some additional work to do.

PR #663, opened 2019-07-19
Fix crash when call hash.

CPWstatic:
close#653

### DIFF
--- a/src/common/filter/FunctionManager.cpp
+++ b/src/common/filter/FunctionManager.cpp
@@ -249,8 +249,26 @@ FunctionManager::FunctionManager() {
         attr.minArity_ = 1;
         attr.maxArity_ = 1;
         attr.body_ = [] (const auto &args) {
-            auto &str = Expression::asString(args[0]);
-            return static_cast<int64_t>(std::hash<std::string>()(str));
+            switch (args[0].which()) {
+                case 0: {
+                    auto v = Expression::asInt(args[0]);
+                    return static_cast<int64_t>(std::hash<int64_t>()(v));
+                }
+                case 1: {
+                    auto v = Expression::asDouble(args[0]);
+                    return static_cast<int64_t>(std::hash<double>()(v));
+                }
+                case 2: {
+                    auto v = Expression::asBool(args[0]);
+                    return static_cast<int64_t>(std::hash<bool>()(v));
+                }
+                case 3: {
+                    auto &v = Expression::asString(args[0]);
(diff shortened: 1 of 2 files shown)


CPWstatic:
Jenkins go.

src/common/filter/FunctionManager.cpp
```diff
@@ -249,8 +249,26 @@ FunctionManager::FunctionManager() {
...
+                case 0: {
+                    auto v = Expression::asInt(args[0]);
+                    return static_cast<int64_t>(std::hash<int64_t>()(v));
+                }
+                case 1: {
+                    auto v = Expression::asDouble(args[0]);
+                    return static_cast<int64_t>(std::hash<double>()(v));
+                }
+                case 2: {
+                    auto v = Expression::asBool(args[0]);
```
laura-ding: is it necessary to hash a bool?

whitewum: > is it necessary to hash a bool?

maybe some code will do that

src/graph/test/YieldTest.cpp
```diff
@@ -60,5 +60,60 @@ TEST_F(YieldTest, Basic) {
...
+        auto code = client->execute(query, resp);
+        ASSERT_EQ(cpp2::ErrorCode::SUCCEEDED, code);
+        std::vector<std::tuple<int64_t>> expected{
+            std::hash<double>{}(123.0)
+        };
+        ASSERT_TRUE(verifyResult(resp, expected));
+    }
+    {
+        cpp2::ExecutionResponse resp;
+        std::string query = "YIELD hash(!0)";
```
whitewum: what if console input some logcial operation?
TRUE AND FALSE OR TRUE

src/graph/test/YieldTest.cpp
```diff
@@ -60,5 +60,60 @@ TEST_F(YieldTest, Basic) {
...
+        auto code = client->execute(query, resp);
+        ASSERT_EQ(cpp2::ErrorCode::SUCCEEDED, code);
+        std::vector<std::tuple<int64_t>> expected{
+            std::hash<int64_t>{}(123 + 456)
+        };
+        ASSERT_TRUE(verifyResult(resp, expected));
+    }
+    {
+        cpp2::ExecutionResponse resp;
+        std::string query = "YIELD hash(123.0)";
```
whitewum: try hash(hash(wahaha))

PR #669, opened 2019-07-22
    Issue#424 to make `Expression::eval` to be able to do error handle

### LINKED ISSUE #424: To make `Expression::eval` to be able to do error handling

monadbobo:
Issue#424 to make `Expression::eval` to be able to do error handle.
    Summary:

    1. When the expression is evaluated, some wrong judgments are added,
       ifthe expression was failed, an error will be replyed to the user.

    2 fixed the crash when using mockStorageServer.

close #424


src/graph/GoExecutor.cpp
```diff
@@ -375,14 +381,18 @@ std::vector<VertexID> GoExecutor::getDstIdsFromResp(RpcResponse &rpcResp) const
     return std::vector<VertexID>(set.begin(), set.end());
 }
 
-
 void GoExecutor::finishExecution(RpcResponse &&rpcResp) {
-    auto outputs = setupInterimResult(std::move(rpcResp));
+    std::unique_ptr<InterimResult> outputs;
+
+    if (!setupInterimResult(std::move(rpcResp), outputs)) {
+        return;
```
laura-ding: do onError_()

src/common/filter/Expressions.cpp
```diff
@@ -562,24 +556,20 @@ std::string PrimaryExpression::toString() const {
...
 }
 
-
-VariantType PrimaryExpression::eval() const {
+OptVariantType PrimaryExpression::eval() const {
     switch (operand_.which()) {
         case kBool:
-            return boost::get<bool>(operand_);
-            break;
+            return OptVariantType(boost::get<bool>(operand_));
```
laura-ding: StatusOr has

src/common/filter/Expressions.cpp
```diff
@@ -662,17 +652,21 @@ std::string FunctionCallExpression::toString() const {
...
-    std::transform(args_.begin(), args_.end(), args.begin(), eval);
-    return function_(args);
-}
 
+    for (auto it = args_.cbegin(); it != args_.cend(); ++it) {
+        auto result = (*it)->eval();
+        if (!result.ok()) {
+            return result;
+        }
+        args.push_back(std::move(result.value()));
```
laura-ding: use emplace_back

monadbobo:
close #424

src/graph/GoExecutor.cpp
monadbobo: Not needed here, because returning false in setupInterimResult (processFinalResult ) means that onError_ has already been called.

src/common/filter/Expressions.cpp
monadbobo: it uesed  push_back was the same.

src/storage/QueryBaseProcessor.inl
```diff
@@ -366,7 +366,7 @@ kvstore::ResultCode QueryBaseProcessor<REQ, RESP>::collectEdgeProps(
                 // TODO(heng): We could remove the lock with one filter one bucket.
                 std::lock_guard<std::mutex> lg(this->lock_);
                 auto& getters = expCtx_->getters();
-                getters.getEdgeProp = [&] (const std::string &prop) -> VariantType {
+                getters.getEdgeProp = [&] (const std::string &prop) -> OptVariantType {
                     auto res = RowReader::getPropByName(reader.get(), prop);
                     CHECK(ok(res));
```
CPWstatic: Could we return an error to replace this ```CHECK```?

src/storage/QueryBaseProcessor.inl
```diff
@@ -375,15 +375,15 @@ kvstore::ResultCode QueryBaseProcessor<REQ, RESP>::collectEdgeProps(
                     return rank;
                 };
                 getters.getSrcTagProp = [&, this] (const std::string& tag,
-                                                   const std::string& prop) -> VariantType {
+                                                   const std::string& prop) -> OptVariantType {
                     auto it = fcontext->tagFilters_.find(std::make_pair(tag, prop));
                     CHECK(it != fcontext->tagFilters_.end());
```
CPWstatic: Could we return an error to replace this ```CHECK```?

PR #671, opened 2019-07-22
Improve random port choosing to listen

dutor:
As title.


src/common/network/NetworkUtils.cpp
```diff
@@ -125,18 +125,18 @@ std::unordered_set<uint16_t> NetworkUtils::getPortsInUse() {
...
-    VLOG(1) << "Dynamic port range is [" << low << ", " << high << "]";
+    CHECK(getLocalPortRange(low, high))
+        << "Failed to get the local port range";
+    VLOG(1) << "Local port range is [" << low << ", " << high << "]";
 
     std::unordered_set<uint16_t> portsInUse = getPortsInUse();
     uint16_t port = 0;
     do {
-        port = folly::Random::rand32(low, static_cast<int32_t>(high) + 1);
+        port = folly::Random::rand32(1024, low);
```
dangleptr: Why pick up a port in [1024, low) ?

dutor: Because the kernel picks up port(when `connect` or `bind(0)`) from the _local port range_. So we choose port that is not in this range. In this way, chances are less to confict.

dangleptr: I think it is useless for conflicts.  The conflict reason is we only pick up port for storage, but not for raft service.  As current logic,  raft port = storage port + 1.   
So to avoid conflicts,  we could implement another interface such as getAvailableRange(len) which could return a available ports len-sized range.

dutor: You picked up an available port, then bind on it, but then the port gets occupied. This is what this PR wants to resolve, but does nothing with raft. Trying to listen on an empheral port is way more likely to conflict than on a non-empheral one.

dangleptr: > You picked up an available port, then try to bind on it, but you found that the port already gets occupied. 

Why this could happen?  We have checked all ports is in use or not.

dutor: Yeah, it **was** available, but now it has been assigned to another connection(likely) or listening port(less likely).

dangleptr: Do you mean after check,  the port being assigned to another connection?  If so,  i think we don't need to take care of it,  CI machine is independent,  the UTs are running one by one, it should not happen.

dutor: Well, I am speechless now.

dangleptr: A HA.  So we should figure out what's the purpose about this PR. 
I think we should handle the port conflicts in UTs in this PR.   And the port conflicts reason is raft port not being checked before using.

dutor: I made some changes on the `getRandomPortToListen`, with which the caller could do some additional checks on the port number, like the port + 1 is also usable.

As for the original changes, I think I have explained pretty much. Sorry, but I tried my best.

dangleptr: The callback added seems good.  But it looks some weird.  
How about putting the port+1 picked up logic in Storage UTs directly.

dutor: That logic scatters all over in our testing code.

dangleptr: Only mockStorageServer will use the logic,   we could put it inside it.   So the interfaces in NetworkUtils will be more clean.

dutor: It is clean.

PR #680, opened 2019-07-24
Sanitized CMAKE files

dutor:
Added macro wrappers on `add_library`, `add_executable` and `add_test` to make the cmake files tight and clean. At the same time, all our own code now implicitly depends on `base_obj`, which in turn depends on all of those auto generated thrift objects.


dutor:
Testing canceled by accident.

Jenkins, go!

PR #681, opened 2019-07-24
[Raft] Add some ut, fix some bugs.

critical27:
Add some raft ut, add some methods to make test easy. There are some bugs in my submission:
1. When followers receives appendLogRequest, we will check whether leader's term is bigger than current term in verifyLeader. Suppose the scenario: leader and followers commit some logs, and one follower crash then reboot, its current term is last wal term, and it won't accept leader's log any more. So I think leader's term greater than or equal to current term is ok.
2. When followers receives appendLogRequest, we will try to commit logs from committedLogId_ + 1 to leader's commit log id, but it seems we can't guarantee that we have that log. So we can only commit to min(lastLogId_, leader's commit log ig).

### DIFF
--- a/src/kvstore/raftex/RaftPart.cpp
+++ b/src/kvstore/raftex/RaftPart.cpp
@@ -292,6 +292,7 @@ void RaftPart::stop() {
         std::unique_lock<std::mutex> lck(raftLock_);
         status_ = Status::STOPPED;
         leader_ = {0, 0};
+        role_ = Role::FOLLOWER;
 
         hosts = std::move(hosts_);
     }
@@ -1171,13 +1172,16 @@ void RaftPart::processAppendLogRequest(
 
     if (req.get_committed_log_id() > committedLogId_) {
         // Commit some logs
-        if (commitLogs(wal_->iterator(committedLogId_ + 1,
-                                      req.get_committed_log_id()))) {
+        // We can only commit logs from firstId to min(lastLogId_, leader's commit log id),
+        // follower can't always commit to leader's commit id because of lack of log
+        LogID lastLogIdCanCommit = std::min(lastLogId_, req.get_committed_log_id());
+        CHECK(committedLogId_ + 1<= lastLogIdCanCommit);
+        if (commitLogs(wal_->iterator(committedLogId_ + 1, lastLogIdCanCommit))) {
             VLOG(2) << idStr_ << "Succeeded committing log "
                               << committedLogId_ + 1 << " to "
-                              << req.get_committed_log_id();
-            committedLogId_ = req.get_committed_log_id();
-            resp.set_committed_log_id(committedLogId_);
+                              << lastLogIdCanCommit;
+            committedLogId_ = lastLogIdCanCommit;
+            resp.set_committed_log_id(lastLogIdCanCommit);
         } else {
             LOG(ERROR) << idStr_ << "Failed to commit log "
                        << committedLogId_ + 1 << " to "
--- a/src/kvstore/raftex/RaftPart.cpp
+++ b/src/kvstore/raftex/RaftPart.cpp
@@ -596,7 +596,8 @@ void RaftPart::replicateLogs(folly::EventBase* eb,
                    prevLogTerm,
                    pHosts = std::move(hosts)] (folly::Try<AppendLogResponses>&& result) mutable {
             VLOG(2) << self->idStr_ << "Received enough response";
-            CHECK(!result.hasException());
+            // TODO: since we kill some peers in ut, this may fail, handle exception if necessary
+            // CHECK(!result.hasException());
 
             self->processAppendLogResponses(*result,
                                             eb,
@@ -673,7 +674,8 @@ void RaftPart::processAppendLogResponses(
             } else {
                 LOG(FATAL) << idStr_ << "Failed to commit logs";
             }
-            VLOG(2) << idStr_ << "Succeeded in committing the logs";
+            VLOG(2) << idStr_ << "Leader succeeded in committing the logs "
+                              << committedId + 1 << " to " << lastLogId;
         }
         // Step 4: Fulfill the promise
         if (iter.hasNonCASLogs()) {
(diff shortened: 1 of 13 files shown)


dangleptr:
Well done.  It is a good starting for raft tests.

critical27:
Jenkins go

src/kvstore/raftex/test/LogAppendTest.cpp
```diff
@@ -29,32 +29,17 @@ TEST(LogAppend, SimpleAppendWithOneCopy) {
     std::vector<HostAddr> allHosts;
     std::vector<std::shared_ptr<RaftexService>> services;
     std::vector<std::shared_ptr<test::TestShard>> copies;
+    std::vector<LogID> lastCommittedLogId;
```
dangleptr: What's the purpose?

src/kvstore/raftex/RaftPart.cpp
```diff
@@ -594,7 +596,8 @@ void RaftPart::replicateLogs(folly::EventBase* eb,
                    prevLogTerm,
                    pHosts = std::move(hosts)] (folly::Try<AppendLogResponses>&& result) mutable {
             VLOG(2) << self->idStr_ << "Received enough response";
-            CHECK(!result.hasException());
+            // TODO: since we kill some peers in ut, this may fail, handle exception if necessary
+            // CHECK(!result.hasException());
```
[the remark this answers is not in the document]
critical27: I mock disconnect from one Host, and remove it from its peer, and it may throw exception like `runtime_error: Not enough futures` because of the Host has been destroyed.

src/kvstore/raftex/RaftPart.cpp
```diff
@@ -1228,7 +1235,7 @@ cpp2::ErrorCode RaftPart::verifyLeader(
     }
 
     // Make sure the remote term is greater than local's
-    if (req.get_current_term() <= term_) {
+    if (req.get_current_term() < term_) {
```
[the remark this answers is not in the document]
critical27: Yeah, it has been handled before when we check the leadership stickness

src/kvstore/raftex/test/LogAppendTest.cpp
critical27: It is used to store a copy's last commit log id outside the Part itself (just like the kLastCommittedIdKey we store it in rocksdb). When a TestShard reboot we can get lastCommittedLogId here.

critical27:
> We need fault injector to mock some errors. Let's design it carefully offline.

Sure, I have not came up with how to mock those errors.

src/kvstore/raftex/RaftPart.cpp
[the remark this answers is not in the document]
dangleptr: if you want to mock host disconnect,  you should stop the host directly instead of removing it from others's peers.

src/kvstore/raftex/RaftPart.cpp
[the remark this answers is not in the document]
dangleptr: Not exactly,  in your code,  if a new leader with the same term comes,   the follower will still follow it.

src/kvstore/raftex/test/LogAppendTest.cpp
dangleptr: If so, why not put it in TestShard?

src/kvstore/raftex/RaftPart.cpp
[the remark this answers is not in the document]
critical27: Em, not only term, but it also check req.get_leader_ip() == leader_.first and req.get_leader_port() == leader_.second. I don't quite get your point. If a new leader comes with the same term, if it is the same leader, everything is fine. But I don't think we can have two leader with same term.

src/kvstore/raftex/test/LogAppendTest.cpp
critical27: Good suggestion

src/kvstore/raftex/RaftPart.cpp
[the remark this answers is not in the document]
critical27: Yeah, you're right. It only throw the exception when I modify the peers. But do you think it is necessary to mock network partition?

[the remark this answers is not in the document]
dangleptr: Of course, we will mock the network partition situation.  We could leave the job in our raft test framework.

src/kvstore/raftex/RaftPart.cpp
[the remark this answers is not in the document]
dangleptr: We can't assume it never happen,   maybe some bugs inside it.

src/kvstore/raftex/test/RaftCase.cpp
```diff
@@ -0,0 +1,258 @@
+/* Copyright (c) 2018 vesoft inc. All rights reserved.
```
dangleptr: 2019

src/kvstore/wal/FileBasedWal.cpp
```diff
@@ -602,7 +602,6 @@ bool FileBasedWal::rollbackToLog(LogID id) {
         }
 
         auto it = walFiles_.upper_bound(id);
```
dangleptr: Separate the changes into another pr.

src/kvstore/raftex/RaftPart.cpp
```diff
@@ -716,17 +718,6 @@ void RaftPart::processAppendLogResponses(
         if (!iter.empty()) {
             appendLogsInternal(std::move(iter), currTerm);
         }
-    } else {
```
dangleptr: Use another pr for it.

src/kvstore/wal/FileBasedWal.cpp
critical27: ok

src/kvstore/raftex/RaftPart.cpp
critical27: ok

PR #682, opened 2019-07-24
Fix use-after-free problem in UT

dangleptr:
All servers depend on metaClient. 

So release meta client after servers.


dangleptr:
Jenkins go

dutor:
@monadbobo Please have a check.

PR #683, opened 2019-07-24
Support variable/pipe reference in WHERE and YIELD

### LINKED ISSUE #432: how to take properties to the next hops?
for example:

vertex:  player:100,   play:101,   team:200

edge1: player: 100  --like--> player 101

edge2:  player 101  ---serve---> team 20

Two hop query:  go from 100 over like yield $^[player].name as p1, $$[player].age as p2 | go from ($-:id ??) over serve yield (how to get p1 and p2?)

dutor:
As title.

This PR will close #432


src/graph/test/GoTest.cpp
```diff
@@ -314,5 +314,99 @@ TEST_F(GoTest, VertexNotExist) {
...
+            {"Chris Paul", "LeBron James", "Ray Allen"},
+            {"Chris Paul", "Carmelo Anthony", "LeBron James"},
+            {"Chris Paul", "Carmelo Anthony", "Dwyane Wade"},
+            {"Chris Paul", "Dwyane Wade", "LeBron James"},
+            {"Chris Paul", "Dwyane Wade", "Carmelo Anthony"},
+        };
+        ASSERT_TRUE(verifyResult(resp, expected));
+    }
+}
+
```
CPWstatic: I think we should add ```GO N STEPS``` to test BackTracker.

dutor:
jenkins go for fun

PR #684, opened 2019-07-24
Fix use-after-free problem in BalancerPlan

dangleptr:
In balancePlan,  after all tasks finished,  the plan will be destroyed.  
So we should save the plan firstly before tasks invoke to avoid the potential risk.


dangleptr:
Jenkins go

PR #689, opened 2019-07-25
Fix failed data_test/go_test.

dangleptr:
The reason is that go_test/data_test will drop space before exit.  And the procedure of drop space  is update information on meta,  then notify storage server asynchronously.  Maybe meta client receive the notification after the storage server stopped,  so it will crash.

To fix it,  we should cut-off the relationship with metaclient before kvstore clean up parts.  But i still want "drop space "could be invoked on storage server,  so i add sleep to ensure it.

### DIFF
--- a/src/kvstore/NebulaStore.cpp
+++ b/src/kvstore/NebulaStore.cpp
@@ -75,7 +75,7 @@ void NebulaStore::init() {
     raftService_ = raftex::RaftexService::createService(ioPool_, raftAddr_.second);
     raftService_->waitUntilReady();
     flusher_ = std::make_unique<wal::BufferFlusher>();
-    CHECK(!!partMan_);
+    CHECK(!!options_.partMan_);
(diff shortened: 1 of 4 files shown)


dangleptr:
Jenkins go

src/kvstore/NebulaStore.cpp
```diff
@@ -127,7 +129,7 @@ void NebulaStore::init() {
     }
 
     LOG(INFO) << "Init data from partManager for " << storeSvcAddr_;
-    auto partsMap = partMan_->parts(storeSvcAddr_);
+    auto partsMap = options_.partMan_->parts(storeSvcAddr_);
```
monadbobo: MayBe we can delete the "partMan_" field in NebulaStore?

dangleptr: Yes, I had removed it.

PR #697, opened 2019-07-26
Issue#192 Support multiple edge types in GO statement

monadbobo:
There are still two issues that need to be implemented.
    1. Since we are not implementing a "null" type, we cannot represent some non-existing attributes.
    2. Due to the existence of 1, there is currently no judgment on the attributes of the edges and vetex.


monadbobo:
I will resubmit this PR.

Issue #192, opened 2019-02-27
Support multiple edge types in GO statement


sherman-the-tank:
Support multiple edge types in VIA clause (GO statement). One special case is to support ```*``` (all edge types)

PR #699, opened 2019-07-29
    Issue#192 Support multiple edge types in GO statement

monadbobo:
Summary:
Implemented over multiple edge types(include over "*" ) for Support all edge type in Go statement.

1. Modification of the storage interface:
   1.1 Modify the GetNeighborsRequest structure, pass a list of edge_type to the storage layer.
   1.2 And add an over_all_edges tag to mark "over *".
       Modify the PropDef structure to represent both tag_id and edge_type
   1.3 Added the EdgeData structure to represent the edge data returned to the client.
2. When "over *", the current implementation was first getting all the types of edges from the meta, and then call the storage interface.
3. When the yield statement does not exist, delete the Deleted the default rename("dst_" to "id").
4. Delete the getOutBound and getInBound interfaces,
   because the edges passed to the storage may have both in and out.
5.  Added a map of edge type to edge name  (toEdgeName) and a map of spaceid to all edgetype for metaclient.
6. Returns the default value of the corresponding type when some properties do not exist.


src/graph/GoExecutor.cpp
```diff
@@ -288,12 +296,12 @@ void GoExecutor::stepOut() {
         return;
     }
     auto returns = status.value();
-    auto future = ectx()->storage()->getNeighbors(spaceId,
-                                                  starts_,
-                                                  edgeType_,
-                                                  !reversely_,
-                                                  "",
-                                                  std::move(returns));
+    auto future  = ectx()->storage()->getNeighbors(spaceId,
```
wadeliuyi: two spaces after future

src/graph/GoExecutor.cpp
```diff
@@ -359,24 +389,40 @@ std::vector<VertexID> GoExecutor::getDstIdsFromResp(RpcResponse &rpcResp) const
...
+    std::vector<std::unique_ptr<YieldColumn>> yc;
+    if (expCtx_->isOverAllEdge() && yields_.empty()) {
+        auto edgeNames = getEdgeNamesFromResp(rpcResp);
+        for (const auto &name : edgeNames) {
+            auto dummy     = new std::string(name + "_id");
+            auto dummy_exp = new EdgeDstIdExpression(dummy);
+            auto ptr       = std::make_unique<YieldColumn>(dummy_exp);
+            dummy_exp->setContext(expCtx_.get());
+            yields_.emplace_back(ptr.get());
+            yc.emplace_back(std::move(ptr));
```
wadeliuyi: why save the ptr to a template vector?

src/graph/GoExecutor.cpp
```diff
@@ -593,13 +654,10 @@ void GoExecutor::processFinalResult(RpcResponse &rpcResp, Callback cb) const {
             continue;
         }
         std::shared_ptr<ResultSchemaProvider> vschema;
-        std::shared_ptr<ResultSchemaProvider> eschema;
+
         if (resp.get_vertex_schema() != nullptr) {
             vschema = std::make_shared<ResultSchemaProvider>(resp.vertex_schema);
         }
```
wadeliuyi: maybe need else to return or print some error info

src/graph/GoExecutor.cpp
```diff
@@ -608,45 +666,54 @@ void GoExecutor::processFinalResult(RpcResponse &rpcResp, Callback cb) const {
                 vreader = RowReader::getRowReader(vdata.vertex_data, vschema);
             }
```
wadeliuyi: ditto

src/parser/parser.yy
```diff
@@ -490,18 +499,50 @@ vid_ref_expression
...
+        $$ = new OverEdge($1, $3, true);
     }
-    | KW_OVER name_label KW_REVERSELY {
-        $$ = new OverClause($2, nullptr, true);
+    ;
+
+over_edges
+    : over_edge {
+        auto edge = new OverEdges();
+	edge->addEdge($1);
```
wadeliuyi: tab and space

src/storage/QueryBaseProcessor.inl
```diff
@@ -104,7 +124,18 @@ cpp2::ErrorCode QueryBaseProcessor<REQ, RESP>::checkAndBuildContexts(const REQ&
...
                 prop.returned_ = true;
-                edgeContext_.props_.emplace_back(std::move(prop));
+                auto it2 = edgeContext_.find(edgeType);
+                if (it2 == edgeContext_.end()) {
+                    if (overAllEdge_) {
+                        std::vector<PropContext> v{std::move(prop)};
+                        edgeContext_.emplace(edgeType, std::move(v));
+                        break;
+                    } else {
+                        CHECK(it2 != edgeContext_.end());
```
wadeliuyi: maybe is it2 == edgeContext_.end() ?

monadbobo: When entering this condition, meaning we must update the it2->second, so the "edgeType" must exist in "edgeContext_".

wadeliuyi: but this check is in the condition "it2 == edgeContext_.end()", so this check is always false.

monadbobo:
close issue #341  #192

src/graph/GoExecutor.cpp
CPWstatic: > why save the ptr to a template vector?

For releasing memory.

src/graph/test/GoTest.cpp
```diff
@@ -78,7 +78,7 @@ TEST_F(GoTest, OneStepOutBound) {
         cpp2::ExecutionResponse resp;
         auto &player = players_["Boris Diaw"];
         auto *fmt = "GO FROM %ld OVER like "
-                    "| GO FROM $-.id OVER like | GO FROM $-.id OVER serve";
+                    "| GO FROM $-.like_id OVER like | GO FROM $-.like_id OVER serve";
```
CPWstatic: ```$-.alias.prop```
This form looks better to me.

src/interface/storage.thrift
```diff
@@ -205,11 +218,9 @@ struct CatchUpDataReq {
 
 
 service StorageService {
-    QueryResponse getOutBound(1: GetNeighborsRequest req)
-    QueryResponse getInBound(1: GetNeighborsRequest req)
+    QueryResponse getBound(1: GetNeighborsRequest req)
```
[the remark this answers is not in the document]
monadbobo: Because when we have more than one edge, there may be both an "in" side and some "out".

src/graph/test/GoTest.cpp
monadbobo: Maybe we need discuss this.

src/storage/QueryBaseProcessor.inl
monadbobo: "it2 != edgeContext.end()" this  means edgeContext_ include edgeType.

wadeliuyi: when overAllEdge_ is false, the logic as follow:
if (it2 == edgeContext_.end()) {
    CHECK(it2 != edgeContext_.end());
}

monadbobo: yes, when overAllEdge_ was false, the edgeContext_ must include edgeType. so maybe i can using the CHECK(false).

src/interface/storage.thrift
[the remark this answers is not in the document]
CPWstatic: I get it.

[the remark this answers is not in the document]
CPWstatic: I get it.

src/parser/parser.yy
```diff
@@ -496,18 +505,50 @@ vid_ref_expression
...
+	$$ = $1;
     }
-    | KW_OVER name_label KW_AS name_label KW_REVERSELY {
-        $$ = new OverClause($2, $4, true);
+    ;
+
+over_clause
+    : KW_OVER MUL {
+        auto edges = new OverEdges();
+	auto s = new std::string("*");
```
laura-ding: alignment

src/parser/parser.yy
```diff
@@ -496,18 +505,50 @@ vid_ref_expression
...
+	auto edge = new OverEdge(s, nullptr, false);
+	edges->addEdge(edge);
+        $$ = new OverClause(edges);
+    }
+    | KW_OVER MUL KW_REVERSELY {
+        auto edges = new OverEdges();
+	auto s = new std::string("*");
```
laura-ding: ditto

src/storage/client/StorageClient.cpp
```diff
@@ -153,20 +148,15 @@ folly::SemiFuture<StorageRpcResponse<cpp2::QueryStatsResponse>> StorageClient::n
...
     return collectResponse(
         evb, std::move(requests),
-        [isOutBound](cpp2::StorageServiceAsyncClient* client,
-                     const cpp2::GetNeighborsRequest& r) {
-            if (isOutBound) {
-                return client->future_outBoundStats(r);
-            } else {
-                return client->future_inBoundStats(r);
-            }
+        [isOutBound](cpp2::StorageServiceAsyncClient* client, const cpp2::GetNeighborsRequest& r) {
```
laura-ding: isOutBound no be used

src/parser/parser.yy
```diff
@@ -436,12 +440,17 @@ go_sentence
...
-            auto *expr = new EdgeDstIdExpression(edge);
-            auto *alias = new std::string("id");
-            auto *col = new YieldColumn(expr, alias);
             auto *cols = new YieldColumns();
-            cols->addColumn(col);
+            for (auto e : $4->edges()) {
+	        if (e->isOverAll()) {
+		    continue;
+		}
+                auto *edge = new std::string(*e->edge());
```
darionyaphet: alignment

src/parser/parser.yy
```diff
@@ -505,18 +514,50 @@ vid_ref_expression
...
+over_edges
+    : over_edge {
+        auto edge = new OverEdges();
+	edge->addEdge($1);
+	$$ = edge;
     }
-    | KW_OVER name_label KW_AS name_label {
-        $$ = new OverClause($2, $4);
+    | over_edges COMMA over_edge {
+        $1->addEdge($3);
```
darionyaphet: alignment

src/parser/parser.yy
```diff
@@ -505,18 +514,50 @@ vid_ref_expression
...
-        $$ = new OverClause($2, $4, true);
+    ;
+
+over_clause
+    : KW_OVER MUL {
+        auto edges = new OverEdges();
+	auto s = new std::string("*");
+	auto edge = new OverEdge(s, nullptr, false);
+	edges->addEdge(edge);
+        $$ = new OverClause(edges);
```
darionyaphet: ditto

src/graph/test/GoTest.cpp
monadbobo: @dutor @sherman-the-tank @laura-ding

src/parser/parser.yy
monadbobo: I will fix this.

src/parser/parser.yy
monadbobo: I will fix this.

src/storage/client/StorageClient.cpp
monadbobo: Yes, I will remove this.

src/parser/parser.yy
monadbobo: I will fix this.

src/parser/parser.yy
monadbobo: I will fix this.

src/parser/parser.yy
monadbobo: I will fix this.

src/interface/storage.thrift
```diff
@@ -41,9 +41,14 @@ enum PropOwner {
     EDGE = 3,
 } (cpp.enum_strict)
 
+union PropId {
```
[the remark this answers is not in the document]
monadbobo: Is PropType more better?

src/interface/storage.thrift
```diff
@@ -61,10 +66,22 @@ struct ResultCode {
...
+    2: common.Schema         schema,
+    3: binary                vertex_data,
+}
+
 struct VertexData {
-    1: common.VertexID vertex_id,
-    2: binary vertex_data, // decode according to vertex_schema.
-    3: binary edge_data,   // decode according to edge_schema.
+    1: common.VertexID       vertex_id,
+    2: list<TagData>         tag_data,
```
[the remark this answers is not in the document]
monadbobo: Because it is possible that a VertexID belongs to a different tag.

src/interface/storage.thrift
```diff
@@ -142,9 +158,10 @@ struct EdgePropRequest {
     1: common.GraphSpaceID space_id,
     // partId => edges
     2: map<common.PartitionID, list<EdgeKey>>(cpp.template = "std::unordered_map") parts,
-    3: common.EdgeType edge_type,
+    3: list<common.EdgeType> edge_types,
```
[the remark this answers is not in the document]
monadbobo: This is needed, because it is more intuitive to pass directly to the storage.

src/storage/QueryBaseProcessor.h
```diff
@@ -105,12 +105,12 @@ class QueryBaseProcessor : public BaseProcessor<RESP> {
 
 protected:
     GraphSpaceID  spaceId_;
-    BoundType     type_;
     std::unique_ptr<ExpressionContext> expCtx_;
     std::unique_ptr<Expression> exp_;
     std::vector<TagContext> tagContexts_;
-    EdgeContext edgeContext_;
+    std::unordered_map<EdgeType, std::vector<PropContext>> edgeContext_;
```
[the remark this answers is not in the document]
monadbobo: Thank you, i will rename this.

src/storage/QueryBaseProcessor.inl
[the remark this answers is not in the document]
monadbobo: Yes, Your writing is more intuitive.

src/interface/storage.thrift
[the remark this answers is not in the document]
dangleptr: Currently,  if user want to fetch different props on different tags,   storage will return them in one line.  For example,   tag1.prop1,  tag2.prop1,  tag2.prop2,  tag3.prop1 ....

And we encode them together,  so query engine could use it directly. 

Of course,  for multi edges,  there is the same problem.  For example,  what's the output for this go sentence now :  "go from id over edge1, edge2 return edge1.prop1, edge2.prop1".   Do we support it now?

src/interface/storage.thrift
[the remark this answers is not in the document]
dangleptr: For getNeighbors,  i think we need it.  Because it represents the over clause. And we should check the consistency with return_columns.   

But for EdgePropRequest,   do we really need it?  The users input maybe "Get props edge1.prop1, edge2.prop2".   That's much simple.

src/interface/storage.thrift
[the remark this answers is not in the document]
dangleptr: Currently,  if user want to fetch different props on different tags,   storage will return them in one line.  For example,   tag1.prop1,  tag2.prop1,  tag2.prop2,  tag3.prop1 ....

And we encode them together,  so query engine could use it directly. 

Of course,  for multi edges,  there is the same problem.  For example,  what's the output for this go sentence now :  "go from id over edge1, edge2 return edge1.prop1, edge2.prop1".   Do we support it now?

src/interface/storage.thrift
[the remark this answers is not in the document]
dangleptr: En.....  Maybe EntryID?  or TagOrEdge ?

src/graph/FetchVerticesExecutor.cpp
```diff
@@ -133,27 +133,27 @@ void FetchVerticesExecutor::processResult(RpcResponse &&result) {
     std::unique_ptr<RowSetWriter> rsWriter;
     auto uniqResult = std::make_unique<std::unordered_set<std::string>>();
     for (auto &resp : all) {
-        if (!resp.__isset.vertices || !resp.__isset.vertex_schema
-                || resp.get_vertices() == nullptr || resp.get_vertex_schema() == nullptr) {
+        if (!resp.__isset.vertices) {
```
darionyaphet: why remove another check ?

src/interface/storage.thrift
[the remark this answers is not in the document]
monadbobo: I think EntryID was more better, thank you.

src/interface/storage.thrift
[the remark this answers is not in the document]
monadbobo: Yes, we can handle this.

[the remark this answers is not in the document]
dangleptr: If so,  why not encode them in one line.  use one schema is enough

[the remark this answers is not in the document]
monadbobo: That's because different tag have different schema.

src/graph/FetchVerticesExecutor.cpp
monadbobo: This is because we have moved the schema to tag_data.

src/interface/storage.thrift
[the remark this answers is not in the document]
monadbobo: yes, we can remove edge_types in EdgePropRequest.

src/interface/storage.thrift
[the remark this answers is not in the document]
dangleptr: This is result set.  It has no relationship with tag schema.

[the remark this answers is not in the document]
monadbobo: When returning the same prop name for multiple tags, there is no good way for the server to get the type of prop returned, so we need "list<TagData>".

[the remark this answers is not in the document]
dangleptr: Not exactly,  we have fixed the problem.  Query engine will use index to get the related tag prop,  and storage return the props as the order passed in.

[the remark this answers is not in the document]
monadbobo: This will cause a problem, assuming you will take tag1.prop1, tag2.prop2, tag1.prop2. If tag2.prop2 does not exist, then the graph server will not know if tag2.prop2 does not exist or tag1.prop2 does not exist.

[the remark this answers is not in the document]
dangleptr: We have talked about this problem.  It should be fixed by null after it is supported in DataMan.

src/graph/test/GoTest.cpp
laura-ding: I agree with CPWstatic's idea

monadbobo: I modified this logic, but there is a problem here. We don't currently support passing "edge.prop/tag.prop" in the pipeline, so we need rename it to use   in the pipeline.

src/interface/storage.thrift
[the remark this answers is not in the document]
monadbobo: It can fix this problem with each tag_data with its own schema, or modify the struct ColumnDef and adding tagid/edgetype to it.

monadbobo:
> Thanks for your great work.
> 
> Just skim the patch about storage part, i want to know is there any possible the filter contains multi edges properties, and how to deal with it.
> 
> There are some inline comments, please check it.
> 
> Very impressive work!

There are bugs in dealing with multiple types of edges, I will fix this problem.

src/interface/storage.thrift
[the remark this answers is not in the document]
dangleptr: Not only this.  For the whole response,  we only have two schema,  the tag result and the edge result.  We have no needs to send schema in each vertex data. It is too cost.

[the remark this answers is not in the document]
monadbobo: I know what you mean. If this is the case, I think it would be better to modify the struct ColumnDef(add EdgeType/TagID).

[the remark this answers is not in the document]
monadbobo: Because if we use index, then the name field in ColumnDef is meaningless.

src/interface/storage.thrift
```diff
@@ -76,8 +91,8 @@ struct ResponseCommon {
 
 struct QueryResponse {
     1: required ResponseCommon result,
-    2: optional common.Schema vertex_schema,   // vertex related props
-    3: optional common.Schema edge_schema,     // edge related props
+    2: optional map<common.TagID, common.Schema>(cpp.template = "std::unordered_map")       vertex_schema,
```
dangleptr: As we talked about offline,  combine all tags schema into one.

src/interface/storage.thrift
```diff
@@ -76,8 +91,8 @@ struct ResponseCommon {
...
+    3: optional map<common.EdgeType, common.Schema>(cpp.template = "std::unordered_map")    edge_schema,
```
dangleptr: Use list instead,  the order is important for query engine.

src/interface/storage.thrift
monadbobo: Later I thought it would be better to do this (the last time I submitted that was the worst). 
The reasons are as follows:
1 In most cases there is only one tag id, then there is almost no difference between using only one schema at this time.
2 If there are multiple tag ids, the disadvantage of this approach is that it will output more schemas, but the advantage is that you don't need to fill in the extra data.
3 For the query engine, use map to save the schema, which is easier to process data.

dangleptr: Well,  it's not a key problem.  Let's go on with your approach.

src/interface/storage.thrift
dangleptr: It's ok for me.  You'd better confirm it with @dutor ,  he has plan to change the logic about getting property by index instead of getting property by name in query engine.

src/storage/QueryBoundProcessor.cpp
```diff
@@ -13,94 +13,204 @@
...
+            }
+
+            edgeTypes.emplace(edgeType);
+            auto it = edgeContexts_.find(edgeType);
+            if (it == edgeContexts_.end()) {
+                PropContext pc;
+                cpp2::PropDef pd;
+                pc.pikType_ = PropContext::PropInKeyType::DST;
+                pc.type_.type = nebula::cpp2::SupportedType::INT;
+                pd.name = "_dst";
```
dangleptr: Why return '_dst' by default?

src/storage/QueryEdgePropsProcessor.cpp
```diff
@@ -39,17 +39,9 @@ kvstore::ResultCode QueryEdgePropsProcessor::collectEdgesProps(
     return ret;
 }
 
-void QueryEdgePropsProcessor::addDefaultProps() {
```
dangleptr: Why remove the default props?

src/storage/QueryEdgePropsProcessor.cpp
```diff
@@ -39,17 +39,9 @@ kvstore::ResultCode QueryEdgePropsProcessor::collectEdgesProps(
...
-    this->edgeContext_.props_.emplace_back("_rank", 1, PropContext::PropInKeyType::RANK);
-    this->edgeContext_.props_.emplace_back("_dst", 2, PropContext::PropInKeyType::DST);
-}
-
 void QueryEdgePropsProcessor::process(const cpp2::EdgePropRequest& req) {
     spaceId_ = req.get_space_id();
-    // By default, _src, _rank, _dst will be returned as the first 3 fields
-    addDefaultProps();
-    int32_t returnColumnsNum = req.get_return_columns().size() + this->edgeContext_.props_.size();
+    //    initContext(req.edge_types, true);
```
dangleptr: Why comment this line?

src/storage/QueryEdgePropsProcessor.cpp
```diff
@@ -59,14 +51,20 @@ void QueryEdgePropsProcessor::process(const cpp2::EdgePropRequest& req) {
...
+    int32_t returnColumnsNum = req.get_return_columns().size() + edgeSize;
     RowSetWriter rsWriter;
     std::for_each(req.get_parts().begin(), req.get_parts().end(), [&](auto& partE) {
         auto partId = partE.first;
         kvstore::ResultCode ret;
         for (auto& edgeKey : partE.second) {
-            ret = this->collectEdgesProps(partId, edgeKey, this->edgeContext_.props_, rsWriter);
-            if (ret != kvstore::ResultCode::SUCCEEDED) {
-                break;
+            for (auto& ec : edgeContexts_) {
```
dangleptr: I can't understand this code.  For each edge key,  you could find the related props in edgeContexts_ and collect them.

src/storage/QueryStatsProcessor.cpp
```diff
@@ -111,9 +111,12 @@ void QueryStatsProcessor::onProcessFinished(int32_t retNum) {
             }
         }
     }
-    for (auto& prop : this->edgeContext_.props_) {
-        CHECK(prop.returned_);
-        props.emplace_back(std::move(prop));
+
+    for (auto& ec : this->edgeContexts_) {
+        auto p = ec.second;
+        for (auto& prop : p) {
```
dangleptr: Add the check back.

src/storage/test/QueryEdgePropsTest.cpp
```diff
@@ -60,58 +60,36 @@ void buildRequest(cpp2::EdgePropRequest& req) {
...
-            EXPECT_EQ(ResultType::SUCCEEDED, it->getInt<int64_t>(1, v));
-            CHECK_EQ(rowNum % 7, v);
-        }
-        {
-            // _dst
-            int64_t v;
-            EXPECT_EQ(ResultType::SUCCEEDED, it->getInt<int64_t>(2, v));
-            CHECK_EQ(10001 + rowNum % 7, v);
-        }
+        EXPECT_EQ(10, it->numFields());
```
dangleptr: Why change the UTs?

src/storage/test/QueryBoundTest.cpp
```diff
@@ -88,22 +88,23 @@ void buildRequest(cpp2::GetNeighborsRequest& req, bool outBound = true) {
         }
```
dangleptr: Add UTs for multi edge

src/storage/QueryEdgePropsProcessor.cpp
monadbobo: Because fetch edge does not require default properties.

src/storage/QueryEdgePropsProcessor.cpp
monadbobo: I need to delete this comment.

src/dataman/RowReader.h
```diff
@@ -89,6 +89,32 @@ class RowReader {
...
+            }
+            case nebula::cpp2::SupportedType::FLOAT:
+            case nebula::cpp2::SupportedType::DOUBLE: {
+                return static_cast<double>(0.0);
+            }
+            case nebula::cpp2::SupportedType::STRING: {
+                return static_cast<std::string>("");
+            }
+            default:
+                LOG(FATAL) << "Unknown type: " << static_cast<int32_t>(vType.type);
```
wadeliuyi: FATAL will crash, maybe other level is better.

src/graph/FetchVerticesExecutor.cpp
```diff
@@ -133,17 +133,30 @@ void FetchVerticesExecutor::processResult(RpcResponse &&result) {
...
             continue;
         }
-        auto vschema = std::make_shared<ResultSchemaProvider>(resp.vertex_schema);
+
+        auto *schema = resp.get_vertex_schema();
+        if (schema == nullptr) {
+            continue;
+        }
+
+        std::unordered_map<TagID, std::shared_ptr<ResultSchemaProvider>> tagSchema;
```
wadeliuyi: maybe no need shared_ptr is better

src/storage/QueryBaseProcessor.inl
```diff
@@ -34,20 +34,40 @@ bool QueryBaseProcessor<REQ, RESP>::validOperation(nebula::cpp2::SupportedType v
...
+    p.emplace_back("_rank", eType, 1, PropContext::PropInKeyType::RANK);
+    p.emplace_back("_dst", eType, 2, PropContext::PropInKeyType::DST);
+}
+
+template <typename REQ, typename RESP>
+void QueryBaseProcessor<REQ, RESP>::initContext(const std::vector<EdgeType>& eTypes,
+                                                bool need_default_props) {
+    std::transform(eTypes.cbegin(), eTypes.cend(),
+                   std::inserter(edgeContexts_, edgeContexts_.end()),
+                   [this, need_default_props](const auto& ec) {
```
wadeliuyi: ec is edge type?

src/tools/storage-perf/StoragePerfTool.cpp
```diff
@@ -214,19 +214,20 @@ class Perf {
...
-                                    VLOG(3) << "request successed!";
-                                }
-                                this->finishedRequests_++;
-                                VLOG(3) << "request successed!";
-                             }).onError([](folly::FutureException&) {
-                                LOG(ERROR) << "request failed!";
-                             });
+        std::vector<EdgeType> e(edgeType_);
+        auto f =
+            client_->getNeighbors(spaceId_, randomVertices(), e, "", randomCols())
```
wadeliuyi: std::move(e)

src/storage/client/StorageClient.cpp
```diff
@@ -94,10 +94,10 @@ folly::SemiFuture<StorageRpcResponse<cpp2::ExecResponse>> StorageClient::addEdge
 folly::SemiFuture<StorageRpcResponse<cpp2::QueryResponse>> StorageClient::getNeighbors(
         GraphSpaceID space,
         std::vector<VertexID> vertices,
-        EdgeType edgeType,
-        bool isOutBound,
+        std::vector<EdgeType> edgeTypes,
```
wadeliuyi: the caller in GoExecutor::stepOut is pass parameter by value, so here maybe const std::vector<EdgeType> &edgeTypes is better, or req.set_edge_types(std::move(edgeTypes));

src/storage/client/StorageClient.cpp
```diff
@@ -112,29 +112,24 @@ folly::SemiFuture<StorageRpcResponse<cpp2::QueryResponse>> StorageClient::getNei
...
+            return client->future_getBound(r);
         });
 }
 
 
 folly::SemiFuture<StorageRpcResponse<cpp2::QueryStatsResponse>> StorageClient::neighborStats(
         GraphSpaceID space,
         std::vector<VertexID> vertices,
-        EdgeType edgeType,
+        std::vector<EdgeType> edgeTypes,
```
wadeliuyi: ditto

src/storage/QueryEdgePropsProcessor.cpp
dangleptr: If so, how to know which edge the returned props belongs to?

src/storage/QueryBaseProcessor.inl
monadbobo: ec is edgeContext

src/storage/test/QueryBoundTest.cpp
monadbobo: Ok, I will add this.

src/graph/FetchVerticesExecutor.cpp
monadbobo: It is more convenient to use shared_ptr.

src/storage/test/QueryEdgePropsTest.cpp
monadbobo: Because for the query engine, there is no need to return the default properties of the  edges.

src/storage/QueryBaseProcessor.inl
```diff
@@ -361,20 +398,26 @@ kvstore::ResultCode QueryBaseProcessor<REQ, RESP>::collectEdgeProps(
...
                 auto& getters = expCtx_->getters();
-                getters.getAliasProp =
-                    [&] (const std::string&, const std::string &prop) -> OptVariantType {
-                        auto res = RowReader::getPropByName(reader.get(), prop);
-                        if (!ok(res)) {
-                            return Status::Error("Invalid Prop");
-                        }
-                        return value(std::move(res));
-                    };
+                getters.getAliasProp = [this, edgeType, &reader](const std::string& edgeName,
```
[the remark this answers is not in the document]
monadbobo: We need to support this later(Maybe we need to reopen a PR to do this).

monadbobo:
Jenkins go.

src/graph/FetchVerticesExecutor.cpp
wadeliuyi: the tagSchema just used in current thread, so raw pointer is better about performance.

src/storage/QueryBaseProcessor.inl
wadeliuyi: I see that pass ec to addDefaultProps directly, and the second parameter is "EdgeType eType", so I think the ec is edge type.

src/storage/QueryBaseProcessor.inl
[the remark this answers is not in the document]
dangleptr: That's ok

src/storage/test/QueryEdgePropsTest.cpp
dangleptr: I don't know why you have this conclusion.  The default props have been talked about.  Don't change this logic if it has no relationship with this pr.  If you have some thoughts about this place,  please open another pr for it.

src/graph/FetchVerticesExecutor.cpp
monadbobo: When the getRowReader function is called below, its argument is the shared pointer.

src/storage/QueryBoundProcessor.cpp
dangleptr: If so,   why not return '_edgeType',  '_rank'?  Because <src, edgeType, rank, dst> identify one edge.

src/interface/storage.thrift
```diff
@@ -142,9 +158,9 @@ struct EdgePropRequest {
     1: common.GraphSpaceID space_id,
     // partId => edges
     2: map<common.PartitionID, list<EdgeKey>>(cpp.template = "std::unordered_map") parts,
-    3: common.EdgeType edge_type,
-    4: binary filter,
-    5: list<PropDef> return_columns,
+    3: binary filter,
+    4: list<PropDef> return_columns,
+    5: bool over_all_edges,
```
dangleptr: Why we have over_all_edges in EdgePropRequest,  the edgeKeys have been specified.

src/parser/parser.yy
```diff
@@ -437,12 +441,17 @@ go_sentence
...
-            auto *col = new YieldColumn(expr, alias);
             auto *cols = new YieldColumns();
-            cols->addColumn(col);
+            for (auto e : $4->edges()) {
+                if (e->isOverAll()) {
+                    continue;
+                }
+                auto *edge  = new std::string(*e->edge());
+                auto *expr  = new EdgeDstIdExpression(edge);
+//                auto *alias = new std::string(*edge + "_id");
```
laura-ding: delete this line

src/storage/QueryBaseProcessor.inl
```diff
@@ -34,20 +34,29 @@ bool QueryBaseProcessor<REQ, RESP>::validOperation(nebula::cpp2::SupportedType v
     return true;
 }
 
+template <typename REQ, typename RESP>
+void QueryBaseProcessor<REQ, RESP>::initEdgeContext(const std::vector<EdgeType>& eTypes) {
+    std::transform(eTypes.cbegin(), eTypes.cend(),
+                   std::inserter(edgeContexts_, edgeContexts_.end()),
+                   [this](const auto& ec) {
```
laura-ding: delete `this`

src/storage/test/QueryEdgePropsTest.cpp
monadbobo: Here are two reasons:
1 The GO statement (example: go from 1 over 2) was to set _dst as the return prop to request the storage to return.
2 For the query edge prop, it does not use the default property returned. @CPWstatic

src/storage/QueryBoundProcessor.cpp
monadbobo: This is just to deal with GO statements, such as "go from 1 over *" , this statement needs to return "_dst" for all edges.

src/interface/storage.thrift
monadbobo: The main reason was that the vertex prop request also calls the QueryBaseProcessor<REQ, RESP>::checkAndBuildContexts method, which uses "req.over_all_edges".

src/storage/test/QueryEdgePropsTest.cpp
dangleptr: To be faster,  rollback the changes about QueryEdgeProps,   it has no relationship with this pr and i don't think it is good enough to be merged in.

src/interface/storage.thrift
dangleptr: To be compatible,  we could do it temporarily,  but it is not a good reason. We should not design our interface as implementation.  Add some comments,  and we go on this pr.

src/storage/QueryBoundProcessor.cpp
dangleptr: I mean for sentence "go from 1 over * ",  we should return the whole edge key,   just '_dst' is not enough

src/storage/QueryBaseProcessor.inl
```diff
@@ -209,25 +226,37 @@ bool QueryBaseProcessor<REQ, RESP>::checkExp(const Expression* exp) {
...
                 VLOG(1) << "No edge requested!";
                 return false;
             }
+
             auto* edgeExp = static_cast<const AliasPropertyExpression*>(exp);
-            const auto* propName = edgeExp->prop();
-            auto schema = this->schemaMan_->getEdgeSchema(spaceId_, edgeContext_.edgeType_);
+
+            // TODO(simon.liu) we need handle rename.
+            auto edgeStatus = this->schemaMan_->toEdgeType(spaceId_, *edgeExp->alias());
```
dangleptr: Could we get the original edgeName from edgeExp?

src/dataman/RowReader.h
monadbobo: Ok.

src/graph/GoExecutor.cpp
```diff
@@ -630,6 +699,8 @@ bool GoExecutor::setupInterimResult(RpcResponse &&rpcResp, std::unique_ptr<Inter
                 case 3:
                     writer << boost::get<std::string>(column);
                     break;
+                case 4:
```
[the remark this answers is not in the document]
monadbobo: This was a mistake, i will fix this.

src/parser/parser.yy
monadbobo: Ok

src/storage/QueryBaseProcessor.inl
monadbobo: Here we can't get the name before alias(as), so for the time being we don't support the alias of edge. If we want to support alias, the change will be bigger, so I think it would be better to re-open a pr to do this.

src/storage/QueryBoundProcessor.cpp
monadbobo: If all the default properties are returned, it is not compatible with the return of a single edge.

For example: "go from 1 over edge1", this sentence will return edge1._dst.

src/storage/test/QueryEdgePropsTest.cpp
monadbobo: Ok, I will rollback this.

src/storage/QueryBaseProcessor.inl
dangleptr: OK.  Let's open another pr for it.

monadbobo:
Jenkins go!

src/graph/GoExecutor.cpp
```diff
@@ -500,29 +579,19 @@ StatusOr<std::vector<storage::cpp2::PropDef>> GoExecutor::getStepOutProps() {
 
 
 StatusOr<std::vector<storage::cpp2::PropDef>> GoExecutor::getDstProps() {
+std::vector<storage::cpp2::PropDef> props;
```
dangleptr: alignment

src/graph/GoExecutor.cpp
```diff
@@ -403,6 +421,28 @@ void GoExecutor::onVertexProps(RpcResponse &&rpcResp) {
...
+std::vector<std::string> GoExecutor::getEdgeNamesFromResp(RpcResponse &rpcResp) const {
+    std::vector<std::string> names;
+    auto spaceId = ectx()->rctx()->session()->space();
+    for (auto &resp : rpcResp.responses()) {
+        auto *vertices = resp.get_vertices();
+        if (vertices == nullptr) {
+            continue;
+        }
+
+        for (auto &vdata : *vertices) {
```
dangleptr: Use edge_schema

src/graph/GoExecutor.cpp
```diff
@@ -668,162 +737,195 @@ void GoExecutor::onEmptyInputs() {
...
-                    }
-                    auto index = tagIter->second;
-                    const nebula::cpp2::ValueType &type = vschema->getFieldType(index);
-                    if (type == CommonConstants::kInvalidValueType()) {
-                        auto msg =
-                            folly::sformat("Tag: {} no schema for the index {}", tagName, index);
-                        LOG(ERROR) << msg;
-                        return Status::Error(msg);
-                    }
-                    auto res = RowReader::getPropByIndex(vreader.get(), index);
```
dangleptr: I notice you change the logic from getPropByIndex to getPropByName,   the performance will be hurt.  Are you OK with it @dutor

src/graph/GoExecutor.cpp
```diff
@@ -403,6 +421,26 @@ void GoExecutor::onVertexProps(RpcResponse &&rpcResp) {
     UNUSED(rpcResp);
 }
 
+std::vector<std::string> GoExecutor::getEdgeNamesFromResp(RpcResponse &rpcResp) const {
+    std::vector<std::string> names;
+    auto spaceId = ectx()->rctx()->session()->space();
+    for (auto &resp : rpcResp.responses()) {
```
dangleptr: Do we need to scan all responses from all hosts?

laura-ding:
jenkins go

laura-ding:
jenkins go

src/graph/GoExecutor.cpp
monadbobo: We don't need it here, so I will modify it.

dutor:
Generally, this looks good to me, thanks for the big effort.

As @dangleptr said, there truly are some changes that have performance impact. But I think we could improve it later on.

CPWstatic:
Jenkins go!

dangleptr:
@monadbobo Please fix the conflicts,   let's go on.

monadbobo:
Jenkins go!

PR #703, opened 2019-07-29
fix wal rollback bug

### LINKED ISSUE #685: [Raft] bugs about term.
I am doing some basic raft test recently, here is a bug which is hard to recur. I will show some key log as follows:

---

One candidate (call it A) receive a askForVoteRequest from B like this:
```
A Recieved a VOTING request from B: B.current_term = 185, B.lastLogId = 440, B.lastLogTerm = 159
```
and A rejected it because its proposed term 195 is bigger than Req.term 185 from B.
```
The partition is currently proposing term 195. The term proposed by the candidate is no greater, so it will be rejected.

A.proposedTerm_ = 195, A.currTerm_ = 179, Term of AskForVoteRequeset from B = 185
```

**But B get a vote from C, and become leader of term 185. When B send appendLogReq to A, strange things happen, because A's term 179 is smaller than B's term 185, it will turn into a follower and apply those request.**

```
A Received logAppend: B.current_term = 185, B.committedLogId = 439, B.lastLogIdSent = 440, B.lastLogTermSent = 159
The current role is Candidate.
The current role is Candidate. Will follow the new leader (which is B)
```

Then A will crash when rolling back wal. The key point is that we use proposed_term to compare in leader election, but current_term to compare in log replication. What do you think about this? Do we need to implement the pre-vote?

The ut is Figure8 in #681, recurs sometimes.

@dangleptr @sherman-the-tank

critical27:
Based on #681.

There is a bug in when we rollback wal, say we have wal of logId [979, 984], and when we want to rollback to any log id in that range, it would lead to crash.

You can always recur in TEST(FileBasedWal, RollbackToZero), and try to rollback to 984 before rollback to 0. 

In fact, we don't need to check `CHECK(it != walFiles_.end())`. If the log id is the last log id in a wal, we get `walFiles_.end()` when try to find coresponding wal files with `walFiles_.upper_bound(id)`, that's ok. It will try to find the wal with biggest id and try to find the log in it.

close #685.


dangleptr:
Jenkins go

PR #706, opened 2019-07-31
Fix bunch of bugs in write test. 

dangleptr:
The bugs across client,  raft,  wal. 

After this patch,  storage cluster works stable during write test.  Now, we could kill one host during test,  then the leader could be changed to other hosts, and client could reconnect to the new leader after a while.   We could restart the killed one again after a while,  it will catch up the data. 

Thanks for @darionyaphet @critical27 @laura-ding help during the test.


src/kvstore/raftex/RaftPart.cpp
```diff
@@ -263,7 +263,15 @@ void RaftPart::start(std::vector<HostAddr>&& peers, bool asLearner) {
...
 
-    committedLogId_ = lastCommittedLogId();
+    auto logIdAndTerm = lastCommittedLogId();
+    committedLogId_ = logIdAndTerm.first;
+    term_ = proposedTerm_ = logIdAndTerm.second;
+
+    if (lastLogId_ < committedLogId_) {
+        lastLogId_ = committedLogId_;
+        lastLogTerm_ = term_;
+        wal_->reset(lastLogId_ + 1);
```
critical27: Maybe reset to committedLogId_ + 1?

src/kvstore/raftex/RaftPart.cpp
```diff
@@ -419,18 +430,21 @@ folly::Future<AppendLogResult> RaftPart::appendLogAsync(ClusterID source,
...
+            firstId = lastLogId_ + 1;
+            termId = term_;
         }
-        firstId = lastLogId_ + 1;
-        termId = term_;
+    }
+    if (res != AppendLogResult::SUCCEEDED) {
+        LOG(ERROR) << idStr_
+                   << "Cannot append logs, clean the buffer";
+        sendingPromise_.setValue(std::move(res));
```
critical27: I think we also need to set cachingPromise to res here. WDYT?

src/kvstore/raftex/RaftPart.cpp
dangleptr: I had removed the param passed in in reset interface. 
The interface just clean up all wall files. So we don't need the logId here.

src/kvstore/raftex/RaftPart.cpp
dangleptr: En...   We could do that,  but the cachingPromise will be set when sending it.

dangleptr:
Jenkins go

src/kvstore/raftex/RaftPart.cpp
```diff
@@ -479,62 +488,60 @@ void RaftPart::appendLogsInternal(AppendLogsIterator iter, TermID termId) {
...
-                LOG(ERROR) << idStr_ << "Failed to write into WAL";
-                res = AppendLogResult::E_WAL_FAILURE;
-                break;
-            }
-            lastId = wal_->lastLogId();
-            VLOG(2) << idStr_ << "Succeeded writing logs ["
-                    << iter.firstLogId() << ", " << lastId << "] to WAL";
-        }  // end of lock scope
+        std::lock_guard<std::mutex> g(raftLock_);
+        if (status_ != Status::RUNNING) {
```
wadeliuyi: maybe we can package the status/role/term check to a function, because processAppendLogResponses has the same logic.

dangleptr:
Jenkins go

src/kvstore/raftex/RaftPart.cpp
dangleptr: Yes,  but i prepared to remove the retry logic in the future.  Let's leave it here now.

src/kvstore/raftex/RaftPart.cpp
```diff
@@ -567,16 +597,14 @@ void RaftPart::replicateLogs(folly::EventBase* eb,
...
-                eb,
-                [=] () -> Future<cpp2::AppendLogResponse> {
-                    return hostPtr->appendLogs(eb,
-                                               currTerm,
-                                               lastLogId,
-                                               committedId,
-                                               prevLogTerm,
-                                               prevLogId);
-                });
+            return via(eb, [=] () -> Future<cpp2::AppendLogResponse> {
```
wadeliuyi: I think we can write more "c++" style code to replace "gen::from|gen::map|gen::as", because it hard to debug in gdb.

src/kvstore/raftex/RaftPart.cpp
```diff
@@ -699,22 +729,23 @@ void RaftPart::processAppendLogResponses(
...
+                        if (casRet.empty()) {
                             // Failed
                             sendingPromise_.setOneSingleValue(
                                 AppendLogResult::E_CAS_FAILURE);
                         }
-                        return res;
+                        return casRet;
                     });
                 logs_.clear();
+                bufferOverFlow_ = false;
```
wadeliuyi: maybe no need logs_.clear(), because it has been moved.

PR #709, opened 2019-08-01
Eliminate duplicate definitions in PrimaryExpression

zhangguoqing:
VariantType vs. Operand

### DIFF
--- a/src/common/filter/Expressions.h
+++ b/src/common/filter/Expressions.h
@@ -577,13 +577,6 @@ private:
 // literal constants: bool, integer, double, string
 class PrimaryExpression final : public Expression {
 public:
-    using Operand = boost::variant<bool, int64_t, double, std::string>;
-    enum Which {
-        kBool = 0,
-        kInt = 1,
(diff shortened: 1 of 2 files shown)


zhangguoqing:
jenkins go

src/common/filter/Expressions.h
```diff
@@ -577,13 +577,13 @@ class SourcePropertyExpression final : public Expression {
 // literal constants: bool, integer, double, string
 class PrimaryExpression final : public Expression {
 public:
-    using Operand = boost::variant<bool, int64_t, double, std::string>;
     enum Which {
```
critical27: There are some macro in base.h like `VAR_STR` and so on, it seems we don't need the enum Which any more.

zhangguoqing: Thanks!  
I have only paid attention to the local and did not notice the definition of VAR_INT64, VAR_DOUBLE, VAR_BOOL, VAR_STR macros in Base.h. 
It has been updated now, please review again, thx.

PR #713, opened 2019-08-01
fix a problem when set flags in all module.

critical27:
1. Fix when we want to update variables in all module, it would fail if one config is immutable. For now, it will update all mutable variables and ignore the immutable variables.
2. Delete all `load_config_interval_secs` (it is used in early implement version), just keep same speed with load_data_interval_secs.

TODO: Maybe we will organize all config into a separate file, and avoid hard code, and config manager will load it and try to register on meta.


sherman-the-tank:
> TODO: Maybe we will organize all config into a separate file, and avoid hard code, and config manager will load it and try to register on meta.

First, can we add this TODO into the code? Messages here intend to be lost when the PR is merged

src/meta/processors/configMan/SetConfigProcessor.cpp
```diff
@@ -21,43 +21,37 @@ void SetConfigProcessor::process(const cpp2::SetConfigReq& req) {
     if (module != cpp2::ConfigModule::ALL) {
         // When we set config of a specified module, check if it exists.
         // If it exists and is mutable, update it.
-        if (!setOneConfig(module, name, type, mode, value, data)) {
```
sherman-the-tank: Why did we remove the result check here?

critical27: If `setOneConfig` return false, which means the config doesn't exist or it is immutable. We just return error at below.

critical27:
> Unit testing passed.

Yeah, I will open a issue to record.

PR #721, opened 2019-08-02
Enhance StoragePerf

darionyaphet:
In this PR, support config data's size per message and multi message per request.


dangleptr:
Not only value size.  Currently,  one thread could sent at most 1000 requests currently.  We should break the limits.   One task could sent N requests once time.

darionyaphet:
I know. I will open another PR to support multi requests

sherman-the-tank:
Can someone explain what's the purpose of this PR?

darionyaphet:
In this PR I have support two features:

 * Support confinable properties's data size when using `storage_perf` to do stress test.
 * Support send multi message per request.

Maybe I should update it's title.  @sherman-the-tank

src/tools/storage-perf/StoragePerfTool.cpp
```diff
@@ -19,28 +18,60 @@ DEFINE_string(method, "getNeighbors", "method type being tested,"
...
+
 class Perf {
 public:
     int run() {
         uint32_t qpsPerThread = FLAGS_qps / FLAGS_threads;
-        uint32_t interval = 1000 / qpsPerThread;
-        CHECK_GT(interval, 0) << "qpsPerThread should not large than 1000, interval " << interval;
+        int32_t radix = qpsPerThread / 1000;
+        int32_t slotSize = sizeof(slots_) / sizeof(int32_t);
+        if (qpsPerThread >= 1000) {
```
critical27: Maybe we don't need this if statement?

src/tools/storage-perf/StoragePerfTool.cpp
```diff
@@ -51,6 +82,7 @@ class Perf {
         CHECK(mClient_->waitForMetadReady());
         client_ = std::make_unique<StorageClient>(threadPool_, mClient_.get());
         time::Duration duration;
+        static uint32_t interval = 1;
```
critical27: I guess you mean 100ms here? According to my understanding, you split 1 second to 10 * 100ms, and set corresponding qps of each slot.

critical27: Ah, I got it.

src/tools/storage-perf/StoragePerfTool.cpp
darionyaphet: Good catch & fixed :)

src/tools/storage-perf/StoragePerfTool.cpp
darionyaphet: 10 ms is a cycle. `slots_` is the sending plan for a cycle. So it should sleep 1 ms.

src/tools/storage-perf/README.md
```diff
@@ -0,0 +1,28 @@
+### Storage Perf Tool
+
+`_build/storage_perf` is the performance tool to test the storage service.
```
dangleptr: Well done

PR #723, opened 2019-08-04
Nebula Service Script

darionyaphet:
In `services.sh`, we could use it to start|stop|restart|status|kill meta, storage and graph daemons.

`USAGE: ./start-all.sh <start|stop|restart|status|kill>`


scripts/start-all.sh
```diff
@@ -0,0 +1,47 @@
...
+
+if [ ! -f ${SCRIPT_PATH}"/metas" ]; then
+  echo "${SCRIPT_PATH}/metas not exist"
+  exit 0
+fi
+
+for host in `cat ${SCRIPT_PATH}/metas`
+do
+  echo "start $host"
+  ssh $host "/usr/local/scripts/nebula.service -c /usr/local/etc/nebula-metad.conf start metad >/dev/null 2>&1 &"
```
CPWstatic: Users need to configure the secret key for ssh manually, right?

darionyaphet: yeap , if you don't setting the SSH password less login you have to input the password.

PR #726, opened 2019-08-05
Meta Server init fail when leader change

andychow01:
When Meta starts, if this is the first time Meta creates a cluster id and at this moment leader change(in kvstore->asyncMultiPut() function), Meta Server start will fail.

### DIFF
--- a/src/meta/ClusterManager.cpp
+++ b/src/meta/ClusterManager.cpp
@@ -12,36 +12,80 @@
 #include "meta/processors/Common.h"
 #include "folly/synchronization/Baton.h"
 
+
+DEFINE_int32(putTryNum, 10, "try num of store clusterId to kvstore");
+
 namespace nebula {
 namespace meta {
 
-using KVStore = nebula::kvstore::KVStore;
-using ResultCode = nebula::kvstore::ResultCode;
-const char* ClusterManager::kClusterIdKey = "metaClusterIdKey";
+const char* ClusterManager::kClusterIdKey = "__meta_cluster_id_key__";
+
+ResultCode ClusterManager::kvPut(KVStore* kvstore, std::string& strClusterId) {
+    ResultCode code;
+    folly::Baton<true, std::atomic> baton;
+    kvstore->asyncMultiPut(kDefaultSpaceId,
+                           kDefaultPartId,
+                           {{kClusterIdKey, strClusterId}},
+                           [&](ResultCode innerCode) {
+                               code = innerCode;
+                               baton.post();
+                           });
+    baton.wait();
+    return code;
+}
 
 bool ClusterManager::loadOrCreateCluId(KVStore* kvstore) {
     std::string strClusterId;
-    kvstore->get(kDefaultSpaceId, kDefaultPartId, kClusterIdKey, &strClusterId);
 
-    bool ret = false;
+    ResultCode getCode = kvstore->get(kDefaultSpaceId,
+                                      kDefaultPartId,
+                                      kClusterIdKey,
+                                      &strClusterId);
+    // TODO: check if leader election success
+    if ((getCode != ResultCode::SUCCEEDED) && (getCode != ResultCode::ERR_KEY_NOT_FOUND)) {
+        LOG(ERROR) << "ClusterManager::loadOrCreateCluId get error!";
+        return false;
+    }
+
     if (!strClusterId.empty()) {
         clusterId_ = folly::to<ClusterID>(strClusterId);
+        LOG(INFO) << "ClusterManager::loadOrCreateCluId load from kvstore, clusterId: "
+                  << clusterId_;
         clusterIdIsSet_ = true;
         return true;
     } else {
         createClusterId();
         strClusterId = folly::stringPrintf("%ld", clusterId_);
-        folly::Baton<true, std::atomic> baton;
-        kvstore->asyncMultiPut(kDefaultSpaceId,
-                               kDefaultPartId,
-                               {{kClusterIdKey, strClusterId}},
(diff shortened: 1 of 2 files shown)


src/daemons/MetaDaemon.cpp
```diff
@@ -117,10 +117,15 @@ int main(int argc, char *argv[]) {
     }
 
     auto *kvstore_ = kvstore.get();
+    auto *nebulaStore = dynamic_cast<nebula::kvstore::NebulaStore*>(kvstore_);
+    if (nebulaStore == nullptr) {
```
wadeliuyi: do we not support cluster id when we not use nebula store?

src/meta/ClusterManager.cpp
```diff
@@ -8,40 +8,81 @@
...
+#include "kvstore/NebulaStore.h"
 #include "meta/processors/Common.h"
 #include "folly/synchronization/Baton.h"
 
 namespace nebula {
 namespace meta {
 
-using KVStore = nebula::kvstore::KVStore;
-using ResultCode = nebula::kvstore::ResultCode;
 const char* ClusterManager::kClusterIdKey = "metaClusterIdKey";
```
wadeliuyi: system key maybe need start with "__"

src/daemons/MetaDaemon.cpp
andychow01: Now, cluster id store in  nebulaStore. Maybe, we can add clusterId support to other store.

src/meta/ClusterManager.cpp
```diff
@@ -8,40 +8,81 @@
...
+                return false;
+            }
+            auto leaderAddrRet = nebulaStore->partLeader(kDefaultSpaceId, kDefaultPartId);
+            if (!ok(leaderAddrRet)) {
+                LOG(ERROR) << "ClusterManager::loadOrCreateCluId partLeader() error!";
+                return false;
+            }
+            auto leaderAddr = value(leaderAddrRet);
+            bool elecCompleted = (leaderAddr != HostAddr(0, 0));
+            if (elecCompleted) {
```
wadeliuyi: no need retry?

src/meta/ClusterManager.cpp
```diff
@@ -8,40 +8,81 @@
...
-                                   ret = code == ResultCode::SUCCEEDED;
-                                   baton.post();
-                               });
-        baton.wait();
-    }
-    return ret;
+        int putTryNum = 3;
+        while (putTryNum > 0) {
+            ResultCode code = kvPut(nebulaStore, strClusterId);
+            --putTryNum;
```
wadeliuyi: move end of the while

src/meta/ClusterManager.cpp
```diff
@@ -8,40 +8,81 @@
...
+                return false;
+            }
+            auto leaderAddr = value(leaderAddrRet);
+            bool elecCompleted = (leaderAddr != HostAddr(0, 0));
+            if (elecCompleted) {
+                LOG(INFO) << "ClusterManager::loadOrCreateCluId elecCompleted";
+                return true;
+            } else {
+                LOG(INFO) << "ClusterManager::loadOrCreateCluId not elecCompleted";
+                sleep(5);
```
wadeliuyi: why just this condition need retry?

src/daemons/MetaDaemon.cpp
wadeliuyi: ClusterManager just dependence on the kvStore's get and put interface, so I think here no need cast.

src/meta/ClusterManager.cpp
andychow01: Because ERR_LEADER_CHANGED implies leader election has not completed(leader election need several seconds). Other error code(i.e. ERR_INVALID_ARGUMENT) normally we need not retry(retry is no use).

src/meta/ClusterManager.cpp
andychow01: yes

src/meta/ClusterManager.cpp
andychow01: yes

src/daemons/MetaDaemon.cpp
andychow01: ClusterManager depends on Nebula::asyncMultiPut(KVStore has not this function), so dynamic_cast is needed.

src/meta/ClusterManager.cpp
andychow01: kvPut function has retry FLAGS_putTryNum times when leader election has not completed.

src/meta/ClusterManager.cpp
```diff
@@ -12,36 +12,80 @@
 #include "meta/processors/Common.h"
 #include "folly/synchronization/Baton.h"
 
+
+DEFINE_int32(putTryNum, 10, "try num of store clusterId to kvstore");
```
wadeliuyi: config use "_" naming like put_try_num

src/meta/ClusterManager.cpp
```diff
@@ -12,36 +12,80 @@
...
+            auto leaderAddr = value(leaderAddrRet);
+            bool elecCompleted = (leaderAddr != HostAddr(0, 0));
+            if (elecCompleted) {
+                LOG(INFO) << "ClusterManager::loadOrCreateCluId elecCompleted";
+                return true;
+            } else {
+                LOG(INFO) << "ClusterManager::loadOrCreateCluId not elecCompleted";
+                sleep(1);
+            }
+            --FLAGS_putTryNum;
```
wadeliuyi: not change the global configure variable directly

Issue #707, opened 2019-07-31
Fix flakey UT nebula_store_test


dangleptr:
http://ci.vesoft.com/job/Nebula-Graph-UnitTest/1597/consoleFull

dangleptr:
```
19:36:19 42: [       OK ] NebulaStoreTest.SimpleTest (1545 ms)
19:36:19 42: [ RUN      ] NebulaStoreTest.PartsTest
19:36:19 42: I0731 19:36:19.837786 122023 RocksEngine.cpp:106] open rocksdb on /tmp/nebula_store_test.4MT3kb/disk1/nebula/0/data
19:36:19 42: I0731 19:36:19.947345 122023 RocksEngine.h:104] Release rocksdb on /tmp/nebula_store_test.4MT3kb/disk1/nebula/0
19:36:19 42: I0731 19:36:19.947736 122023 RocksEngine.cpp:106] open rocksdb on /tmp/nebula_store_test.4MT3kb/disk2/nebula/0/data
19:36:20 42: I0731 19:36:20.045956 122023 RocksEngine.h:104] Release rocksdb on /tmp/nebula_store_test.4MT3kb/disk2/nebula/0
19:36:20 42: I0731 19:36:20.046301 122023 NebulaStore.cpp:51] Start the raft service...
19:36:20 42: I0731 19:36:20.046525 122023 RaftexService.cpp:65] Init thrift server for raft service.
19:36:20 42: I0731 19:36:20.046612 122151 RaftexService.cpp:92] Starting the Raftex Service
19:36:20 42: E0731 19:36:20.048321 122151 ThriftServer.cpp:440] Got an exception while setting up the server: No threads available
19:36:20 42: E0731 19:36:20.048389 122151 RaftexService.cpp:83] Setup the Raftex Service failed, error: No threads available
19:36:20 42: I0731 19:36:20.049808 122164 Acceptor.cpp:453] Dropping all connections from Acceptor=0x7f5586c7d300 in thread 0x7f558788f000
19:36:20 42: I0731 19:36:20.049891 122164 Acceptor.cpp:453] Dropping all connections from Acceptor=0x7f5586c7df00 in thread 0x7f558788f000
19:36:20 42: I0731 19:36:20.049966 122165 Acceptor.cpp:453] Dropping all connections from Acceptor=0x7f5586c7d600 in thread 0x7f5586cad280
[...]

Issue #727, opened 2019-08-05
crashed by PartsTest in nebula_store_test


monadbobo:
compiled with addresssanitizer.

stack:

> *** Aborted at 1564996893 (unix time) try "date -d @1564996893" if you are using GNU date ***
> PC: @          0x1ecbce5 apache::thrift::Cpp2Worker::create()
> *** SIGSEGV (@0x58000ae) received by PID 15159 (TID 0x7f927dbc8700) from PID 92274862; stack trace: ***
>     @          0x23395e9 google::(anonymous namespace)::FailureSignalHandler()
>     @     0x7f928582af70 (unknown)
>     @          0x1ecbce5 apache::thrift::Cpp2Worker::create()
>     @          0x1ecbff0 apache::thrift::ThriftAcceptorFactory::newAcceptor()
>     @          0x2198b47 wangle::ServerWorkerPool::threadStarted()
>     @          0x223580c folly::ThreadPoolExecutor::addThreads()
>     @          0x2237f68 folly::ThreadPoolExecutor::setNumThreads()
>     @          0x1eb29af apache::thrift::ThriftServer::setup()
>     @          0x1d217c0 nebula::raftex::RaftexService::setup()
>     @          0x1d21f87 nebula::raftex::RaftexService::serve()
>     @          0x1d20f00 _ZZN6nebula6raftex13RaftexService5startEvENKUlvE_clEv
>     @          0x1d247ed _ZSt13__invoke_implIvZN6nebula6raftex13RaftexService5startEvEUlvE_JEET_St14__invoke_otherOT0_DpOT1_
>     @          0x1d243a6 _ZSt8__invokeIZN6nebula6raftex13RaftexService5startEvEUlvE_JEENSt15__invoke_resultIT_JDpT0_EE4typeEOS5_DpOS6_
>     @          0x1d24f16 _ZNSt6thread8_InvokerISt5tupleIJZN6nebula6raftex13RaftexService5startEvEUlvE_EEE9_M_invokeIJLm0EEEEDTcl8__invokespcl10_S_declvalIXT_EEEEESt12_Index_tupleIJXspT_EEE
>     @          0x1d24ea4 _ZNSt6thread8_InvokerISt5tupleIJZN6nebula6raftex13RaftexService5startEvEUlvE_EEEclEv
[...]

PR #728, opened 2019-08-05
issue #727 crashed by PartsTest in nebula_store_test.

monadbobo:
sumarry:

fixed coredump for sharing the same iothreadpool in the two
different NebulaStore object.

This crash will only appear if you compiled with addresssanitizer.

### DIFF
--- a/src/kvstore/test/NebulaStoreTest.cpp
+++ b/src/kvstore/test/NebulaStoreTest.cpp
@@ -19,8 +19,6 @@ DECLARE_uint32(heartbeat_interval);
 namespace nebula {
 namespace kvstore {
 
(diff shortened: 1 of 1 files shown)


monadbobo:
jenkins go!

darionyaphet:
LGTM 👍 

Can I take it as we shouldn't share something  between test case ?

dangleptr:
> LGTM 👍
> 
> Can I take it as we shouldn't share something between test case ?

Not exactly.  The ioThreadPool here will be used as io pool for raft service.  So after one UT finished,  raft service will stop the pool in its dtor.  It will cause problems when we use the pool again in next UT.

dangleptr:
Close #707

PR #729, opened 2019-08-05
Changed the WAL to call write() on every log message

### LINKED ISSUE #531: Data loss due to WAL async flusher
Currently,  wal is only hold in cache,  and it will be flushed in background thread.  So data maybe lost when process crashed.  

The problem is very critical.  We should fix it firstly.

sherman-the-tank:
We moved away from the memory buffered version. In this PR, every log message will be written to the system cache to prevent message loss due to the crash

fixed #531


sherman-the-tank:
jenkins go

src/kvstore/wal/FileBasedWal.cpp
```diff
@@ -48,9 +44,17 @@ FileBasedWal::FileBasedWal(const folly::StringPiece dir,
 
     scanAllWalFiles();
     if (!walFiles_.empty()) {
+        firstLogId_ = walFiles_.begin()->second->firstId();
         auto& info = walFiles_.rbegin()->second;
-        lastLogId_ = info->lastId();
-        lastLogTerm_ = info->lastTerm();
+        if (info->lastId() <= 0 && walFiles_.rbegin()++ != walFiles_.rend()) {
```
wadeliuyi: what is the condition mean "walFiles_.rbegin()++ != walFiles_.rend()"? It look like useless about ++.

src/kvstore/wal/FileBasedWal.cpp
```diff
@@ -311,125 +309,81 @@ void FileBasedWal::prepareNewFile(LogID startLogId) {
...
+            close(fd);
+            return 0;
+        }
 
-    // Rollover if required
-    if (buffer->needToRollover()) {
-        closeCurrFile();
-    }
+        if (id == logId) {
+            // Found
```
wadeliuyi: close(fd)

src/kvstore/wal/FileBasedWal.cpp
```diff
@@ -449,32 +403,46 @@ bool FileBasedWal::appendLogInternal(BufferPtr& buffer,
...
+    } else if (currInfo_->size() + strBuf.size() > maxFileSize_) {
+        // Need to roll over
+        closeCurrFile();
 
-    // Create a new buffer if needed
-    if (!buffer) {
-        std::unique_lock<std::mutex> guard(buffersMutex_);
-        buffer = createNewBuffer(id, guard);
+        std::lock_guard<std::mutex> g(walFilesMutex_);
+        prepareNewFile(id);
```
wadeliuyi: if here need a lock to protected the current file, I think we should protected the file every where we use it.

src/kvstore/wal/FileBasedWal.cpp
```diff
@@ -449,32 +403,46 @@ bool FileBasedWal::appendLogInternal(BufferPtr& buffer,
...
     }
 
-    DCHECK_EQ(
-        id,
-        static_cast<int64_t>(buffer->firstLogId() + buffer->numLogs()));
-    buffer->push(term, cluster, std::move(msg));
+    ssize_t bytesWritten = write(currFd_, strBuf.data(), strBuf.size());
```
wadeliuyi: what is the effect about performance that call write every time? and the os cache will lost when power off.

src/kvstore/wal/FileBasedWal.cpp
```diff
@@ -311,125 +309,81 @@ void FileBasedWal::prepareNewFile(LogID startLogId) {
...
-    // Dump the buffer to file
-    auto lastLog = buffer->accessAllLogs(accessFn);
-
-    // Dump the rest if any
-    if (!cord.empty()) {
-        dumpCord(cord, firstIdInCord, lastLog.first, lastLog.second);
-    }
+        // Read the message length
+        int32_t len;
+        if (pread(fd, &len, sizeof(int32_t), pos + sizeof(LogID) + sizeof(TermID))
```
wadeliuyi: maybe we should read the termId from end to begin, because there are duplicate log id in file when we call "FileBasedWal::appendLogs" failed.

src/kvstore/wal/FileBasedWal.cpp
sherman-the-tank: Good catch!! 👍  Shame on myself. Supposed to be
```
++walFiles_.rbegin() != walFiles_.rend()
```

src/kvstore/wal/FileBasedWal.cpp
sherman-the-tank: 👍

src/kvstore/wal/FileBasedWal.cpp
sherman-the-tank: The lock actually protected the walFiles_ (list of WAL info), not the file itself

src/kvstore/wal/FileBasedWal.cpp
sherman-the-tank: Good question. The implementation here is based on the assumption that binary crash happens way more often then physical machine crash

There is definitely some sacrifice of the performance here, but it's acceptable

src/kvstore/wal/FileBasedWal.cpp
sherman-the-tank: There will be no dup log id in the same file. When rolling back, a new WAL file will be created to overwrite the log ids in the previous file

sherman-the-tank:
@wadeliuyi Awesome comments 👍 I'll update the PR

src/kvstore/wal/FileBasedWal.cpp
wadeliuyi: because I find that there is reopen current file in prepareNewFile and rollbackToLog, so if we just lock here, current thread unlock walFilesMutex_ and begin write data to currFd_, then the other thread call rollbackToLog and reopen current file, maybe there is some problem.

src/kvstore/wal/FileBasedWal.cpp
```diff
@@ -311,125 +309,81 @@ void FileBasedWal::prepareNewFile(LogID startLogId) {
 }
```
wadeliuyi: maybe we need process the os error more safely and friendly when open file failed, because this error is very common when cpu has a high pressure, crash directly is danger.

sherman-the-tank: Agreed.

For simplicity, we used CHECK and LOG(FATAL) too much.

sherman-the-tank: We can address this in a separate PR

src/kvstore/wal/FileBasedWal.cpp
sherman-the-tank: Good point

We don't expect ```rollbackToLog()``` and ```appendLog()``` to be called simultaneously

wadeliuyi: so maybe we can design it lock free?

src/kvstore/wal/FileBasedWal.cpp
```diff
@@ -48,9 +44,17 @@ FileBasedWal::FileBasedWal(const folly::StringPiece dir,
...
     scanAllWalFiles();
     if (!walFiles_.empty()) {
+        firstLogId_ = walFiles_.begin()->second->firstId();
         auto& info = walFiles_.rbegin()->second;
-        lastLogId_ = info->lastId();
-        lastLogTerm_ = info->lastTerm();
+        if (info->lastId() <= 0 && walFiles_.rbegin()++ != walFiles_.rend()) {
+            auto it = walFiles_.rbegin();
+            it++;
+            lastLogId_ = info->firstId() - 1;
```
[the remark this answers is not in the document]
sherman-the-tank: I think your question might be "Why we set the lastLogId_ to one less than firstId()"

This code block (info->lastId() <= 0) means the current WAL file is empty. The firstId() is extracted from the file name, which will be the id for the first log in this empty file.

In the case of rollback, we actually don't physically remove the logs from the end of the WAL file. In stead, we create a new file, which first id is the id being rolled back. So we cannot use the last log id of a WAL file as the lastLogId_ of the WAL

src/kvstore/wal/FileBasedWal.cpp
```diff
@@ -312,124 +309,82 @@ void FileBasedWal::prepareNewFile(LogID startLogId) {
...
-                        " need to wait for vacancy";
-        // TODO: Output a counter here
-        // Need to wait for a vacant slot
-        slotReadyCV_.wait(guard, [self = shared_from_this()] {
-            return self->buffers_.size() < self->policy_.numBuffers;
-        });
-    }
+BufferPtr FileBasedWal::getLastBuffer(LogID id, size_t expectedToWrite) {
+    std::unique_lock<std::mutex> g(buffersMutex_);
+    if (!buffers_.empty()) {
```
dangleptr: Why we still need a list of buffers?   
One buffer is enough now.

src/kvstore/wal/FileBasedWal.cpp
```diff
@@ -546,132 +498,75 @@ bool FileBasedWal::rollbackToLog(LogID id) {
...
-        while (it != walFiles_.end()) {
-            // Need to remove the file
-            VLOG(2) << "Removing file " << it->second->path();
-            unlink(it->second->path());
-            it = walFiles_.erase(it);
-        }
+    //------------------------------
+    // 2. Roll back in-memory buffers
+    //------------------------------
+    {
```
dangleptr: We don't need rollback the buffers.   Just clean them all.

src/kvstore/wal/FileBasedWal.cpp
sherman-the-tank: Totally agreed.

Since the current implementation already implemented buffer list, I just reused it. In order to keep the PR small, I suggest to switch to the ring buffer in a separate PR. What do you think?

src/kvstore/wal/FileBasedWal.cpp
sherman-the-tank: Yes, cleaning all of them is simple, but it might hurt the perf. For example, we have to read from the file

src/kvstore/wal/FileBasedWal.cpp
monadbobo: Why don't we use RocksDB to save the WAL?

src/kvstore/wal/FileBasedWal.cpp
```diff
@@ -450,69 +405,68 @@ bool FileBasedWal::appendLogInternal(BufferPtr& buffer,
...
+        prepareNewFile(id);
+    } else if (currInfo_->size() + strBuf.size() > maxFileSize_) {
+        // Need to roll over
+        closeCurrFile();
 
-    // Create a new buffer if needed
-    if (!buffer) {
-        std::unique_lock<std::mutex> guard(buffersMutex_);
-        buffer = createNewBuffer(id, guard);
+        std::lock_guard<std::mutex> g(walFilesMutex_);
```
wadeliuyi: as what you say that "We don't expect rollbackToLog() and appendLog() to be called simultaneously", and I check the code, and I think we can design it lock free.

src/kvstore/wal/FileBasedWal.cpp
```diff
@@ -48,10 +44,17 @@ FileBasedWal::FileBasedWal(const folly::StringPiece dir,
 
     scanAllWalFiles();
     if (!walFiles_.empty()) {
-        auto& info = walFiles_.rbegin()->second;
         firstLogId_ = walFiles_.begin()->second->firstId();
-        lastLogId_ = info->lastId();
-        lastLogTerm_ = info->lastTerm();
+        auto& info = walFiles_.rbegin()->second;
+        if (info->lastId() <= 0 && ++walFiles_.rbegin() != walFiles_.rend()) {
```
wadeliuyi: maybe check the condition walFiles_.size() >= 2 is faster.

src/kvstore/wal/FileBasedWal.cpp
```diff
@@ -450,69 +405,68 @@ bool FileBasedWal::appendLogInternal(BufferPtr& buffer,
...
-        buffer = createNewBuffer(id, guard);
+        std::lock_guard<std::mutex> g(walFilesMutex_);
+        prepareNewFile(id);
     }
 
-    DCHECK_EQ(
-        id,
-        static_cast<int64_t>(buffer->firstLogId() + buffer->numLogs()));
-    buffer->push(term, cluster, std::move(msg));
+    ssize_t bytesWritten = write(currFd_, strBuf.data(), strBuf.size());
```
monadbobo: Maybe we need a configuration to decide if we need fsync data. Just like  [sysvar_innodb_flush_log_at_trx_commit](https://dev.mysql.com/doc/refman/8.0/en/innodb-parameters.html#sysvar_innodb_flush_log_at_trx_commit) MySQL.

dutor: We could use _gather output_ to do this, aka. `pwritev`, which could save us one copy operation, which means a lot here.

Of course we could do it in a separate PR. Please let me take care of it.

src/kvstore/wal/FileBasedWal.cpp
sherman-the-tank: We cannot make it lock free, the reason is ```FileBasedWalIterator``` will access the file list, which could conflict with either ```appendLogs()``` or ```rollbackToLog()```

src/kvstore/wal/FileBasedWal.cpp
sherman-the-tank: haha, you are right 👍 

Don't remember why I chose this ugly condition to check :-(

src/kvstore/wal/FileBasedWal.cpp
sherman-the-tank: Sure, we could add this later. I think fsync and DIO are two choices

src/kvstore/wal/FileBasedWal.cpp
```diff
@@ -312,124 +309,82 @@ void FileBasedWal::prepareNewFile(LogID startLogId) {
...
-    };
-    if (!cord.applyTo(cb)) {
-        LOG(FATAL) << "Failed to flush the wal file";
-    } else {
-        // Succeeded writing all buffered content, adjust the file size
-        currInfo_->setSize(currInfo_->size() + cord.size());
-        currInfo_->setLastId(lastId);
-        currInfo_->setLastTerm(lastTerm);
+TermID FileBasedWal::readTermId(const char* path, LogID logId) {
+    int32_t fd = open(path, O_EXCL | O_RDONLY | O_CLOEXEC | O_LARGEFILE, 0644);
```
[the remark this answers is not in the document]
sherman-the-tank: Agreed 👍

src/kvstore/wal/FileBasedWal.cpp
sherman-the-tank: Ok, I'll leave it out of this PR

sherman-the-tank:
jenkins go

PR #732, opened 2019-08-06
Fix crash problem when stop storage/meta daemon

dangleptr:
For storage daemon,  we have many components inside.  And they depend on each other,  especially the ioThreadPool and thrift workers which are shared by both raft  thrift service and storage thrift service.  

As implement by thrift,  we should stop the services in order carefully,  otherwise it will be crashed.  To control each component easily,  we extract them into one class named StorageServer.  

For meta daemon,  we don't care about the performance.  So we have no needs to share the io and workers.


src/daemons/MetaDaemon.cpp
```diff
@@ -45,6 +45,68 @@ static std::unique_ptr<apache::thrift::ThriftServer> gServer;
 static void signalHandler(int sig);
 static Status setupSignalHandler();
 
+std::unique_ptr<nebula::kvstore::KVStore> initKV(std::vector<nebula::HostAddr> peers,
+                                                 nebula::HostAddr localhost) {
+    auto partMan
```
wadeliuyi: std::vector<nebula::HostAddr> &peers and nebula::HostAddr &localhost

src/storage/StorageServer.cpp
```diff
@@ -0,0 +1,166 @@
...
+    if (metaClient_) {
+        metaClient_->stop();
+    }
+    if (kvstore_) {
+        kvstore_.reset();
+    }
+    if (tfServer_) {
+        tfServer_->stop();
+    }
+    stopped_ = true;
```
wadeliuyi: move to line 148, maybe need change to atomic_bool, else set true here is useless

src/kvstore/NebulaStore.cpp
```diff
@@ -52,7 +53,9 @@ bool NebulaStore::init() {
     LOG(INFO) << "Start the raft service...";
     bgWorkers_ = std::make_shared<thread::GenericThreadPool>();
     bgWorkers_->start(FLAGS_num_workers);
-    raftService_ = raftex::RaftexService::createService(ioPool_, handlersPool_, raftAddr_.second);
+    raftService_ = raftex::RaftexService::createService(ioPool_,
+                                                        handlersPool_,
```
wadeliuyi: maybe worker is more readable.

src/storage/StorageServer.cpp
```diff
@@ -0,0 +1,166 @@
...
+    }
+    if (webStatus_ == Status::RUNNING) {
+        nebula::WebService::stop();
+        webStatus_ = Status::STOPPED;
+    }
+    if (metaClient_) {
+        metaClient_->stop();
+    }
+    if (kvstore_) {
+        kvstore_.reset();
```
wadeliuyi: maybe here need some comment that why we can reset kvstore before tfServer that tfServer is dependence on kvstore.

dangleptr: That's a good question!  And it is some hard to explain. 
In short,  after we stop ioThreadPool  in kvstore dtor,   it will cut-off the relationship with acceptors in storage thrift service.

src/kvstore/NebulaStore.cpp
dangleptr: Good suggestion.

src/storage/StorageServer.cpp
dangleptr: what's the meaning?  stopped_ has been atomic_bool yet.

src/daemons/MetaDaemon.cpp
dangleptr: Good point.

dangleptr:
Jenkins go

src/storage/StorageServer.cpp
```diff
@@ -0,0 +1,178 @@
...
+
+    WebService::registerHandler("/status", [] {
+        return new StorageHttpStatusHandler();
+    });
+    WebService::registerHandler("/download", [this] {
+        auto* handler = new storage::StorageHttpDownloadHandler();
+        handler->init(hdfsHelper_.get(), webWorkers_.get(), kvstore_.get(), dataPaths_);
+        return handler;
+    });
+    nebula::WebService::registerHandler("/ingest", [&] {
```
darionyaphet: please using this for capture variables

src/daemons/MetaDaemon.cpp
```diff
@@ -100,63 +167,23 @@ int main(int argc, char *argv[]) {
...
+    if (!initComponents(kvstore.get())) {
+        LOG(ERROR) << "Init components failed";
         return EXIT_FAILURE;
     }
-    std::unique_ptr<nebula::hdfs::HdfsHelper> helper =
-        std::make_unique<nebula::hdfs::HdfsCommandHelper>();
 
-    std::unique_ptr<nebula::thread::GenericThreadPool> pool =
-        std::make_unique<nebula::thread::GenericThreadPool>();
+    LOG(INFO) << "Start http service";;
```
darionyaphet: please remove duplicate semicolon

src/storage/StorageServer.cpp
```diff
@@ -0,0 +1,178 @@
...
+    }
+    return true;
+}
+
+void StorageServer::stop() {
+    if (stopped_) {
+        LOG(INFO) << "All services has been stopped";
+        return;
+    }
+    if (webStatus_ == Status::RUNNING) {
```
darionyaphet: Stop WebService at first, then close MetaClient and reset KVStore and ThriftService at last 👍

darionyaphet:
Generally  LGTM ! please review the comment 👍

src/storage/StorageServer.cpp
wadeliuyi: the stopped_ look like to stop multi thread to entry this function, but it useless if set the value at the last, maybe we should use exchange.

wadeliuyi: the stopped_ look like to stop multi thread to entry this function, but it useless if set the value at the last, maybe we should use exchange.

dangleptr: @wadeliuyi Not exactly.  stopped_ here is used to protect the server from being stopped twice.

src/daemons/MetaDaemon.cpp
dangleptr: Good catch

src/storage/StorageServer.cpp
dangleptr: Good catch

src/storage/StorageServer.cpp
wadeliuyi: I am not sure the case when user send a SIGTERM signal to storage service, the storage service will process this signal in some thread context, maybe it is not main context. then the main context exit from gServer->start(), and call gServer's dctor, maybe this is a multi thread env.

dangleptr: I got your point.  Yes,  let's move stopped_ = true before thrift stop

Issue #736, opened 2019-08-07
Remove add/remove hosts steps


dangleptr:
Now we discover hosts automatically. 
To avoid one host without registered disturb our cluster,  we support clusterID.

Now it's time to remove the  "add/remove hosts"  steps.
There are two actions:
1.  Set the flag --hosts_whitelist_enabled=false by default.  We need to adjust many UTs in this step. 
2.  Remove the code about add/remove hosts.

dangleptr:
The first one action is in progress,  i will submit a pr recently.  The pr make a big refactor about clusterId manager and active host manager to make the logic correct on multi replicas cluster.

sherman-the-tank:
Before we move further, let's have a quick brain storm

dangleptr:
Now,   the next action is to remove add/remove hosts code.  It is not in hurry,  put the task here.

PR #739, opened 2019-08-08
Make metad works stable on multi replicas cluster

dangleptr:
Main changes:
1.  Refactor clusterId and ActiveHostMan
2. make FLAGS_hosts_whitelist_enabled to be false by default,  so we don't need add host anymore.
3.  Fix some bugs.


dangleptr:
Jenkins go

src/daemons/MetaDaemon.cpp
```diff
@@ -69,9 +76,50 @@ std::unique_ptr<nebula::kvstore::KVStore> initKV(std::vector<nebula::HostAddr> p
...
     if (!(kvstore->init())) {
-        LOG(ERROR) << "nebula store init failed";
+        LOG(ERROR) << "Nebula store init failed";
         return nullptr;
     }
+
+    LOG(INFO) << "Waiting for the leader elected...";
+    nebula::HostAddr leader;
+    while (true) {
+        auto ret = kvstore->partLeader(0, 0);
```
wadeliuyi: what mean about (0, 0)?

dangleptr: For meta,  we only have one space (0) and one part (0)

src/meta/client/MetaClient.cpp
```diff
@@ -999,27 +1008,26 @@ StatusOr<SchemaVer> MetaClient::getNewestEdgeVerFromCache(const GraphSpaceID& sp
 
 
 folly::Future<StatusOr<bool>> MetaClient::heartbeat() {
+    if (clusterId_.load() == 0) {
+        clusterId_ = ClusterIdMan::getClusterIdFromFile(FLAGS_cluster_id_path);
```
critical27: clusterId_.store()?

src/meta/client/MetaClient.cpp
```diff
@@ -999,27 +1008,26 @@ StatusOr<SchemaVer> MetaClient::getNewestEdgeVerFromCache(const GraphSpaceID& sp
...
-            && !clusterMan_->isClusterIdSet()) {
-            clusterMan_->setClusterId(resp.get_clusterId());
-            if (!clusterMan_->clusterIdDumped()) {
-                if (!clusterMan_->dumpClusterId()) {
-                    LOG(ERROR) << "meta client clusterId dump failed!";
-                    return false;
-                }
+        if (clusterId_.load() == 0) {
+            LOG(INFO) << "Persisit the cluster Id from metad " << resp.get_cluster_id();
+            if (ClusterIdMan::persistInFile(resp.get_cluster_id(), FLAGS_cluster_id_path)) {
```
critical27: Do we need return false when persist failed?

sherman-the-tank:
Again, I think we need more discussion on using clusterId to prevent unexpected hosts joining the cluster. I'll suggest to hold on a little so that we have enough time to think through all scenarios

Let's have a quick offline discussion

src/daemons/MetaDaemon.cpp
wadeliuyi: maybe we can replace this magic number to some readable word, that coder understand it more easy.

sherman-the-tank: Or make the previous reply as a comment in front of this line

src/common/base/Status.h
```diff
@@ -60,6 +60,10 @@ class Status final {
         return code() == rhs.code();
     }
 
+    bool operator!=(const Status &rhs) const {
```
sherman-the-tank: Why do we need this? My understanding is the compiler will convert ```!=``` to ```! ==()```

src/daemons/MetaDaemon.cpp
```diff
@@ -30,9 +30,9 @@ using nebula::Status;
 DEFINE_int32(port, 45500, "Meta daemon listening port");
 DEFINE_bool(reuse_port, true, "Whether to turn on the SO_REUSEPORT option");
 DEFINE_string(data_path, "", "Root data path");
-DEFINE_string(peers, "", "It is a list of IPs split by comma,"
-                         "the ips number equals replica number."
-                         "If empty, it means replica is 1");
+DEFINE_string(meta_server_addrs, "", "It is a list of IPs split by comma,"
+                                     "the ips number equals replica number."
```
sherman-the-tank: This line conflicts with the next sentence. If the number of IPs equals to the number of replicas, it means the IP list includes the local IP as well. So when the IP list is empty, the number of replica should be zero

src/daemons/MetaDaemon.cpp
sherman-the-tank: > For meta, we only have one space (0) and one part (0)

Put this as a comment in the code?

src/meta/ActiveHostsMan.cpp
```diff
@@ -6,150 +6,52 @@
...
-            updateHostInfo({host.ip, host.port}, info);
+        HostInfo info = HostInfo::decode(iter->val());
+        if (now - info.lastHBTimeInSec_ < expiredTTL) {
+            hosts.emplace_back(host.ip, host.port);
         }
         iter->next();
     }
-}
-
-void ActiveHostsMan::cleanExpiredHosts() {
```
sherman-the-tank: We still need this to remove the ghost hosts that are dead for too long, don't we?

src/daemons/MetaDaemon.cpp
dangleptr: Good suggestion.  I will fix it.

src/common/base/Status.h
dangleptr: No.  The compiler will not do that.

src/meta/ActiveHostsMan.cpp
dangleptr: Yes.  I am planing to do it in compaction.

src/meta/client/MetaClient.cpp
dangleptr: They are the same thing.

src/meta/client/MetaClient.cpp
dangleptr: Good point.

src/daemons/MetaDaemon.cpp
dangleptr: Good catch.

dangleptr:
#736

PR #749, opened 2019-08-09
Implement groupby

### LINKED ISSUE #676: Support limit and group by
Subtask of #492
- LIMIT syntax: 
LIMIT size/LIMIT skip, size
- GROUP BY syntax
  - syntax
  GROUP BY COUNT(propName), SUM(propName) ……
  - functions
 COUNT、COUNT DISTINCT、 SUM、 AVG、 MIN、MAX、STD
BIT_AND(), BIT_OR(), and BIT_XOR()

laura-ding:
Closes #676


src/parser/test/ParserTest.cpp
```diff
@@ -1098,4 +1098,39 @@ TEST(Parser, ConfigOperation) {
...
 }
 
+TEST(Parser, GroupBy) {
+    // All fun succeed
+    {
+        GQLParser parser;
+        std::string query = "GO FROM 1 OVER work "
+                            "YIELD $$.company.name, $^.person.name "
+                            "| GROUP BY $$.company.name "
+                            "YIELD $$.company.name as name, "
```
whitewum: use replace all `as` to AS
let's follow nGQL style

src/parser/test/ParserTest.cpp
```diff
@@ -1098,4 +1098,39 @@ TEST(Parser, ConfigOperation) {
...
+    // All fun error, empty group name
+    {
+        GQLParser parser;
+        std::string query = "GO FROM 1 OVER work "
+                            "YIELD $$.company.name, $^.person.name "
+                            "| GROUP BY "
+                            "YIELD $$.company.name as name, "
+                            "COUNT($^.person.name )";
+        auto result = parser.parse(query);
+        ASSERT_FALSE(result.ok());
```
whitewum: if it's syntax error, better to assert syntax error than just false

src/parser/test/ParserTest.cpp
```diff
@@ -1098,4 +1098,39 @@ TEST(Parser, ConfigOperation) {
...
+                            "MIN($^.person.name ), "
+                            "F_STD($^.person.name ), "
+                            "F_BIT_AND($^.person.name ), "
+                            "F_BIT_OR($^.person.name ), "
+                            "F_BIT_XOR($^.person.name )";
+
+        auto result = parser.parse(query);
+        ASSERT_TRUE(result.ok()) << result.status();
+    }
+    // All fun error, empty group name
```
whitewum: I think currently we don't support in the following order,

YIELD xxx, sum(YYY) GROUP BY xxx 

So let parser test make sure of it (in case any further questions).

src/graph/test/GroupByLimitTest.cpp
```diff
@@ -0,0 +1,237 @@
...
+        ASSERT_EQ(cpp2::ErrorCode::E_EXECUTION_ERROR, code);
+    }
+}
+
+TEST_F(GroupByLimitTest, GroupByLimitTest) {
+}
+
+TEST_F(GroupByLimitTest, GroupByOrderByLimitTest) {
+}
+}  // namespace graph
```
whitewum: so you will add some case including 
group by, order by, and limit
would you?

src/graph/test/GroupByLimitTest.cpp
```diff
@@ -0,0 +1,237 @@
...
+        cpp2::ExecutionResponse resp;
+        auto &player = players_["Paul Gasol"];
+        auto *fmt = "GO FROM %ld OVER like "
+                    "YIELD $$.player.age as age, "
+                    "like._dst as id "
+                    "| GROUP BY $-.id "
+                    "YIELD $-.id as id, "
+                    "SUM(age) as age |"
+                    "GO FROM $-.id OVER serve "
+                    "YIELD $$.team.name as name";
```
whitewum: YIELD $$.team.name AS name, $-.age AS sumAge to test propogation $- is correct.

src/graph/AggregateFunction.h
```diff
@@ -0,0 +1,409 @@
...
+    cpp2::ColumnValue getResult() override {
+        cpp2::ColumnValue result;
+        if (elems_.empty()) {
+            result.set_integer(0);
+            return result;
+        }
+        double acc = 0.0;
+        double avg = avg_.getResult().get_double_precision();
+        std::for_each(std::begin(elems_), std::end(elems_), [&](const double elem) {
+            acc += (elem - avg) *(elem - avg);
```
whitewum: good math
but space: * (

src/parser/test/ParserTest.cpp
laura-ding: This is the parser module, and the result is not ok, it means syntax error

src/graph/test/GroupByLimitTest.cpp
laura-ding: yeah

laura-ding:
jenkins go

src/graph/AggregateFunction.h
```diff
@@ -0,0 +1,451 @@
...
+            dSum = sum.get_double_precision();
+        }
+        sum.set_double_precision(dSum / count_);
+        return sum;
+    }
+
+private:
+    Sum                 sum_;
+    uint64_t            count_{0};
+    bool                has_{false};
```
liuyu85cn: if there is a count_ , is 'has_' really useful?

src/graph/GroupByExecutor.cpp
```diff
@@ -0,0 +1,435 @@
...
+        onResult_(std::move(outputs));
+    }
+
+    DCHECK(onFinish_);
+    onFinish_();
+}
+
+
+Status GroupByExecutor::groupingData() {
+    // key : group col vals, val: cal funptr
```
CPWstatic: Would you please make the annotation much more detailed and readable. We could use full word to explain the data structure here.

src/graph/YieldExecutor.cpp
```diff
@@ -266,32 +369,67 @@ Status YieldExecutor::executeConstant() {
...
+            cols.emplace_back(std::move(val));
+            i++;
+        }
+        row.set_columns(std::move(cols));
+        auto rsWriter = std::make_unique<RowSetWriter>(schema);
+        auto status = getAggResultWriter(std::move(row), rsWriter.get());
         if (!status.ok()) {
             return status;
         }
+        finishExecution(std::move(rsWriter));
```
CPWstatic: It seems we can put them into single block. And it's better to make it as a function.

PR #754, opened 2019-08-12
Fixed incorrect rate metric in StatsManager

dutor:
This bug seems due to improper support to customized clock by `folly::MultiLevelTimeSeries`


src/common/stats/StatsManager.inl
```diff
@@ -19,9 +19,9 @@ StatsManager::VT StatsManager::readValue(StatsHolder& stats,
         case StatsMethod::COUNT:
             return stats.count(level);
         case StatsMethod::AVG:
-            return stats.avg(level);
+            return stats.template avg<VT>(level);
```
dangleptr: Why the changes?

dutor: The default type argument here is `double` while `VT` is of type `int64_t`, thus introducing two meaningless type conversions.

PR #758, opened 2019-08-13
Replace DCHECK_NOTNULL with DCHECK in TraverseExecutor.

CPWstatic:
```DCHECK_NOTNULL``` not work well when compile release version.


CPWstatic:
Jenkins go!

PR #772, opened 2019-08-15
Improved nebula.service

### LINKED ISSUE #767: minor bug of nebula.service to start service
if one process crash while other two remains alive,
use nebula.service start all, the return is

image

The  alive process prevent graphd start

dutor:
This closes #767

### DIFF
--- a/scripts/nebula.service
+++ b/scripts/nebula.service
@@ -119,7 +119,8 @@ function start_daemon {
     pid_file=$(get_pid_file_from_config ${config})
     if [[ -f ${pid_file} ]]; then
         if is_process_running ${pid_file}; then
-            ERROR_AND_EXIT "Process already running: $(cat ${pid_file})"
+            ERROR "${daemon_name} already running: $(cat ${pid_file})"
+            return 1
         fi
     fi
 


scripts/nebula.service
```diff
@@ -119,7 +119,8 @@ function start_daemon {
     pid_file=$(get_pid_file_from_config ${config})
     if [[ -f ${pid_file} ]]; then
         if is_process_running ${pid_file}; then
-            ERROR_AND_EXIT "Process already running: $(cat ${pid_file})"
+            ERROR "${daemon_name} already running: $(cat ${pid_file})"
+            return 1
```
darionyaphet: Return not zero-value means found error.

dutor: Yes, that is what I want.

whitewum:
I suggest wait a while for restart service. otherwise fail to restart every time....
or even worse
image

whitewum:
Maybe it's not the script's problem, But I DO THINK sleep 1 in stop and start will prevend a lot starting and stoping errors (meta-refresh, storage-stop-time, etc).
So Please consider @dutor @sherman-the-tank @dangleptr

dutor:
> Maybe it's not the script's problem, But I DO THINK sleep 1 in stop and start will prevend a lot starting and stoping errors (meta-refresh, storage-stop-time, etc).
> So Please consider @dutor @sherman-the-tank @dangleptr

It does nothing with the scripts themselves, and we won't wait for anything for any time in vain.

FYI. While restarting, we actually wait for exit after stopping. Meanwhile, the waiting operation is not just a `sleep N`. And, if there were an event-driven solution to wait for exit, I won't use the blindly sleeping scheme.

PS. Whenever there is a bug or something, we shall try to address it in essence.

dutor:
I’m going to merge this PR. If there is anything worth discussing, please contact me offline.

PR #776, opened 2019-08-15
Update README.md

whitewum:
update docker address


whitewum:
Close since @jude-zhu also works on this doc

PR #792, opened 2019-08-16
Fix graph failed to register config to meta.

### LINKED ISSUE #780: [2 bugs] about show variables
1. the service has started for about an hour, but why `show variables graph` is empty
2. why `show variables graph` two times return different?  -- I can't reproduce the syntax error

image
image

grep don't return any `update config in cache `

image

critical27:
1. Let meta client retry to register configs to meta server.
2. Add `gflags.json` to avoid hard code config mode, and all default config mode is `IMMUTABLE`.
3. Add related command in complete.json

close #780

### DIFF
--- a/share/resources/gflags.json
+++ b/share/resources/gflags.json
@@ -41,8 +41,6 @@
         "helpshort",
         "alsologtoemail",
         "version",
-        "graphd_gflags_json",
-        "metad_gflags_json",
-        "storaged_gflags_json"
+        "gflags_mode_json"
     ]
 }
--- a/share/resources/gflags.json
+++ b/share/resources/gflags.json
@@ -32,6 +32,17 @@
         "stop_logging_if_full_disk",
         "symbolize_stacktrace",
         "vmodule",
-        "v"
+        "v",
+        "help",
+        "helpon",
+        "helpxml",
+        "helpfull",
+        "helpmatch",
+        "helpshort",
+        "alsologtoemail",
+        "version",
+        "graphd_gflags_json",
+        "metad_gflags_json",
+        "storaged_gflags_json"
     ]
 }
--- a/share/resources/gflags.json
+++ b/share/resources/gflags.json
@@ -1,0 +1,37 @@
+{
+    "IMMUTABLE": [
+    ],
+    "REBOOT": [
+    ],
+    "MUTABLE": [
+        "load_data_interval_secs"
+    ],
+    "IGNORED": [
+        "logging",
+        "flagfile",
+        "fromenv",
+        "tryfromenv",
+        "undefok",
+        "tab_completion_columns",
+        "tab_completion_word",
+        "alsologtostderr",
+        "drop_log_memory",
+        "log_backtrace_at",
+        "log_dir",
+        "log_link",
+        "log_prefix",
+        "logbuflevel",
+        "logbufsecs",
+        "logemaillevel",
+        "logfile_mode",
+        "logmailer",
+        "logtostderr",
+        "max_log_size",
+        "minloglevel",
+        "stderrthreshold",
+        "stop_logging_if_full_disk",
+        "symbolize_stacktrace",
+        "vmodule",
+        "v"
+    ]
+}
--- a/src/meta/GflagsManager.cpp
+++ b/src/meta/GflagsManager.cpp
@@ -34,9 +34,11 @@ std::unordered_map<std::string, cpp2::ConfigMode> GflagsManager::parseConfigJson
         LOG(ERROR) << "Load gflags json failed";
         return configModeMap;
     }
-    std::vector<std::string> keys = {"IMMUTABLE", "REBOOT", "MUTABLE", "IGNORED"};
-    std::vector<cpp2::ConfigMode> modes = {cpp2::ConfigMode::IMMUTABLE, cpp2::ConfigMode::REBOOT,
-                                           cpp2::ConfigMode::MUTABLE, cpp2::ConfigMode::IGNORED};
+    static std::vector<std::string> keys = {"IMMUTABLE", "REBOOT", "MUTABLE", "IGNORED"};
+    static std::vector<cpp2::ConfigMode> modes = {cpp2::ConfigMode::IMMUTABLE,
+                                                  cpp2::ConfigMode::REBOOT,
+                                                  cpp2::ConfigMode::MUTABLE,
+                                                  cpp2::ConfigMode::IGNORED};
     for (size_t i = 0; i < keys.size(); i++) {
         std::vector<std::string> values;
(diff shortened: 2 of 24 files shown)


src/meta/client/MetaClient.cpp
```diff
@@ -1055,6 +1056,9 @@ MetaClient::regConfig(const std::vector<cpp2::ConfigItem>& items) {
 
 folly::Future<StatusOr<std::vector<cpp2::ConfigItem>>>
 MetaClient::getConfig(const cpp2::ConfigModule& module, const std::string& name) {
+    if (!configReady_) {
+        registerCfg();
```
laura-ding: getConfig() used in graphd's executor, and registerCfg() is a sync call, in graphd executor should not exist sync call. Can we do` if configReady_ is not true, direct return error, like the schema?`

src/meta/GflagsManager.cpp
```diff
@@ -23,19 +25,59 @@ std::string GflagsManager::gflagsValueToThriftValue<std::string>(
...
 
-void GflagsManager::declareGflags() {
-    // declare all gflags in ClientBasedGflagsManager
+std::unordered_map<std::string, cpp2::ConfigMode> GflagsManager::parseConfigJson() {
+    auto path = fs::FileUtils::readLink("/proc/self/exe").value();
+    auto json = fs::FileUtils::dirname(path.c_str()) + "/../share/resources/gflags.json";
+    std::unordered_map<std::string, cpp2::ConfigMode> configModeMap;
+    Configuration conf;
+    if (!conf.parseFromFile(json).ok()) {
+        return configModeMap;
```
laura-ding: add error log

src/meta/client/MetaClient.cpp
critical27: Yeah, it do block graphd. The reason I add this logic is that if graphd register failed during initialization, that would take another `load_data_interval_secs`, during that time users can't show/get/update, do you think it's ok?

laura-ding: yes

critical27:
Jenkins go

src/meta/GflagsManager.cpp
```diff
@@ -23,19 +25,60 @@ std::string GflagsManager::gflagsValueToThriftValue<std::string>(
...
+std::unordered_map<std::string, cpp2::ConfigMode> GflagsManager::parseConfigJson() {
+    auto path = fs::FileUtils::readLink("/proc/self/exe").value();
+    auto json = fs::FileUtils::dirname(path.c_str()) + "/../share/resources/gflags.json";
+    std::unordered_map<std::string, cpp2::ConfigMode> configModeMap;
+    Configuration conf;
+    if (!conf.parseFromFile(json).ok()) {
+        LOG(ERROR) << "Load gflags json failed";
+        return configModeMap;
+    }
+    std::vector<std::string> keys = {"IMMUTABLE", "REBOOT", "MUTABLE", "IGNORED"};
```
wadeliuyi: static local variable?

src/meta/GflagsManager.cpp
```diff
@@ -23,19 +25,60 @@ std::string GflagsManager::gflagsValueToThriftValue<std::string>(
...
-        cpp2::ConfigMode mode = cpp2::ConfigMode::REBOOT;
         VariantType value;
         std::string valueStr;
 
+        // default config type would be immutable
+        cpp2::ConfigMode mode = cpp2::ConfigMode::IMMUTABLE;
+        if (configModeMap.find(name) != configModeMap.end()) {
+            mode = configModeMap[name];
+        }
+        // ignore some useless gflags
```
wadeliuyi: find and [] will traverse the map twice.

auto iter = configModeMap.find(name);
if (iter != configModeMap.end()) {
    mode = iter->second;
}

src/meta/GflagsManager.cpp
```diff
@@ -23,19 +25,60 @@ std::string GflagsManager::gflagsValueToThriftValue<std::string>(
...
+        if (mode == cpp2::ConfigMode::IGNORED) {
```
wadeliuyi: mode default is cpp2::ConfigMode::IMMUTABLE, if configModeMap.find(name) != configModeMap.end() is false, this condition never true, is it right to go the next logic?

src/meta/GflagsManager.cpp
critical27: nice catch.

src/meta/GflagsManager.cpp
critical27: good suggestion

src/meta/GflagsManager.cpp
critical27: The configModeMap is load from json, if found in json, replace the default mode. So for those configs we want to ignore, the mode will set to  cpp2::ConfigMode::IGNORED, and we can just ignore those gflags.

share/resources/gflags.json
```diff
@@ -0,0 +1,37 @@
+{
```
laura-ding: add gflags.json into share/resources/CMakeLists.txt

critical27:
Jenkins go

src/meta/GflagsManager.cpp
```diff
@@ -23,19 +25,63 @@ std::string GflagsManager::gflagsValueToThriftValue<std::string>(
...
+            continue;
+        }
+        if (module == cpp2::ConfigModule::META) {
+            // all config of meta is immutable for now
+            mode = cpp2::ConfigMode::IMMUTABLE;
+        }
+
         // TODO: all int32 and uint32 are converted to int64
         if (type == "uint32" || type == "int32" || type == "int64") {
```
dangleptr: Add uint64

share/resources/gflags.json
dangleptr: That's great.  One question,  how to distinguish different module,  storage/graph/meta  has different gflags.

src/meta/GflagsManager.cpp
critical27: Nice catch.

share/resources/gflags.json
critical27: For now it seems we can apply theres to all storage/graph/meta, if we do need to distinguish modules of a gflag, I will add another layer to indicate modules.

dangleptr: Yes, we need it.   Maybe we will start graph/meta/storage on the same server

share/resources/gflags.json
```diff
@@ -0,0 +1,37 @@
...
+    "IMMUTABLE": [
```
dangleptr: Could we start all services on one host with just one gflags.json?

critical27: Yeah, all services will try to read this file.

Issue #712, opened 2019-08-01
crash when using "FIND" command


jude-zhu:
![1](https://user-images.githubusercontent.com/51590253/62279115-08514900-b47c-11e9-9925-9585ced3d45a.jpg)

monadbobo:
The find sentence has not been implemented yet, so I think we need to return to the user an unimplemented error.

sherman-the-tank:
Yes, we should never crash

laura-ding:
Has been solved

PR #794, opened 2019-08-16
Return error msg for find and match sentence

### LINKED ISSUE #777: graphd crash at again
dup with #712

laura-ding:
fix #777  and #712


src/graph/test/DataTest.cpp
```diff
@@ -359,6 +359,23 @@ TEST_F(DataTest, InsertVertex) {
     // TODO: Test insert multi tags, and delete one of them then check other existent
 }
 
+TEST_F(DataTest, FindTest) {
+    {
```
darionyaphet: why don't support some `SUCCEEDED` example ?

laura-ding: Because they have not to be implemented

PR #795, opened 2019-08-16
Implement snapshot logic

dangleptr:
subtask of #182


src/kvstore/raftex/RaftPart.cpp
```diff
@@ -1313,6 +1334,50 @@ cpp2::ErrorCode RaftPart::verifyLeader(
...
+        return;
+    }
+    if (status_ != Status::WAITING_SNAPSHOT) {
+        LOG(INFO) << idStr_ << "Begin to receive the snapshot";
+        status_ = Status::WAITING_SNAPSHOT;
+        wal_->reset();
+    }
+    // TODO(heng): Maybe we should save them into one sst firstly?
+    // And before commit the snapshot, we should clean up all data.
+    auto ret = commitSnapshot(req.get_rows(),
```
monadbobo: Before installing snapshot, We also need to check the following  three  conditions:

1. Whether it is stale term(req.term and current_term).
2. Whether more than one leader (req.leaderId and leader_id).
3. If the current node's leader is not set at this time, then we need to set the leader.

dangleptr: Actually,  i have considered this problem.  For snapshot,   we don't care about it is from leader or follower,  because the snapshot is the same.

dangleptr:
Jenkins go

dangleptr:
Jenkins go

src/kvstore/raftex/SnapshotManager.cpp
```diff
@@ -0,0 +1,116 @@
...
+
+folly::Future<Status> SnapshotManager::sendSnapshot(std::shared_ptr<RaftPart> part,
+                                                    const HostAddr& dst) {
+    folly::Promise<Status> p;
+    auto fut = p.getFuture();
+    executor_->add([this, p = std::move(p), part, dst] () mutable {
+        auto spaceId = part->spaceId_;
+        auto partId = part->partId_;
+        auto termId = part->term_;
+        auto commitLogIdAndTerm = part->lastCommittedLogId();
```
critical27: I think `commitLogIdAndTerm` may differs from the log term and id in `appendLogs` when we need to send snapshot, which is `committedLogId_` and `logTermToSend_`. Do you think there may be potential data lose?

src/kvstore/LogEncoder.cpp
```diff
@@ -13,6 +13,26 @@ namespace kvstore {
...
+std::string encodeKV(const folly::StringPiece& key,
+                     const folly::StringPiece& val) {
+    uint32_t ksize = key.size();
+    uint32_t vsize = val.size();
+    std::string str;
+    str.reserve(sizeof(uint32_t) * 2 + ksize + vsize);
+    str.append(reinterpret_cast<const char*>(&ksize), sizeof(ksize));
+    str.append(reinterpret_cast<const char*>(&vsize), sizeof(vsize));
+    str.append(key.data(), ksize);
+    str.append(key.data(), vsize);
```
wadeliuyi: is val.data()?

src/kvstore/Part.cpp
```diff
@@ -40,7 +40,8 @@ Part::Part(GraphSpaceID spaceId,
            std::shared_ptr<folly::IOThreadPoolExecutor> ioPool,
            std::shared_ptr<thread::GenericThreadPool> workers,
            wal::BufferFlusher* flusher,
-           std::shared_ptr<folly::Executor> handlers)
+           std::shared_ptr<folly::Executor> handlers,
+           std::shared_ptr<raftex::SnapshotManager> snapshotMan)
```
wadeliuyi: maybe the parameter is too many, is there some way to compact it?

src/kvstore/Part.cpp
```diff
@@ -253,6 +255,30 @@ bool Part::commitLogs(std::unique_ptr<LogIterator> iter) {
...
+                                                 TermID committedLogTerm,
+                                                 bool finished) {
+    auto batch = engine_->startBatchWrite();
+    int64_t count = 0;
+    int64_t size = 0;
+    for (auto& row : rows) {
+        count++;
+        size += row.size();
+        auto kv = decodeKV(row);
+        CHECK_EQ(ResultCode::SUCCEEDED, batch->put(kv.first, kv.second));
```
wadeliuyi: maybe need change use too many micro CHECK, it is not safe crash directly.

src/kvstore/Part.cpp
```diff
@@ -253,6 +255,30 @@ bool Part::commitLogs(std::unique_ptr<LogIterator> iter) {
...
+    }
+    if (finished) {
+        std::string commitMsg;
+        commitMsg.reserve(sizeof(LogID) + sizeof(TermID));
+        commitMsg.append(reinterpret_cast<char*>(&committedLogId), sizeof(LogID));
+        commitMsg.append(reinterpret_cast<char*>(&committedLogTerm), sizeof(TermID));
+        batch->put(folly::stringPrintf("%s%d", kCommitKeyPrefix, partId_), commitMsg);
+    }
```
wadeliuyi: define a function to do this?

src/kvstore/Part.cpp
darionyaphet: yep. We are replace it with check status :)

src/kvstore/RocksEngine.cpp
```diff
@@ -32,7 +32,7 @@ class RocksWriteBatch : public WriteBatch {
     rocksdb::DB* db_{nullptr};
 
 public:
-    explicit RocksWriteBatch(rocksdb::DB* db) : db_(db) {}
+    explicit RocksWriteBatch(rocksdb::DB* db) : batch_(4096), db_(db) {}
```
wadeliuyi: why is 4096?

src/kvstore/raftex/Host.cpp
```diff
@@ -106,37 +106,22 @@ folly::Future<cpp2::AppendLogResponse> Host::appendLogs(
...
-            }
-        } else {
-            // Otherwise, logId has to be greater
-            if (logId < logIdToSend_) {
-                LOG(INFO) << idStr_ << "The log has been sended";
-                cpp2::AppendLogResponse r;
-                r.set_error_code(cpp2::ErrorCode::SUCCEEDED);
-                return r;
-            }
+        if (logId <= lastLogIdSent_) {
```
wadeliuyi: a niubility change.
but I have a question, which case will cause logId <= lastLogIdSent_? and if appear the case logId == lastLogIdSent_, we return success directly, and this will cause collectNSucceeded() return success directly in "RaftPart::replicateLogs", and the current thread will commit the log,   but in the other thread do really thing that replica log to other part by network and but failed, then the log just exist in leader?

src/kvstore/raftex/RaftPart.cpp
```diff
@@ -1148,22 +1175,52 @@ void RaftPart::processAppendLogRequest(
...
+            lastLogId_ = wal_->lastLogId();
+            lastLogTerm_ = wal_->lastLogTerm();
+            resp.set_last_log_id(lastLogId_);
+            resp.set_last_log_term(lastLogTerm_);
+            resp.set_error_code(cpp2::ErrorCode::SUCCEEDED);
+        } else {
+            LOG(ERROR) << idStr_ << "Failed to append logs to WAL";
+            resp.set_error_code(cpp2::ErrorCode::E_WAL_FAIL);
+        }
+        return;
```
wadeliuyi: current node is waiting snapshot, why not return cpp2::ErrorCode::E_WAITING_SNAPSHOT?

src/kvstore/raftex/RaftPart.cpp
```diff
@@ -1148,22 +1175,52 @@ void RaftPart::processAppendLogRequest(
...
+        VLOG(2) << idStr_ << "Writing log [" << firstId
+                << ", " << firstId + numLogs - 1 << "] to WAL";
+        LogStrListIterator iter(firstId,
+                                req.get_log_term(),
+                                req.get_log_str_list());
+        if (wal_->appendLogs(iter)) {
+            CHECK_EQ(firstId + numLogs - 1, wal_->lastLogId());
+            lastLogId_ = wal_->lastLogId();
+            lastLogTerm_ = wal_->lastLogTerm();
+            resp.set_last_log_id(lastLogId_);
```
wadeliuyi: why can we append log to WAL and update the lastLogId_ and lastLogTerm_ if a node is WAITING_SNAPSHOT? if some record is failed in processSendSnapshotRequest, it will exist a hole of log entry in this node.

src/kvstore/raftex/RaftPart.cpp
```diff
@@ -1148,22 +1175,52 @@ void RaftPart::processAppendLogRequest(
...
+            resp.set_last_log_term(lastLogTerm_);
+            resp.set_error_code(cpp2::ErrorCode::SUCCEEDED);
```
wadeliuyi: if return success but processSendSnapshotRequest failed, the leader will believe that majority part save the new log, but this node can not commit the log because there is a log hole in this part.
if we have 3 node, one is leader, one is wait snapshot and return success, the other is follower, but leader replica log to follow failed, the leader will commit the log, but the node wait snapshot can not commit the log, and if lead is down, the cluster will lost this log.

yixinglu:
close #815

src/kvstore/raftex/SnapshotManager.cpp
```diff
@@ -0,0 +1,116 @@
...
+                              commitLogIdAndTerm.second,
+                              localhost,
+                              std::move(data),
+                              totalSize,
+                              totalCount,
+                              dst,
+                              finished);
+                // TODO(heng): we send request one by one to avoid too large memory occupied.
+                try {
+                    auto resp  = std::move(f).get();
```
wadeliuyi: block the executor?

src/kvstore/raftex/SnapshotManager.cpp
```diff
@@ -0,0 +1,116 @@
...
+DEFINE_int32(snapshot_io_threads, 4, "Threads number for snapshot");
+DEFINE_int32(snapshot_send_retry_times, 3, "Retry times if send failed");
+DEFINE_int32(snapshot_send_timeout_ms, 60000, "Rpc timeout for sending snapshot");
+
+namespace nebula {
+namespace raftex {
+
+SnapshotManager::SnapshotManager() {
+    executor_.reset(new folly::IOThreadPoolExecutor(FLAGS_snapshot_worker_threads));
+    ioThreadPool_.reset(new folly::IOThreadPoolExecutor(FLAGS_snapshot_io_threads));
```
wadeliuyi: why use private thread pool? 
maybe executor is threadmanager is better, because it is not process the network io.

src/kvstore/raftex/SnapshotManager.cpp
wadeliuyi: I think it is ok, because the log id is continuous even is commitLogIdAndTerm is different with real term and id.

src/kvstore/LogEncoder.cpp
dangleptr: Good catch.    Let me add a UT for it.

src/kvstore/Part.cpp
dangleptr: Good suggestion!  After balance completed,  i will refactor the code here.

src/kvstore/Part.cpp
dangleptr: Good point.

src/kvstore/Part.cpp
dangleptr: Good point

src/kvstore/RocksEngine.cpp
dangleptr: Just  experience dependence,  maybe use a flag here is better.

src/kvstore/raftex/Host.cpp
dangleptr: > which case will cause logId <= lastLogIdSent_?

Currently,  quorum policy means there are some hosts requests maybe not in the order what we want.  Especially, for learner,  we can't ensure the order.  

> but in the other thread do really thing that replica log to other part by network and but failed, then the log just exist in leader?

Not exactly,  lastLogIdSent_ is used to record the last log id the leader send to some follower.  So if logId <= lastLogIdSent_,  it means the log has been received by the follower.

src/kvstore/raftex/RaftPart.cpp
dangleptr: >  if some record is failed in processSendSnapshotRequest, it will exist a hole of log entry in this node.

Not exactly,  if some snapshot request failed,   the leader will mark this snapshot send failed,  and after some while,  the follower will reset the status and clean up all WALs.  Then,  the leader could send snapshot again.

src/kvstore/raftex/RaftPart.cpp
dangleptr: You are right,  it has problem under this situation,  but not the case you described. 

Actually,  the log is still exists on two hosts (one is on leader,  the other is on the follower receiving snapshot), it is not lost.   After the leader down,   no leader will be elected,  because one follower misses log and the other misses the snapshot.   I will handle this case in the future.  Currently, we could add a TODO comment here.

src/kvstore/raftex/SnapshotManager.cpp
dangleptr: Because the threads in executor will be blocked during sending snapshot.

src/kvstore/raftex/SnapshotManager.cpp
dangleptr: @critical27 You are right,   maybe the committedLogId is less than the real one in the snapshot.   It will not loss the data,  but maybe some record will be committed twice.   Let me add a TODO comment here firstly.

src/kvstore/raftex/Host.cpp
wadeliuyi: But in the code and raft logic, the logid is Monotonically increasing, and the lastLogIdSent_ is always set by the prevLogId which is less than logId, even if the host is a learner. so is it some thing wrong when in case " logId <= lastLogIdSent_"

src/kvstore/raftex/Host.cpp
```diff
@@ -300,6 +291,47 @@ void Host::appendLogsInternal(folly::EventBase* eb,
...
+                VLOG(2) << self->idStr_ << "Log stale, reset lastLogIdSent " << self->lastLogIdSent_
+                        << " to the followers lastLodId" << resp.get_last_log_id();
+                {
+                    std::lock_guard<std::mutex> g(self->lock_);
+                    self->lastLogIdSent_ = resp.get_last_log_id();
+                    self->lastLogTermSent_ = resp.get_last_log_term();
+                    self->setResponse(resp);
+                }
+                self->noMoreRequestCV_.notify_all();
+                return resp;
```
wadeliuyi: maybe the lambda no need return a response object.

src/kvstore/raftex/RaftPart.cpp
```diff
@@ -977,10 +986,15 @@ void RaftPart::statusPolling() {
         VLOG(2) << idStr_ << "Need to send heartbeat";
         sendHeartbeat();
     }
-    wal_->cleanWAL();
+    if (needToCleanupSnapshot()) {
```
wadeliuyi: maybe else if is better

src/kvstore/raftex/RaftPart.cpp
wadeliuyi: you are right, I miss the logic.

src/kvstore/raftex/RaftPart.cpp
wadeliuyi: the function "processSendSnapshotRequest " is execute in other thread and failed, and after some time this node will clear the WAL, then the log lost?

src/kvstore/raftex/SnapshotManager.cpp
wadeliuyi: maybe can define a thread manager here which owned by SnapshotManager

src/kvstore/raftex/SnapshotManager.cpp
wadeliuyi: I have understand the problem

src/kvstore/raftex/Host.cpp
dangleptr: > lastLogIdSent_ is always set by the prevLogId which is less than logId

Not exactly,  the lastLogIdSent_ only be updated when received response from follower.  So the logId leader want to send maybe less than lastLogIdSent_

src/kvstore/raftex/RaftPart.cpp
dangleptr: Yes,  but it is a long time (currently, it is 5 minutes).   We should figure out a way to make sure the leader will be elected in this time.

src/kvstore/raftex/SnapshotManager.cpp
dangleptr: ThreadManager is not a good thread pool implementation.  I think IOThreadPool is better, because it holds a event loop inside.

src/kvstore/raftex/RaftPart.cpp
dangleptr: We can't use 'else if'.   Because as follower,  check leader election and check snapshot  are two actions independently.

src/kvstore/raftex/Host.cpp
dangleptr: Yes,  you are right.  Let me remove all unnecessary response in the callback.  
Good point.

src/kvstore/raftex/Host.cpp
wadeliuyi: I see it update immediate in Host::appendLogs before call appendLogsInternal.

src/kvstore/raftex/RaftPart.cpp
wadeliuyi: and I see the code that in function "RaftPart::needToSendHeartbeat" and "RaftPart::needToStartElection" it check condition "status_ == Status::RUNNING", it is different status with "status_ == Status::WAITING_SNAPSHOT", and this function never change the status, so I think need use else if.

and more, the state machine of raft is too complex now, is there some method to make it easy?

src/kvstore/raftex/RaftPart.cpp
```diff
@@ -1368,6 +1427,54 @@ cpp2::ErrorCode RaftPart::verifyLeader(
...
+        resp.set_error_code(cpp2::ErrorCode::E_NOT_READY);
+        return;
+    }
+    if (status_ != Status::WAITING_SNAPSHOT) {
+        LOG(INFO) << idStr_ << "Begin to receive the snapshot";
+        wal_->reset();
+        cleanup();
+        lastLogId_ = committedLogId_ = 0;
+        lastTotalCount_ = 0;
+        lastTotalSize_ = 0;
```
wadeliuyi: code follow exist three place, do it in a inner function?
wal_->reset();
        cleanup();
        lastLogId_ = committedLogId_ = 0;
        lastTotalCount_ = 0;
        lastTotalSize_ = 0;

src/kvstore/raftex/Host.cpp
dangleptr: The logic has been updated in this pr

src/kvstore/raftex/RaftPart.cpp
dangleptr: Yes.  Currently, the logic is the same with using "else if".  But they are two independent actions.

src/kvstore/raftex/RaftPart.cpp
dangleptr: Good point.

src/kvstore/raftex/Host.cpp
```diff
@@ -270,35 +259,74 @@ void Host::appendLogsInternal(folly::EventBase* eb,
...
+                    }
+                }
+                if (newReq) {
+                    self->appendLogsInternal(eb, newReq);
+                } else {
+                    self->noMoreRequestCV_.notify_all();
+                }
+                return;
+            }
+            case cpp2::ErrorCode::E_LOG_STALE: {
```
critical27: 👍

src/kvstore/raftex/RaftPart.cpp
```diff
@@ -1122,8 +1149,13 @@ void RaftPart::processAskForVoteRequest(
 
     // If the partition used to be a leader, need to fire the callback
     if (oldRole == Role::LEADER) {
-        // Need to invoke the onLostLeadership callback
         LOG(INFO) << idStr_ << "Was a leader, need to do some clean-up";
+        if (wal_->lastLogId() > lastLogId_) {
```

src/kvstore/raftex/RaftPart.cpp
```diff
@@ -1029,10 +1038,15 @@ void RaftPart::statusPolling() {
         VLOG(2) << idStr_ << "Need to send heartbeat";
         sendHeartbeat();
     }
-    wal_->cleanWAL();
+    if (needToCleanupSnapshot()) {
+        LOG(INFO) << idStr_ << "Clean up the snapshot";
+        cleanupSnapshot();
+    }
+    wal_->cleanWAL(FLAGS_wal_ttl);
```
critical27: :thumb:

dangleptr:
Jenkins go

src/kvstore/raftex/RaftPart.cpp
dangleptr: Not the same,  but i still add some checks here to make the logic more simple

src/kvstore/Part.cpp
```diff
@@ -254,24 +256,60 @@ bool Part::commitLogs(std::unique_ptr<LogIterator> iter) {
...
     }
-
     return engine_->commitBatchWrite(std::move(batch)) == ResultCode::SUCCEEDED;
 }
 
+std::pair<int64_t, int64_t> Part::commitSnapshot(const std::vector<std::string>& rows,
+                                                 LogID committedLogId,
+                                                 TermID committedLogTerm,
+                                                 bool finished) {
+    auto batch = engine_->startBatchWrite();
```
monadbobo: when previouse snapshot is under installing, we should check if this is the same snapshot and resume it, otherwise drop previous snapshot.

As described in the paper:

> * If lastIndex is larger than lastest snapshot's, save snapshot file and Raft state(lastIndex, lastTerm, lastConfig). discard any existing or partial snapshot.
> * If existing log entry has same index and term as lastIndex and lastTerm,discard log up through lastIndex(but retain any following entries) and reply.
>

src/kvstore/raftex/RaftPart.cpp
```diff
@@ -1368,6 +1431,68 @@ cpp2::ErrorCode RaftPart::verifyLeader(
...
+        resp.set_error_code(cpp2::ErrorCode::E_NOT_READY);
+        return;
+    }
+    if (UNLIKELY(role_ != Role::FOLLOWER && role_ != Role::LEARNER)) {
+        LOG(ERROR) << idStr_ << "Bad role " << roleStr(role_);
+        resp.set_error_code(cpp2::ErrorCode::E_BAD_STATE);
+        return;
+    }
+    if (UNLIKELY(leader_ != HostAddr(req.get_leader_ip(), req.get_leader_port())
+            || term_ != req.get_term())) {
```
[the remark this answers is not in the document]
dangleptr: To be simple,  i use a more strict condition check here.

src/kvstore/raftex/RaftPart.cpp
```diff
@@ -1368,6 +1431,68 @@ cpp2::ErrorCode RaftPart::verifyLeader(
...
+                  << "The part has been stopped, skip the request";
+        resp.set_error_code(cpp2::ErrorCode::E_BAD_STATE);
+        return;
+    }
+    if (UNLIKELY(status_ == Status::STARTING)) {
+        LOG(ERROR) << idStr_ << "The partition is still starting";
+        resp.set_error_code(cpp2::ErrorCode::E_NOT_READY);
+        return;
+    }
+    if (UNLIKELY(role_ != Role::FOLLOWER && role_ != Role::LEARNER)) {
```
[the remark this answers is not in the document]
dangleptr: For CANDIDATE,  just return ERROR.

src/kvstore/Part.cpp
dangleptr: Currently,  if the leader changed when sending snapshot,  the new leader's snapshot request will be rejected due to lastTotalCount and lastTotalSize not matched.  And after some while, the follower will clean up all data and new leader will send snapshot again.

src/kvstore/raftex/RaftPart.cpp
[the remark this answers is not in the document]
monadbobo: For CANDIDATE, I think we should change the role to FOLLOWER.

src/kvstore/raftex/SnapshotManager.h
```diff
@@ -0,0 +1,65 @@
...
+#include "thrift/ThriftClientManager.h"
+#include <folly/futures/Future.h>
+#include <folly/Function.h>
+#include <folly/executors/IOThreadPoolExecutor.h>
+
+namespace nebula {
+namespace raftex {
+
+using SnapshotCallback = folly::Function<void(std::vector<std::string>&& rows,
+                                              int64_t totalCount,
```
darionyaphet: why we need total count and total size

src/kvstore/raftex/RaftPart.cpp
```diff
@@ -1042,6 +1056,19 @@ void RaftPart::statusPolling() {
...
+    return status_ == Status::WAITING_SNAPSHOT &&
+           role_ != Role::LEADER &&
+           lastSnapshotRecvDur_.elapsedInSec() >= FLAGS_raft_snapshot_timeout;
+}
+
+void RaftPart::cleanupSnapshot() {
+    LOG(INFO) << idStr_ << "Clean up the snapshot";
+    std::lock_guard<std::mutex> g(raftLock_);
+    reset();
+    status_ = Status::RUNNING;
```
darionyaphet: STOPPED ?

src/kvstore/raftex/RaftPart.cpp
```diff
@@ -1368,6 +1431,68 @@ cpp2::ErrorCode RaftPart::verifyLeader(
...
+                                          cpp2::SendSnapshotResponse& resp) {
+    VLOG(1) << idStr_ << "Receive snapshot, total rows " << req.get_rows().size()
+            << ", total count received " << req.get_total_count()
+            << ", total size received " << req.get_total_size()
+            << ", finished " << req.get_done();
+    std::lock_guard<std::mutex> g(raftLock_);
+    // Check status
+    if (UNLIKELY(status_ == Status::STOPPED)) {
+        LOG(ERROR) << idStr_
+                  << "The part has been stopped, skip the request";
```
darionyaphet: indent

src/kvstore/raftex/RaftPart.cpp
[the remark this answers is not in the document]
dangleptr: We don't need to do that because leader still send heartbeat when sending snapshot.

src/kvstore/raftex/RaftPart.cpp
dangleptr: No,   the logic here is to clean up snapshot state,  and reset status to be running from wait_snapshot

src/kvstore/raftex/RaftPart.cpp
dangleptr: Good catch.  I will fix it in #881

src/kvstore/raftex/Host.cpp
```diff
@@ -353,8 +381,25 @@ Host::prepareAppendLogRequest() const {
...
+        req->set_sending_snapshot(true);
+        if (!sendingSnapshot_) {
+            LOG(INFO) << idStr_ << "Can't find log " << lastLogIdSent_ + 1
+                      << " in wal, send the snapshot";
+            sendingSnapshot_ = true;
+            part_->snapshot_->sendSnapshot(part_, addr_).then([this] (Status&& status) {
+                if (status.ok()) {
+                    LOG(INFO) << idStr_ << "Send snapshot succeeded!";
+                } else {
+                    LOG(INFO) << idStr_ << "Send snapshot failed!";
```
darionyaphet: ERROR is better ？

src/kvstore/raftex/Host.cpp
```diff
@@ -353,8 +381,25 @@ Host::prepareAppendLogRequest() const {
...
+                    // TODO(heng): we should tell the follower i am failed.
+                }
+                sendingSnapshot_ = false;
+            });
+        } else {
+            LOG(INFO) << idStr_ << "The snapshot req is in queue, please wait for a moment";
```
darionyaphet: Currently, it's not queued but discarded.

src/meta/test/BalanceIntegrationTest.cpp
```diff
@@ -93,15 +87,15 @@ TEST(BalanceIntegrationTest, DISABLED_LeaderBalanceTest) {
         serverContexts.emplace_back(std::move(sc));
     }
 
-    ret = mClient->createSpace("storage", partition, replica).get();
+    auto ret = mClient->createSpace("storage", partition, replica).get();
     ASSERT_TRUE(ret.ok());
-    sleep(FLAGS_load_data_interval_secs + 1);
-    sleep(FLAGS_raft_heartbeat_interval_secs);
+    sleep(FLAGS_load_data_interval_secs + FLAGS_raft_heartbeat_interval_secs + 3);
```
darionyaphet: FLAGS_load_data_interval_secs + FLAGS_raft_heartbeat_interval_secs + 1 ?

src/kvstore/raftex/Host.cpp
dangleptr: The task has been putted into event loop inside IOThread in SnapshotManager

src/meta/test/BalanceIntegrationTest.cpp
dangleptr: +3 is due to the UT occasionally failed.

src/kvstore/raftex/Host.cpp
dangleptr: Sounds good. Fix it in next pr

darionyaphet:
LGTM & thanks for you explain.

PR #800, opened 2019-08-19
Extend CAS in Raft

dangleptr:
1.  Support more atomic operations not only cas
2.  We could pass the atomic operator into raft,  and it will be invoked atomically.


src/kvstore/raftex/RaftPart.cpp
```diff
@@ -737,14 +739,14 @@ void RaftPart::processAppendLogResponses(
                     firstLogId,
                     currTerm,
                     std::move(logs_),
-                    [this] (const std::string& log) -> std::string {
-                        auto casRet = compareAndSet(log);
-                        if (casRet.empty()) {
+                    [this] (AtomicOp op) -> std::string {
+                        auto opRet = op();
```
critical27: According to my understanding, in AtomicOp, user just define whatever they want to do atomically, and return the result as a string.
Why don't we just pass the `AtomicOp` instead of `folly::Function<std::string(AtomicOp op)> opCB`, it seems a little bit weird.

dangleptr: Because we will set promise value,  if we implement the logic directly in Iterator,  you have to passed the part into it.  This is not what we want.

PR #806, opened 2019-08-20
Do self-checking in ProcessUtils::isPidAvailable

dutor:
When launched as a daemon, there are chances that the pid held by the pid file happens to equal to the pid of the current process.


yixinglu:
LGTM

Issue #796, opened 2019-08-17
Export the statistics of RocksDB


monadbobo:
Export the [statistics](https://github.com/facebook/rocksdb/wiki/Statistics) of RocksDB.

It's output is same with mysql's SHOW ENGINE "innodb/rocksdb" STATUS directive, so we also need to provide syntax for nGQL.

monadbobo:
I think we should provide such a grammar:

`SHOW ENGINE STATUS ip/servername/all`

@sherman-the-tank @dutor @laura-ding @CPWstatic @dangleptr @jude-zhu

jude-zhu:
https://github.com/vesoft-inc/nebula/pull/2262

PR #834, opened 2019-08-22
Issue #796 export the statistics of RocksDB.

monadbobo:
Summary:

1. Provide new syntax to export the statistics for all storage nodes.
     show engine status;
2. It will output five columns(Ip, Port, SpaceId, Path, EngineStatus).
3. Added two configuration options(rocksdb_stats_level and rocksdb_stats_dump_period_sec).

### DIFF
--- a/src/graph/ShowExecutor.cpp
+++ b/src/graph/ShowExecutor.cpp
@@ -6,6 +6,7 @@
 
 #include "graph/ShowExecutor.h"
 #include "network/NetworkUtils.h"
+#include "storage/client/StorageClient.h"
 
 namespace nebula {
 namespace graph {
@@ -63,6 +64,9 @@ void ShowExecutor::execute() {
         case ShowSentence::ShowType::kShowCreateEdge:
             showCreateEdge();
             break;
+        case ShowSentence::ShowType::kShowEngineStatus:
+            showEngineStatus();
+            break;
         case ShowSentence::ShowType::kUnknown:
             onError_(Status::Error("Type unknown"));
             break;
@@ -112,7 +116,6 @@ void ShowExecutor::showHosts() {
     };
     std::move(future).via(runner).thenValue(cb).thenError(error);
 }
-
 
 void ShowExecutor::showSpaces() {
     auto future = ectx()->getMetaClient()->listSpaces();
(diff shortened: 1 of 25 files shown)


src/kvstore/RocksEngineConfig.cpp
```diff
@@ -42,6 +42,22 @@ DEFINE_int32(rocksdb_batch_size,
...
+
+    VLOG(3) << "Invalid value for --" << flagname << ": " << value;
+    return false;
+}
+
+DEFINE_int32(rocksdb_stats_level, 0,
+             "Statistics Level for RocksDB. Default is 0 (kExceptHistogramOrTimers)");
+DEFINE_validator(rocksdb_stats_level, &validateRocksdbStatsLevel);
+
+DEFINE_int32(rocksdb_stats_dump_period_sec, 0, "DBOptions::stats_dump_period_sec for RocksDB");
```
critical27: I remember this option only control the interval for dumping stats into the LOG file, and if we want to get the stats in string, need to call `db->GetProperty()` to get the string version.

src/graph/ShowExecutor.cpp
```diff
@@ -456,6 +459,63 @@ void ShowExecutor::showCreateEdge() {
...
 
+void ShowExecutor::showEngineStatus() {
+    auto future  = ectx()->getMetaClient()->listHosts();
+    auto *runner = ectx()->rctx()->runner();
+
+    auto cb = [this](auto &&resp)
+        -> folly::SemiFuture<
+            nebula::storage::StorageRpcResponse<nebula::storage::cpp2::StatisticsResp>> {
+        if (!resp.ok()) {
+            throw std::runtime_error(std::move(resp).status().toString());
```
laura-ding: suggest doing onError_() instead of throw exception

src/graph/ShowExecutor.cpp
```diff
@@ -456,6 +459,63 @@ void ShowExecutor::showCreateEdge() {
...
+        std::vector<HostAddr> v;
+        v.reserve(retShowHosts.size());
+
+        std::transform(retShowHosts.cbegin(), retShowHosts.cend(), std::back_inserter(v),
+                       [](auto &h) { return h.first; });
+
+        return ectx()->storage()->getStatistics(v);
+    };
+
+    auto statusHandle = [this](auto &&resp) {
```
laura-ding: missing handle failed

jude-zhu:
close #796

dangleptr:
It is a good feature. Maybe we could introduce it inside storage-2.0. @critical27 @panda-sheep @darionyaphet 
For 1.x,  we close it now. Thanks @monadbobo for your contribution.

PR #836, opened 2019-08-22
Fixed occasional errors in FileBasedWalTest

### LINKED ISSUE #828: file_based_wal_test scheduling delay
**Describe the bug**
When I running `file_based_wal_test`, it maybe happen an exception:

```
Expected: numFilesAfterGC
Which is: 6
To be equal to: totalFilesNum - expiredFilesNum
Which is: 5
total 9, numFilesAfterGC 6, expiredFilesNum 4
```

This may caused by the `GC` scheduling delay.

critical27:
We need to wait until all ongoing wals has been flushed out, sometimes in CI env, it would failed.

close #828


jude-zhu:
Jenkins go

darionyaphet:
LGTM !  Maybe we should append some comment to explain the idea next time. Just suggestion!

critical27:
Jenkins go

src/kvstore/wal/FileBasedWal.cpp
```diff
@@ -282,6 +283,10 @@ void FileBasedWal::closeCurrFile() {
     currInfo_->setMTime(::time(nullptr));
     DCHECK_EQ(currInfo_->size(), FileUtils::fileSize(currInfo_->path()))
         << currInfo_->path() << " size does not match";
+    struct utimbuf timebuf;
+    timebuf.modtime = currInfo_->mtime();
```
darionyaphet: make sure it same with file last time

critical27:
For god sake, it finally pass ut.

src/kvstore/wal/test/FileBasedWalTest.cpp
```diff
@@ -362,19 +362,23 @@ TEST(FileBasedWal, TTLTest) {
                            folly::stringPrintf("Test string %02d", i)));
     }
     // Wait for the buffer flushed.
-    sleep(1);
+    while (wal->onGoingBuffersNum_ > 0) {
```
dangleptr: Do we still need the while scope?  After flusher removed,  onGoingBuffersNum_ will always be zero.

src/kvstore/wal/FileBasedWal.cpp
critical27: Yeah, I add a check of return value;

src/kvstore/wal/test/FileBasedWalTest.cpp
critical27: Good suggestion.

src/kvstore/wal/FileBasedWal.cpp
dangleptr: Well done!

Issue #417, opened 2019-05-22
Build docker image from Dockerfile 


darionyaphet:
Docker can build images automatically by reading the instructions from a `Dockerfile`.

A Dockerfile is a text document that contains all the commands a user could call on the command line to assemble an image. 

Using `docker build` users can create an automated build that executes several command-line instructions in succession.

sherman-the-tank:
We need to figure out where to store these docker files and some other scripts

darionyaphet:
refer to [pingcap/tidb](https://github.com/pingcap/tidb), we can put it under the root directory.

Issue #416, opened 2019-05-22
Using Kubernetes to manage docker containers


darionyaphet:
Kubernetes is an open-source system for automating deployment, scaling, and management of containerized applications. It groups containers that make up an application into logical units for easy management and discovery.

knightXun:
We are closely finishing this, see pr: #1473

PR #838, opened 2019-08-22
Support docker file

darionyaphet:


yixinglu:
I suggest we should split 3 nebula services to 3 docker images and follow the [recommendation](https://devops.stackexchange.com/questions/447/why-it-is-recommended-to-run-only-one-process-in-a-container): **one service per container**.

You maybe think there are many same dependencies for 3 nebula services. We can create a base image described in [this document](https://docs.docker.com/develop/develop-images/baseimages/). And 3 nebula services are `FROM` the base image and there's only one image layer stored in user disk.

For easy usage, we can use `docker-compose` to deploy our nebula services with a `docker-compose.yaml` file.

[Best practices for writing Dockerfiles](https://docs.docker.com/develop/develop-images/dockerfile_best-practices/) is an excellent reference document.

Dockerfile
```diff
@@ -0,0 +1,32 @@
...
+RUN yum -y install wget \
+                   openssh-clients \
+                   openssh-server net-tools
+
+RUN wget -O nebula-1.0.0-beta.el7-5.x86_64.rpm \ 
+    https://github.com/vesoft-inc/nebula/releases/download/v1.0.0-beta/nebula-1.0.0-beta.el7-5.x86_64.rpm
+
+RUN rpm -ivh nebula-1.0.0-beta.el7-5.x86_64.rpm
+
+RUN rm nebula-1.0.0-beta.el7-5.x86_64.rpm
```
yixinglu: We should extract a `NEBULA_VERSION` env variable to avoid updating the package name each release.

yixinglu:
refere to #416

Dockerfile
darionyaphet: Good point

darionyaphet:
> refere to #416

Thanks for you comment. It's really a good suggestion.  Actually, we use `docker` as a demo to show how to use `Nebula`.

This PR should refer to [Build docker image from Dockerfile](https://github.com/vesoft-inc/nebula/issues/417)  .

Not [Using Kubernetes to manage docker containers](https://github.com/vesoft-inc/nebula/issues/416) .

yixinglu:
@darionyaphet 
I think 3 nebula services could use a same docker image now. later we can build them separately to single image. The `Dockerfile` is like the following codes:

```
FROM centos:7.6.1810

MAINTAINER darion.wang "darion.wang@vesoft.com"

ARG NEBULA_VERSION

RUN yum -y install wget openssh-clients openssh-server net-tools

RUN wget -O nebula.el7-5.x86_64.rpm \
  https://github.com/vesoft-inc/nebula/releases/download/v${NEBULA_VERSION}/nebula-${NEBULA_VERSION}.el7-5.x86_64.rpm && \
  rpm -ivh nebula.el7-5.x86_64.rpm && \
  rm nebula.el7-5.x86_64.rpm

WORKDIR /usr/local/nebula
```

we build `nebula-graph` image each release by 
```
$ docker build -t vesoft/nebula-graph:latest -f Dockerfile --build-arg NEBULA_VERSION=1.0.0-beta .
```

Given the following `docker-compose.yaml`, users can start nebula services with one shell command:

```shell
$ docker-compose up -d
```

```
# docker-compose.yaml
version: '3.2'
services:
  metad:
    image: vesoft/nebula-graph:latest
    expose:
      - 45500
    entrypoint:
      - ./bin/nebula-metad
      - --flagfile=./etc/nebula-metad.conf
      - --daemonize=false
      - --meta_server_addrs=172.28.1.1:45500
      - --local_ip=172.28.1.1
      - --port=45500
    networks:
      nebula-net:
        ipv4_address: 172.28.1.1

  graphd:
    image: vesoft/nebula-graph:latest
    ports:
      - 6699:3699
    entrypoint:
      - ./bin/nebula-graphd
      - --flagfile=./etc/nebula-graphd.conf
      - --daemonize=false
      - --meta_server_addrs=172.28.1.1:45500
      - --port=3699
    networks:
      nebula-net:
        ipv4_address: 172.28.1.2

  storaged:
    image: vesoft/nebula-graph:latest
    expose:
      - 44500
    entrypoint:
      - ./bin/nebula-storaged
      - --flagfile=./etc/nebula-storaged.conf
      - --daemonize=false
      - --meta_server_addrs=172.28.1.1:45500
      - --local_ip=172.28.1.3
      - --port=44500
    networks:
      nebula-net:
        ipv4_address: 172.28.1.3

networks:
  nebula-net:
    ipam:
      driver: default
      config:
        - subnet: 172.28.0.0/16
```

We can improve `scripts/nebula.service` bash script and let it accept some parameters from outside to simplify above docker-compose entrypoint option.

Dockerfile
jude-zhu: each 'RUN' command will create a new layer

PR #841, opened 2019-08-23
Added some built-in functions for string operations

darionyaphet:


src/common/filter/FunctionManager.cpp
```diff
@@ -243,6 +243,156 @@ FunctionManager::FunctionManager() {
...
+    }
+    {
+        auto &attr = functions_["substr"];
+        attr.minArity_ = 3;
+        attr.maxArity_ = 3;
+        attr.body_ = [] (const auto &args) {
+            auto value   = Expression::asString(args[0]);
+            auto start   = Expression::asInt(args[1]);
+            auto length  = Expression::asInt(args[2]);
+            return value.substr(start, length);
```
critical27: Maybe need to handle exception if start > value.size()

darionyaphet: Good catch. 

According to `MySQL` , when the start is bigger than value's size or length is less than or equal to 0 . It will return a empty string.


```
mysql> select substr("123", 1, 1);
+---------------------+
| substr("123", 1, 1) |
+---------------------+
| 1                   |
+---------------------+
1 row in set (0.00 sec)

mysql> select substr("123", 1, 0);
+---------------------+
| substr("123", 1, 0) |
+---------------------+
|                     |
+---------------------+
1 row in set (0.00 sec)

mysql> select substr("123", 1, -1);
+----------------------+
| substr("123", 1, -1) |
+----------------------+
|                      |
+----------------------+
1 row in set (0.00 sec)

mysql> select substr("123", 5, 1);
+---------------------+
| substr("123", 5, 1) |
+---------------------+
|                     |
+---------------------+
1 row in set (0.00 sec)

mysql> select substr("123", -1, 1);
+----------------------+
| substr("123", -1, 1) |
+----------------------+
| 3                    |
+----------------------+
1 row in set (0.00 sec)

mysql> select substr("123", -1, 0);
+----------------------+
| substr("123", -1, 0) |
+----------------------+
|                      |
+----------------------+
1 row in set (0.00 sec)

mysql> select substr("123", -1, -1);
+-----------------------+
| substr("123", -1, -1) |
+-----------------------+
|                       |
+-----------------------+
1 row in set (0.00 sec)
```

When

critical27: Reasonable.

dutor:
Jenkins,go!

PR #850, opened 2019-08-27
Use smart pointer make function to construct objects

yixinglu:
Construct objects with smart pointer **make** function rather than `new` way.

references:
- [Quick Q: Differences between std::make_unique and std::unique_ptr with new](https://isocpp.org/blog/2019/06/quick-q-differences-between-stdmake-unique-and-stdunique-ptr-with-new)


laura-ding:
jenkins go

PR #853, opened 2019-08-27
Added support to singular forms for STEPS and VALUES

dutor:
As title


darionyaphet:
LGTM !

Do you mind support some test case for the update ？

Issue #852, opened 2019-08-27
Use big-endian for data version,  otherwise the key order is messed up


dangleptr:
It is a critical bug,  it will cause the result incorrect.

dangleptr:
#710 is caused by this problem.

Issue #710, opened 2019-08-01
Fail to get the latest version data


zlcook:
**Describe the bug(__must be provided__)**
Inserting same edges with different property values, However,  can't get the latest version data at query time.

**How To Reproduce(__must be provided__)**
1. add the following code to `nebula/src/graph/test/DataTest.cpp`
2. compile and run the test case multiples time.
```
TEST_F(DataTest, InsertSameEdges) {
    {
        cpp2::ExecutionResponse resp;
        std::string cmd = "INSERT VERTEX person(name, age) "
                          "VALUES hash(\"Tony\"):(\"Tony\", 18), hash(\"Mack\"):(\"Mack\", 19)";
        auto code = client_->execute(cmd, resp);
        ASSERT_EQ(cpp2::ErrorCode::SUCCEEDED, code);
    }
    // Insert the same edges with different property value
    {
        cpp2::ExecutionResponse resp;
        std::string cmd = "INSERT EDGE schoolmate(likeness) VALUES "
                          "hash(\"Tony\")->hash(\"Mack\")@111:(80)";
        auto code = client_->execute(cmd, resp);
        ASSERT_EQ(cpp2::ErrorCode::SUCCEEDED, code);
    }
    {
        cpp2::ExecutionResponse resp;
        std::string cmd = "INSERT EDGE schoolmate(likeness) VALUES "
                          "hash(\"Tony\")->hash(\"Mack\")@111:(9000)";
        auto code = client_->execute(cmd, resp);
        ASSERT_EQ(cpp2::ErrorCode::SUCCEEDED, code);
    }
    {
        cpp2::ExecutionResponse resp;
        std::string cmd = "INSERT EDGE schoolmate(likeness) VALUES "
                          "hash(\"Tony\")->hash(\"Mack\")@111:(123456789)";
        auto code = client_->execute(cmd, resp);
        ASSERT_EQ(cpp2::ErrorCode::SUCCEEDED, code);
    }
    // Get result
    {
        cpp2::ExecutionResponse resp;
        std::string cmd = "GO FROM hash(\"Tony\") OVER schoolmate "
                          "YIELD $$.person.name, schoolmate.likeness";
        auto code = client_->execute(cmd, resp);
        ASSERT_EQ(cpp2::ErrorCode::SUCCEEDED, code);
        using valueType = std::tuple<std::string, int64_t>;
        // get the latest result
        std::vector<valueType> expected{
            {"Mack", 123456789},
        };
[...]

PR #854, opened 2019-08-27
[Fix] Insert multi version tag and edge's order error

darionyaphet:
Please see: 
[Fail to get the latest version data](https://github.com/vesoft-inc/nebula/issues/710) and  [Use big-endian for data version, otherwise the key order is messed up](https://github.com/vesoft-inc/nebula/issues/852)


dangleptr:
close #710
close #852

src/storage/AddEdgesProcessor.cpp
```diff
@@ -16,12 +16,18 @@ void AddEdgesProcessor::process(const cpp2::AddEdgesRequest& req) {
     auto spaceId = req.get_space_id();
     auto version =
         std::numeric_limits<int64_t>::max() - time::WallClock::fastNowInMicroSec();
+    // Switch version to big-endian, make sure the key is in ordered.
+    version = folly::Endian::big(version);
```
wadeliuyi: can we use PC's time to version in Distributed Systems?

darionyaphet: Use `max - timestamp` is to make sure the version is in descending order and the latest version is the first one when you scan rocksdb with prefix.

PR #855, opened 2019-08-27
Report error when wrong column name was given in order_by.

CPWstatic:
We should report error in such situation.


CPWstatic:
Jenkins go!

PR #858, opened 2019-08-28
Add logic xor and arithmetic ^

### LINKED ISSUE #636: nGQL expressions support 'or' 'and' 'not'
Just like mysql, nGQL logic operators support _"And"_, _"Or"_ and _"Not"_ etc.

laura-ding:
close #636


laura-ding:
jenkins go

src/common/filter/Expressions.cpp
```diff
@@ -839,10 +842,22 @@ OptVariantType ArithmeticExpression::eval() const {
...
+            if (isArithmetic(l) && isArithmetic(r)) {
+                if (isDouble(l) || isDouble(r)) {
+                    return fmod(asDouble(l), asDouble(r));
+                }
                 return OptVariantType(asInt(l) % asInt(r));
             }
             break;
+        case XOR:
+            if (isArithmetic(l) && isArithmetic(r)) {
+                if (isDouble(l) || isDouble(r)) {
```
critical27: Do we really need to support xor for double? It is a little bit weird that we round up double and then do xor.

laura-ding: I see mysql is supported

darionyaphet: What scenarios are applicable for xor with double ?

darionyaphet:
LGTM ! Could you support some document to explain XOR operator ?

src/common/filter/Expressions.cpp
dutor: @critical27 @darionyaphet 
Personally, I have not seen this usage.

But from the perspective of semantic of XOR itself, or other bit ops like bitwise AND, they are just literally performing the ops bitwisely. It's up to the user to the determine whether the op is meaningful.

Issue #631, opened 2019-07-15
Create schema with default values


Amber1990Zhang:
`INSERT VERTEX player(name) VALUES 101:("Vicenta"); -- age is set to default value 0`
![image](https://user-images.githubusercontent.com/42762957/61204165-19c9ef80-a71f-11e9-9218-5b89402b3a38.png)
Docs says code above will set `age` to default value `0`, but actually nebula keeps setting the same error shown in the picture.
So I wonder if I can set default values for schema upon building it, thus the error won't be set again.

darionyaphet:
```
(user@127.0.0.1) [nba]> INSERT VERTEX player(name) VALUES -8379929135833483044: ("Amar'e Stoudemire")
[ERROR (-8)]: Wrong number of props
```

I test it and catch a error message.

dangleptr:
Typically,  if you want to a default value for one column.  You should specify it when creating the schema.  Currently, we haven't supported it yet.  But it is our design.

PR #860, opened 2019-08-28
Support default value when create tag and edge. 

darionyaphet:
#631


src/graph/InsertEdgeExecutor.cpp
```diff
@@ -125,6 +137,78 @@ StatusOr<std::vector<storage::cpp2::Edge>> InsertEdgeExecutor::prepareEdges() {
             values.emplace_back(ovalue.value());
         }
 
+        if (expressions.size() != schema_->getNumFields()) {
```
wadeliuyi: this function is too long, maybe should separate it to some sub function?

src/graph/InsertEdgeExecutor.cpp
```diff
@@ -125,6 +137,78 @@ StatusOr<std::vector<storage::cpp2::Edge>> InsertEdgeExecutor::prepareEdges() {
...
+                                variantType = folly::to<bool>(defaultValues_[fieldName]);
+                            } catch (const std::exception& ex) {
+                                LOG(ERROR) << "Conversion to bool failed: "
+                                           << defaultValues_[fieldName];
+                                return Status::Error("Type Conversion Failed");
+                            }
+
+                            VLOG(3) << "Insert " << fieldName << ":" << variantType
+                                    << " at " << insertPosition;
+                            values.insert(values.begin() + insertPosition, std::move(variantType));
```
wadeliuyi: in this function, maybe list for values is better than vector.

src/graph/InsertEdgeExecutor.cpp
```diff
@@ -125,6 +137,78 @@ StatusOr<std::vector<storage::cpp2::Edge>> InsertEdgeExecutor::prepareEdges() {
...
+                            break;
+                        default:
+                            LOG(ERROR) << "Unknow type";
+                            return Status::Error("Unknow type");
+                    }
+                    insertPosition++;
+                } else if (*props_[propsIndex] == fieldName) {
```
wadeliuyi: why here we can visit props_ by order, but in function check we visit it by find_if ?

src/graph/InsertVertexExecutor.cpp
```diff
@@ -111,22 +135,83 @@ StatusOr<std::vector<storage::cpp2::Vertex>> InsertVertexExecutor::prepareVertic
...
             auto schema = schemas_[index];
 
-            // props's number should equal to value's number
-            if ((valuePos + propSize) > values.size()) {
-                LOG(ERROR) << "Input values number " << values.size()
-                           << ", props number " << propSize;
-                return Status::Error("Wrong number of value");
+            auto schemaNumFields = schema->getNumFields();
+            size_t propsIndex = 0;
+            for (size_t schemaIndex = 0; schemaIndex < schemaNumFields; schemaIndex++) {
```
wadeliuyi: maybe can do this in sub function

src/graph/InsertVertexExecutor.cpp
```diff
@@ -111,22 +135,83 @@ StatusOr<std::vector<storage::cpp2::Vertex>> InsertVertexExecutor::prepareVertic
...
+                        case nebula::cpp2::SupportedType::STRING:
+                            variantType = defaultValues_[fieldName];
+                            values.insert(values.begin() + insertPosition, std::move(variantType));
+                            break;
+                        default:
+                            LOG(ERROR) << "Unknow type";
+                            return Status::Error("Unknow type");
+                    }
+                    insertPosition++;
+                } else if (*props[propsIndex] == fieldName) {
```
wadeliuyi: else if (propsIndex < props.size()) {
    if (*props[propsIndex] == fieldName) {
        propsIndex++;
    }
    insertPosition++;
}

src/graph/InsertVertexExecutor.cpp
```diff
@@ -111,22 +135,83 @@ StatusOr<std::vector<storage::cpp2::Vertex>> InsertVertexExecutor::prepareVertic
...
+                    insertPosition++;
+                } else {
+                    insertPosition++;
+                }
             }
 
             RowWriter writer(schema);
             auto valueIndex = valuePos;
-            for (auto fieldIndex = 0u; fieldIndex < schema->getNumFields(); fieldIndex++) {
+            for (auto fieldIndex = 0u; fieldIndex < schemaNumFields; fieldIndex++) {
```
wadeliuyi: no auto when define a variable

src/meta/client/MetaClient.cpp
```diff
@@ -1140,6 +1140,50 @@ folly::Future<StatusOr<bool>> MetaClient::balanceLeader() {
     return future;
 }
 
+folly::Future<StatusOr<std::string>> MetaClient::getTagDefaultValue(GraphSpaceID spaceId,
+                                                                    TagID tagId,
+                                                                    const std::string& field) {
+    cpp2::GetReq req;
+    static std::string defaultKey = "__default__";
```
wadeliuyi: the default value maybe can cached in client.

src/meta/processors/schemaMan/CreateEdgeProcessor.cpp
```diff
@@ -49,7 +49,65 @@ void CreateEdgeProcessor::process(const cpp2::CreateEdgeReq& req) {
     data.emplace_back(MetaServiceUtils::schemaEdgeKey(req.get_space_id(), edgeType, 0),
                       MetaServiceUtils::schemaEdgeVal(req.get_edge_name(), req.get_schema()));
 
-    LOG(INFO) << "Create Edge  " << req.get_edge_name() << ", edgeType " << edgeType;
+    auto columns = req.get_schema().get_columns();
+    for (auto& column : columns) {
+        if (column.__isset.default_value) {
```
wadeliuyi: do it in sub function

src/graph/InsertEdgeExecutor.cpp
darionyaphet: I have extracted some common parts.

src/meta/client/MetaClient.cpp
darionyaphet: Yep Good Point ~

src/graph/InsertEdgeExecutor.cpp
```diff
@@ -125,6 +137,78 @@ StatusOr<std::vector<storage::cpp2::Edge>> InsertEdgeExecutor::prepareEdges() {
...
+                    propsIndex++;
+                    insertPosition++;
+                } else {
+                    insertPosition++;
```
critical27: Why we can increase the position here? It seems you handle the situation when `props_[propsIndex]` match with `fieldName` in if statement, and mismatch in previous else if. Forgive me, a little bit hard to comprehend.

whitewum:
Question: How to modify the default value?  for the first time, I set int as 0; and the second time, I hope a int is -1?

darionyaphet:
> Question: How to modify the default value? for the first time, I set int as 0; and the second time, I hope a int is -1?

According to `MySQL`, it could use `ALTER TABLE` to modify the default value's data type and value. 

Maybe there are some related works  for this PR for example: `show create table with default info` and `alter default value`

src/graph/InsertEdgeExecutor.cpp
```diff
@@ -150,31 +147,84 @@ StatusOr<std::vector<storage::cpp2::Edge>> InsertEdgeExecutor::prepareEdges() {
...
+                        break;
+                    case nebula::cpp2::SupportedType::STRING:
+                        value = defaultValues_[fieldName];
+                        break;
+                    default:
+                        LOG(ERROR) << "Unknow type";
+                        return Status::Error("Unknow type");
+                }
+                VLOG(3) << "Supplement default value : " << fieldName << " : " << value;
+                if (schemaType.type == nebula::cpp2::SupportedType::TIMESTAMP) {
```
critical27: How about reuse same code with Line 166

src/graph/InsertVertexExecutor.cpp
```diff
@@ -124,55 +142,101 @@ StatusOr<std::vector<storage::cpp2::Vertex>> InsertVertexExecutor::prepareVertic
...
                     }
-                    writeVariantType(writer, timestamp.value());
+
                 } else {
-                    writeVariantType(writer, value);
-                }
+                    // fetch default value from cache
+                    VariantType value;
+                    auto type = schema->getFieldType(schemaIndex).type;
+                    switch (type) {
```
critical27: It seems this piece is same with InsertEdge, could we reuse them?

darionyaphet: Thanks & I will move it into `Executor`

Issue #826, opened 2019-08-21
Int value range is incorrect


moneytoyang:
数类
Nebula支持两种基础数据类型 -- int64 & double

int64的范围是[-9223372036854775808, 9223372036854775807]。double没有上限和下限。

注意：在基于int64的计算中不存在溢出。

case:

```
(user@127.0.0.1) [myspace]> create tag  player(name string, age int);
Execution succeeded (Time spent: 3024/4232 us)


(user@127.0.0.1) [myspace]> show tags;
==========
| Name   |
==========
| player |
----------
Got 1 rows (Time spent: 2356/3419 us)



(user@127.0.0.1) [myspace]> desc tag player;
==================
| Field | Type   |
==================
| name  | string |
------------------
| age   | int    |
------------------
Got 2 rows (Time spent: 1188/2033 us)


(user@127.0.0.1) [myspace]> insert vertex player(name, age) values 1001:("yangzi", 9223372036854775808);
[ERROR (-7)]: syntax error near `9223372036854775808'
(user@127.0.0.1) [myspace]> insert vertex player(name, age) values 1001:("yangzi", 9223372036854775809);
[ERROR (-7)]: syntax error near `9223372036854775809'
(user@127.0.0.1) [myspace]> insert vertex player(name, age) values 1001:("yangzi", 92233720368547758010);
[ERROR (-7)]: syntax error near `92233720368547758010'
(user@127.0.0.1) [myspace]> insert vertex player(name, age) values 1001:("yangzi", 922337203685477580100);
[ERROR (-7)]: syntax error near `922337203685477580100'

(user@127.0.0.1) [myspace]> insert vertex player(name, age) values 1001:("yangzi", 9223372036854775807);
Execution succeeded (Time spent: 1708/2260 us)


(user@127.0.0.1) [myspace]> fetch prop on player 1001;
======================================
| player.name | player.age           |
======================================
| yangzi      | 9223372036854775807 |
--------------------------------------
Got 1 rows (Time spent: 2341/3002 us)


##########################################################
(user@127.0.0.1) [myspace]> insert vertex player(name, age) values 1001:("yangzi", -9223372036854775808);
[ERROR (-7)]: syntax error near `9223372036854775808'

(user@127.0.0.1) [myspace]> insert vertex player(name, age) values 1002:("yangzi", -9223372036854775807);
Execution succeeded (Time spent: 1400/1974 us)

(user@127.0.0.1) [myspace]> fetch prop on player 1002;
[...]

PR #862, opened 2019-08-28
fix Int value range is incorrect 

moneytoyang:
fix int value range is incorrect #826


src/parser/parser.yy
```diff
@@ -451,8 +451,18 @@ go_sentence
 
 step_clause
     : %empty { $$ = new StepClause(); }
-    | INTEGER KW_STEPS { $$ = new StepClause($1); }
-    | KW_UPTO INTEGER KW_STEPS { $$ = new StepClause($2, true); }
+    | INTEGER KW_STEPS {
+        if ((uint64_t)$1 == 9223372036854775808ULL) {
+            LOG(ERROR) << "Integer overflow";
```
laura-ding: We need to handle exceptions

src/parser/scanner.lex
```diff
@@ -285,8 +285,11 @@ IP_OCTET                    ([0-9]|[1-9][0-9]|1[0-9][0-9]|2[0-4][0-9]|25[0-5])
                                         yyterminate();
                                     }
                                 }
-                                int64_t val = 0;
+                                uint64_t val = 0;
```
laura-ding: Hex and octal do not need to be processed

src/parser/parser.yy
steppenwolfyuetong: Where is the exception?

moneytoyang: Where is the exception?

src/parser/scanner.lex
moneytoyang: why？ Don't they overflow?

src/parser/parser.yy
moneytoyang: Where is the exception?

laura-ding:
I have fixed this problem on p #1181, so close this

Issue #185, opened 2019-02-26
Push stats info to 3rd-party stats-base, such as Prometheus or openTSDB


sherman-the-tank:
Actively push the stats to external monitoring services

Shylock-Hg:
I think we'd better to integrate Prometheus for it's the standard in fact.
The Prometheus is widely used in production, easy to deploy and has rich documents,
active community.
We could integrate it to nebula by the [3rd SDK](https://github.com/jupp0r/prometheus-cpp)
but need 1. fork and change the http server component to `proxygen`, 2. add this
to third-party dependencies.Then we can expose the HTTP endpoint such as `/metrics`
in application to Prometheus server to scrape.

jude-zhu:
ref to https://github.com/vesoft-inc/nebula-stats-exporter

knightXun:
OpenTSDB is too heavy. I want to use thanos.

PR #872, opened 2019-08-30
Add basic metrics in storage service

### LINKED ISSUE #287: Add some metrics in storage service.
At least, we should support QPS,   latency (AVG/ P99/ P999) for all interfaces

critical27:
1. Add basic metrics for data operation in storage service
2. Add a storage integrity test tool in storage-perf.

close #287 

Now we support following metrics: `get_bound, bound_stats, vertex_props, edge_props, add_vertex, add_edge, del_vertex, update_vertex, update_edge`.

We can get qps or latency from these metrics (add suffix), for example: `vertex_props_qps` or `vertex_props_latency`. 

You can get a single metrics like this: 
`curl -G "http://127.0.0.1:50005/get_stats?stats=vertex_props_qps.rate.60`

If you want get multiple metrics like this:
`curl -G "http://127.0.0.1:50005/get_stats?stats=vertex_props_qps.rate.60,vertex_props_latency.avg.60,vertex_props_latency.p99.60&returnjson"`
We can get qps, avg latency and p99 latency in a minutue.

You can get all metrics like this:
`curl -G "http://127.0.0.1:50005/get_stats?stats"`

---
The storage integrity tool is used for verifying data integrity in storage. It only works for single replica for now. You can use it like this:
`./storage_integrity --meta_server_addrs=192.168.8.5:60001 --first_vertex_id=100 --width=1000 --height=1000`

`meta_server_addrs` is the meta server addresss.
`first_vertex_id` is the first vertex id which we inserts.
`width` and `height` is the width and height of the big matrix.


src/storage/StorageServiceHandler.h
```diff
@@ -74,6 +85,14 @@ class StorageServiceHandler final : public cpp2::StorageServiceSvIf {
...
     kvstore::KVStore* kvstore_ = nullptr;
     meta::SchemaManager* schemaMan_;
+    StorageStats getOutBoundQpsStat_;
+    StorageStats getInBoundQpsStat_;
+    StorageStats outBoundStatsQpsStat_;
+    StorageStats inBoundStatsQpsStat_;
+    StorageStats vertexPropsQpsStat_;
+    StorageStats edgePropsQpsStat_;
+    StorageStats addVerticesQpsStat_;
+    StorageStats addEdgesQpsStat_;
```
wadeliuyi: maybe should move all of the property of status to each class which use it, It is not a property of StorageServiceHandler, and just used them to construct sub processor, and it is a property of each sub processor more suitable

src/tools/storage-perf/StorageIntegrityTool.cpp
```diff
@@ -0,0 +1,263 @@
...
+        FLAGS_load_data_interval_secs = 10;
+        auto metaAddrsRet = nebula::network::NetworkUtils::toHosts(FLAGS_meta_server_addrs);
+        if (!metaAddrsRet.ok() || metaAddrsRet.value().empty()) {
+            LOG(ERROR) << "Can't get metaServer address, status: " << metaAddrsRet.status()
+                       << ", FLAGS_meta_server_addrs: " << FLAGS_meta_server_addrs;
+            return false;
+        }
+        threadPool_ = std::make_shared<folly::IOThreadPoolExecutor>(FLAGS_io_threads);
+        mClient_ = std::make_unique<meta::MetaClient>(threadPool_, metaAddrsRet.value());
+        CHECK(mClient_->waitForMetadReady());
```
wadeliuyi: return false is better
if (!mClient_->waitForMetadReady()) {
print some log
return false;
}

src/storage/StorageServiceHandler.h
critical27: The reason is partly because of each time we create a new instance of processor to handle a request. And we have to maintain the same stat id for all of them.

critical27:
Jenkins go, please don't abort - -!

src/tools/storage-perf/StorageIntegrityTool.cpp
critical27: Good suggestion.

src/tools/storage-perf/StorageIntegrityTool.cpp
```diff
@@ -0,0 +1,271 @@
...
+            auto tagIter = vschema->find(tagId_);
+            DCHECK(tagIter != vschema->end());
+            auto tagProvider = std::make_shared<ResultSchemaProvider>(tagIter->second);
+
+            for (auto& vdata : results[0].vertices) {
+                auto iter = std::find_if(vdata.tag_data.begin(), vdata.tag_data.end(),
+                    [this] (const auto& tagData) {
+                        return tagData.tag_id == tagId_;
+                    });
+                auto tagReader = RowReader::getRowReader(iter->data, tagProvider);
```
laura-ding: need to judge `iter` is the end()

critical27: Good catch

critical27:
> how to get the result of statsmanger? I have no see, will use Http?

Yeah, u r right

dangleptr:
How to get the metrics?  Could you describe it in the summary?

src/tools/storage-perf/StorageIntegrityTool.cpp
```diff
@@ -0,0 +1,274 @@
...
+            decltype(v.tags) tags;
+            storage::cpp2::Tag tag;
+            tag.set_tag_id(tagId_);
+            std::string props;
+            RowWriter writer;
+            writer << prev[i];
+            tag.set_props(writer.encode());
+            tags.emplace_back(std::move(tag));
+
+            v.set_tags(tags);
```
darionyaphet: using `std::move()` ?

src/tools/storage-perf/StorageIntegrityTool.cpp
```diff
@@ -0,0 +1,274 @@
...
+DEFINE_int32(partition_num, 1024, "partititon for space");
+DEFINE_string(space_name, "test_space", "the space name");
+DEFINE_string(tag_name, "test_tag", "the tag name");
+DEFINE_string(prop_name, "test_prop", "the property name");
+
+DEFINE_int64(first_vertex_id, 1, "The smallest vertex id");
+DEFINE_uint64(width, 100, "width of matrix");
+DEFINE_uint64(height, 1000, "height of matrix");
+
+DECLARE_int32(load_data_interval_secs);
```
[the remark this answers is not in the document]
critical27: Because it has been defined in MetaClient

jude-zhu:
relate to #185

PR #882, opened 2019-09-02
Added docs on how to make docs in PDF

Amber1990Zhang:
doc-pdf


darionyaphet:
so this doc.pdf is export from ? Please update the title and summary. :)

whitewum:
I prefer update the script, not the binary.

darionyaphet:
If you make sure it's ready, please remove the comments.

docs/merge-docs-into-one.py
```diff
@@ -0,0 +1,86 @@
...
+        elif os.path.isfile(path):
+            pass
+            # print(file_path + ":it's a normal file")
+    write_file(source_dir, dst_md_file)
+    dir_list.sort()
+    for sub_dir in dir_list:
+        write_all(sub_dir, dst_md_file)
+
+if __name__ == "__main__":
+    file = '/Users/zhangying/Desktop/output.md' # Output file
```
critical27: We'd better pass the output file as a parameter

Priceincoding: Yes we can definitely do that. However, for reusability it is better to set up constants in the file so that a specific user do not need to repeat the path parameters. I will add both methods.

docs/merge-docs-into-one.py
```diff
@@ -0,0 +1,69 @@
...
+
+    # Output file
+    targetFile = UNDEFINED
+    # Source file to be merged
+    sourceDir = UNDEFINED
+
+    # argument 1 will be the source, argument 2 will be the target
+    if len(sys.argv) != 3:
+        if targetFile == UNDEFINED or sourceDir == UNDEFINED:
+            sys.exit('error: targetFile and sourceDir not given in either scripts or commandline')
```
whitewum: print help msg for ./xxx or ./xxx -h

Amber1990Zhang: Both the index and pdf can be generated by pandoc, I'll take care that

Amber1990Zhang:
> 1. please paste a screenshot of the index.md (is it auto generated? or man-made?)
> 2. please give some description of how to install `pandoc` and how to use it to generate a pdf.
>    This is help others to use this tool
> 
> BTW, mv the file to a dir docs/tools/merge-all.py

I'll add a README.md file to detail how to use the tool.

docs/merge-docs-into-one.py
```diff
@@ -0,0 +1,69 @@
+import os, sys
+
+# This script is to help you merge all the docs into one mardown file, then you can convert it into one pdf file and down load.
+def get_all_md_file_path(file_path):
+    if not os.path.isdir(file_path):
+        print(file_path+" is not a dir")
+    file_list_in_md = os.listdir(file_path)
```
darionyaphet: you should use `pylint` to formatted the source code

whitewum:
it's better to generate section number. But it's ok right now

PR #890, opened 2019-09-03
update compile

Amber1990Zhang:
Update compile in get-started


docs/manual-CN/get-started.md
```diff
@@ -133,7 +112,7 @@ nebula 遵循 c++14 标准，依赖第三方库：
 
 * 配置 nebula-metad.conf
 
-   在 nebula 安装目录下，运行以下命令：
+   在 nebula 安装目录(/usr/local/nebula/)下，运行以下命令：
```
darionyaphet: this is the default install path

Amber1990Zhang: I just want to make it clear here, thanks for your suggestion.

PR #891, opened 2019-09-03
Generate thrift code for Python

laura-ding:


darionyaphet:
I have tested this update, it seems couldn't generation the Python's source code.

Do you test it ?

laura-ding:
@darionyaphet I have tested it, it can generate Python's source code. it should put ahead before generate go

PR #902, opened 2019-09-06
update Importer doc

jude-zhu:
update importer doc
at this moment, send users fb-thrift and graphd-client packages. When our mvn repo is ready, update the **Download JARs** section accordingly.


whitewum:
Currently, it's a single thread tool to import the median size csv file. You can refer to xxx if you have a data warehouse such as Spark.

darionyaphet:
> Currently, it's a single thread tool to import the median size csv file. You can refer to xxx if you have a data warehouse such as Spark.

That's true. If you have a large dataset, loading the SST file maybe better.

Issue #901, opened 2019-09-06
Support UUID


dangleptr:
UUID will return the unique id in the whole cluster for each string key.  

To avoid conflicts, we should keep the string to id map inside storage.

If so,  we will introduce another type key for our system,  and it is unfixed length.  We should redesign our key format to support it (of course,  index key should be supported too).

PR #914, opened 2019-09-09
Support key type

darionyaphet:
This pull request is subtask of #901 

Currently, the key format as follows:

The first byte used to describe the key type (data, index uuid or system key), and the latter three byte is PartitionID and the max Partition ID is `2^24 - 1`.  

And the secondary high position in tagID and edgeType is use to distinguish the vertex and edge (0 is vertex and 1 is edge).

Vertex key format:
`type(1) + partId(3) + vertexId(8) + tagId(4) + version(8)`

Edge key format:
`type(1) + partId(3) + srcId(8) + edgeType(4) + edgeRank(8) + dstId(8) + version(8)`


src/common/base/NebulaKeyUtils.h
```diff
@@ -13,13 +13,20 @@ namespace nebula {
...
- * partId(4) + vertexId(8) + tagId(4) + version(8)
+ * type(1) + partId(3) + vertexId(8) + tagId(4) + version(8)
  *
  * EdgeKeyUtils:
- * partId(4) + srcId(8) + edgeType(4) + edgeRank(8) + dstId(8) + version(8)
+ * type(1) + partId(3) + srcId(8) + edgeType(4) + edgeRank(8) + dstId(8) + version(8)
  *
  * */
 
+enum NebulaKeyType : char {
```
dangleptr: kData
kIndex
kUUID

src/common/base/NebulaKeyUtils.h
```diff
@@ -13,13 +13,20 @@ namespace nebula {
 
 /**
  * VertexKeyUtils:
- * partId(4) + vertexId(8) + tagId(4) + version(8)
+ * type(1) + partId(3) + vertexId(8) + tagId(4) + version(8)
  *
  * EdgeKeyUtils:
- * partId(4) + srcId(8) + edgeType(4) + edgeRank(8) + dstId(8) + version(8)
+ * type(1) + partId(3) + srcId(8) + edgeType(4) + edgeRank(8) + dstId(8) + version(8)
```
dangleptr: tagId/edgeType support vertex/edge

src/common/base/NebulaKeyUtils.cpp
```diff
@@ -13,8 +13,12 @@ std::string NebulaKeyUtils::vertexKey(PartitionID partId, VertexID vId,
                                       TagID tagId, TagVersion tv) {
     std::string key;
     key.reserve(kVertexLen);
-    key.append(reinterpret_cast<const char*>(&partId), sizeof(PartitionID))
-       .append(reinterpret_cast<const char*>(&vId), sizeof(VertexID))
+    for (int32_t n = 0; n < 3; n++) {
```
dangleptr: (PartId >> 8) | (type << 24)

src/common/base/NebulaKeyUtils.cpp
```diff
@@ -27,9 +31,13 @@ std::string NebulaKeyUtils::edgeKey(PartitionID partId,
                                     EdgeRanking rank,
                                     VertexID dstId,
                                     EdgeVersion ev) {
+    static int32_t edgeMask = 0x40000000;
     std::string key;
     key.reserve(kEdgeLen);
-    key.append(reinterpret_cast<const char*>(&partId), sizeof(PartitionID))
+    type |= edgeMask;
+
+    int32_t item = (partId >> 8) | (NebulaKeyType::kData);
```
critical27: This maybe be wrong. You discards the most significant byte when `>> 8`, and `bit or kData` at the least significant byte (which may overwrite the bits in partId). I think `int32_t item = (partId << 8) | (NebulaKeyType::kData);` is ok.

darionyaphet: It's reasonable. 

`int32_t` encode with little-encode but right shift and left shift maybe not dependence on encoded.

For example: `partID` is `0x0000000F` , `(partID << 8)` is `0x00000F00` .

src/kvstore/Part.cpp
```diff
@@ -59,7 +60,8 @@ Part::Part(GraphSpaceID spaceId,
 
 std::pair<LogID, TermID> Part::lastCommittedLogId() {
     std::string val;
-    ResultCode res = engine_->get(folly::stringPrintf("%s%d", kCommitKeyPrefix, partId_), &val);
+    auto key = folly::stringPrintf("%d%s%d", NebulaKeyType::kSystem, kCommitKeyPrefix, partId_);
```
dangleptr: The format should be type + partId + "committed"

src/common/base/NebulaKeyUtils.h
```diff
@@ -65,34 +84,41 @@ class NebulaKeyUtils final {
...
+        return type > 0 ? type & edgeMask : type;
     }
 
     static EdgeRanking getRank(const folly::StringPiece& rawKey) {
         CHECK_EQ(rawKey.size(), kEdgeLen);
         auto offset = sizeof(PartitionID) + sizeof(VertexID) + sizeof(EdgeType);
-        return readInt<EdgeRanking>(rawKey.data() + offset,
-                                    rawKey.size() - offset);
+        return readInt<EdgeRanking>(rawKey.data() + offset, sizeof(EdgeRanking));
     }
```
dangleptr: isDateKey and isIndexKey functions should be updated.

src/kvstore/Part.cpp
darionyaphet: we should move commitKey into NebulaKeyUtils

critical27:
Maybe need to update `kSystemParts` as well

src/common/base/NebulaKeyUtils.cpp
```diff
@@ -11,9 +11,13 @@ namespace nebula {
 // static
 std::string NebulaKeyUtils::vertexKey(PartitionID partId, VertexID vId,
                                       TagID tagId, TagVersion tv) {
+    static int32_t tagMask = 0xBFFFFFFF;
```
dutor: 1. For such constants, it would be better to place them in some well-known place, if they are or will be used in multiple places. 
  2. In my humble opinion, constants used as bit masks should be of type unsigned.
  3. For your information, declaring `tagMask`  like this will occupy memory spaces, in the global data segment. `constexpr` is what should use instead.

src/kvstore/RocksEngine.cpp
```diff
@@ -337,16 +330,16 @@ void RocksEngine::removePart(PartitionID partId) {
...
 
     std::vector<PartitionID> parts;
     while (iter->valid()) {
         auto key = iter->key();
-        CHECK_EQ(key.size(), prefixLen + sizeof(PartitionID));
-        parts.emplace_back(
-           *reinterpret_cast<const PartitionID*>(key.data() + key.size() - sizeof(PartitionID)));
+        CHECK_EQ(key.size(), sizeof(PartitionID));
+        auto partId = *reinterpret_cast<const PartitionID*>(key.data());
+        partId = partId >> 8;
```
dutor: Please note that we should pay more attentions to do right shifting operations on a signed integer, since there are chances that the sign extention is not what you want. In my opinion, it's always a bad idea to make an integer like this signed.

src/common/filter/Expressions.h
```diff
@@ -10,7 +10,7 @@
 #include "base/StatusOr.h"
 #include "base/Status.h"
 #include <boost/variant.hpp>
-#include  <boost/unordered_set.hpp>
+#include <boost/unordered_set.hpp>
```
dutor: Why not `std::unordered_set`?

src/common/base/NebulaKeyUtils.h
```diff
@@ -103,12 +136,15 @@ class NebulaKeyUtils final {
...
 
     static bool isDataKey(const folly::StringPiece& key) {
-        return !key.empty() && key[0] != kSysPrefix;
+        static int32_t typeMask = 0x000000FF;
+        static int32_t len = static_cast<int32_t>(sizeof(NebulaKeyType));
```
dutor: Again, `constexpr` and no `static`.

darionyaphet: But I found `kVertexLen` and `kEdgeLen` are be modified by `static constexpr` ?

src/kvstore/RocksEngine.cpp
darionyaphet: When we encoded the partition ID, using left shift firstly. 

Actually, partition ID is unsigned integer and it can't be a negative number so I think right shift maybe ok :)

src/common/base/NebulaKeyUtils.h
```diff
@@ -13,13 +13,21 @@ namespace nebula {
...
- * partId(4) + srcId(8) + edgeType(4) + edgeRank(8) + dstId(8) + version(8)
+ * type(1) + partId(3) + srcId(8) + edgeType(4) + edgeRank(8) + dstId(8) + version(8)
  *
  * */
 
+enum NebulaKeyType : uint32_t {
+    kData              = 0x00000001,
+    kIndex             = 0x00000002,
+    kUUID              = 0x00000003,
+    kSystemCommit      = 0x00000004,
```
dangleptr: Just use one kSystem to represent key inside our system. 
And you could distinguish commit key and part key by suffix.

src/kvstore/Part.cpp
dangleptr: I think we don't need to do it because the commitKey is only used in this place.

src/kvstore/RocksEngine.cpp
```diff
@@ -330,14 +330,20 @@ void RocksEngine::removePart(PartitionID partId) {
 
 std::vector<PartitionID> RocksEngine::allParts() {
     std::unique_ptr<KVIterator> iter;
-    static const std::string prefixStr = NebulaKeyUtils::systemPartsPrefix();
+    static const std::string prefixStr = NebulaKeyUtils::systemPrefix();
```
critical27: Maybe we can scan with prefix = kSystem + kSystemPart.

darionyaphet: But I put partition ID with kSystem

darionyaphet:
jenkins go

PR #916, opened 2019-09-09
Support multi address and execution retry

darionyaphet:
* Support multi address when create `GraphClient` .
* Support execution retry, the default is three times.
* When the insert execution failed, the record will append to a file.
* Update line limit from 160 to 100.


src/client/java/src/main/java/com/vesoft/nebula/graph/client/GraphClientImpl.java
```diff
@@ -147,16 +156,20 @@ public int execute(String stmt) {
             return ErrorCode.E_DISCONNECTED;
         }
 
-        try {
-            ExecutionResponse executionResponse = client.execute(sessionId_, stmt);
-            if (executionResponse.getError_code() != ErrorCode.SUCCEEDED) {
-                LOGGER.error("execute error: " + executionResponse.getError_msg());
+        int retry = executionRetry;
+        while (retry-- > 0) {
```
laura-ding: actual retry times is retry-1

src/client/java/src/main/java/com/vesoft/nebula/graph/client/GraphClientImpl.java
```diff
@@ -50,31 +50,38 @@
...
         checkArgument(timeout > 0);
-        checkArgument(retry > 0);
-        if (!InetAddresses.isInetAddress(host) || (port <= 0 || port >= 65535)) {
-            throw new IllegalArgumentException(String.format("%s:%d is not a valid address", host, port));
-        }
+        checkArgument(connectionRetry > 0);
+        addresses.forEach(address -> {
+            String host = address.getHost();
+            int port = address.getPort();
+            if (!InetAddresses.isInetAddress(host) || (port <= 0 || port >= 65535)) {
```
laura-ding: the handle to check address and port may be better put in  `fetchOptionValue`

darionyaphet: `fetchOptionValue` is in Importer and this is `GraphClient`

PR #917, opened 2019-09-09
Use ring buffer in wal

critical27:
Replace a list of in memory buffers with a single ring buffer in wal.


src/kvstore/Part.cpp
```diff
@@ -179,9 +179,10 @@ bool Part::commitLogs(std::unique_ptr<LogIterator> iter) {
     LogID lastId = -1;
     TermID lastTerm = -1;
     while (iter->valid()) {
-        lastId = iter->logId();
-        lastTerm = iter->logTerm();
-        auto log = iter->logMsg();
+        auto logEntry = iter->logEntry();
+        lastId = std::get<0>(logEntry);
+        lastTerm = std::get<1>(logEntry);
```
wadeliuyi: move to after check log.empty()

src/kvstore/wal/FileBasedWalIterator.cpp
```diff
@@ -181,56 +85,88 @@ LogIterator& FileBasedWalIterator::operator++() {
...
+                   << " is out of the wal files range";
+        currId_ = lastId_ + 1;
+        return;
     }
 
-    return *this;
+    nextFirstId_ = getFirstIdInNextFile();
+    CHECK_LE(currId_, idRanges_.front().second);
+
+    if (!idRanges_.empty()) {
```
wadeliuyi: check idRanges_.empty() line 128, so here never empty, no need check

src/kvstore/wal/FileBasedWalIterator.cpp
```diff
@@ -240,75 +176,69 @@ LogID FileBasedWalIterator::logId() const {
...
 
+LogEntry FileBasedWalIterator::logEntry() {
+    LogEntry logEntry;
+    // Try to get log from ring buffer first, if failed to read from buffer,
+    // try to read from wal files
+    // TODO: need to figure out when can we stop reading from wal file anymore once hit ring buffer
+    if (!ringBuffer_->getLogEntry(currId_, logEntry)) {
+        if (!needToReadWalFile_) {
+            loadWalFiles();
+            needToReadWalFile_ = true;
```
wadeliuyi: change needToReadWalFile_  to hasBeenReadWalFile_?

src/kvstore/wal/FileBasedWalIterator.cpp
```diff
@@ -240,75 +176,69 @@ LogID FileBasedWalIterator::logId() const {
...
-
-LogID FileBasedWalIterator::getFirstIdInNextBuffer() const {
-    auto it = buffers_.begin();
-    ++it;
-    if (it == buffers_.end()) {
-        return buffers_.front()->lastLogId() + 1;
-    } else {
-        return (*it)->firstLogId();
+                 static_cast<ssize_t>(currMsgLen_));
+        return {logId, term, cluster, logMsg};
```
wadeliuyi: maybe here can std::move(logMsg)

src/kvstore/wal/FileBasedWalIterator.cpp
critical27: Good catch

src/kvstore/wal/FileBasedWalIterator.cpp
critical27: The variable is false typically because we read log from buffer. If it is out of ring buffer, then we need to read from wal files, so I think needToReadWalFile_ make more sense

src/kvstore/wal/FileBasedWalIterator.cpp
critical27: Good catch

src/kvstore/wal/FileBasedWalIterator.cpp
wadeliuyi: needToReadWalFile_  to chinese is "是否需要读wal", when it is false, "不需要读wal", but the logic is read wal file.

src/kvstore/Part.cpp
wadeliuyi: if the log is empty, can less a memory access.

src/kvstore/wal/FileBasedWalIterator.cpp
critical27: Fair enough

src/kvstore/wal/InMemoryLogBuffer.h
```diff
@@ -9,57 +9,71 @@
...
-//
-// In-memory buffer (thread-safe)
-//
 class InMemoryLogBuffer final {
 public:
-    explicit InMemoryLogBuffer(LogID firstLogId)
-        : firstLogId_(firstLogId) {}
+    InMemoryLogBuffer() {
+        logs_.reserve(1 << FLAGS_wal_buffer_size_exp);
+        mask_ = logs_.capacity() - 1;
```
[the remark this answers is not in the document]
dutor: Quoted from [cppreference::vector::reserve](https://en.cppreference.com/w/cpp/container/vector/reserve):
> Increase the capacity of the vector to a value that's greater or equal to new_cap.

src/kvstore/Part.cpp
```diff
@@ -179,9 +179,10 @@ bool Part::commitLogs(std::unique_ptr<LogIterator> iter) {
     LogID lastId = -1;
     TermID lastTerm = -1;
     while (iter->valid()) {
-        lastId = iter->logId();
-        lastTerm = iter->logTerm();
-        auto log = iter->logMsg();
+        auto logEntry = iter->logEntry();
```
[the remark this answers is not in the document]
dutor: In the near future, we could utilize `folly::IOBuf` in such scenarios, which is designed for zero-copy. I will take care of this.

dangleptr:
Has been implemented inside storage 2.0. Close it now.

PR #921, opened 2019-09-10
Add ENABLE_TESTING to enable to build unit tests optionally

laura-ding:


CMakeLists.txt
```diff
@@ -506,6 +511,13 @@ macro(nebula_link_libraries target)
     )
 endmacro(nebula_link_libraries)
 
+function(nebula_add_subdirectory dir_name)
+    if ((NOT ENABLE_TESTING) AND (${dir_name} STREQUAL test))
```
yixinglu: I think it's better not to ignore the entire `test` directory, but rather to provide a macro like `nebula_add_test` to ignore a single UT.

laura-ding: In addition to binary, there are also objects. We specify the `test` directory as the directory of tests

yixinglu: Do all tests have to be placed in the test directory? How to disable one or more of these tests ?

dutor: @yixinglu To assume that all unit tests are placed in _**test**_ directory is not a strong assumption.

For the time being, we have not demands to disable the building of individual tests.

yixinglu: Got it!

dutor:
jenkins go

PR #924, opened 2019-09-10
Add graph data simulation generation tool

zhangguoqing:


zhangguoqing:
A ratable graph is generated according to the configuration evaluated by the  actual business graph structure. And the graph data is randomly generated  according to the schema definition of the tag and the edge.

zhangguoqing:
Jenkins go

dangleptr:
Close it now.  We have a perf tool for LDBC dataset already which is written by @dutor 
Please consider https://github.com/dutor/nebula-bench

PR #934, opened 2019-09-12
Issue #932 fixed the crashed when the tag prop not exists.

### LINKED ISSUE #932: graphd crash when the node no exists
image

image

monadbobo:
close #932


monadbobo:
Jenkins go!

PR #935, opened 2019-09-12
Modify the toHostAddr interface to support host resolved (gethostbyname).

### LINKED ISSUE #931: The meta_server_addrs in the configuration should support the host name.
The meta_server_addrs in the configuration should support the host name.

monadbobo:
close #931


monadbobo:
Jenkins go!

src/common/network/NetworkUtils.cpp
```diff
@@ -215,15 +215,32 @@ std::string NetworkUtils::intToIPv4(IPv4 ip) {
     return buf;
 }
 
-StatusOr<HostAddr> NetworkUtils::toHostAddr(folly::StringPiece ip, int32_t port) {
+StatusOr<std::vector<HostAddr>> NetworkUtils::toHostAddr(folly::StringPiece ip, int32_t port) {
```
dutor: Given that now we don't have IPv6 support in mind, semanticly, we shall not return multiple addresses for a single hostname. So we could check on the list returned from `gethostbyname` to only yield the one that meets our criteria.

src/common/network/NetworkUtils.cpp
```diff
@@ -215,15 +215,32 @@ std::string NetworkUtils::intToIPv4(IPv4 ip) {
...
-        return Status::Error("Bad ip format:%s", ip.start());
+    std::string str = ip.toString();
+    if (ipv4ToInt(str, ipV4)) {
+        addrs.emplace_back(std::move(ipV4), port);
+        return addrs;
     }
-    return std::make_pair(ipV4, port);
+
+    struct hostent* h;
+    h = gethostbyname(str.c_str());
```
dutor: According to the man pages, `gethostbyname` is already obsolete, while the alternative interface is `getaddrinfo`.

src/common/network/NetworkUtils.cpp
monadbobo: This has nothing to do with ipv6, the host was likely to return multiple ip addresses.

src/common/network/NetworkUtils.cpp
monadbobo: In some older systems does not provide getaddrinfo, so here we need to detect the system for getaddrinfo support.So I will rewrite it here.

src/common/network/NetworkUtils.cpp
dutor: Multiple IP for single hostname? I have no idea.

src/common/network/NetworkUtils.cpp
dutor: `getaddrinfo` was introduced since glibc 2.3.4, I don't think it's a problem for us.

monadbobo: Ok, then I will ignore gethostbyname.

src/common/network/NetworkUtils.cpp
monadbobo: Because gethostbyname/getaddrinfo may take dns query, it is very common to return multiple ip addresses. Even if it is a local host, a host corresponding to multiple IP addresses is more convenient for operation and maintenance.

dutor: Fair enough

dutor:
Wait a minute.

Maybe It would be better to change the name of the interface accordingly to indicates that this function will return multiple IP addresses. Agree?

monadbobo:
> Wait a minute.
> 
> Maybe It would be better to change the name of the interface accordingly to indicates that this function returns multiple IP addresses. Agree?

Maybe we can split "toHostAddr(folly::StringPiece ip, int32_t port)" to two functions, getHostByName and getHostByAddr, and then call them in "toHostAddr(folly::StringPiece ipPort)"?

src/common/network/NetworkUtils.cpp
```diff
@@ -215,28 +215,52 @@ std::string NetworkUtils::intToIPv4(IPv4 ip) {
     return buf;
 }
 
-StatusOr<HostAddr> NetworkUtils::toHostAddr(folly::StringPiece ip, int32_t port) {
-    IPv4 ipV4;
-    if (!ipv4ToInt(ip.toString(), ipV4)) {
-        return Status::Error("Bad ip format:%s", ip.start());
+StatusOr<std::vector<HostAddr>> NetworkUtils::resolveHost(const std::string &ip, int32_t port) {
```
CPWstatic: Would it be better to change **_ip_** with _**hostname**_?

PR #938, opened 2019-09-12
Fix the problem of displaying the _dst attribute of the edge.

### LINKED ISSUE #933: The "go from" doesn't return expected node
I insert an edge
```
> INSERT EDGE like(likeness) VALUES 101 -> 102:(95.1)
```

When querying
```
>  go from 101 over like
======================
| like._dst          |
======================
| 66                 |
----------------------
```

image

monadbobo:
this problem was caused by the conversion of decimal to hexadecimal when the console prints the VID type.
close #933


monadbobo:
Jenkins go!

PR #940, opened 2019-09-12
Fixed buffer overflow in Scanner

dutor:
There are buffer overflows when the length of string literals exceed limit. For the time being, we take this as _syntax error_. In future, we shall make the limit configurable at runtime, either through system configuration or various types, like _varchar_.


src/common/filter/test/ExpressionTest.cpp
```diff
@@ -735,4 +735,42 @@ TEST_F(ExpressionTest, InvalidExpressionTest) {
 #undef TEST_EXPR
 }
 
+
+TEST_F(ExpressionTest, StringLengthLimitTest) {
```
monadbobo: Perhaps this testcase should be better placed in parser/test/ParserTest.cpp?

dutor: Emmm, I have been trying to figure out a way to test `Expression` individually, instead of testing from inside a complete statement. But it seems no easy way to do this with Bison, unless we allow to accept an ordinary expression as a whole statement in the gramar. Any idea?

dutor:
Jenkins go

dutor:
Jenkins go!

PR #948, opened 2019-09-18
Improved build concurrency

dutor:
At compile time, we actually depend on the _**codegen**_ of the thrift files, but no need to _**compile**_ the generated files first.
With this PR, the overall build time could be reduced by nearly 30% in a 56 processors environment.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -380,9 +380,6 @@ macro(nebula_add_test)
     endif()
 endmacro()
 
-macro(nebula_build_after_base obj_name)
-endmacro()
-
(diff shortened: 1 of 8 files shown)


CMakeLists.txt
```diff
@@ -381,12 +381,23 @@ macro(nebula_add_test)
 endmacro()
 
 macro(nebula_build_after_base obj_name)
-    add_dependencies(${obj_name} base_obj)
 endmacro()
```
yixinglu: delete the unused macro `nebula_build_after_base`.

dutor: My bad

PR #957, opened 2019-09-19
Use manual configuration when generate go source code.

CPWstatic:
The default configuration of _**thrift_import**_ is out of date. It would not work well. So I replace it with a manual configuration.

### DIFF
--- a/cmake/ThriftGenerate.cmake
+++ b/cmake/ThriftGenerate.cmake
@@ -84,7 +84,7 @@ add_custom_command(
     --gen "mstch_cpp2:include_prefix=${include_prefix},process_in_event_base,stack_arguments"
     --gen "py"
     --gen "java:hashcode"
-    --gen "go"
+    --gen "go:thrift_import=github.com/facebook/fbthrift/thrift/lib/go/thrift"
     -o "." "${file_path}/${file_name}.thrift"
   DEPENDS "${file_path}/${file_name}.thrift"
   COMMENT "Generating thrift files for ${file_name}"


CPWstatic:
Jenkins go!

cmake/ThriftGenerate.cmake
```diff
@@ -84,7 +84,7 @@ add_custom_command(
     --gen "mstch_cpp2:include_prefix=${include_prefix},process_in_event_base,stack_arguments"
     --gen "py"
     --gen "java:hashcode"
-    --gen "go"
+    --gen "go:thrift_import=github.com/facebook/fbthrift/thrift/lib/go/thrift"
```
darionyaphet: why add the dependency when compile and generation source code？

CPWstatic: The generated code depends on it. It has a default value, but out of date.

CPWstatic:
Jenkins go!

cmake/ThriftGenerate.cmake
darionyaphet: But we use glide to manage the dependency.

CPWstatic: It is not about the dependency management. I mean the configuration of **thrift_import** is explicitly insert into the generated code. like:
```import "github.com/facebook/fbthrift/thrift/lib/go/thrift"```

dutor: @darionyaphet The default import path is _github.com/facebook/fbthrift-go_, which seems no longer exist. You could check on the output of `/path/to/the/fbthrift/generator --help`, i.e. `/opt/nebula/third-party/fbthrift/bin/thrift1 --help`

PR #958, opened 2019-09-19
Support uuid in storage service

critical27:
`UUID` key format: type(1) + partId(3) + name.
The value is the corresponding vertex id.

Waiting on #914.
Subtask of #901.


src/storage/GetUUIDProcessor.h
```diff
@@ -0,0 +1,75 @@
...
+        auto partId = req.get_part_id();
+        auto name = req.get_name();
+        auto key = NebulaKeyUtils::uuidKey(partId, name.c_str());
+        std::string val;
+        VertexID vId;
+        auto ret = kvstore_->get(spaceId, partId, key, &val);
+        // try to get the corresponding vertex id
+        if (ret != kvstore::ResultCode::SUCCEEDED) {
+            // need to generate new vertex id of this uuid
+            std::hash<std::string> hashFunc;
```
dangleptr: We'd better use different hash function with partition when generating UUID.

src/storage/GetUUIDProcessor.h
```diff
@@ -0,0 +1,75 @@
...
+            auto hashValue = hashFunc(name);
+            auto now = time::WallClock::fastNowInMicroSec();
+            vId = (hashValue & hashMask) | (now & timeMask);
+            val.append(reinterpret_cast<char*>(&vId), sizeof(VertexID));
+            std::vector<kvstore::KV> data;
+            data.emplace_back(key, val);
```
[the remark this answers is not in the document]
critical27: Good catch.

src/storage/GetUUIDProcessor.h
critical27: 👍

src/storage/GetUUIDProcessor.h
```diff
@@ -0,0 +1,76 @@
...
+        auto key = NebulaKeyUtils::uuidKey(partId, name.c_str());
+        std::string val;
+        VertexID vId;
+        auto ret = kvstore_->get(spaceId, partId, key, &val);
+        // try to get the corresponding vertex id
+        if (ret != kvstore::ResultCode::SUCCEEDED) {
+            // need to generate new vertex id of this uuid
+            MurmurHash2 hashFunc;
+            auto hashValue = hashFunc(name);
+            auto now = time::WallClock::fastNowInMicroSec();
```
dangleptr: for int32 we’d better use seconds

src/common/base/NebulaKeyUtils.cpp
```diff
@@ -68,6 +68,16 @@ std::string NebulaKeyUtils::systemPartKey(PartitionID partId) {
     return key;
 }
 
+// static
+std::string NebulaKeyUtils::uuidKey(PartitionID partId, const folly::StringPiece& name) {
+    std::string key;
+    key.reserve(sizeof(PartitionID) + name.size());
+    int32_t item = (partId << 8) | (NebulaKeyType::kUUID);
+    key.append(reinterpret_cast<const char*>(&item), sizeof(int32_t))
+       .append(name.str());
```
dangleptr: Extra copy here

src/storage/GetUUIDProcessor.h
critical27: If we use second, I suppose we would have high conflict rate? Cuz If we generate many uuid in a partition at the exact same second, the time will be all the same.

src/common/base/NebulaKeyUtils.cpp
critical27: Good catch

src/storage/GetUUIDProcessor.h
dangleptr: we can’t use int32 to represent micro seconds, it has more conflicts

critical27: Ok

PR #961, opened 2019-09-20
Support UUID

darionyaphet:


src/storage/client/StorageClient.cpp
```diff
@@ -377,5 +377,11 @@ PartitionID StorageClient::partId(GraphSpaceID spaceId, int64_t id) const {
     return s;
 }
 
+int64_t StorageClient::getUUID(GraphSpaceID space, const std::string& field) {
+    LOG(INFO) << "Called GetUUID";
+    UNUSED(space);
+    return static_cast<int64_t>(std::hash<std::string>()(field));;
```
CPWstatic: Is this a temporary solution?

darionyaphet: yep this is rely on [Support uuid in storage service](https://github.com/vesoft-inc/nebula/pull/958)

darionyaphet:
#958

src/common/filter/Expressions.cpp
```diff
@@ -619,6 +621,19 @@ const char* FunctionCallExpression::decode(const char *pos, const char *end) {
...
 }
 
+std::string UUIDExpression::toString() const {
+    return folly::stringPrintf("uuid(%s)", field_->c_str());
+}
+
+OptVariantType UUIDExpression::eval() const {
+     auto client = context_->storageClient();
+     auto space = context_->space();
+     return client->getUUID(space, *field_);
```
critical27: The interface in StorageClient is async, maybe you need to get future here.

darionyaphet: Sure！ `StorageClient`  will be update when #958 merged :)

src/graph/InsertEdgeExecutor.cpp
```diff
@@ -68,11 +71,18 @@ Status InsertEdgeExecutor::check() {
 
 
 StatusOr<std::vector<storage::cpp2::Edge>> InsertEdgeExecutor::prepareEdges() {
+    expCtx_ = std::make_unique<ExpressionContext>();
```
wadeliuyi: the expCtx_ is make in prepare, why need remake the expCtx_ here?

src/common/filter/Expressions.cpp
```diff
@@ -619,6 +621,19 @@ const char* FunctionCallExpression::decode(const char *pos, const char *end) {
     return pos;
 }
 
+std::string UUIDExpression::toString() const {
```
dangleptr: Why we need UUIDExpression?  Why not implement it in FunctionManager?

dangleptr:
@CPWstatic Do you talked about it with @darionyaphet 

I am not sure why not make it a special UDF instead of an expression.

src/common/filter/Expressions.cpp
darionyaphet: Because when `uuid` is called, some context should be referenced such as space ID and storage client.  FunctionManager is a singleton use Expression maybe better.

src/common/filter/Expressions.cpp
dutor: Noted.

src/common/filter/Expressions.cpp
```diff
@@ -619,6 +621,24 @@ const char* FunctionCallExpression::decode(const char *pos, const char *end) {
...
+    return folly::stringPrintf("uuid(%s)", field_->c_str());
+}
+
+OptVariantType UUIDExpression::eval() const {
+     auto client = context_->storageClient();
+     auto space = context_->space();
+     auto uuidResult = client->getUUID(space, *field_).get();
```
dangleptr: Please talk about the sync way with @dutor ,  is it ok in Expression?

dutor: @dangleptr IMO, such a way to do evaluation asynchronously is more generic in function call.

dutor: BTW. From discussion above, @darionyaphet will introduce the async way in another PR.

PR #962, opened 2019-09-20
Support fetch test case in parser

darionyaphet:


src/parser/test/ParserTest.cpp
```diff
@@ -715,6 +715,133 @@ TEST(Parser, DeleteEdge) {
...
+        GQLParser parser;
+        std::string query = "$var = GO FROM 1 over e1; "
+                            "FETCH PROP ON person $var.id YIELD person.name, person.age";
+        auto result = parser.parse(query);
+        ASSERT_TRUE(result.ok()) << result.status();
+    }
+    {
+        GQLParser parser;
+        std::string query = "FETCH PROP ON player 1,2,3 "
+                            "YIELD DISTINCT person.name, person.age";
```
CPWstatic: There is no syntax error, but the semantics are wrong. I suggest change the tag of person to player.

src/parser/test/ParserTest.cpp
```diff
@@ -715,6 +715,133 @@ TEST(Parser, DeleteEdge) {
...
+                            "YIELD transfer._src AS s, serve._dst AS d | "
+                            "FETCH PROP ON transfer $-.s -> $-.d YIELD transfer.amount";
+        auto result = parser.parse(query);
+        ASSERT_TRUE(result.ok()) << result.status();
+    }
+    {
+        GQLParser parser;
+        std::string query = "$var = GO FROM 12345 OVER transfer "
+                            "YIELD transfer._src AS s, edu._dst AS d; "
+                            "FETCH PROP ON serve $var.s -> $var.d YIELD transfer.amount";
```
CPWstatic: Same problem as mentioned above.

darionyaphet: This case is refer to [fetch-syntax](https://github.com/vesoft-inc/nebula/blob/master/docs/manual-EN/statement-syntax/data-manipulation-statements/fetch-syntax.md).  But no mistakes were found.

src/parser/test/ParserTest.cpp
darionyaphet: Nice

src/parser/test/ParserTest.cpp
CPWstatic: I mean you are about to fetch prop on serve while yield from transfer.

PR #964, opened 2019-09-23
Do implicit type casting in relational expression.

### LINKED ISSUE #530: logic && bug in where clause
nebula> GO FROM 100 OVER like WHERE likeness >= 0 && likeness <= 100;                                                                                                                                              

======                                                                                                                                                                                                             | id |                                                                                                                                                                                                             ======                                                                                                                                                                                                             Got 0 rows (Time spent: 1512/2474 us)          

BUT: 
The likeness value is 90.02 : 

nebula> GO FROM 100 OVER like yield like.likeness                                                                                                                                                                  =================                                                                                                                                                                                                  | like.likeness |                                                                                                                                                                                                  =================                                                                                                                                                                                                  |     90.020000 |                                                                                                                                                                                                  -----------------                                                                                                                                                                                                  Got 1 rows (Time spent: 1296/1971 us)    


and try || , the result is correct.

CPWstatic:
Close #530.
The **boost::variant** will not compare the content if the type do not match. So we should do implicit type casting.

### DIFF
--- a/src/common/filter/Expressions.cpp
+++ b/src/common/filter/Expressions.cpp
@@ -933,6 +933,13 @@ OptVariantType RelationalExpression::eval() const {
 
     auto l = left.value();
     auto r = right.value();
+    if (l.which() != r.which()) {
+        auto s = implicitCasting(l, r);
+        if (!s.ok()) {
+            return s;
+        }
+    }
+
     switch (op_) {
         case LT:
             return OptVariantType(l < r);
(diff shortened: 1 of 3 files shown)


src/common/filter/Expressions.cpp
```diff
@@ -933,6 +933,13 @@ OptVariantType RelationalExpression::eval() const {
 
     auto l = left.value();
     auto r = right.value();
+    if (l.which() != r.which()) {
+        auto s = implicitCasting(l, r);
+        if (!s.ok()) {
+            return std::move(s);
```
dutor: FYI. In principle, there is no need for `std::move` here.

CPWstatic:
Jenkins go!

PR #966, opened 2019-09-23
Added exception handling for  folly::to

### LINKED ISSUE #833: Exception handling in folly::to
I found that folly::to function may throw an exception.

For example, in function FileBasedWal::scanAllWalFiles:
   ```
      try {
             startIdFromName = folly::to<int64_t>(parts[0]);
         } catch (const std::exception& ex) {
             LOG(ERROR) << "Ignore bad file name \"" << fn << "\"";
             continue;
         }
```

But in many places, the exception thrown by the following::to is not processed， for example：in  function HBaseStore::decode

Unhandled exception  will cause server crash.

### PR DESCRIPTION
**Pull request by darionyaphet**
closes #833

### CODE REVIEW

File: src/kvstore/NebulaStore.cpp [hunk 1]
```diff
@@ -57,7 +57,14 @@ bool NebulaStore::init() {
             for (auto& dir : dirs) {
                 LOG(INFO) << "Scan path \"" << path << "/" << dir << "\"";
                 try {
-                    auto spaceId = folly::to<GraphSpaceID>(dir);
+                    GraphSpaceID spaceId;
+                    try {
+                        spaceId = folly::to<GraphSpaceID>(dir);
+                    } catch (const std::exception& ex) {
+                        LOG(ERROR) << "Data path invalid: " << dir;
```
dutor: Maybe we should log out the verbose info of the exception.

src/kvstore/plugins/hbase/HBaseStore.cpp
```diff
@@ -80,28 +80,33 @@ std::string HBaseStore::encode(GraphSpaceID spaceId,
...
+                    break;
+                case cpp2::SupportedType::BOOL:
+                    values.emplace_back(folly::to<bool>(value));
+                    break;
+                default:
+                    LOG(ERROR) << "Type Error : " << value;
+                    break;
+            }
+        } catch (const std::exception& ex) {
+            LOG(ERROR) << "encoding failed: " << fieldName << ":" << value;
```
dutor: ditto.

yixinglu:
Jenkins go

Issue #910, opened 2019-09-09
Package IMPORTER assembled with its dependencies. 


dangleptr:
IMPORTER should be an independent tool.

PR #969, opened 2019-09-23
Support IMPORTER assembled with its dependencies

darionyaphet:
#910 Usage case: 

```
java -jar target/importer-1.0.0-beta.jar args
```


jude-zhu:
Jenkins go

darionyaphet:
thanks ~

PR #975, opened 2019-09-25
Try to introduce concurrency to unit testing

dutor:
This patch reduces the chances of listen port conflicts, so that we can run unit tests concurrently, i.e `ctest --output-on-failure -jN`. By doing this, the duration of unit testing could be reduced to 2 minutes from 12.

### DIFF
--- a/src/common/network/NetworkUtils.cpp
+++ b/src/common/network/NetworkUtils.cpp
@@ -114,13 +114,56 @@ bool NetworkUtils::getDynamicPortRange(uint16_t& low, uint16_t& high) {
 
 std::unordered_set<uint16_t> NetworkUtils::getPortsInUse() {
     static const std::regex regex("[^:]+:[^:]+:([0-9A-F]+).+");
-    fs::FileUtils::FileLineIterator iter("/proc/net/tcp", &regex);
     std::unordered_set<uint16_t> inUse;
-    while (iter.valid()) {
-        auto &sm = iter.matched();
-        inUse.emplace(std::stoul(sm[1].str(), NULL, 16));
-        ++iter;
+    {
+        fs::FileUtils::FileLineIterator iter("/proc/net/tcp", &regex);
+        while (iter.valid()) {
+            auto &sm = iter.matched();
+            inUse.emplace(std::stoul(sm[1].str(), NULL, 16));
+            ++iter;
+        }
     }
+    {
+        fs::FileUtils::FileLineIterator iter("/proc/net/tcp6", &regex);
+        while (iter.valid()) {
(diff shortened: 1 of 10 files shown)


src/common/stats/test/StatsManagerCrossLevelTest.cpp
```diff
@@ -0,0 +1,53 @@
...
+
+TEST(StatsManager, CrossLevelTest) {
+    auto statId = StatsManager::registerHisto("stat03", 1, 1, 100);
+    std::vector<std::thread> threads;
+    for (int i = 0; i < 10; i++) {
+        threads.emplace_back([statId, i] () {
+            for (int k = i * 10 + 10; k >= i * 10 + 1; k--) {
+                StatsManager::addValue(statId, k);
+                if (k > i * 10 + 1) {
+                    sleep(7);
```
laura-ding: why sleep 7s?  this test will cost more than 1 min

dutor: Yes, it's costly. But the durations are predefined, i.e. 60, 600, etc. Here for the following assertions, we are testing on the minimal duration of 60s.

src/common/network/NetworkUtils.cpp
```diff
@@ -114,13 +114,56 @@ bool NetworkUtils::getDynamicPortRange(uint16_t& low, uint16_t& high) {
```
yixinglu: Could you update [`.github/workflows/ci.yaml`](https://github.com/vesoft-inc/nebula/blob/master/.github/workflows/ci.yaml#L32) and add `-j$(nproc)` option to `ctest --output-on-failure` command

src/common/network/NetworkUtils.cpp
```diff
@@ -135,9 +178,15 @@ uint16_t NetworkUtils::getAvailablePort() {
 
     std::unordered_set<uint16_t> portsInUse = getPortsInUse();
     uint16_t port = 0;
-    do {
-        port = folly::Random::rand32(low, static_cast<int32_t>(high) + 1);
-    } while (portsInUse.find(port) != portsInUse.end());
+    while (true) {
+        port = folly::Random::rand32(1025, low);
```
yixinglu: Need some comments. Why to select a port from range [1025, low]?

src/common/network/NetworkUtils.cpp
dutor: sure

src/common/network/NetworkUtils.cpp
dutor: Yeah, it should have been.

PR #977, opened 2019-09-26
Support general storage in StorageService

darionyaphet:
This PR will support put and get in storage engine.


src/storage/GetProcessor.cpp
```diff
@@ -0,0 +1,68 @@
...
+    folly::Promise<std::pair<PartitionID, kvstore::ResultCode>> promise;
+    auto future = promise.getFuture();
+
+    executor_->add([this, p = std::move(promise), part, keys] () mutable {
+        std::vector<std::string> values;
+        auto ret = this->kvstore_->multiGet(space_, part, keys, &values);
+        if (ret == kvstore::ResultCode::SUCCEEDED) {
+            std::lock_guard<std::mutex> lg(this->lock_);
+            for (int32_t i = 0; i < static_cast<int32_t>(keys.size()); i++) {
+                pairs_.emplace(keys[i], values[i]);
```
critical27: We can use move here

src/storage/PutProcessor.cpp
```diff
@@ -0,0 +1,29 @@
...
+
+void PutProcessor::process(const cpp2::PutRequest& req) {
+    const auto& pairs = req.get_parts();
+    auto space = req.get_space_id();
+    callingNum_ = pairs.size();
+    CHECK_NOTNULL(kvstore_);
+
+    std::for_each(pairs.begin(), pairs.end(), [&](auto& value) {
+        auto part = value.first;
+        std::vector<kvstore::KV> data;
```
critical27: Maybe we can collect all pairs and put only once

darionyaphet: Seems `doPut` would process each partition at one time.

I will use folly::collectAll().via().then() to process put in parallel. Similar with GetProcessor

src/daemons/BenchmarkDaemon.cpp
```diff
@@ -0,0 +1,120 @@
+/* Copyright (c) 2019 vesoft inc. All rights reserved.
```
dangleptr: Let's move it into tools dir,  and use another name "SimpleKVVerifyTool"

darionyaphet: Should I move it as test case or benchmark ?

dangleptr: No,  it is not a UT or a benchmark

zhangguoqing:
close #973

PR #983, opened 2019-09-26
test partition performance for index

boshengchen:


bright-starry-sky:
I don't think this test needs to be added to the regression test. Because it runs too long, it will affect other tests.

dangleptr:
Don't close it.  The perf is valuable,  add it into our repo,  of course we don't need to run it as UT

src/kvstore/test/CMakeLists.txt
```diff
@@ -23,6 +23,13 @@ nebula_add_test(
     LIBRARIES ${THRIFT_LIBRARIES} ${ROCKSDB_LIBRARIES} wangle gtest
 )
 
+nebula_add_test(
```
dangleptr: Use nebula_add_executable, we don't need to add it as a UT

bright-starry-sky:
let me review it . I will open a new PR if necessary.

bright-starry-sky:
Close this one. We will tracking it via #1807

PR #984, opened 2019-09-26
Support python client

darionyaphet:


src/client/py/nebula/client.py
```diff
@@ -0,0 +1,61 @@
+#!/usr/bin/env python
+
+# Copyright (c) 2018 vesoft inc. All rights reserved.
```
zhangguoqing: 2019 yet?

src/client/py/nebula/client.py
```diff
@@ -0,0 +1,61 @@
...
+from thrift import Thrift
+from thrift.transport import TSocket
+from thrift.transport import TTransport
+from thrift.protocol import TBinaryProtocol
+
+from graph import GraphService
+
+class AuthException(Exception):
+    def __init__(message):
+        Exception.__init__ (self, message)
```
zhangguoqing: How about ```Exception.__init__(self, message)```?   Why leave a blank space?

src/client/py/nebula/client.py
```diff
@@ -0,0 +1,61 @@
...
+        self.protocol = TBinaryProtocol.TBinaryProtocol(self.transport)
+        self.transport.open()
+        self.client = GraphService.Client(self.protocol)
+
+    def authenticate(self, user, password):
+        authResponse = self.client.authenticate(user, password)
+        if authResponse.error_code == 0:
+            return authResponse.session_id
+        else:
+            raise AuthException("Auth failed")
```
zhangguoqing: How about below?
```
       if authResponse.error_code:
            raise AuthException("Auth failed")
        else:
            return authResponse.session_id
```

src/client/py/nebula/client.py
```diff
@@ -0,0 +1,61 @@
...
+
+
+client = GraphClient("127.0.0.1", 13699)
+session = client.authenticate("user", "password")
+print "Session: " + str(session)
+
+executionResponse = client.execute(session, "show spaces")
+print executionResponse
+
+client.signout(session)
```
zhangguoqing: Why not move them to the tests.py?

src/interface/meta.thrift
```diff
@@ -7,6 +7,7 @@
 namespace cpp nebula.meta
 namespace java com.vesoft.nebula.meta
 namespace go nebula.meta
+namespace py nebula.meta
```
zhangguoqing: I'm not sure this line is necessary?

darionyaphet: I will move it into example later :)

src/client/py/nebula/client.py
darionyaphet: I will move it into example later :)

yixinglu:
Move this client to the repo [vesoft-inc/nebula-python](https://github.com/vesoft-inc/nebula-python)

PR #1001, opened 2019-09-29
Format before commit and correct cron job schedule time

yixinglu:
- Correct cron job UTC schedule time
- Use `clang-format` to format changes file list before commit
- Ignore files in `.gitignore` for docker
- Change threads of packaging

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -2,4 +2,32 @@
 
 LICENSES/
 docs/
-_build/
+
+*.swp
+*.so
+*.DS_Store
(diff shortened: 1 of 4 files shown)


yixinglu:
> I noticed that you enabled `clang-format` to do atomatic formating. I have researched and tried this tool recently. Generally, that's an excellent formating tool. But even if after some tweaking, there are still some minor points that need further discussion before enabling it. Let's talk about this offline.

OK,

PR #1003, opened 2019-09-29
Support uuid in async mode

darionyaphet:


src/common/filter/Expressions.cpp
```diff
@@ -628,12 +628,24 @@ std::string UUIDExpression::toString() const {
...
+         }
+         ret = OptVariantType(resp.value().get_id());
+     };
+
+     auto error = [&ret] (auto &&e) {
+         LOG(ERROR) << "Exception caught: " << e.what();
+         ret = OptVariantType(Status::Error("Get UUID Failed"));
+     };
+
+     std::move(future).via(context_->executor()).thenValue(cb).thenError(error).wait();
```
dutor: This is not the way to do async. Actually, by performing `wait`, you are blocking the calling thread.

Issue #844, opened 2019-08-26
Add `meta_server_addrs` option to default meta configuration file


yixinglu:
Now in [`etc/nebula-metad.conf.default`](https://github.com/vesoft-inc/nebula/blob/master/etc/nebula-metad.conf.default) default configuration file, miss the important option `meta_server_addrs` which configures how many peers this metad service has. The option default value is `""`.

laura-ding:
has complete it, so close it.

PR #1005, opened 2019-09-29
Automatically update the java version

laura-ding:
- When release version, the script will uniformly modify the version of java.
- You can change the java version by this

```
cmake -DNEBULA_BUILD_VERSION=${version}  ${other_args} ..
```


etc/nebula-metad.conf.default
```diff
@@ -15,6 +15,9 @@
 --logbufsecs=0
 
 ########## networking ##########
+# It is a list of IPs split by comma, the ips number equals replica number. If empty, it means replica is 1
+# The format looks like ip1, ip2, ip3
```
yixinglu: ip:port

laura-ding: good catch

yixinglu:
reference #844

laura-ding:
jenkins go

etc/nebula-metad.conf.default
```diff
@@ -15,6 +15,8 @@
 --logbufsecs=0
 
 ########## networking ##########
+# List of meta server addresses, the format looks like ip1:port1, ip2:port2, ip3:port3
```
zhangguoqing: It's good to add the 'meta_server_addrs' configure item to the nebula-metad.conf.

yixinglu:
`meta_server_addrs` option has been added to `nebula-meta.conf` in #1111

PR #1016, opened 2019-10-08
Support error code in NGQLException

darionyaphet:


src/client/java/src/main/java/com/vesoft/nebula/graph/client/NGQLException.java
```diff
@@ -7,6 +7,7 @@
 package com.vesoft.nebula.graph.client;
 
 public class NGQLException extends Exception {
+    private int code;
```
zlcook: add `getCode` method to access `code`.

src/client/java/src/main/java/com/vesoft/nebula/graph/client/NGQLException.java
```diff
@@ -15,5 +16,10 @@ public NGQLException() {
     public NGQLException(String message) {
         super(message);
     }
```
zlcook: `SQL Exception(String message)` and `SQLException()` is unnecessary.

jude-zhu:
Jenkins go

PR #1023, opened 2019-10-09
Fixed memory leak in parser

monadbobo:
fixed the memory leak.


src/parser/parser.yy
```diff
@@ -1188,6 +1188,7 @@ update_item
     }
     | alias_ref_expression ASSIGN expression {
         $$ = new UpdateItem($1, $3);
+        delete $1;
```
dutor: What the hell in `UpdateItem::UpdateItem(Expression*, Expression*)`…………

PR #1025, opened 2019-10-09
Fix thread block when error returned in MetaClient

### LINKED ISSUE #1017: [Critical] Methods in MetaClient block for seconds on errors
1. There are retry logics in RPC involved methods of MetaClient, i.e. `getResponse(...)`
  2. In order to retry, the current scheme utilizes `::sleep`, in a sync way, which will block a networking thread which is typically the current calling thread of the async callback.

To reproduce this issue, one could try to execute some statement like `DESCRIBE TAG nonexistent_tag_name`.

critical27:
Remove sleep when error returned in MetaClient retry logic.

Thanks for @dutor's help.

close #1017


dutor:
> Typically, we should wait for some time before retry.
> You could use AsyncTimeout in folly to avoid sleep.

IMO. as for retrying, we should classify the causes of failures. For example, it's not appropriate to retry for `DESCRIBE TAG typo_tag_name`

src/meta/client/MetaClient.cpp
```diff
@@ -286,6 +286,31 @@ MetaClient::reverse(const PartsAlloc& parts) {
...
+         typename RpcResponse,
+         typename Response>
+void MetaClient::retry(Request req,
+                       RemoteFunc remoteFunc,
+                       RespGenerator respGen,
+                       folly::Promise<StatusOr<Response>> pro,
+                       bool toLeader,
+                       int32_t retry,
+                       int32_t retryLimit) {
+    folly::EventBase retryEvb;
```
dangleptr: Why not reuse current evb inside ioThreadPool?

src/meta/client/MetaClient.cpp
```diff
@@ -286,6 +286,31 @@ MetaClient::reverse(const PartsAlloc& parts) {
...
+    auto observer = folly::AsyncTimeout::make(retryEvb,
+            [req = std::move(req), remoteFunc = std::move(remoteFunc),
+             respGen = std::move(respGen), pro = std::move(pro), toLeader, retry,
+             retryLimit, this] () mutable noexcept {
+        getResponse(std::move(req), std::move(remoteFunc), std::move(respGen),
+                    std::move(pro), toLeader, retry + 1, retryLimit);
+    });
+    observer->scheduleTimeout(
+            std::chrono::milliseconds(FLAGS_meta_client_retry_interval_secs * 1000));
+    retryEvb.loop();
```
dangleptr: Is it still blocking the thread?

src/meta/client/MetaClient.cpp
```diff
@@ -321,35 +346,34 @@ void MetaClient::getResponse(Request req,
                     updateActive();
                 }
                 if (retry < retryLimit) {
-                    sleep(FLAGS_meta_client_retry_interval_secs);
-                    getResponse(std::move(req), std::move(remoteFunc), std::move(respGen),
-                                std::move(p), toLeader, retry + 1, retryLimit);
+                    retryResponse(std::move(req), std::move(remoteFunc), std::move(respGen),
```
dangleptr: Where is retryResponse definition?

src/meta/client/MetaClient.cpp
dutor: Seems so.

src/meta/client/MetaClient.cpp
critical27: My bad, didn't push the commit, fixed.

src/meta/client/MetaClient.cpp
critical27: Yeah, it would block the thread, but it is a stack allocated evb, it won't block our io thread pool

src/meta/client/MetaClient.cpp
critical27: I've tried before. Quoted from folly:
>   Your code just tried to loop over an event base from inside another event base loop. Since libevent is not reentrant, this leads to undefined behavior in opt builds. Please fix immediately. For the common case of an inner function that needs to do some synchronous computation on an event-base, replace getEventBase() by a new, stack-allocated EvenBase.

dutor: > I've tried before. Quoted from folly:
> 
> > Your code just tried to loop over an event base from inside another event base loop. Since libevent is not reentrant, this leads to undefined behavior in opt builds. Please fix immediately. For the common case of an inner function that needs to do some synchronous computation on an event-base, replace getEventBase() by a new, stack-allocated EvenBase.

Actually, by reusing the present evb, you dont loop by yourself, you are just to add a timer event to the current event loop.

critical27: > Actually, by reusing the present evb, you dont loop by yourself, you are just to add a timer event to the current event loop.

Yeah, but when I reuse the present evb (aka evb from ioThreadPool), the retry timer event was never fired, it really confuses me.

dangleptr: You could write the code in this way 
```
evb->runAfterDelay([req = std::move(req),
                                    remoteFunc = std::move(remoteFunc),
                                    respGen = std::move(respGen),
                                    pro = std::move(pro),
                                    toLeader,
                                    retry,
                                    retryLimit,
                                    this]) mutable {
      getResponse(std::move(req), 
                             std::move(remoteFunc),
                             std::move(respGen),
                             std::move(pro),
                             toLeader,
                             retry + 1,
                             retryLimit);
},  FLAGS_meta_client_retry_interval_secs * 1000);
```

src/meta/client/MetaClient.cpp
```diff
@@ -321,35 +321,52 @@ void MetaClient::getResponse(Request req,
...
-                p.setValue(respGen(std::move(resp)));
+                pro.setValue(respGen(std::move(resp)));
                 return;
             } else if (resp.code == cpp2::ErrorCode::E_LEADER_CHANGED) {
                 HostAddr leader(resp.get_leader().get_ip(), resp.get_leader().get_port());
                 {
                     folly::RWSpinLock::WriteHolder holder(hostLock_);
                     leader_ = leader;
                 }
+                if (retry < retryLimit) {
```
dangleptr: Why move the retry logic code in the "else if" scope. 
All failures should be retry.

dutor: I can hardly agree. For example, it's not appropriate to retry on `DESCRIBE TAG typo_tag_name`

dangleptr: make sense

critical27: I think the other failures should be handled in upper stream, for example, `create space existed`, it seems no necessary to retry.

dangleptr: make sense

PR #1033, opened 2019-10-10
[doc]Updating openssl in CentOS 7.0

wfystx:
Summary:
```[ERROR]libcrypto.so.10(OPENSSL_1.0.2)(64bit) is needed by vs-cmake-3.11.4-2.el7.centos.x86_64```
Should upgrade openssl from 1.0.1 to 1.0.2

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,3 +26,4 @@ Testing/
 target/
 cluster.id
 pids/
+.idea/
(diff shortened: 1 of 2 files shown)


laura-ding:
Thanks very much. Centos7.0 is not yet supported, and this step does not support compilation. I recommend not writing this.

PR #1035, opened 2019-10-10
fixed the memory leak.

monadbobo:


src/storage/test/UpdateEdgeTest.cpp
```diff
@@ -145,10 +145,12 @@ TEST(UpdateEdgeTest, Set_Filter_Yield_Test) {
     decltype(req.return_columns) tmpColumns;
     tmpColumns.emplace_back(Expression::encode(val1));
     tmpColumns.emplace_back(Expression::encode(val2));
-    std::string name = folly::to<std::string>(3002);
-    std::string prop = "tag_3002_col_2";
-    auto* sourcePropExp = new SourcePropertyExpression(&name, &prop);
+    delete val1;
+    delete val2;
```
CPWstatic: Actually, I don't think we need to create object on heap.

dutor: > Actually, I don't think we need to create object on heap.

I second this. Most of these `delete`s are not supposed to be there.

PR #1039, opened 2019-10-11
Fixed an undefined behavior.

monadbobo:
1. Fixed an undefined behavior.
2. Replace static_cast with dynamic_cast.


src/graph/FetchExecutor.cpp
```diff
@@ -34,11 +34,11 @@ Status FetchExecutor::prepareYield() {
         // such as YIELD 1+1, it has not type in schema, the type from the eval()
         colTypes_.emplace_back(nebula::cpp2::SupportedType::UNKNOWN);
         if (col->expr()->isAliasExpression()) {
-            colNames_.emplace_back(*static_cast<InputPropertyExpression*>(col->expr())->prop());
+            colNames_.emplace_back(*dynamic_cast<AliasPropertyExpression*>(col->expr())->prop());
             continue;
         } else if (col->expr()->isTypeCastingExpression()) {
             // type cast
-            auto exprPtr = static_cast<TypeCastingExpression*>(col->expr());
+            auto exprPtr = dynamic_cast<TypeCastingExpression*>(col->expr());
```
wadeliuyi: dynamic_cast will return nullptr to exprPtr, so need check it?

monadbobo: I will fix this.

PR #1043, opened 2019-10-11
fix a bug in DeleteVertexExecutor

zlcook:
* before: deleting a vertex without edges will cause the following exception
```
#0  0x0000000001a28d47 in nebula::storage::StorageRpcResponse<nebula::storage::cpp2::ExecResponse>::completeness (this=0x7f390db69148) at /export/nebula/src/storage/client/StorageClient.h:53
53	        return (totalReqsSent_ - failedReqs_) * 100 / totalReqsSent_;
```


jude-zhu:
Jenkins go

PR #1046, opened 2019-10-11
Support comparison function in().

### LINKED ISSUE #1040: support \in in the WHERE clause
like SQL IN, nGQL should support IN such as:

    WHERE column_name IN (value1,value2,...);

The Set ops (Intersect, Union, Minus) can't work in this condition.

CPWstatic:
I notice that _**in()**_ is a comparison function in Mysql. So I learn from the design and implement it with a function.
The syntax would looks like:
```
YIELD 1 in(1,2,3);
```

Or you may reference it in filter:
```
WHERE 1 in(1,2,3)
```

Definitely, you can reference the input , src and dest ref:
```
WHERE $-.id in(1,2,3)
WHERE $$.team.name in("Spurs","Suns")
WHERE $^.player.name in("Tim Duncan", "Tony Parker")
```

close #1040

### DIFF
--- a/src/graph/test/GoTest.cpp
+++ b/src/graph/test/GoTest.cpp
@@ -626,5 +626,69 @@ TEST_F(GoTest, NotExistTagProp) {
         ASSERT_NE(cpp2::ErrorCode::SUCCEEDED, code);
     }
 }
+
+TEST_F(GoTest, inCall) {
+    {
+        cpp2::ExecutionResponse resp;
+        auto &player = players_["Boris Diaw"];
+        auto *fmt = "GO FROM %ld OVER serve "
+                    "WHERE $$.team.name in (\"Hawks\", \"Suns\") "
+                    "YIELD $^.player.name, serve.start_year, serve.end_year, $$.team.name";
+        auto query = folly::stringPrintf(fmt, player.vid());
+        auto code = client_->execute(query, resp);
+        ASSERT_EQ(cpp2::ErrorCode::SUCCEEDED, code) << *resp.get_error_msg();
+
+        std::vector<std::string> expectedColNames{
+            {"$^.player.name"}, {"serve.start_year"}, {"serve.end_year"}, {"$$.team.name"}
+        };
+        ASSERT_TRUE(verifyColNames(resp, expectedColNames));
+
+        std::vector<std::tuple<std::string, int64_t, int64_t, std::string>> expected = {
+            {player.name(), 2003, 2005, "Hawks"},
+            {player.name(), 2005, 2008, "Suns"},
+        };
+        ASSERT_TRUE(verifyResult(resp, expected));
+    }
+    {
+        cpp2::ExecutionResponse resp;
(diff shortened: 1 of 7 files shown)


src/graph/test/GoTest.cpp
```diff
@@ -626,5 +626,69 @@ TEST_F(GoTest, NotExistTagProp) {
...
     }
 }
+
+TEST_F(GoTest, inCall) {
+    {
+        cpp2::ExecutionResponse resp;
+        auto &player = players_["Boris Diaw"];
+        auto *fmt = "GO FROM %ld OVER serve "
+                    "WHERE $$.team.name in (\"Hawks\", \"Suns\") "
+                    "YIELD $^.player.name, serve.start_year, serve.end_year, $$.team.name";
```
whitewum: Great

src/graph/test/GoTest.cpp
```diff
@@ -626,5 +626,69 @@ TEST_F(GoTest, NotExistTagProp) {
...
+        std::vector<std::tuple<std::string, int64_t, int64_t, std::string>> expected = {
+            {player.name(), 2003, 2005, "Hawks"},
+            {player.name(), 2005, 2008, "Suns"},
+        };
+        ASSERT_TRUE(verifyResult(resp, expected));
+    }
+    {
+        cpp2::ExecutionResponse resp;
+        auto *fmt = "GO FROM %ld OVER like YIELD like._dst AS id"
+                    "| GO FROM $-.id OVER serve WHERE $-.id in(%ld, 123)";
```
whitewum: Great

src/graph/test/GoTest.cpp
```diff
@@ -626,5 +626,69 @@ TEST_F(GoTest, NotExistTagProp) {
...
+            {"serve._dst"}
+        };
+        ASSERT_TRUE(verifyColNames(resp, expectedColNames));
+
+        std::vector<std::tuple<int64_t>> expected = {
+            {teams_["Spurs"].vid()},
+            {teams_["Hornets"].vid()},
+        };
+        ASSERT_TRUE(verifyResult(resp, expected));
+    }
```
yixinglu: Add failed UT for only one parameter in `IN()` function

src/graph/test/GoTest.cpp
```diff
@@ -626,5 +626,69 @@ TEST_F(GoTest, NotExistTagProp) {
         ASSERT_NE(cpp2::ErrorCode::SUCCEEDED, code);
     }
 }
+
+TEST_F(GoTest, inCall) {
+    {
+        cpp2::ExecutionResponse resp;
+        auto &player = players_["Boris Diaw"];
+        auto *fmt = "GO FROM %ld OVER serve "
+                    "WHERE $$.team.name in (\"Hawks\", \"Suns\") "
```
yixinglu: Want `NOT IN` test case

yixinglu:
fix #1040

src/graph/test/GoTest.cpp
whitewum: Right.

PR #1047, opened 2019-10-11
another way of install without auto download packages

laura-ding:


yixinglu:
Shall we move the `build_deps.sh` to a new folder named `build` in project root directory ?

laura-ding:
@yixinglu I think it's  more intuitive under `nebula`

PR #1050, opened 2019-10-12
Added support for local commands in console

dutor:
All user inputs prefixed with ":" are considered as local commands, which won't be sent on wire to the remote server.

In this PR, I added:
  * `:exit`, `:quit` to quit from console
  * `:sh`, `:bash` to execute any command in a sub-shell
  * `:history` to display the input history

It is worth to mention that, with `:sh sleep N`, we are able to do schema manipulation(e.g. create tag) in a batch mode.

### DIFF
--- a/src/console/CmdProcessor.cpp
+++ b/src/console/CmdProcessor.cpp
@@ -5,8 +5,10 @@
  */
 
 #include "base/Base.h"
+#include <readline/history.h>
 #include "console/CmdProcessor.h"
 #include "time/Duration.h"
+#include "process/ProcessUtils.h"
 
 namespace nebula {
 namespace graph {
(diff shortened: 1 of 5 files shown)


whitewum:
to run local bash commands?

dutor:
> to run local bash commands?

Yes.

src/console/CmdProcessor.cpp
```diff
@@ -486,5 +479,77 @@ const std::string& CmdProcessor::getSpaceName() const {
...
+}
+
+
+bool CmdProcessor::LocalCommandProcessor::doExit(folly::StringPiece) const {
+    return false;
+}
+
+
+bool CmdProcessor::LocalCommandProcessor::doShell(folly::StringPiece args) const {
+    ::system(args.toString().c_str());
```
laura-ding: handle the return from system

dangleptr:
I think it is a good feature.  We could release it in 1.x.  
wdyt?  @sherman-the-tank @whitewum @jude-zhu

PR #1055, opened 2019-10-12
Added some new compiler Option(Warning flags).

monadbobo:
1. Added some new compiler Option(Warning flags).
2. fixed the compile warning.


monadbobo:
Jenkins Go!

dangleptr:
It's a good improvement.  I think we could take it in release 1.x.  
@monadbobo Could you rebase it on current master?

src/storage/StorageServer.cpp
```diff
@@ -59,9 +59,9 @@ std::unique_ptr<kvstore::KVStore> StorageServer::getStoreInstance() {
         }
         return nbStore;
     } else if (FLAGS_store_type == "hbase") {
-        LOG(FATAL) << "HBase store has not been implemented";
```
Shylock-Hg, line 62: Why change this ?

CPWstatic:
The compiling part had been refactored since nebula 2.0.

PR #1073, opened 2019-10-14
Fixed the crash that might occur when stopping the raft service.

monadbobo:


src/kvstore/raftex/Host.h
```diff
@@ -132,6 +127,8 @@ class Host final : public std::enable_shared_from_this<Host> {
 
     LogID committedLogId_{0};
     std::atomic_bool sendingSnapshot_{false};
+
+    thrift::ThriftClientManager<cpp2::RaftexServiceAsyncClient> connManager_;
```
dangleptr: We should share the same instance of ThriftClientManager between different hosts.

monadbobo: We may need a better way to fix this. So I will close this PR first.

PR #1090, opened 2019-10-16
[Urgent!!] Added support for GO REVERSELY

### LINKED ISSUE #812: Go supports reverse query.
Currently,  `Go` does not support query by inbound.  We need this feature in actual projects.

dutor:
close #812

### DIFF
--- a/src/graph/GoExecutor.cpp
+++ b/src/graph/GoExecutor.cpp
@@ -239,6 +239,12 @@ Status GoExecutor::prepareOver() {
         LOG(FATAL) << "Over clause shall never be null";
     }
 
+    isReversely_ = clause->isReversely();
+
+    if (isReversely()) {
+        edgeHolder_ = std::make_unique<EdgeHolder>();
+    }
+
     auto edges = clause->edges();
     for (auto e : edges) {
         if (e->isOverAll()) {
@@ -252,12 +258,10 @@ Status GoExecutor::prepareOver() {
             return edgeStatus.status();
         }
 
-        if (e->isReversely()) {
-            edgeTypes_.push_back(-edgeStatus.value());
-            return Status::Error("`REVERSELY' not supported yet");
-        }
-
         auto v = edgeStatus.value();
+        if (isReversely()) {
+            v = -v;
+        }
         edgeTypes_.push_back(v);
 
         if (e->alias() != nullptr) {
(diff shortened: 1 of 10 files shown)


src/graph/test/GoTest.cpp
```diff
@@ -690,5 +691,175 @@ TEST_F(GoTest, is_inCall) {
...
+            { players_["Tiago Splitter"].vid() },
+            { players_["Dejounte Murray"].vid() },
+            { players_["Shaquile O'Neal"].vid() },
+        };
+        ASSERT_TRUE(verifyResult(resp, expected));
+    }
+    {
+        cpp2::ExecutionResponse resp;
+        auto query = "GO FROM hash('Tim Duncan') OVER like REVERSELY "
+                      "YIELD $$.player.name";
```
CPWstatic: We reference _**$$**_ to indicate the vertex which can go to 'Tim Duncan' here, while we indicate the same thing by referencing _**_src**_ in last test case. Does this logic make sense to users?

yixinglu:
go-reversely-bug

yixinglu:
```text
(user@127.0.0.1) [wb]> GO FROM 1009 OVER contain REVERSELY YIELD contain._src AS db_vid | \
                    -> GO FROM $-.db_vid OVER contain WHERE contain._dst == 25085 OR contain._dst == 471803 YIELD contain._dst AS tbl_vid
[ERROR (-8)]: The specific vid column is not type of VID.
(user@127.0.0.1) [wb]> GO FROM 1009 OVER contain REVERSELY YIELD contain._src AS db_vid
==========
| db_vid |
==========
| 870    |
----------
Got 1 rows (Time spent: 5734/6491 us)
```

src/graph/GoExecutor.cpp
```diff
@@ -452,6 +447,135 @@ void GoExecutor::onStepOutResponse(RpcResponse &&rpcResp) {
...
 
 
+void GoExecutor::maybeFinishExecution(RpcResponse &&rpcResp) {
+    auto requireDstProps = expCtx_->hasDstTagProp();
+    auto requireEdgeProps = !expCtx_->aliasProps().empty();
+
+    // Non-reversely traversal, no properties required on destination nodes
+    // Or, Reversely traversal but no properties on edge and destination nodes required.
+    if ((!requireDstProps && !isReversely()) ||
+            (isReversely() && !requireDstProps && !requireEdgeProps)) {
```
dangleptr: For reversely traversal,   requireDstProps  maybe renamed as requireSrcProps. 
The variable name here makes me confused. 

PS.  we should add TODO here,  because the in-edges props will be stored with the dst Id in the future.

src/graph/GoExecutor.cpp
```diff
@@ -452,6 +447,135 @@ void GoExecutor::onStepOutResponse(RpcResponse &&rpcResp) {
...
+        finishExecution(std::move(rpcResp));
+        return;
+    }
+
+    auto dstids = getDstIdsFromResp(rpcResp);
```
dangleptr: Very confused.  In reversely traversal,  it is srcIds.

whitewum:
@dutor @dangleptr @CPWstatic @laura-ding @monadbobo 
please help with this pr.
It is really really urgent!!

src/graph/GoExecutor.cpp
```diff
@@ -872,21 +1004,27 @@ bool GoExecutor::processFinalResult(RpcResponse &rpcResp, Callback cb) const {
                         if (saveTypeFlag) {
                             colTypes.back() = iter->getSchema()->getFieldType(prop).type;
```
CPWstatic: We need save the property type when go reversely.

dutor:
Closed since this patch had been merged from with #1349

PR #1094, opened 2019-10-16
Adjusted default # of threads of graphd to # of CPU cores

dutor:
As title.


src/daemons/GraphDaemon.cpp
```diff
@@ -110,10 +110,24 @@ int main(int argc, char *argv[]) {
         return EXIT_FAILURE;
     }
 
+    if (FLAGS_num_netio_threads == 0) {
+        FLAGS_num_netio_threads = std::thread::hardware_concurrency();
+    }
     if (FLAGS_num_netio_threads <= 0) {
         LOG(WARNING) << "Number of networking IO threads should be greater than zero";
         return EXIT_FAILURE;
     }
```
whitewum: combine these two conditions ?

My thought: FLAGS_num_netio_threads <= 0 means all cores

src/daemons/GraphDaemon.cpp
```diff
@@ -110,10 +110,24 @@ int main(int argc, char *argv[]) {
...
+    LOG(INFO) << "Number of networking IO threads: " << FLAGS_num_netio_threads;
+
+    if (FLAGS_num_worker_threads == 0) {
+        FLAGS_num_worker_threads = std::thread::hardware_concurrency();
+    }
+    if (FLAGS_num_worker_threads <= 0) {
+        LOG(WARNING) << "Number of worker threads should be greater than zero";
+        return EXIT_FAILURE;
+    }
```
whitewum: ditto?

src/daemons/GraphDaemon.cpp
dutor: `0` is to be a legal option and the only option to indicate _number of cores_, while a negative value is not. The second check is to cover the negative AND the failure case(`hard_concurrency` has chances to fail)

etc/nebula-graphd.conf.default
```diff
@@ -36,10 +36,10 @@
 --session_idle_timeout_secs=60000
 # The number of threads to accept incoming connections
 --num_accept_threads=1
```
yixinglu: Miss this option?

src/graph/GraphFlags.cpp
```diff
@@ -16,7 +16,7 @@ DEFINE_int32(session_reclaim_interval_secs, 10, "Period we try to reclaim expire
 DEFINE_int32(num_netio_threads, 0,
                 "Number of networking threads, 0 for number of physical CPU cores");
 DEFINE_int32(num_accept_threads, 1, "Number of threads to accept incoming connections");
```
yixinglu: ditto

etc/nebula-graphd.conf.default
dutor: > Miss this option?

Nope. One single thread is enough to handle incomming connections.

PR #1095, opened 2019-10-16
[Raft] add election weight, improve snapshot

### LINKED ISSUE #1081: Leader can't be elected when logId and termId has conflicts.
We could fix it with different weight according to different logId and termId.

critical27:
1. Add election weight, close #1081
2. Reset invalid wal after snapshot received
3. Only clean outdate wal when the partition is not sending snapshot to its peer.
4. Fix crash in callback of snapshot


src/kvstore/raftex/RaftPart.cpp
```diff
@@ -1002,6 +1004,10 @@ typename RaftPart::Role RaftPart::processElectionResponses(
     for (auto& r : results) {
         if (r.second.get_error_code() == cpp2::ErrorCode::SUCCEEDED) {
             ++numSucceeded;
+        } else if (r.second.get_error_code() == cpp2::ErrorCode::E_LOG_STALE) {
+            VLOG(2) << idStr_ << "My last log id is less than " << hosts[r.first]
```
dangleptr: We could use LOG(INFO) here.

src/kvstore/raftex/RaftPart.cpp
```diff
@@ -1002,6 +1004,10 @@ typename RaftPart::Role RaftPart::processElectionResponses(
...
+                    << ", increase my election interval.";
+            ++weight_;
```
dangleptr: We could be more aggressive.  For example,  weight *= 2

src/kvstore/raftex/RaftPart.cpp
critical27: Good suggestion.

src/kvstore/raftex/RaftPart.cpp
critical27: Make sense.

critical27:
Jenkins go

src/kvstore/raftex/Host.cpp
```diff
@@ -190,6 +190,10 @@ void Host::appendLogsInternal(folly::EventBase* eb,
             {
                 std::lock_guard<std::mutex> g(self->lock_);
                 self->setResponse(r);
+                self->lastLogIdSent_++;
```
dangleptr: why NOT  self->lastLogIdSent_ =  self->logIdToSend_

src/kvstore/raftex/RaftPart.cpp
```diff
@@ -1169,6 +1178,15 @@ void RaftPart::cleanupSnapshot() {
     status_ = Status::RUNNING;
 }
 
+bool RaftPart::needToCleanWal() {
+    for (auto& host : hosts_) {
```
dangleptr: std::lock_guard<std::mutex> g(raftLock_);

critical27: Good catch

src/kvstore/raftex/Host.cpp
critical27: Resonable!

critical27:
Jenkins go

src/kvstore/raftex/RaftPart.cpp
```diff
@@ -1169,6 +1178,16 @@ void RaftPart::cleanupSnapshot() {
...
+        std::lock_guard<std::mutex> g(raftLock_);
```
dangleptr: hosts_ should be protected by raftLock_

critical27: My bad. Hold on please, there is another place to update.

src/kvstore/raftex/Host.cpp
```diff
@@ -401,14 +402,14 @@ Host::prepareAppendLogRequest() {
...
+                .then([idStr = this->idStr_] (Status&& status) {
                 if (status.ok()) {
-                    LOG(INFO) << idStr_ << "Send snapshot succeeded!";
+                    LOG(INFO) << idStr << "Send snapshot succeeded!";
                 } else {
-                    LOG(INFO) << idStr_ << "Send snapshot failed!";
+                    LOG(INFO) << idStr << "Send snapshot failed!";
                     // TODO(heng): we should tell the follower i am failed.
                 }
-                sendingSnapshot_ = false;
```
wadeliuyi: why delete this logic?

src/kvstore/raftex/Host.cpp
```diff
@@ -190,6 +190,7 @@ void Host::appendLogsInternal(folly::EventBase* eb,
             {
                 std::lock_guard<std::mutex> g(self->lock_);
                 self->setResponse(r);
+                self->lastLogIdSent_ = self->logIdToSend_;
```
wadeliuyi: when send log failed, why update the last sent log id?

src/kvstore/raftex/SnapshotManager.cpp
```diff
@@ -63,23 +67,31 @@ folly::Future<Status> SnapshotManager::sendSnapshot(std::shared_ptr<RaftPart> pa
...
             LOG(WARNING) << part->idStr_ << "Send snapshot failed!";
             p.setValue(Status::Error("Send snapshot failed"));
+            onFinished(host);
             return;
         });
     });
     return fut;
 }
 
+void SnapshotManager::onFinished(std::shared_ptr<Host> host) {
```
wadeliuyi: modify the host member in snapshotmanager is bad, it is hard understand and the module coupling is complex.

src/kvstore/raftex/Host.cpp
critical27: It is moved into SnapshotManager, during sending snapshot, the Host maybe destructed.

src/kvstore/raftex/Host.cpp
critical27: That would cause leader trying to send snapshot and scanning rocskdb.

src/kvstore/raftex/SnapshotManager.cpp
critical27: We must extend Host's life cycle after the snapshot has been sent out, because the Host may be destructed during this phase

PR #1096, opened 2019-10-16
Function is in

### LINKED ISSUE #1040: support \in in the WHERE clause

CPWstatic:
As we discussed this offline, we will make **in** as an operator. But it requires type of **list** which will not be implemented in a short time. So we would implement it as a special function at present which will not make compatibility issues.
The syntax would looks like:

YIELD udf_is_in(1,1,2,3);
Or you may reference it in filter:

WHERE udf_is_in(1,1,2,3)
Definitely, you can reference the input , src and dest ref:

WHERE udf_is_in($-.id,1,2,3)
WHERE udf_is_in($$.team.name,"Spurs","Suns")
WHERE udf_is_in($^.player.name,"Tim Duncan", "Tony Parker")
close #1040

### DIFF
--- a/src/common/filter/Expressions.cpp
+++ b/src/common/filter/Expressions.cpp
@@ -538,6 +538,10 @@ const char* PrimaryExpression::decode(const char *pos, const char *end) {
 std::string FunctionCallExpression::toString() const {
     std::string buf;
     buf.reserve(256);
+    if (expr_ != nullptr) {
+        buf += expr_->toString();
+        buf += " ";
+    }
     buf += *name_;
     buf += "(";
     for (auto &arg : args_) {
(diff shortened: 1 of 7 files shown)


CPWstatic:
> Q1: about not_in ?
> 
> where udf_is_in(z, a, b, c) == false? or where !udf_is_in(z, a, b, c)
> 
> are they both OK (for now)?
> 
> Q2: what if : udf_is_in(z, a+z, b-z, c*c)

Both supported.

CPWstatic:
Jenkins go!

PR #1105, opened 2019-10-17
Fix in docker couldn't get system version

laura-ding:


build_dep.sh
```diff
@@ -170,25 +175,15 @@ function addAlias {
...
-        return 2
-    fi
-    result=`cat /proc/version|grep el6`
-    if [[ -n $result ]]; then
-        return 3
-    fi
-    result=`cat /proc/version|grep ubuntu1~18`
-    if [[ -n $result ]]; then
-        return 4
+    if [[ -e /etc/redhat-release ]]; then
```
yixinglu: why not check `/etc/os-release` to get os info?

laura-ding: Centos6 does not have this file

laura-ding:
Have been solved

PR #1110, opened 2019-10-17
fix key format of key-value

critical27:
1. Add partition id for key of kv pairs.
2. Skip compaction filter if kv_mode enabled.


src/storage/GetProcessor.cpp
```diff
@@ -36,10 +36,14 @@ GetProcessor::asyncProcess(PartitionID part,
                            const std::vector<std::string>& keys) {
     folly::Promise<std::pair<PartitionID, kvstore::ResultCode>> promise;
     auto future = promise.getFuture();
+    std::vector<std::string> kvKeys;
+    std::transform(keys.begin(), keys.end(), std::back_inserter(kvKeys),
```
dangleptr: We'd better reserve enough space before using kvKeys to avoid extra malloc.

critical27: Good suggestion.

PR #1111, opened 2019-10-17
Update the metad conf template

### LINKED ISSUE #768: Update our meta conf template
It is stale.  We should update it now. 

1.  ADD meta_server_addrs flag.  
2.  Tell our users,  metad can't use multi data paths.

zhangguoqing:
1. ADD meta_server_addrs flag.
2. Tell users, metad can't use multi data paths.

close #768


dangleptr:
Jenkins go

PR #1116, opened 2019-10-17
Null value support

sherman-the-tank:
This is the first PR to add support for Null value. It also sets up the framework for the default value. This PR mainly modify the data serialization module.

The PR to add the supoort for the default value in nGQL will follow


src/dataman2/DataCommon.h
```diff
@@ -0,0 +1,48 @@
+/* Copyright (c) 2018 vesoft inc. All rights reserved.
```
zlcook: 2019

src/dataman2/RowReader.cpp
```diff
@@ -0,0 +1,992 @@
...
+
+
+bool RowReader::isNull(const int64_t index) const noexcept {
+    if (index < 0) {
+        LOG(ERROR) << "Field index " << index << " is out of the range";
+        return true;
+    }
+    size_t flagIdx = index >> 3;
+    DCHECK_GT(nullFlags_.size(), flagIdx);
+    return nullFlags_[flagIdx] & (static_cast<uint8_t>(0x80) >> (index & 0x0000000000000007));
```
critical27: 👍

src/dataman2/RowWriter.cpp
```diff
@@ -0,0 +1,383 @@
...
+
+
+void RowWriter::writeNullValue(int64_t colIndex, NullValue::NullType type) noexcept {
+    // set the Null flag
+    size_t byteIdx = colIndex >> 3;
+    nullFlags_.resize(byteIdx + 1, 0);
+    nullFlags_.back() |= ((uint8_t)0x80 >> (colIndex & 0x00000007));
+
+    // Output the Null value type
+    cord_ << static_cast<int8_t>(type);
```
critical27: I think the `nullFlags_` is enough for indicate whether a value is null, so I guess the type is used for NaN and other special values?

dangleptr:
Implement it in 2.0

PR #1121, opened 2019-10-22
Fix link dynamic library

laura-ding:
Fix when there is a dynamic library, first go to link the dynamic library


laura-ding:
jenkins go

Issue #1077, opened 2019-10-15
When stopping the raft service, it may cause a crash.


monadbobo:
When stopping the raft service, it may cause a crash.

Below is the output of AddressSanitizer:

```
READ of size 8 at 0x616000330080 thread T145 (ThriftIO0)
    #0 0xb1230b in auto nebula::thrift::ThriftClientManager<nebula::raftex::cpp2::RaftexServiceAsyncClient>::client(std::pair<int, int> const&, folly::EventBase*, bool, unsigned int)::{lambda(auto:1*)#2}::operator()<nebula::raftex::cpp2::RaftexServiceAsyncClient>(nebula::raftex::cpp2::RaftexServiceAsyncClient*) const src/common/thrift/ThriftClientManager.inl:62
    #1 0xb30eaf in std::_Sp_counted_deleter<nebula::raftex::cpp2::RaftexServiceAsyncClient*, nebula::thrift::ThriftClientManager<nebula::raftex::cpp2::RaftexServiceAsyncClient>::client(std::pair<int, int> const&, folly::EventBase*, bool, unsigned int)::{lambda(auto:1*)#2}, std::allocator<void>, (__gnu_cxx::_Lock_policy)2>::_M_dispose() /usr/include/c++/8/bits/shared_ptr_base.h:471
    #2 0x9200e8 in std::_Sp_counted_base<(__gnu_cxx::_Lock_policy)2>::_M_release() /usr/include/c++/8/bits/shared_ptr_base.h:155
    #3 0x91da33 in std::__shared_count<(__gnu_cxx::_Lock_policy)2>::~__shared_count() /usr/include/c++/8/bits/shared_ptr_base.h:728
    #4 0xafec69 in std::__shared_ptr<nebula::raftex::cpp2::RaftexServiceAsyncClient, (__gnu_cxx::_Lock_policy)2>::~__shared_ptr() /usr/include/c++/8/bits/shared_ptr_base.h:1167
    #5 0xafecca in std::shared_ptr<nebula::raftex::cpp2::RaftexServiceAsyncClient>::~shared_ptr() /usr/include/c++/8/bits/shared_ptr.h:103
    #6 0xb2aa01 in std::pair<std::pair<std::pair<int, int>, folly::EventBase*> const, std::shared_ptr<nebula::raftex::cpp2::RaftexServiceAsyncClient> >::~pair() /usr/include/c++/8/bits/stl_pair.h:208
[...]

PR #1127, opened 2019-10-22
Issue #1077 Fixed the crash that might occur when stopping the raft service

monadbobo:
When the ThriftClientManager object needs to be shared, if the EventBase object was released,
the corresponding host was removed from the ThriftClientManager.


src/common/thrift/ThriftClientManager.inl
```diff
@@ -63,6 +63,11 @@ std::shared_ptr<ClientType> ThriftClientManager<ClientType>::client(
             delete p;
         });
     });
+
+    if (shared_) {
```
dangleptr: Why remove the related host-evb pair only in share mode?

monadbobo: Because in the shared mode, the clientMap_ object will be released after the eventbase object is released, so we need to delete the value corresponding to clientMap_ while the eventBase is released.

dangleptr: I mean could we just delete the entry in clientMap_ whether in shared mode or not?  Because the clientMap_ is threadLocal,  it means it should be destroyed before related evb released.  WDYT?

monadbobo: Yes, I am going to modify it.

PR #1130, opened 2019-10-23
Fix dependencies build script and CI package failure

yixinglu:
- When installing 3rdparty dependencies in docker container, `build_deps.sh` script will return no-zero result in `getSystemVer` function and result in failure of building of docker image.
- Fix packaging error in CI.


build_dep.sh
```diff
@@ -115,22 +115,30 @@ function addAlias {
...
-    if [[ -e /etc/issue ]]; then
+        if [[ -n `cat /etc/redhat-release|grep Fedora` ]]; then
+            echo 1
+        elif [[ -n `cat /etc/redhat-release|grep "release 7."` ]]; then
+            echo 2
+        elif [[ -n `cat /etc/redhat-release|grep "release 6."` ]]; then
+            echo 3
+        else
+            echo 0
+        fi
```
laura-ding: and `return 0`

PR #1140, opened 2019-10-23
Test the performance of multiple partitions

bright-starry-sky:
This one is improve of PR #983.

Test result :
1，A small number of partitions is faster than a large number of partitions. But performance will be stable if the number of partitions is about 100.
2，Multithreaded scanning is faster than single-threaded scanning.


bright-starry-sky:
This one duplicated #1807  and #1807 was merged. 
Close this one.

PR #1151, opened 2019-10-25
Check variable existence in fetch

### LINKED ISSUE #1132: 'fetch prop' returns column id not found
```
GO FROM 200 over like | FETCH PROP ON student $- YIELD student.name
[ERROR (-8)]: Column `id' not found

```

same with user-defined variables.

So, must the tag have `id` prop?

### LINKED ISSUE #1135: Variable `var' not defined when user-define variable result is empty
The result of $var is empty.

CPWstatic:
Close #1132, close #1135


CPWstatic:
Jenkins go!

CPWstatic:
> Do you add one UT case to cover this?

The result is not wrong, but an unreadable hint was given.

PR #1156, opened 2019-10-28
Update the load_data_interval_secs to be 1 second

dangleptr:
Our users always complain the load_data_interval_secs is too long (2 minutes by default).


dangleptr:
Jenkins go

PR #1160, opened 2019-10-29
Forward the dependencies to the latest version

dangleptr:
This pr is to record the changes when i upgrade all our dependencies. 
The following dependencies are compiled on centos 7.5
```
drwxrwxr-x. 4 heng heng  30 Oct 24 19:54 autoconf-2.69
drwxrwxr-x. 4 heng heng  30 Oct 25 10:26 automake-1.16
drwxrwxr-x. 4 heng heng  30 Oct 25 11:35 automake-1.16.1
drwxrwxr-x. 5 heng heng  41 Oct 21 22:44 bison-3.4
drwxrwxr-x. 4 heng heng  32 Oct 17 14:04 boost-1.71.0
drwxrwxr-x. 6 heng heng  54 Oct 17 18:27 bzip2
drwxrwxr-x. 5 heng heng  41 Oct 25 14:14 cmake-3.15.4
drwxrwxr-x. 4 heng heng  34 Oct 25 14:49 double-conversion-3.1.5
drwxrwxr-x. 5 heng heng  43 Oct 22 00:20 fbthrift-2019.10.14.00
drwxrwxr-x. 4 heng heng  32 Oct 21 16:51 fizz-2019.10.14.00
drwxrwxr-x. 6 heng heng  56 Oct 24 19:19 flex-2.6.4
drwxrwxr-x. 4 heng heng  34 Oct 21 22:50 fmt-6.0.0
drwxrwxr-x. 4 heng heng  32 Oct 21 16:46 folly-2019.10.14.00
drwxrwxr-x. 8 heng heng  84 Oct 17 10:57 gcc-8.3.0
drwxrwxr-x. 6 heng heng  56 Oct 25 12:01 gettext-0.20.1
drwxrwxr-x. 5 heng heng  43 Oct 17 14:35 gflags-v2.2.2
drwxrwxr-x. 4 heng heng  34 Oct 17 14:47 glog-v0.4.0
drwxrwxr-x. 5 heng heng  45 Oct 16 19:33 gmp-6.1.2
drwxrwxr-x. 4 heng heng  30 Oct 25 12:09 gperf-3.1
drwxrwxr-x. 4 heng heng  34 Oct 17 17:15 gtest-1.8.1
drwxrwxr-x. 5 heng heng  41 Oct 24 16:33 help2man-1.47.9
drwxr-xr-x. 6 heng heng  56 Oct 25 15:13 jemalloc-5.2.1
drwxrwxr-x. 2 heng heng  40 Oct 16 19:40 lib
[...]

PR #1161, opened 2019-10-29
Update the interfaces about folly::future and fbthrift constructor

dangleptr:
Some interfaces have been changed in the latest version of folly and fbthrift. 
Let's modify them to avoid broken under latest dependencies.  

Subtask of #1160


src/kvstore/raftex/RaftPart.cpp
```diff
@@ -775,7 +775,7 @@ void RaftPart::replicateLogs(folly::EventBase* eb,
             return resp.get_error_code() == cpp2::ErrorCode::SUCCEEDED
                     && !hosts[index]->isLearner();
         })
-        .then(executor_.get(), [self = shared_from_this(),
+        .via(executor_.get()).then([self = shared_from_this(),
```
darionyaphet: Should using `thenValue` to replace `then` ?

dangleptr: Because we need to handle the exception in folly::Try,  so we should use "then" in this place.

zhangguoqing:
src/meta/test/TestUtils.h +131
allHosts.emplace_back(apache::thrift::FragileConstructor::FRAGILE, i, i);

maybe some others in ```grep FRAGILE -rn ./src/```

dangleptr:
Jenkins go

PR #1162, opened 2019-10-29
Add geo interface in native client.

CPWstatic:
Seems not required anymore.

### DIFF
--- a/src/tools/native-client/pom.xml
+++ b/src/tools/native-client/pom.xml
@@ -35,6 +35,11 @@
             <version>4.12</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>io.sgr</groupId>
+            <artifactId>s2-geometry-library-java</artifactId>
+            <version>1.0.1</version>
+        </dependency>
     </dependencies>
 
     <build>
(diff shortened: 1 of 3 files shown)


CPWstatic:
Jenkins go!

src/tools/native-client/pom.xml
```diff
@@ -35,6 +35,11 @@
             <version>4.12</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
```
darionyaphet: why don't move `S2` into `Spark Generator` ？

darionyaphet:
`native client` is a lightly warpper for cpp interface. The `dimension transformation` seems not matching

CPWstatic:
> Could we change the class name?
> native_client is not reasonable. "NabulaUtils" maybe better.

Agree.

dangleptr:
Do we still need it?  @CPWstatic

darionyaphet:
Maybe we should support this method in `nebula java` ?

dangleptr:
> Maybe we should support this method in `nebula java` ?

Sounds good. Let's reimplement it in nebula-java.  Close it.  @darionyaphet

PR #1163, opened 2019-10-29
Push new tag of docker image when releasing nebula version

yixinglu:
- Add new `release` event listener for docker image building jobs
- Fix docker context in building script
- Add docker image address in readme


.github/workflows/docker.yaml
```diff
@@ -3,6 +3,9 @@ name: docker
 on:
   schedule:
     - cron: '0 18 * * *'
+  release:
```
Shylock-Hg: When will be triggered?

yixinglu: next time when release will be created.

.github/workflows/docker.yaml
```diff
@@ -24,10 +27,20 @@ jobs:
...
         run: |
           docker build -t ${IMAGE_NAME} -f docker/Dockerfile.${{ matrix.service }} .
           docker push ${IMAGE_NAME}
         shell: bash
+      - name: build release image
+        if: github.event_name == 'release'
+        env:
+          IMAGE_NAME: ${{ secrets.DOCKER_USERNAME }}/nebula-${{ matrix.service }}
+        run: |
+          TAG=$(git describe --exact-match --abbrev=0 --tags)
```
Shylock-Hg: Will the `TAG` empty?

yixinglu: No. Because this job will be triggered by release event which will create a new tag on the current commit.

yixinglu: No. Because this job will be triggered by release event which will create a new tag on the current commit.

PR #1164, opened 2019-10-29
fix ut fail when mockStorageServer takes longer time to open rocksdb

critical27:
Wait until all leaders have been elected


zhangguoqing:
LGTM!

PR #1167, opened 2019-10-29
Add slow op tracker and some debug logs

dangleptr:


src/kvstore/raftex/RaftPart.cpp
```diff
@@ -775,18 +782,23 @@ void RaftPart::replicateLogs(folly::EventBase* eb,
             return resp.get_error_code() == cpp2::ErrorCode::SUCCEEDED
                     && !hosts[index]->isLearner();
         })
-        .then(executor_.get(), [self = shared_from_this(),
+        .then(executor_.get(),
```
liuyu85cn: 这个接口要改thenValue吗?

dangleptr: I have finished it in #1161

PR #1168, opened 2019-10-29
Downgrade java client jdk version from 1.8 to 1.6

darionyaphet:


Shylock-Hg:
Why change this?

darionyaphet:
> Why change this?

Because some user still using `JDK6` and we should support it

zhangguoqing:
LGTM!  But, don't you want to downgrade the native-client too?

PR #1171, opened 2019-10-29
Check DT_UNKNOWN in file_utils.

CPWstatic:
We found that _**readdir**_ doesn't work well on CentOS7.1. So we would check the path type by calling the **lstat** when **readdir** returns path type with **DT_UNKNOWN**. I have tested it in CentOS7.1.


CPWstatic:
Jenkins go!

PR #1172, opened 2019-10-30
Clear the code associated with Add/Remove Hosts

### LINKED ISSUE #736: Remove add/remove hosts steps

### LINKED ISSUE #782: [bug] insert empty prop errors
I add two hosts (in fact they are the same machine)

If I add vertex with property, it's ok

If I add vertex with no property. It's internal error.

image

image

I'm not sure it's about two hosts or empty prop, but it's easy to reproduce.

zhangguoqing:
close #736
close #782


zhangguoqing:
Jenkins go!

src/meta/processors/partsMan/ListHostsProcessor.cpp
```diff
@@ -45,19 +47,24 @@ StatusOr<std::vector<cpp2::HostItem>> ListHostsProcessor::allHostsWithStatus(
...
+        if (now - info.lastHBTimeInSec_ < FLAGS_removed_threshold_sec) {
+            if (now - info.lastHBTimeInSec_ < FLAGS_expired_threshold_sec) {
+                item.set_status(cpp2::HostStatus::ONLINE);
+            } else {
+                item.set_status(cpp2::HostStatus::OFFLINE);
+            }
+            hostItems.emplace_back(item);
         } else {
-            item.set_status(cpp2::HostStatus::OFFLINE);
+            removeHostsKey.emplace_back(iter->key());
```
dangleptr: That's cool.

src/meta/test/MetaClientTest.cpp
```diff
@@ -503,178 +495,6 @@ TEST(MetaClientTest, HeartbeatTest) {
     ASSERT_EQ(1, ActiveHostsMan::getActiveHosts(sc->kvStore_.get()).size());
 }
 
-class TestMetaService : public cpp2::MetaServiceSvIf {
```
dangleptr: Why remove the UTs

dangleptr: The UTs are used to test "retry" logic,   maybe we could use another interface instead of "addHosts"

zhangguoqing: Because the `addHosts` interface has been removed, I also deleted the UT here.  And which interface do we want to use instead of it ?

src/meta/processors/partsMan/ListHostsProcessor.cpp
zhangguoqing: thx. :)

src/meta/test/MetaClientTest.cpp
dangleptr: For example, you could use "addTag" instead

zhangguoqing: Thx.  After reading the code carefully, I understand the role of those UTs, and have chosen the `heartbeat` instead of `addHosts`.

zhangguoqing:
Jenkins go!

PR #1177, opened 2019-10-30
Improve spark sst file generation job

spacewalkman:
Some improvements:

Improve hive table to sst files generation strategy as following:
1. Upgrade spark to 2.4.3 to use its external shuffle service to increase stability.
2. Fix FNV hash overflow bug
3. Aggregate all tag and edge type to query in 1 round instead of one-by-one, which will cause explosion of sst files with small size(like 100k+)
4. Persist RDD to disk to avoid nasty OOM exception and heavy tuning job when running on a restricted cluster.
5. Some refactoring to mapping configuration class


spacewalkman:
Jenkins go!

dangleptr:
We don't need the SST generator anymore. So close it.

PR #1194, opened 2019-10-31
Scan through the last wal in case wal is cut off.

critical27:


src/kvstore/wal/FileBasedWal.cpp
```diff
@@ -238,7 +232,28 @@ void FileBasedWal::scanAllWalFiles() {
 
         // We now get all necessary info
         close(fd);
-        walFiles_.insert(std::make_pair(startIdFromName, info));
+    }
+
+    if (!walFiles_.empty()) {
+        auto it = walFiles_.rbegin();
+        if (it->second->lastId() == 0) {
```
dangleptr: Could we unify them?

src/kvstore/wal/FileBasedWal.cpp
```diff
@@ -377,6 +392,80 @@ TermID FileBasedWal::readTermId(const char* path, LogID logId) {
...
+               + sizeof(ClusterID)
+               + sizeof(int32_t)
+               + head
+               + sizeof(int32_t);
+        ++curLogId;
+    }
+
+    if (pos < 0 && pos < FileUtils::fileSize(path)) {
+        LOG(WARNING) << "Invalid wal " << path << ", truncate from offset " << pos;
+        ftruncate(fd, pos);
```
dangleptr: "ftruncate" only be used on file opened in write mode.  And you should check the return value for it.

src/kvstore/wal/FileBasedWal.cpp
```diff
@@ -238,7 +232,28 @@ void FileBasedWal::scanAllWalFiles() {
...
+        if (it->second->lastId() == 0) {
+            // if last wal is empty, scan previous wal
+            auto expectedLastWalId = it->second->firstId() - 1;
+            unlink(it->second->path());
+            walFiles_.erase(it->first);
+            it = walFiles_.rbegin();
+            scanLastWal(it->second, it->second->firstId(), expectedLastWalId);
+        } else {
+            // scan last wal, if it is illegal, scan the previous wal
+            scanLastWal(it->second, it->second->firstId(), -1);
```
dangleptr: Too verbose. 
Why you need the expectedLastWalId?

critical27: That's because we didn't truncate when we rollback, and it makes verbose. Say we have a wal `1000.wal` with `firstId = 1000, lastId =1200`, and we rollback to 1100, it still has all these wal. `expectedLastWalId` is used to mark the last valid wal. 

I'will try to simplify the logic.

src/kvstore/wal/FileBasedWal.cpp
critical27: Good catch.

Issue #1192, opened 2019-10-31
Traverse multiple edges can't get edge property


Amber1990Zhang:
As title.
![image](https://user-images.githubusercontent.com/42762957/67941707-35df9700-fc11-11e9-9775-fc329cc048bb.png)

whitewum:
add one more prove: 
go from 100 over like yield like.likeness;
go from 100 over serve yield serve.end_year

Amber1990Zhang:
> try:
> go from 100 over like yield like.likeness;
> go from 100 over serve yield serve.end_year
> try:
> go from 100 over * yield *

I tried your suggestions and they all works, but traverse multiple edges won't.

PR #1197, opened 2019-11-01
Fixed the bug when using multi edge

monadbobo:
Issue #1192  Fixed the bug where the default value of the edge prop could not be obtained correctly.


monadbobo:
close #1192

PR #1201, opened 2019-11-04
Fixed a compile warning when using the Clang 8.0.

monadbobo:
Fixed a compile warning when using the Clang 8.0.


dutor:
Thanks, but I wonder what kind of warning was triggered.

Issue #1203, opened 2019-11-04
graphd crash yield not existing node prop for debug builds


jude-zhu:
following the [get-started.md](https://github.com/vesoft-inc/nebula/blob/master/docs/manual-EN/1.overview/2.quick-start/1.get-started.md) example, then run the following query:

```
delete vertex 200; INSERT EDGE like(likeness) VALUES 201 -> 200:(85.6);go from 201 over like yield $$.student.name as name
```

`graphd` crashed:

![image](https://user-images.githubusercontent.com/51590253/68112968-97637680-ff2d-11e9-9a11-d1581814bbd7.png)

p.s. I built the code under debug version

PR #1204, opened 2019-11-04
Issue #1203 fixed the possible crash.

monadbobo:
The crash is caused by the use of two meaningless DCHECK in debug mode.


monadbobo:
close #1203

Issue #1206, opened 2019-11-05
Support id in show tags and edges


darionyaphet:
```
(user@127.0.0.1) [t]> show tags
=================
| ID | Name     |
=================
| 2  | course   |
-----------------
| 3  | building |
-----------------
| 4  | student  |
-----------------
Got 3 rows (Time spent: 1784/2755 us)

(user@127.0.0.1) [t]> show edges
===============
| ID | Name   |
===============
| 5  | like   |
---------------
| 6  | select |
---------------
Got 2 rows (Time spent: 1519/2359 us)
```

CPWstatic:
What circumstances it is that we need this internal id?

darionyaphet:
Currently `Spark SST Generator` need manual assignment the tag and edge type .

PR #1207, opened 2019-11-05
Support id in show tags and show edges

darionyaphet:
```
(user@127.0.0.1) [t]> show tags
=================
| ID | Name     |
=================
| 2  | course   |
-----------------
| 3  | building |
-----------------
| 4  | student  |
-----------------
Got 3 rows (Time spent: 1784/2755 us)

(user@127.0.0.1) [t]> show edges
===============
| ID | Name   |
===============
| 5  | like   |
---------------
| 6  | select |
---------------
Got 2 rows (Time spent: 1519/2359 us)
```


zhangguoqing:
LGTM!  It's very useful, and close #1206

CPWstatic:
Did you run tests offline? It seems that UT would failed if you just change the return value without modify the corresponding UT.

dangleptr:
Jenkins go

src/graph/ShowExecutor.cpp
```diff
@@ -218,18 +218,19 @@ void ShowExecutor::showTags() {
         auto value = std::move(resp).value();
         resp_ = std::make_unique<cpp2::ExecutionResponse>();
         std::vector<cpp2::RowValue> rows;
-        std::vector<std::string> header{"Name"};
+        std::vector<std::string> header{"ID", "Name"};
         resp_->set_column_names(std::move(header));
 
-        std::set<std::string> uniqueTags;
+        std::map<nebula::cpp2::TagID, std::string> tagItems;
```
laura-ding: I think it’s better to put the name as the key in the map , it will order the name. Because users usually remember the name firstly, then through name to get the id. and the returns put the name in the first column.

Amber1990Zhang: I think to order name is a good idea but no need to put the names in the first column. List number comes first is more appreciated to me.

darionyaphet: Currently ordered by `ID` is same order as created

PR #1210, opened 2019-11-05
Support spark job using nebula client

darionyaphet:
`SparkClientGenerator` is a simple spark job used to write data into `Nebula Graph` parallel.

The insert statement is generated according to the configuration file and executed through thrift client.

Data Source could be `HDFS` (`Parquet`, `JSON` and `ORC File`) or `Hive`.

In the configuration file, the tag section is similar to the edge section, which is used to specify the generation rules.

The `type` specified the data source type, and it should be `parquet`, `json`, `ORC`, `csv` or `hive`.

If the `type` is a file, the `path` is use to point out the `HDFS` path, otherwise it use `exec` to execute SQL in `hive`.

`fields` describe the mapping between the data field and the Nebula field.

In the tag section, use `vertex` to describe the vertex key, and in the edge section, `source` is the starting point, `target` is the ending point and `ranking` is the weight which is optional.


bright-starry-sky:
Great job！a man of great versatility！

jude-zhu:
Jenkins go

PR #1218, opened 2019-11-06
Support remove scan prefix in general storage

darionyaphet:


src/storage/kv/PrefixProcessor.cpp
```diff
@@ -0,0 +1,58 @@
...
+PrefixProcessor::asyncProcess(PartitionID part, std::string prefix) {
+    folly::Promise<PartitionCode> promise;
+    auto future = promise.getFuture();
+
+    auto encodedPrefix = NebulaKeyUtils::generalKey(part, prefix);
+    executor_->add([this, p = std::move(promise), part,
+                    encodedPrefix = std::move(encodedPrefix)] () mutable {
+        std::unique_ptr<kvstore::KVIterator> iter;
+        auto ret = this->kvstore_->prefix(space_, part, encodedPrefix, &iter);
+        if (ret == kvstore::ResultCode::SUCCEEDED) {
```
critical27: Maybe we need to add a row limit in case OOM

src/storage/kv/PrefixProcessor.cpp
```diff
@@ -0,0 +1,58 @@
...
+            std::lock_guard<std::mutex> lg(this->lock_);
```
critical27: How about each thread has a map, and merge them at last

dangleptr:
Let's release it in storage 2.0 .

PR #1221, opened 2019-11-07
add gflags to control max_edge_returned_per_vertex

liuyu85cn:


src/storage/test/QueryBoundTest.cpp
```diff
@@ -225,7 +225,9 @@ TEST(QueryBoundTest, OutBoundSimpleTest) {
     checkResponse(resp, 30, 12, 10001, 7, true);
 }
 
-TEST(QueryBoundTest, inBoundSimpleTest) {
+TEST(QueryBoundTest, MaxEdgesReturenedTest) {
```
dangleptr: Don't change the original UT,  add a new one.

src/storage/QueryBaseProcessor.cpp
```diff
@@ -8,6 +8,7 @@
 
 DEFINE_int32(max_handlers_per_req, 10, "The max handlers used to handle one request");
 DEFINE_int32(min_vertices_per_bucket, 3, "The min vertices number in one bucket");
+DEFINE_int32(max_edge_returned_per_vertex, 1000, "The max edge number returnred searching vertex");
```
CPWstatic: The default value should be a bigger one, I'd like it is the maximum of int32. That means we will not cut-off anything by default. And users would config it with an appropriate value in config-file, such as 1000 or 5000, etc.

liuyu85cn: thx, change to INT_MAX

dangleptr:
Jenkins go

PR #1227, opened 2019-11-07
Support to create hard link for current WAL

dangleptr:
It will be used when creating snapshot.


src/kvstore/wal/FileBasedWal.cpp
```diff
@@ -591,6 +591,23 @@ std::unique_ptr<LogIterator> FileBasedWal::iterator(LogID firstLogId,
...
+        LOG(INFO) << idStr_ << "Create link failed, there is no wal files!";
+        return false;
+    }
+    auto it = walFiles_.rbegin();
+    if (link(it->second->path(), newPath) != 0) {
+        LOG(INFO) << idStr_ << "Create link failed for " << it->second->path()
+                  << " on " << newPath << ", error:" << strerror(errno);
+        return false;
+    }
+    LOG(INFO) << idStr_ << "Create link success for " << it->second->path()
```
bright-starry-sky: I have a question, do we need a prepareNewFile after the hard link created?

bright-starry-sky:
So fast, great work！

src/kvstore/wal/FileBasedWal.cpp
bright-starry-sky: Because I'm worried that the link file will change with the source file.

bright-starry-sky:
Jenkins go

bright-starry-sky:
UT failed as below :

image

bright-starry-sky:
Jenkins go

src/kvstore/wal/FileBasedWal.cpp
dangleptr: We don't need to call prepareNewFile here.  Because a new wal file will be created when next log appended.

dangleptr: > Because I'm worried that the link file will change with the source file.

The WAL becomes immutable after closed.

bright-starry-sky:
Approved this one. Thanks for your meticulous work!

dangleptr:
Jenkins go

PR #1231, opened 2019-11-07
fix Dockerfile.graphd port 3699 typo error

jievince:


docker/Dockerfile.graph
```diff
@@ -1,24 +0,0 @@
-FROM vesoft/nebula-dev:centos7 as builder
```
yixinglu: Don't delete this file now. Some documents still reference the `vesoft/nebula-graph` image. And this dockerfile is used to build that image.

jievince: ok

yixinglu:
Jenkins go

Issue #174, opened 2019-02-26
Complete GO statement functionality (on Query Engine)


sherman-the-tank:
Make GO statement fully functional (without pushing filtering/aggregation to storage servers)
- [ ] REVERSELY
- [ ] UPTO

PR #1233, opened 2019-11-07
Return variable if not empty

CPWstatic:
The **return** sentence is used for returning variable if it is not empty.

We have CN-docs with pr #1246


CPWstatic:
Jenkins go!

src/graph/ReturnExecutor.cpp
```diff
@@ -0,0 +1,79 @@
...
+            LOG(ERROR) << "Get rows failed: " << ret.status();
+            onError_(std::move(ret).status());
+            return;
+        }
+        auto rows = ret.value();
+        if (rows.empty()) {
+            onFinish_(Executor::ProcessControl::kNext);
+            return;
+        }
+        resp_->set_rows(std::move(ret).value());
```
monadbobo: std::move(rows)

src/parser/ProcessControlSentences.cpp
```diff
@@ -0,0 +1,25 @@
...
+ */
+
+#include "parser/ProcessControlSentences.h"
+
+namespace nebula {
+std::string ReturnSentence::toString() const {
+    std::string buf;
+    buf.reserve(256);
+
+    buf += "RETURN $";
```
laura-ding: if var_ is nullptr, here  has not ”RETURN $"

src/parser/parser.yy
```diff
@@ -1721,13 +1723,26 @@ maintain_sentence
     | balance_sentence { $$ = $1; }
     ;
 
+return_sentence
+    : KW_RETURN VARIABLE {
```
laura-ding: why?

src/parser/ProcessControlSentences.cpp
CPWstatic: Actually the var_ shall never be nullptr, or fatal error might occurred somewhere.

zhangguoqing:
LGTM!

jude-zhu:
relate to #174

PR #1236, opened 2019-11-08
Improve heartbeat between metad and storaged by MetaClient

### LINKED ISSUE #1060: [Bug]Graphd no response
I deployed cluster with 3 hosts, then killed 2, then tried to query, but graphd is timeout with no response, I think there may be a bug here:
image

### LINKED ISSUE #1173: Improve heartbeat of storage.
We can add more info in heartbeat from storage to meta, such as leader/follower/learner count and so on. So we can avoid block in meta service.

After #1086 merged in, I will change the interface.

zhangguoqing:
1. Add the LastUpdateTimeMan to record the latest metadata change(insert, update, delete)
   time on the MetaServer side.
2. Enrich the information(lastUpdateTime, partitions leaderDist) of `heartbeat` exchange
   between MetaServer and MetaClient at storage node.
3. It could optimize the load data logic inside meta client by lastUpdateTime, which makes
   metadata synchronization smarter and more timely.
4. Remove sendHeartBeat_ and load_data_interval_secs from MetaClient class, and add the
    bool inStoraged_ to indicate whether a metaclient is in storaged.

close #1173
close #1060


dangleptr:
Excellent!

src/kvstore/PartManager.cpp
```diff
@@ -155,5 +155,14 @@ void MetaServerBasedPartManager::onPartUpdated(const PartMeta& partMeta) {
     UNUSED(partMeta);
 }
 
+void MetaServerBasedPartManager::onLeaderDist(std::unordered_map<GraphSpaceID,
+        std::vector<PartitionID>>& leaderIds) {
```
dangleptr: alignment.

src/kvstore/PartManager.h
```diff
@@ -164,6 +166,9 @@ class MetaServerBasedPartManager : public PartManager, public meta::MetaChangedL
 
      void onPartUpdated(const PartMeta& partMeta) override;
 
+     void onLeaderDist(std::unordered_map<GraphSpaceID,
```
dangleptr: Maybe "fetchLeaderInfo" is a better name?

src/meta/ActiveHostsMan.cpp
```diff
@@ -16,11 +16,11 @@ namespace meta {
 
 kvstore::ResultCode ActiveHostsMan::updateHostInfo(kvstore::KVStore* kv,
                                                    const HostAddr& hostAddr,
-                                                   const HostInfo& info) {
+                                                   const TimeInfo& info) {
     CHECK_NOTNULL(kv);
     std::vector<kvstore::KV> data;
     data.emplace_back(MetaServiceUtils::hostKey(hostAddr.first, hostAddr.second),
-                      HostInfo::encode(info));
+                      TimeInfo::encode(info));
```
dangleptr: Why change the name?

src/meta/ActiveHostsMan.cpp
```diff
@@ -45,8 +45,8 @@ std::vector<HostAddr> ActiveHostsMan::getActiveHosts(kvstore::KVStore* kv, int32
     auto now = time::WallClock::fastNowInMilliSec();
     while (iter->valid()) {
         auto host = MetaServiceUtils::parseHostKey(iter->key());
-        HostInfo info = HostInfo::decode(iter->val());
-        if (now - info.lastHBTimeInMilliSec_ < threshold) {
+        TimeInfo info = TimeInfo::decode(iter->val());
+        if (now - info.lastTime_ < threshold) {
```
dangleptr: ditto

src/meta/ActiveHostsMan.cpp
zhangguoqing: The ActiveHostsMan and LastUpdateTimeMan are all need to record the time, which is read(decode)/write(encode) the kvstore of metad.  So I extract the struct and change its name from `HostInfo` to `TimeInfo`.

src/meta/processors/admin/HBProcessor.cpp
```diff
@@ -39,16 +39,39 @@ void HBProcessor::process(const cpp2::HBReq& req) {
...
             resp_.set_leader(toThriftHost(nebula::value(leaderRet)));
         }
     }
-    onFinished();
+
+    int64_t lastUpdateTime = LastUpdateTimeMan::get(this->kvstore_);
+    resp_.set_last_update_time_in_ms(lastUpdateTime);
+
+    if (!data.empty()) {
+        doPut(std::move(data), false);
```
critical27: Why don't we refresh `last_update_time` here?

zhangguoqing: The `last_update_time` is used to record some changes(insert, update, delete) that require meta client synchronization(eg. config, schema, parts map).  Here the `doPut` operation only update the storaged host time info(keep it online)  by heartbeat. The information does not  need to sync by metaclient.

zhangguoqing:
Jenkins go!

dangleptr:
I have another thought about last_update_time.  Could we use "lastCommittedLogId" for meta instead?

src/meta/client/MetaClient.cpp
```diff
@@ -1123,6 +1114,31 @@ StatusOr<SchemaVer> MetaClient::getNewestEdgeVerFromCache(const GraphSpaceID& sp
...
+                    if (it == spaceIt->second.end()) {
+                        return true;
+                    }
+                }
+            }
+        }
+    }
+    return false;
+}
+
```
bright-starry-sky: I have a question，Is the above code equal to this:
```
bool MetaClient::diffLeader(const LeaderIds& oldLeaderIds, const LeaderIds& newLeaderIds) {
      return (oldLeaderIds != newLeaderIds);
  }
```

src/meta/processors/admin/HBProcessor.cpp
```diff
@@ -39,16 +39,39 @@ void HBProcessor::process(const cpp2::HBReq& req) {
...
+    auto ret = kvstore::ResultCode::SUCCEEDED;
+    if (req.host.ip != -1 && req.host.port != -1) {
+        TimeInfo hbInfo(time::WallClock::fastNowInMilliSec());
+        ret = ActiveHostsMan::updateHostInfo(this->kvstore_, host, hbInfo);
+
+        if (req.__isset.leader_parts) {
+            nebula::cpp2::HostAddr storageNode;
+            storageNode.set_ip(req.host.ip);
+            storageNode.set_port(req.host.port);
+            const auto& leaderIds = std::move(req).get_leader_parts();
```
bright-starry-sky: Seems not need std::move, What do you think?

src/meta/client/MetaClient.cpp
zhangguoqing: using LeaderIds = std::unordered_map<GraphSpaceID, std::vector<PartitionID>>;
so, the answer maybe `No`.

zhangguoqing:
> I have another thought about last_update_time. Could we use "lastCommittedLogId" for meta instead?

Good proposal, so the synchronization point is more accurate. Recording updates with time is an unreliable way.

src/meta/processors/admin/HBProcessor.cpp
zhangguoqing: OK!

src/meta/processors/admin/HBProcessor.cpp
critical27: 👍

zhangguoqing:
Jenkins go!

src/meta/client/MetaClient.cpp
bright-starry-sky: Wow , Just a discussion, I'm not sure it's appropriate at here. :smiley_cat:

```
    typedef std::unordered_map<int64_t, std::vector<int64_t>> map;
    map a = { {1, {1,2,3}}, {2, {1,2,3}}, {3, {1,2,3}}};
    map b = { {3, {1,2,3}}, {2, {1,2,3}}, {1, {1,2,3}} };
    map c = { {3, {2,3}}, {2, {1,2,3}}, {1, {1,2,3}} };

  if (a==b) {
      LOG(INFO) << "a and b are equal";    <--- hit
  }

  if (a!=c) {
      LOG(INFO) << "a and c are not equal";     <--- hit
  }
```

zhangguoqing: FYI. https://en.cppreference.com/w/cpp/container/unordered_map/operator_cmp
You are right! Then, I have learnt the `unordered_map` and cleared some useless code in the PR about it.
Thx.

zhangguoqing:
Jenkins go!

zhangguoqing:
Jenkins go!

jude-zhu:
close #284

src/meta/client/MetaClient.cpp
critical27: 👍

src/meta/client/MetaClient.cpp
```diff
@@ -1124,17 +1116,31 @@ StatusOr<SchemaVer> MetaClient::getNewestEdgeVerFromCache(const GraphSpaceID& sp
...
     cpp2::HBReq req;
+    req.set_in_storaged(inStoraged_);
     nebula::cpp2::HostAddr thriftHost;
     thriftHost.set_ip(localHost_.first);
     thriftHost.set_port(localHost_.second);
     req.set_host(std::move(thriftHost));
     req.set_cluster_id(clusterId_.load());
+
+    LeaderIds leaderIds;
+    if (inStoraged_ && listener_ != nullptr) {
```
zhangguoqing: It should be in storage node, that can report the leader distribution info.

zhangguoqing:
Jenkins go!

zhangguoqing:
Jenkins go!

dangleptr:
> > I have another thought about last_update_time. Could we use "lastCommittedLogId" for meta instead?
> 
> Good proposal, so the synchronization point is more accurate. Recording updates with time is an unreliable way.
> But this needs to expose a private method of Part class.
> FYI. https://github.com/vesoft-inc/nebula/blob/master/src/kvstore/Part.h#L83
> 
> ```
> private:
>   // Methods inherited from RaftPart
>   std::pair<LogID, TermID> lastCommittedLogId() override;
> ```
> 
> Another problem is that this lastCommittedLogId represents newest put/remove KV operation on metad, but not all changes(eg. update ActiveHostsMan, some customKV) require synchronization of metaclient.

Make sense.  It seems we should use lastUpdateTime.

src/parser/test/ParserTest.cpp
```diff
@@ -1323,18 +1323,6 @@ TEST(Parser, ConfigOperation) {
         auto result = parser.parse(query);
         ASSERT_TRUE(result.ok()) << result.status();
     }
-    {
-        GQLParser parser;
-        std::string query = "UPDATE CONFIGS load_data_interval_secs=120";
```
critical27: We'd better keep this test. It is still used in console. We could change `load_data_interval_secs` to `v`.

src/meta/MetaServiceUtils.cpp
```diff
@@ -20,11 +20,20 @@ const std::string kIndexTable  = "__index__";   // NOLINT
 const std::string kUsersTable  = "__users__";    // NOLINT
 const std::string kRolesTable  = "__roles__";    // NOLINT
 const std::string kConfigsTable = "__configs__"; // NOLINT
+const std::string kLastUpdateTimeTable = "__last_update_time__"; // NOLINT
+const std::string kLeaderDistTable = "__leader_dist__"; // NOLINT
```
dangleptr: Can't we save the leader information in parts table or hosts table?

src/parser/test/ParserTest.cpp
zhangguoqing: OK!

src/meta/MetaServiceUtils.cpp
zhangguoqing: OK, the leader information may be more appropriate to put into the hosts table.

src/meta/ActiveHostsMan.cpp
zhangguoqing: The `ActiveHostsMan` and `LastUpdateTimeMan` have been moved  to the file ActiveMan.h.

src/kvstore/PartManager.h
zhangguoqing: done!

src/kvstore/PartManager.cpp
zhangguoqing: done!

src/meta/MetaServiceUtils.cpp
zhangguoqing: done!

zhangguoqing:
Now, the PR does not store the leader parts information(`leader_partIds`) of heartbeat from storage node.

src/meta/client/MetaClient.cpp
```diff
@@ -26,12 +25,12 @@ MetaClient::MetaClient(std::shared_ptr<folly::IOThreadPoolExecutor> ioThreadPool
...
                        HostAddr localHost,
                        ClusterID clusterId,
-                       bool sendHeartBeat)
+                       bool inStoraged)
     : ioThreadPool_(ioThreadPool)
     , addrs_(std::move(addrs))
     , localHost_(localHost)
     , clusterId_(clusterId)
-    , sendHeartBeat_(sendHeartBeat) {
+    , inStoraged_(inStoraged) {
```
dangleptr: So how to update the cache on query engine side.

dangleptr:
> Now, the PR does not store the leader parts information(`leader_partIds`) of heartbeat from storage node.

That's fine.  Let's open another pr for it.

src/meta/client/MetaClient.cpp
zhangguoqing: Now, sendHeartBeat is enabled by default, either on the query engine or the storage engine side. Whether to update the local cache is determined after each heartbeat based on the last update time of the metad. Only hosts where inStoraged is true will be recorded on the meta server.

src/meta/client/MetaClient.cpp
```diff
@@ -60,17 +59,17 @@ MetaClient::~MetaClient() {
...
-        loadCfg();
+    auto ret = heartbeat().get();
+    if (!ret.ok() && ret.status() != Status::LeaderChanged()) {
+        LOG(ERROR) << "Heartbeat failed, status:" << ret.status();
+        ready_ = false;
+        return ready_;
+    }
+
+    bool ldRet = loadData();
+    bool lcRet = loadCfg();
```
critical27: When `FLAGS_local_config` is true, we don't load config from meta any more, this mode is used for local test (optimize configs).

src/meta/client/MetaClient.cpp
```diff
@@ -122,11 +115,15 @@ void MetaClient::heartBeatThreadFunc() {
...
     }
-}
 
-void MetaClient::loadDataThreadFunc() {
-    loadData();
-    addLoadDataTask();
+    // if MetaServer has some changes, refesh the localCache_
+    if (localLastUpdateTime_ < metadLastUpdateTime_) {
+        bool ldRet = loadData();
+        bool lcRet = loadCfg();
```
critical27: ditto

src/meta/client/MetaClient.cpp
zhangguoqing: done!

src/meta/client/MetaClient.cpp
zhangguoqing: done!

src/meta/client/MetaClient.cpp
```diff
@@ -1323,17 +1318,31 @@ StatusOr<SchemaVer> MetaClient::getNewestEdgeVerFromCache(const GraphSpaceID& sp
...
     nebula::cpp2::HostAddr thriftHost;
     thriftHost.set_ip(localHost_.first);
     thriftHost.set_port(localHost_.second);
     req.set_host(std::move(thriftHost));
     req.set_cluster_id(clusterId_.load());
+
+    if (inStoraged_ && listener_ != nullptr) {
+        std::unordered_map<GraphSpaceID, std::vector<PartitionID>> leaderIds;
+        listener_->fetchLeaderInfo(leaderIds);
+        if (leaderIds_ != leaderIds) {
```
critical27: I've been thinking about this, if we want to use the leader info in heartbeat later, we'd better include it in every request. Otherwise, we can't judge the difference from stale old data and new data. 

But for this PR, it is not important, let me take care of it in another PR.

src/meta/processors/BaseProcessor.h
```diff
@@ -141,7 +142,7 @@ class BaseProcessor {
     /**
      * General put function.
      * */
-    void doPut(std::vector<kvstore::KV> data);
+    void doPut(std::vector<kvstore::KV> data, bool refresh = true);
```
dangleptr: IIRC,  for balance procedure,  it will not call doPut interface inside BaseProcessor.

And in most cases, we don't want the lastUpdateTime being updated.  

So we'd better control it by each processor not in the BaseProcessor.

src/meta/client/MetaClient.cpp
zhangguoqing: OK, I will make thie report the leader info in every HB request.

src/meta/processors/BaseProcessor.h
zhangguoqing: done!

src/meta/client/MetaClient.cpp
zhangguoqing: done!

src/meta/processors/admin/BalanceTask.cpp
```diff
@@ -152,6 +152,9 @@ void BalanceTask::invoke() {
                     ret_ = Result::FAILED;
                 } else {
                     status_ = Status::REMOVE_PART_ON_SRC;
+                    if (kv_ != nullptr) {
```
dangleptr: We don't need to update the lateUpdateTime for each task. 

We could update the time when all balance plan finished.

zhangguoqing: done!

dangleptr:
@Amber1990Zhang Please update documents about oad_data_interval_secs,  we don't need it anymore.

Issue #1032, opened 2019-10-10
Stop balance plan in progress


dangleptr:
We should support stopping running balance plan.

PR #1238, opened 2019-11-08
Balance stop

critical27:
1. Support stop balance. For all balance tasks which has not started yet, we mark them as stopped. For those tasks which has started, we don't interrupt them.

We can run balance data stop in console. Just like balance data, it will return running balance plan id (if no valid plan running, an error is returned). You can check the status by balance data id. Once remaining tasks has finished, you can trigger the plan by balance data to continue all tasks stopped previously.

2. Modify leader balance only consider balancing between hosts with valid partitions.


src/meta/processors/admin/Balancer.cpp
```diff
@@ -446,24 +465,27 @@ Balancer::buildLeaderBalancePlan(HostLeaderMap* hostLeaderMap, GraphSpaceID spac
...
+    std::unordered_set<HostAddr> activeHosts;
+    for (const auto& host : *hostLeaderMap) {
+        // only balance leader between hosts which have valid partition
+        if (!allHostParts[host.first].empty()) {
+            activeHosts.emplace(host.first);
+            leaderHostParts[host.first] = std::move((*hostLeaderMap)[host.first][spaceId]);
+        }
+    }
+
+    size_t avgLoad = leaderParts / activeHosts.size();
```
dangleptr: Check the activeHosts size

src/meta/processors/admin/Balancer.cpp
```diff
@@ -78,14 +89,18 @@ bool Balancer::recovery() {
...
+            {
+                std::lock_guard<std::mutex> lg(lock);
+                running_ = false;
+                plan_.reset();
+            }
         };
         if (!plan_->recovery()) {
             LOG(ERROR) << "Can't recovery plan " << corruptedPlans[0];
-            plan_->onFinished_();
+            running_ = false;
```
dangleptr: why not call onFinished_ directly?

src/meta/processors/admin/Balancer.cpp
```diff
@@ -78,14 +89,18 @@ bool Balancer::recovery() {
         CHECK_EQ(1, corruptedPlans.size());
         plan_ = std::make_unique<BalancePlan>(corruptedPlans[0], kv_, client_.get());
         plan_->onFinished_ = [this] () {
-            bool expected = true;
-            auto &running = this->running_;   // Get the reference before the captured `this' lost
-            plan_.reset();
-            CHECK(running.compare_exchange_strong(expected, false));
+            // Get the reference before the captured `this' lost
+            auto &lock = this->lock_;
```
dangleptr: We don't need to capture this->lock_ as your logic.

src/meta/processors/admin/Balancer.cpp
```diff
@@ -126,17 +141,22 @@ Status Balancer::buildBalancePlan() {
         }
     }
     plan_->onFinished_ = [this] () {
-        bool expected = true;
-        auto &running = this->running_;   // Get the reference before the captured `this' lost
-        plan_.reset();
-        CHECK(running.compare_exchange_strong(expected, false));
+        // Get the reference before the captured `this' lost
+        auto &lock = this->lock_;
```
dangleptr: ditto

src/meta/processors/admin/Balancer.cpp
```diff
@@ -126,17 +141,22 @@ Status Balancer::buildBalancePlan() {
...
+        {
+            std::lock_guard<std::mutex> lg(lock);
+            running_ = false;
+            plan_.reset();
+        }
     };
     if (plan_->tasks_.empty()) {
-        plan_->onFinished_();
+        running_ = false;
```
dangleptr: why change the logic.

critical27:
Jenkins go

src/meta/processors/admin/Balancer.cpp
```diff
@@ -126,17 +141,22 @@ Status Balancer::buildBalancePlan() {
...
+        plan_.reset();
         return Status::Balanced();
     }
     if (!plan_->saveInStore()) {
-        plan_->onFinished_();
+        running_ = false;
```
dangleptr: ditto

src/meta/processors/admin/Balancer.cpp
```diff
@@ -446,24 +465,27 @@ Balancer::buildLeaderBalancePlan(HostLeaderMap* hostLeaderMap, GraphSpaceID spac
     std::unordered_map<HostAddr, std::vector<PartitionID>> allHostParts;
     getHostParts(spaceId, allHostParts, totalParts);
 
-    size_t avgLoad = leaderParts / hostLeaderMap->size();
+    std::unordered_set<HostAddr> activeHosts;
```
dangleptr: What's the purpose?

src/meta/processors/admin/Balancer.cpp
critical27: Good catch.

src/meta/processors/admin/Balancer.cpp
critical27: Previously we balance leader between all active hosts, but when we add a new machine (it has no valid partition), we should not consider it.

src/meta/processors/admin/Balancer.cpp
critical27: I has a lock in onFinished_, and recovery has been locked outside.

src/meta/processors/admin/Balancer.cpp
```diff
@@ -19,21 +19,26 @@ namespace nebula {
...
             return Status::Error("Can't do balance because there is one corruptted balance plan!");
         }
         if (plan_ == nullptr) {
             LOG(INFO) << "There is no corrupted plan need to recovery, so create a new one";
             auto status = buildBalancePlan();
+            if (!status.ok()) {
+                finish();
+            }
             if (plan_ == nullptr) {
                 LOG(ERROR) << "Create balance plan failed, status " << status.toString();
```
dangleptr: Call finish() here.

critical27: Make sense.

critical27:
Jenkins go

src/kvstore/raftex/RaftPart.cpp
```diff
@@ -1377,7 +1377,8 @@ void RaftPart::processAppendLogRequest(
     lastMsgRecvDur_.reset();
 
     if (req.get_sending_snapshot() && status_ != Status::WAITING_SNAPSHOT) {
-        LOG(INFO) << idStr_ << "Begin to wait for the snapshot";
+        LOG(INFO) << idStr_ << "Begin to wait for the snapshot"
```
darionyaphet: "Begin to wait for the snapshot " ?

critical27: Yeah, if leader found its peer need to get snapshot, it will send a request with `sending_snapshot = true`, and follower will change its own status to `WAIT_SNAPSHOT`.

darionyaphet:
Great work！

jude-zhu:
close #1032

Issue #1027, opened 2019-10-10
Support shrinking cluster


dangleptr:
Currently, we could stop storage instance one by one to shrink the cluster. However, it is not convenient,  especially when we want to stop multi storage instances once. 

We can support this feature by another command "Dismiss ip1, ip2, ip3...". 

After dismissing the hosts,  they will be tagged as dismissed.  Moreover, we could rerun "balance data" to move away the parts on them.

PR #1243, opened 2019-11-08
Support shrink storage by specifying host

critical27:
Support balance data between current host and the host we specified in console (we can  exclude some host).

We can use in console like this:
* Remove current machine
`BALANCE DATA REMOVE 192.168.0.1:50000,192.168.0.1:50001`


critical27:
Jenkins go

src/meta/client/MetaClient.cpp
```diff
@@ -1156,11 +1160,25 @@ folly::Future<StatusOr<bool>> MetaClient::heartbeat() {
...
+        std::vector<nebula::cpp2::HostAddr> tHostDel;
+        tHostDel.reserve(hostDel.size());
+        std::transform(hostDel.begin(), hostDel.end(),
+                       std::back_inserter(tHostDel), [](const auto& h) {
+            nebula::cpp2::HostAddr th;
+            th.set_ip(h.first);
+            th.set_port(h.second);
+            return th;
+        });
+        req.set_host_del(tHostDel);
```
wadeliuyi: maybe std::move(tHostDel) is better

src/meta/client/MetaClient.cpp
```diff
@@ -1156,11 +1160,25 @@ folly::Future<StatusOr<bool>> MetaClient::heartbeat() {
     return future;
 }
 
-folly::Future<StatusOr<int64_t>> MetaClient::balance(bool isStop) {
+folly::Future<StatusOr<int64_t>> MetaClient::balance(std::vector<HostAddr> hostDel,
```
wadeliuyi: because it never stolen the memory of hostDel in the function body, so maybe std::vector<HostAddr> &hostDel is better, this will not call right copy function.

src/meta/processors/admin/Balancer.cpp
```diff
@@ -123,16 +123,20 @@ bool Balancer::getAllSpaces(std::vector<GraphSpaceID>& spaces, kvstore::ResultCo
     return true;
 }
 
-Status Balancer::buildBalancePlan() {
+cpp2::ErrorCode Balancer::buildBalancePlan(std::vector<HostAddr> hostDel) {
```
wadeliuyi: genTask parameter is left reference, so maybe std::vector<HostAddr> &hostDel is better

critical27:
Jenkins go

critical27:
Jenkins go

jude-zhu:
close #1027

Issue #1232, opened 2019-11-07
Strengthen Update vertex/edge function


zlcook:
**Describe the bug(__must be provided__)**
Currently, `Update` vertex/edge function exists the following problems:
* Update  syntax doesn't support `now()` function
* Updating the no-existing properties or no-existing tag doesn't cause an error message
* Updating a no-existing vertex doesn't cause an error message

zhangguoqing:
OK, I will fix it as soon as possible.

PR #1253, opened 2019-11-11
Strengthen the update vertex/edge

zhangguoqing:
colse #1232


zhangguoqing:
Jenkins go!

zhangguoqing:
Jenkins go!

joker-kick:
good

src/graph/UpdateEdgeExecutor.cpp
```diff
@@ -249,6 +249,26 @@ void UpdateEdgeExecutor::execute() {
             return;
         }
         auto rpcResp = std::move(resp).value();
+        for (auto& code : rpcResp.get_result().get_failed_codes()) {
```
wadeliuyi: A little quetion, is there return many failed codes? if it is, we just process the first failed code in the follow logic, because we return at each case.

src/graph/UpdateEdgeExecutor.cpp
```diff
@@ -105,7 +105,7 @@ Status UpdateEdgeExecutor::prepareSet() {
         updateItems_.emplace_back(std::move(updateItem));
     }
     if (updateItems_.empty()) {
-        status = Status::Error("There must be some correct update items.");
+        status = Status::SyntaxError("There must be some correct update items.");
     }
```
bright-starry-sky: Why should we care? I don't think that will happen. WDYT?

bright-starry-sky: This problem has been considered in the parser.yy file, correct? 
```

update_list
    : update_item {
        $$ = new UpdateList();
        $$->addItem($1);
    }
    | update_list COMMA update_item {
        $$ = $1;
        $$->addItem($3);
    }
    ;

update_item
    : name_label ASSIGN expression {
        $$ = new UpdateItem($1, $3);
    }
    | alias_ref_expression ASSIGN expression {
        $$ = new UpdateItem($1, $3);
        delete $1;
    }
    ;
```

src/graph/UpdateEdgeExecutor.cpp
zhangguoqing: Normally there will only be one code in failed_codes, because the current update operation only happens on one partition. Here only the first code is checked and the prompt is returned.

src/graph/UpdateEdgeExecutor.cpp
zhangguoqing: Good catch, thx.

PR #1230, opened 2019-11-07
[doc] Fix docker deployment and csv importer related documents

yixinglu:
- Fix document format error
- Reference nebula-docker-compose repo in document about docker cluster doployment
- Fix nebula-importer url
- Update default storage/meta log level for debug mode

PR #1256, opened 2019-11-11
Change graphd/metad/storaged's vlog level config with default value 0.

CPWstatic:
Logging can have a significant impact on performance, and I found that it resulted in at least a 20 times performance degradation in my test cases. When users try nebula, they usually try out the default configurations and know nothing about them, so I think it makes more sense to set the vlog level with 0 by default.


yixinglu:
I have updated this in #1230

PR #1263, opened 2019-11-13
Fixed the build warning when comliling with gcc 7.4.0

sherman-the-tank:
The erroes are caused by

-Werror=sign-compare
-Werror=suggest-override

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -128,7 +128,6 @@ endif()
 # third-party
 if(NOT ${NEBULA_THIRDPARTY_ROOT} STREQUAL "")
     message(STATUS "Specified NEBULA_THIRDPARTY_ROOT: " ${NEBULA_THIRDPARTY_ROOT})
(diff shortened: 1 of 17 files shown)


CMakeLists.txt
```diff
@@ -195,6 +193,9 @@ add_compile_options(-Wall)
 add_compile_options(-Werror)
 add_compile_options(-Wunused-parameter)
 add_compile_options(-Wshadow)
+# Disable the complain anout comparison between signed and unsigned integers
+# TODO (sye) Need to figure out a better solution
+add_compile_options(-Wno-error=sign-compare)
```
monadbobo: We should decide whether to add this option based on the compiler version.

sherman-the-tank: I prefer not to add this option at the global level at all. We can fix all warnings in our code. The trouble is caused by folly. This is just a temporary solution. That's why I added a TODO here

sherman-the-tank:
Go jenkins

sherman-the-tank:
Jenkins go

dutor:
Closed for now.
If this issue still exists, please reopen it.

PR #1265, opened 2019-11-14
Fix bug when ingesting files

dangleptr:
There are some other problems when using multiple disks.   
Let's open another pr for it,  @darionyaphet   Let's use #1266 to track it.


dangleptr:
Jenkins go

PR #1268, opened 2019-11-14
Add ConcurrentLRUCache

dangleptr:


src/common/base/ConcurrentLRUCache.h
```diff
@@ -0,0 +1,109 @@
...
+            lru_->insert(key, val);
+        }
+
+        StatusOr<V> get(const K& key) {
+            std::lock_guard<std::mutex> guard(lock_);
+            auto v = lru_->get(key);
+            if (v == boost::none) {
+                return Status::Error();
+            }
+            return std::move(v).value();
```
bright-starry-sky: Ask a question，do we need insert it to cache immediate If this key is not exist in cache ?

dangleptr: Agreed,  maybe we need another method  putIfAbsent

bright-starry-sky:
LGTM

bright-starry-sky:
Jenkins go

src/common/base/ConcurrentLRUCache.h
```diff
@@ -0,0 +1,130 @@
...
+     * Otherwise, just return the value for the existed key.
+     * */
+    StatusOr<V> putIfAbsent(const K& key, const V& val, int32_t hint = -1) {
+        return buckets_[bucketIndex(key, hint)].putIfAbsent(key, val);
+    }
+
+    void clear() {
+        for (auto i = 0; i < bucketsNum_; i++) {
+            buckets_[i].clear();
+        }
```
wadeliuyi: is it need buckets_.clear() at here?

src/common/base/test/ConcurrentLRUCacheTest.cpp
```diff
@@ -0,0 +1,79 @@
+/* Copyright (c) 2018 vesoft inc. All rights reserved.
```
darionyaphet: 2019 ?

src/common/base/ConcurrentLRUCache.h
```diff
@@ -0,0 +1,130 @@
...
+        void insert(const K& key, const V& val) {
+            std::lock_guard<std::mutex> guard(lock_);
+            lru_->insert(key, val);
+        }
+
+        StatusOr<V> get(const K& key) {
+            std::lock_guard<std::mutex> guard(lock_);
+            auto v = lru_->get(key);
+            if (v == boost::none) {
+                return Status::Error();
```
darionyaphet: `Status::Error(folly::stringPrintf("%s not found", key))` . would be better ?

src/common/base/ConcurrentLRUCache.h
```diff
@@ -0,0 +1,130 @@
...
+ * attached with Common Clause Condition 1.0, found in the LICENSES directory.
+ */
+
+#ifndef COMMON_BASE_CONCURRENTLRUCACHE_H_
+#define COMMON_BASE_CONCURRENTLRUCACHE_H_
+
+#include "base/Base.h"
+#include "base/StatusOr.h"
+#include <boost/compute/detail/lru_cache.hpp>
+#include <gtest/gtest_prod.h>
```
darionyaphet: why include `gtest`?

src/common/base/ConcurrentLRUCache.h
dangleptr: When we call clear,  what we want is to clear the content in cache, not destroy the buckets.

src/common/base/test/ConcurrentLRUCacheTest.cpp
dangleptr: Good catch

src/common/base/ConcurrentLRUCache.h
dangleptr: It will hurt the performance because miss cache is normal in practice.

src/common/base/ConcurrentLRUCache.h
dangleptr: Because we want to use marco "FRIEND_TEST"

PR #1235, opened 2019-11-07
[doc] add glog curl command 

whitewum:
change v & minloglevel at runtime by http command curl.

It is quit useful for on-site debuging

@Amber1990Zhang help to change to Eng and add to readme.md TOC.

PR #1273, opened 2019-11-15
support expression in update configs

critical27:
1. Support expression in update configs, e.g. 
`update configs storage:rocksdb_column_family_options={
    target_file_size_base=100*1024*1024
}`
2. Temporarily forbid modify `write_buffer_size` in `update configs`, it would cause rocksdb crash. Remove some configs which can't be set dynamically.
3. Remove some unused Status.
4. Support some mutable configs in console, such as `minlevel`, `v`, etc.


src/parser/MutateSentences.cpp
```diff
@@ -162,6 +162,22 @@ std::string UpdateItem::toString() const {
...
+    buf += "=";
+    auto ret = value_->eval();
+    if (!ret.ok()) {
+        return ret.status();
+    }
+    if (Expression::isString(ret.value())) {
+        return Status::Error("Invalid string config value");
+    }
+
+    buf += Expression::toString(ret.value());
```
bright-starry-sky: Like this ！

whitewum:
Changing the log level is the most frequently used command. Definitely, we should provide a document about this operation.

could you give an example doc. So I can abandon #1235

critical27:
> Changing the log level is the most frequently used command. Definitely, we should provide a document about this operation.
> 
> could you give an example doc. So I can abandon #1235

Maybe we can still keep #1235, because if user have run register configs of `minloglevel` before, it is immutable, we still can't update it in console unless clear data on meta.

Anyway, I'll give a brief sample of update `minloglevel` in console to @Amber1990Zhang

critical27:
Jenkins go

PR #1274, opened 2019-11-15
Fixed possible crash after storage service failed to start.

monadbobo:


bright-starry-sky:
Good job, An additional question，What's kind of sign type occur?  Why `setupSignalHandler` can't catch it ?

src/daemons/StorageDaemon.cpp
```diff
@@ -109,6 +109,7 @@ int main(int argc, char *argv[]) {
                                                                       paths);
     if (!gStorageServer->start()) {
         LOG(ERROR) << "Storage server start failed";
+        gStorageServer->stop();
```
dangleptr: Well done.  I want to know why the daemon crash?

monadbobo: You can read the [documentation](https://en.cppreference.com/w/cpp/thread/thread/~thread) for the thread destructor.

> If *this has an associated thread (joinable() == true), std::terminate() is called.

dangleptr: which thread?

monadbobo: > nebula::WebService::wsThread_.

image

dangleptr: I see.  Good catch. 
But i think a better way to solve the problem is to refactor web service.  It has lots of static members inside.  It is hard to handle the lifecycle.  Maybe we should remove them. WDYT?

Of course, we could open another pr for it.

dangleptr:
> Good job, An additional question，What's kind of sign type occur? Why `setupSignalHandler` can't catch it ?

Good question. I want to know it too.

dangleptr:
> > Good job, An additional question，What's kind of sign type occur? Why `setupSignalHandler` can't catch it ?
> 
> Good question. I want to figure it out too.
I tried it locally,  std::terminate() can't be catched by signal(),  we should use std::set_terminate instead.   @bright-starry-sky

bright-starry-sky:
> > Good job, An additional question，What's kind of sign type occur? Why `setupSignalHandler` can't catch it ?
> 
> Good question. I want to figure it out too.
> I tried it locally, std::terminate() can't be catched by signal(), we should use std::set_terminate instead. @bright-starry-sky

Wow, The truth is out！Thanks ! So the terminate should be strengthened.

PR #1284, opened 2019-11-18
Fixed the crash that might occur when stopping the raft service

monadbobo:
1. When the ThriftClientManager object needs to be shared, if the EventBase object was released, the corresponding host was removed from the ThriftClientManager.
2. fixed the compile warning by Clang.


monadbobo:
close #1077

src/common/thrift/ThriftClientManager.h
```diff
@@ -35,7 +35,7 @@ class ThriftClientManager final {
         std::shared_ptr<ClientType>                 // Async thrift client
     >;
 
-    folly::ThreadLocal<ClientMap> clientMap_;
+    ClientMap                     clientMap_;
```
dangleptr: Why not use static clientMap_ directly?

monadbobo: Whether to use static here is the same, because we use the ThriftClientManager object as thread_local.

dangleptr: I mean why not use “static folly::ThreadLocal clientMap”

monadbobo: If we used "static folly::ThreadLocal clientMap", all used clients (meta/storage) will share a clientMap. For example, adminclient and storageclient will likely use the same StorageServiceAsyncClient object. Is this necessary?

dangleptr: I think it's ok for connections.   The same thread share the same connection.  wdyt?

PR #1287, opened 2019-11-19
Fix race condition between rollback and iterator in WAL module

dangleptr:


src/kvstore/wal/FileBasedWalIterator.cpp
```diff
@@ -16,9 +16,10 @@ FileBasedWalIterator::FileBasedWalIterator(
     std::shared_ptr<FileBasedWal> wal,
     LogID startId,
     LogID lastId)
-        : wal_(std::move(wal))
+        : wal_(wal)
         , currId_(startId) {
-    if (lastId >= 0) {
+    holder_ = std::make_unique<folly::RWSpinLock::ReadHolder>(wal_->rollbackLock_);
```
monadbobo: It may be better to use the allocation on the stack here.

src/kvstore/wal/FileBasedWalIterator.cpp
```diff
@@ -16,9 +16,10 @@ FileBasedWalIterator::FileBasedWalIterator(
...
+    if (lastId >= 0 && lastId <= wal_->lastLogId()) {
```
monadbobo: Why is the condition of "lastId <= wal _-> lastLogId() " added here?

critical27: Cause we can't have bigger log id than last log id in wal.

monadbobo: > Cause we can't have bigger log id than last log id in wal.

Why does this happen？

critical27: > > Cause we can't have bigger log id than last log id in wal.
> 
> Why does this happen？

For example, leader want to send some latest log, and acquire a iterator, however the wal is rollbacked.

critical27:
close #943

PR #1294, opened 2019-11-19
Support vertex cache on storage side

dangleptr:
Hope it could be helpful in your case.  @CPWstatic


src/storage/AddVerticesProcessor.cpp
```diff
@@ -30,11 +32,16 @@ void AddVerticesProcessor::process(const cpp2::AddVerticesRequest& req) {
...
             const auto& tags = v.get_tags();
             std::for_each(tags.begin(), tags.end(), [&](auto& tag) {
-                VLOG(4) << "PartitionID: " << partId << ", VertexID: " << v.get_id()
+                VLOG(3) << "PartitionID: " << partId << ", VertexID: " << v.get_id()
                         << ", TagID: " << tag.get_tag_id() << ", TagVersion: " << version;
                 auto key = NebulaKeyUtils::vertexKey(partId, v.get_id(),
                                                      tag.get_tag_id(), version);
                 data.emplace_back(std::move(key), std::move(tag.get_props()));
+                if (FLAGS_enable_vertex_cache && vertexCache_ != nullptr) {
+                    vertexCache_->evict(std::make_pair(v.get_id(), tag.get_tag_id()), partId);
```
critical27: vertexCache_->insert?

src/storage/StorageServiceHandler.cpp
```diff
@@ -27,6 +27,9 @@
     processor->process(req); \
     return f;
 
+DEFINE_int32(vertex_cache_num, 16 * 1000 * 1000, "Total keys inside the cache");
+DEFINE_int32(vertex_cache_bucket_exp, 4, "Total buckets number is 1 << cache_bucket_exp");
```
darionyaphet: 👍

src/storage/AddVerticesProcessor.cpp
darionyaphet: it looks as if...

src/common/base/ConcurrentLRUCache.h
```diff
@@ -188,15 +214,16 @@ class LRU {
                 // cache is full, evict the least recently used item
                 evict();
             }
+            VLOG(3) << "Insert key " << key << ", val " << value;
             // insert the new item
             list_.push_front(key);
-            map_[key] = std::make_tuple(value, list_.begin());
-            VLOG(3) << "Insert key " << key << ", val " << value;
+            map_.emplace(std::move(key), std::make_tuple(std::move(value), list_.begin()));
```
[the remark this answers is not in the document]
dangleptr: Because key, value here is left reference.

src/storage/AddVerticesProcessor.cpp
dangleptr: When updating vertex (insert/update/delete),  we just invalid the key in cache.

bright-starry-sky: Like this, Is it will be insert to the cache on the next hit?

src/storage/QueryBaseProcessor.inl
```diff
@@ -358,6 +361,18 @@ kvstore::ResultCode QueryBaseProcessor<REQ, RESP>::collectVertexProps(
                             const std::vector<PropContext>& props,
                             FilterContext* fcontext,
                             Collector* collector) {
+    if (FLAGS_enable_vertex_cache && vertexCache_ != nullptr) {
+        auto result = vertexCache_->get(std::make_pair(vId, tagId), partId);
+        if (result.ok()) {
+            auto v = std::move(result).value();
+            auto reader = RowReader::getTagPropReader(this->schemaMan_, v, spaceId_, tagId);
+            this->collectProps(reader.get(), "", props, fcontext, collector);
```
bright-starry-sky: I'm still not sure what the purpose for empty of key.  😊

dangleptr: Because we don't have any properties on key for a vertex.

src/storage/AddVerticesProcessor.cpp
dangleptr: Yes.  The cache only be updated when reading.

PR #1299, opened 2019-11-21
update storage/meta metrics

critical27:
1. modify register stats.
2. add heartbeat stats in meta

[supported metric of meta/storage](https://www.yuque.com/nebulagraph/euuoip/otv64x)

### DIFF
--- a/src/common/stats/StatsManager.h
+++ b/src/common/stats/StatsManager.h
@@ -65,7 +65,9 @@ public:
     static void setReportInfo(HostAddr addr, int32_t interval);
 
     // Both register methods return the index to the internal data structure.
-    // This index will be used by addValue() methods
+    // This index will be used by addValue() methods.
+    // Both register methods is not thread safe, and user need to register stats one
+    // by onebefore calling addValue, readStats, readHisto and so on.
     static int32_t registerStats(folly::StringPiece counterName);
     static int32_t registerHisto(folly::StringPiece counterName,
                                  VT bucketSize,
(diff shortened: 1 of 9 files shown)


critical27:
Jenkins go ... plz

src/common/stats/StatsManager.h
```diff
@@ -65,7 +65,9 @@ class StatsManager final {
     static void setReportInfo(HostAddr addr, int32_t interval);
 
     // Both register methods return the index to the internal data structure.
-    // This index will be used by addValue() methods
```
Shylock-Hg: Why change this?The metrics is read/write concurrency by server handlers.

critical27: The `register` method is not for read/write, only for server to reigster stats.

src/common/stats/StatsManager.h
```diff
@@ -65,7 +65,9 @@ class StatsManager final {
...
+    // This index will be used by addValue() methods.
+    // Both register methods is not thread safe, and user need to register stats one
+    // by onebefore calling addValue, readStats, readHisto and so on.
     static int32_t registerStats(folly::StringPiece counterName);
```
dangleptr: You better make it to be private,  and expose a method named "init"  to do "register stats"

critical27: Because the stats of graph/meta/storage is not the same, and we have to specify a counter is a stat or histogram type, so maybe we'd better register them in each module. 

@laura-ding, WDYT?

laura-ding: image
This is graphd/metad/storaged that can be used together, you can put it in StatsManager as a common function. @critical27

critical27: > image
> This is graphd/metad/storaged that can be used together, you can put it in StatsManager as a common function. @critical27

Good suggestion, we could do it in #1301

jude-zhu:
relate to #185

PR #1303, opened 2019-11-21
Fixed possible data loss in raft.

monadbobo:
When in the vote phase, the rollbacklog operation cannot be performed because the leader has not been selected.

### DIFF
--- a/src/kvstore/raftex/RaftPart.cpp
+++ b/src/kvstore/raftex/RaftPart.cpp
@@ -1287,7 +1287,6 @@ void RaftPart::processAskForVoteRequest(
     }
     // Ok, no reason to refuse, we will vote for the candidate
     LOG(INFO) << idStr_ << "The partition will vote for the candidate";
-    resp.set_error_code(cpp2::ErrorCode::SUCCEEDED);
 
     Role oldRole = role_;
     TermID oldTerm = term_;
@@ -1300,14 +1299,20 @@ void RaftPart::processAskForVoteRequest(
     lastMsgRecvDur_.reset();
     weight_ = 1;
 
+    if (req.get_last_log_term() < wal_->lastLogTerm()) {
+        resp.set_error_code(cpp2::ErrorCode::E_TERM_OUT_OF_DATE);
+        return;
+    }
+
+    if (req.get_last_log_id() < wal_->lastLogId()) {
+        resp.set_error_code(cpp2::ErrorCode::E_LOG_STALE);
+        return;
+    }
+    resp.set_error_code(cpp2::ErrorCode::SUCCEEDED);
+
     // If the partition used to be a leader, need to fire the callback
     if (oldRole == Role::LEADER) {
         LOG(INFO) << idStr_ << "Was a leader, need to do some clean-up";
-        if (wal_->lastLogId() > lastLogId_) {
-            LOG(INFO) << idStr_ << "There is one log " << wal_->lastLogId()
-                      << " i did not commit when i was leader, rollback to " << lastLogId_;
-            wal_->rollbackToLog(lastLogId_);
-        }
         // Need to invoke the onLostLeadership callback
         bgWorkers_->addTask(
             [self = shared_from_this(), oldTerm] {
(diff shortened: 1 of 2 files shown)


src/kvstore/raftex/RaftPart.cpp
```diff
@@ -1300,14 +1299,20 @@ void RaftPart::processAskForVoteRequest(
...
+    }
+    resp.set_error_code(cpp2::ErrorCode::SUCCEEDED);
+
     // If the partition used to be a leader, need to fire the callback
     if (oldRole == Role::LEADER) {
         LOG(INFO) << idStr_ << "Was a leader, need to do some clean-up";
-        if (wal_->lastLogId() > lastLogId_) {
-            LOG(INFO) << idStr_ << "There is one log " << wal_->lastLogId()
-                      << " i did not commit when i was leader, rollback to " << lastLogId_;
-            wal_->rollbackToLog(lastLogId_);
```
critical27: I agree with that we should not rollback when a host was leader before, at least according to standard raft, new leader would correct these log for old leader. 

But there is a problem that if we do rollback here, say ABC, A was leader of Term 1 and has one more log than B and C(e.g. it has not been sent out), and then B is elected as Term 2. If A didn't rollback the extra log, A will never accept any log from B, because we restrict the wal log id must be consecutive.

dangleptr:
Could you explain the case in detail?

monadbobo:
> Could you explain the case in detail?

Suppose three nodes, node1 was the leader, then write data, node2 writes successfully, and node3 fails to write for some reason, then starts election at this time, node3 sends a vote request, then node1 should send election failure when accepting this request to node3.

src/kvstore/raftex/RaftPart.cpp
monadbobo: So here we compare term and id as a whole.

dangleptr:
> > Could you explain the case in detail?
> 
> Suppose three nodes, node1 was the leader, then write data, node2 writes successfully, and node3 fails to write for some reason, then starts election at this time, node3 sends a vote request, then node1 should send election failure when accepting this request to node3.

Yes,  is there any problems currently?

monadbobo:
> > > Could you explain the case in detail?
> > 
> > 
> > Suppose three nodes, node1 was the leader, then write data, node2 writes successfully, and node3 fails to write for some reason, then starts election at this time, node3 sends a vote request, then node1 should send election failure when accepting this request to node3.
> 
> Yes, is there any problems currently?

Our current implementation encounters this situation and will roll back the log of node1.According to my understanding, before the leader is elected, it is not allowed to roll back the log because there may be election failure.

dangleptr:
> Our current implementation encounters this situation and will roll back the log

Who rollback the log? Node1 or Node3?

monadbobo:
> > Our current implementation encounters this situation and will roll back the log
> 
> Who rollback the log? Node1 or Node3?

Node1 .

dangleptr:
If node1 rollback the log, it means node3 lastogID >= its lastLogId, but as your description, it should NOT happen.

Do you encounter the case in real cluster? What’s the phenomenon？

monadbobo:
> If node1 rollback the log, it means node3 lastogID >= its lastLogId, but as your description, it should NOT happen.
> 
> Do you encounter the case in real cluster? What’s the phenomenon？

If node1 is rolled back, it means that the lastLogId_ of node1's WAL is larger than the lastLogId_ of the corresponding RaftPart/voteRequest.


This is my pressure test write will cause the check failure in the FileBasedWalIterator constructor and then crash (I put the patch of the PR you submitted will still crash #1287 , and I determine the problem by printing the log due to rollback).

dangleptr:
It’s a valuable case need to be digged. 
Could you post the coredump here,
 maybe it is another problem different with the case in #1287

monadbobo:
> It’s a valuable case need to be digged.
> Could you post the coredump here,
> maybe it is another problem different with the case in #1287

After my pressure test, this modification will cause other problems, so I will modify it and submit it again.

dangleptr:
> > It’s a valuable case need to be digged.
> > Could you post the coredump here,
> > maybe it is another problem different with the case in #1287
> 
> After my pressure test, this modification will cause other problems, so I will modify it and submit it again.

Fine,  the coredump and log will be helpful for digging.   Thanks for taking care of it.

PR #1314, opened 2019-11-22
Add EnumArray

dangleptr:
It will save lots of space occupied when the Enum has only few values.

### DIFF
--- a/src/common/base/EnumArray.h
+++ b/src/common/base/EnumArray.h
@@ -172,7 +172,9 @@ public:
         DCHECK(idx < kTotalBits) << "idx " << idx << ", total " << kTotalBits;
         auto bIdx = blockIndex(idx);
         auto offset = bitOffset(idx);
-        uint64_t mask = v << offset;
+        uint64_t mask =  ((1ul << kLogBitsOfEnum) - 1) << offset;
+        blocks_[bIdx] &= (~mask);
+        mask = v << offset;
         blocks_[bIdx] |= mask;
         VLOG(3) << "idx " << idx
                 << ", v " << static_cast<int32_t>(v)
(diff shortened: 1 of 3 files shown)


src/common/base/EnumArray.h
```diff
@@ -0,0 +1,249 @@
...
+        };
+
+        typedef Stat<number, count> Next;
+    };
+
+    template<size_t n>
+    class isPowerTwo {
+    public:
+        enum {
+            ret = (n != 0) && !(n & (n - 1)),
```
wadeliuyi: this mean ret is true just when n=1,2,4,8,16... like 0x100000...0000?

src/common/base/EnumArray.h
```diff
@@ -0,0 +1,249 @@
...
+    }
+    ~EnumArray() = default;
+
+    void put(size_t idx, E e) {
+        auto v = Transformer::toValue(e);
+        DCHECK(idx < kTotalBits) << "idx " << idx << ", total " << kTotalBits;
+        auto bIdx = blockIndex(idx);
+        auto offset = bitOffset(idx);
+        uint64_t mask = v << offset;
+        blocks_[bIdx] |= mask;
```
wadeliuyi: is need set 0 to blocks_[bIdx]? if call put with same idx but diffirent e, then maybe can not get correct e?

src/common/base/EnumArray.h
dangleptr: Ha,  it just handles the unsigned integer.

src/common/base/EnumArray.h
dangleptr: Good point!  We need to handle it.

PR #1318, opened 2019-11-22
Support badge

darionyaphet:
support star and fork badge


whitewum:
image

darionyaphet:
@whitewum  should reflush it.

Shylock-Hg:
But the star and fork are just in upper right corner.

PR #1319, opened 2019-11-22
Enhance go executor with multi jobs.

CPWstatic:
We found that the strategy of split the work is too much simple, and it doesn't bring too much benefit to us in high load situation. So it is not required for meituan-poc for now. I am about to close it first, and we might take this enhancement in the future.


src/storage/QueryBaseProcessor.inl
```diff
@@ -440,7 +430,7 @@ kvstore::ResultCode QueryBaseProcessor<REQ, RESP>::collectEdgeProps(
                             << prop << ", value " << it->second;
                     return it->second;
                 };
-                auto value = exp_->eval();
+                auto value = exp_->eval(getters);
```
dangleptr: If you could ensure eval is thread-safe,  you could remove the lock directly

Issue #1320, opened 2019-11-23
Memory usage


ownthink:
![image](https://user-images.githubusercontent.com/13991775/69477534-5373db00-0e22-11ea-8e48-0a71907beada.png)

nebula-storaged.conf：
![image](https://user-images.githubusercontent.com/13991775/69477680-fe38c900-0e23-11ea-85de-1d8ef3685663.png)

partition：
CREATE SPACE test(partition_num=1024, replica_factor=1);

edge：
100 million

**Your Environments(__must be provided__)**
  * OS: CentOS Linux release 7.4.1708

whitewum:
![image](https://user-images.githubusercontent.com/50101159/69478512-5674c880-0e2e-11ea-893b-f6b7c1af8169.png)

whitewum:
This is due to raft memory usage.

zhangguoqing:
> This is due to raft memory usage.

nebula-storaged.conf  FYI.
```
--part_man_type=memory
--part_type=simple
```

critical27:
See #1324, we have adjust default memory usage.

PR #1324, opened 2019-11-23
[doc] change default partition num and add hardware requirement

whitewum:
1024 * 32MB(raft) is too large to try in a laptop


whitewum:
close #1320

dutor:
Jenkins go

PR #1330, opened 2019-11-25
[hot-fix] update default configs

critical27:
1. update some default value of configs, such as default_parts_num
2. update default template conf
3. make some irrelevant gflags ignored.


etc/nebula-metad.conf.default
```diff
@@ -34,6 +34,6 @@
 
 ########## Misc #########
 # The default number of parts when a space is created
---default_parts_num=1024
+--default_parts_num=100
```
darionyaphet: What's wrong with too many partitions?

critical27: In previous default configs, we have 8M * 4 raft wal buffer for a partition, that would be 32M * 1024 = 32G (if we have import lots of data). So I change to 8M * 2 * 100 = 1.6G instead.

PR #1339, opened 2019-11-26
[WIP] Fetch graphd tcp connections by procfs

yixinglu:
Wait for `/metrics` http endpoint


knightXun:
how about macos and windows? These operator-systems do not provide procfs

dangleptr:
Let's release it in 2.0.  Close it now.  Thanks @yixinglu

PR #1341, opened 2019-11-27
[hot-fix] speed up storage init

### LINKED ISSUE #1315: Storaged process is running, but 44500 port is not be listened for a long time
**Describe the bug(__must be provided__)**
(1). I start services, and the storaged 44500 port is not been listened:
image
(2). loop up the listening port, and port 44500 is not in the listening list:
```bash
ss -lntp | grep nebula
```
image
(3). 
image
(4).
image

**Your Environments(__must be provided__)**
  on the test machine root@192.168.8.205

critical27:
Since we need to scan last wal when we reopen part, this would take quite a while. So we use the worker thread pool to speed up.

close #1315


src/kvstore/NebulaStore.cpp
```diff
@@ -76,31 +75,81 @@ bool NebulaStore::init() {
...
+                            raftService_->addPartition(part);
+                            part->start(std::move(peers), false);
+                            LOG(INFO) << "Load part " << spaceId << ", " << partId << " from disk";
+
+                            {
+                                folly::RWSpinLock::WriteHolder holder(&lock_);
+                                auto iter = spaces_.find(spaceId);
+                                CHECK(iter != spaces_.end());
+                                iter->second->parts_.emplace(partId, part);
+                            }
```
critical27: This piece is same with `newPart`

jievince:
This worked well for me. It can decrease the time used to storage initialization from hours to 7 minutes when the max wal file size is about 100Mb.

src/kvstore/NebulaStore.cpp
```diff
@@ -76,31 +75,81 @@ bool NebulaStore::init() {
...
+                            auto partMeta = options_.partMan_->partMeta(spaceId, partId);
+                            std::vector<HostAddr> peers;
+                            for (auto& h : partMeta.peers_) {
+                                if (h != storeSvcAddr_) {
+                                    peers.emplace_back(getRaftAddr(h));
+                                    VLOG(1) << "Add peer " << peers.back();
+                                }
+                            }
+                            raftService_->addPartition(part);
+                            part->start(std::move(peers), false);
```
bright-starry-sky: I have a question，What's the difference between this code block and method newPart?

critical27: No difference, only add a counter.

bright-starry-sky: In addition to this. LGTM! Well done！

PR #1353, opened 2019-11-28
Update operations return leader info when leader change event happens

zlcook:
As title

### DIFF
--- a/src/storage/BaseProcessor.h
+++ b/src/storage/BaseProcessor.h
@@ -98,6 +98,18 @@ protected:
         }
     }
 
+    void handleLeaderChanged(GraphSpaceID spaceId, PartitionID partId) {
+        auto addrRet = kvstore_->partLeader(spaceId, partId);
+        if (ok(addrRet)) {
+            auto leader = value(std::move(addrRet));
+            this->pushResultCode(cpp2::ErrorCode::E_LEADER_CHANGED, partId, leader);
+        } else {
+            LOG(ERROR) << "Fail to get part leader, spaceId: " << spaceId
+                       << ", partId: " << partId << ", ResultCode: " << error(addrRet);
+            this->pushResultCode(to(error(addrRet)), partId);
+        }
+    }
+
     nebula::cpp2::HostAddr toThriftHost(const HostAddr& host) {
         nebula::cpp2::HostAddr tHost;
         tHost.set_ip(host.first);
--- a/src/storage/UpdateEdgeProcessor.cpp
+++ b/src/storage/UpdateEdgeProcessor.cpp
@@ -364,16 +364,23 @@ void UpdateEdgeProcessor::process(const cpp2::UpdateEdgeRequest& req) {
             return std::string("");
         },
         [this, partId, edgeKey, req] (kvstore::ResultCode code) {
-            this->pushResultCode(this->to(code), partId);
-            if (code == kvstore::ResultCode::SUCCEEDED) {
-                onProcessFinished(req.get_return_columns().size());
-            } else {
-                LOG(ERROR) << "Failure update edge, spaceId: " << this->spaceId_
+            while (true) {
+                if (code == kvstore::ResultCode::SUCCEEDED) {
+                    onProcessFinished(req.get_return_columns().size());
+                    break;
+                }
+                LOG(ERROR) << "Fail to update edge, spaceId: " << this->spaceId_
                            << ", partId: " << partId
                            << ", src: " << edgeKey.get_src()
                            << ", edge_type: " << edgeKey.get_edge_type()
                            << ", dst: " << edgeKey.get_dst()
                            << ", ranking: " << edgeKey.get_ranking();
+                if (code == kvstore::ResultCode::ERR_LEADER_CHANGED) {
+                    handleLeaderChanged(this->spaceId_, partId);
+                    break;
+                }
+                this->pushResultCode(to(code), partId);
+                break;
             }
             this->onFinished();
         });
--- a/src/storage/UpdateVertexProcessor.cpp
+++ b/src/storage/UpdateVertexProcessor.cpp
@@ -341,12 +341,19 @@ void UpdateVertexProcessor::process(const cpp2::UpdateVertexRequest& req) {
             return std::string("");
         },
         [this, partId, vId, req] (kvstore::ResultCode code) {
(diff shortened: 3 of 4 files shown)


zlcook:
Jenkins go

src/storage/UpdateEdgeProcessor.cpp
```diff
@@ -364,10 +364,22 @@ void UpdateEdgeProcessor::process(const cpp2::UpdateEdgeRequest& req) {
...
         },
         [this, partId, edgeKey, req] (kvstore::ResultCode code) {
-            this->pushResultCode(this->to(code), partId);
+            cpp2::ResultCode thriftResult;
+            thriftResult.set_code(to(code));
+            thriftResult.set_part_id(partId);
+            if (code == kvstore::ResultCode::ERR_LEADER_CHANGED) {
+                nebula::cpp2::HostAddr leader;
+                auto addrRet = kvstore_->partLeader(this->spaceId_, partId);
+                CHECK(ok(addrRet));
```
monadbobo: Here should return failure.

src/storage/UpdateVertexProcessor.cpp
```diff
@@ -341,10 +341,22 @@ void UpdateVertexProcessor::process(const cpp2::UpdateVertexRequest& req) {
...
         },
         [this, partId, vId, req] (kvstore::ResultCode code) {
-            this->pushResultCode(this->to(code), partId);
+            cpp2::ResultCode thriftResult;
+            thriftResult.set_code(to(code));
+            thriftResult.set_part_id(partId);
+            if (code == kvstore::ResultCode::ERR_LEADER_CHANGED) {
+                nebula::cpp2::HostAddr leader;
+                auto addrRet = kvstore_->partLeader(this->spaceId_, partId);
+                CHECK(ok(addrRet));
```
monadbobo: the same as above

critical27:
Maybe we can reuse the method `pushResultCode(cpp2::ErrorCode code, PartitionID partId, HostAddr leader)` in BaseProcessor.

src/storage/UpdateVertexProcessor.cpp
```diff
@@ -341,10 +341,22 @@ void UpdateVertexProcessor::process(const cpp2::UpdateVertexRequest& req) {
...
+                auto addr = value(std::move(addrRet));
+                leader.set_ip(addr.first);
+                leader.set_port(addr.second);
+                thriftResult.set_leader(std::move(leader));
+            }
             if (code == kvstore::ResultCode::SUCCEEDED) {
                 onProcessFinished(req.get_return_columns().size());
             } else {
+                this->codes_.emplace_back(std::move(thriftResult));
```
Shylock-Hg: Why not use `pushResultCode`.

Shylock-Hg: And better use `if ... else if ... else` statements.

zlcook: Ok

src/storage/UpdateEdgeProcessor.cpp
zlcook: > Here should return failure.

Why?

monadbobo: We can't crash the process at will, we must handle the errors that can be handled.

PS: What I mean here is that you need to determine whether addrRet is legal.

zlcook: > We can't crash the process at will, we must handle the errors that can be handled.
> 
> PS: What I mean here is that you need to determine whether addrRet is legal.

You are right

zlcook:
Jenkins go

src/storage/UpdateVertexProcessor.cpp
```diff
@@ -341,12 +341,25 @@ void UpdateVertexProcessor::process(const cpp2::UpdateVertexRequest& req) {
...
+                    auto addrRet = kvstore_->partLeader(this->spaceId_, partId);
+                    if (ok(addrRet)) {
+                        auto leader = value(std::move(addrRet));
+                        this->pushResultCode(to(code), partId, leader);
+                        break;
+                    }
+                    LOG(ERROR) << "Fail to get part leader, spaceId: " << this->spaceId_
+                               << ", partId: " << partId << ", ResultCode: " << error(addrRet);
+                }
+                this->pushResultCode(to(code), partId);
```
yixinglu: Why not extract these same logics to a function like `handleLeaderChanged`?

zlcook: > Why not extract these same logics to a function like `handleLeaderChanged` in base class?

No reason to refuse you...

zlcook:
Jenkins Go

jude-zhu:
Jenkins Go

src/storage/BaseProcessor.h
```diff
@@ -98,6 +98,17 @@ class BaseProcessor {
         }
     }
 
+    void handleLeaderChanged(GraphSpaceID spaceId, PartitionID partId) {
+        auto addrRet = kvstore_->partLeader(spaceId, partId);
+        if (ok(addrRet)) {
+            auto leader = value(std::move(addrRet));
+            this->pushResultCode(cpp2::ErrorCode::E_LEADER_CHANGED, partId, leader);
+        } else {
+            LOG(ERROR) << "Fail to get part leader, spaceId: " << spaceId
```
dangleptr: Be careful.  If we can't get the leader information (which is normal),  we should return the empty leader information and errorcode Leader_change. 

By the way,  could we unify the duplicate code about leader change logic with handleAsync in BaseProcessor

zlcook: Oh my fault

src/storage/BaseProcessor.h
```diff
@@ -98,6 +98,18 @@ class BaseProcessor {
...
+                       << ", partId: " << partId << ", ResultCode: " << error(addrRet);
+            this->pushResultCode(cpp2::ErrorCode::E_LEADER_CHANGED, partId);
```
dangleptr: Return the real error code (error(addrRet)) instead of E_LEADER_CHANGED here.

zlcook: > Return the real error code (error(addrRet)) instead of E_LEADER_CHANGED here.
Good catched

zlcook:
Everybody go on

PR #1355, opened 2019-11-28
Added the wal_sync configuration option

monadbobo:
summary:

This option is mainly used to control whether fsync needs to be called every time
after the WAL of the raft writes data(default was false).


monadbobo:
Jenkins go.

monadbobo:
Jenkins go.

src/common/fs/FileUtils.cpp
```diff
@@ -530,6 +530,13 @@ void FileUtils::Iterator::openFileOrDirectory() {
     status_ = Status::OK();
 }
 
+int FileUtils::sync(int fd) {
+#ifdef HAVE_FDATASYNC
+    return fdatasync(fd);
```
dangleptr: Be careful!  We should NOT use "fdatasync" here.

monadbobo: ok, I will fix this.

src/kvstore/wal/FileBasedWal.cpp
```diff
@@ -280,9 +280,17 @@ void FileBasedWal::closeCurrFile() {
...
-    CHECK_EQ(fsync(currFd_), 0) << strerror(errno);
+    if (lastLogId_ > firstLogId_) {
+        if (::fsync(currFd_) == -1) {
+            LOG(WARNING) << "sync wal \"" << currInfo_->path()
+                         << "\" failed, error: " << strerror(errno);
+        }
+    }
     // Close the file
-    CHECK_EQ(close(currFd_), 0) << strerror(errno);
+    if (::close(currFd_) == -1) {
```
critical27: Maybe we can discard CHECK of `utime` below as well. WDYT?

monadbobo: ok, i will fix this.

src/kvstore/wal/FileBasedWal.cpp
```diff
@@ -283,20 +283,33 @@ void FileBasedWal::closeCurrFile() {
         return;
     }
 
-    CHECK_EQ(fsync(currFd_), 0) << strerror(errno);
+    bool synced = false;
+    if (lastLogId_ > firstLogId_) {
```
critical27, line 287: Why not fsync every time?
BTW, I think if fsync failed, maybe crash would be better, WDYT?

Issue #458, opened 2019-05-29
Support "create index" interface in storage layer


dangleptr:
Subtask #71 

After receiving the command,  storage will record current timestamp,  and auto-generate index when inserting edges/vertexes.  Meanwhile start a background thread to build index for old data.

darionyaphet:
**Index Key Structure :** 

`SpaceID` + `Type` + `IndexID` + `index-values` + `RowKey`

`SpaceID` and `Type` are use to avoid naming conflict.

Using `IndexID` could save some space and it's fixed size.

Should I save `index` in another `ColumnFamily` ?

I'm  afraid the index key will same with another data's key, although the possibility is very small.

dangleptr:
What's the type meaning?

boshengchen:
I have some questions as below ：
1 . ` index(col1, col2) `is equal to` index(col1) + index(col2)` ?  I mean，a index(col1, col2) exists. I want to filter col2 by where clause , whether a new index(col2) needs to be created?
2.  What is the purpose of RowKey？ Is it used to remove edges and vertexes? Or just for key unique?

darionyaphet:
`type` is use to explain the index is on `vertex` or `edge`. It should be very small.

This's use to support prefix scan for `show tag indexes` and `show edge indexes`.

darionyaphet:
`index(col1, col2)` and `index(col1) + index(col2)` are not equipment .

`index(col1, col2)` could support `col1` and `col1+col2`. This is same with `MySQL`

darionyaphet:
Because using `Index ID` to replace the `Index Name` , `type` is useless .

dangleptr:
> SpaceID + IndexID + index-values + RowKey
 * SpaceId is useless,  because each rocksdb instance belongs to one space.
 * We should use a prefix to indicate this is index key.  For example, use  "_i"  as prefix.

darionyaphet:
Because we append `RowKey` at the end of `IndexKey`, it must not be duplicated.

Is there any use for prefix ?

dangleptr:
Example1:
the index key on Tag1, col1 
**"_i" + Tag1 + Col1 + value + vertexId**
To support "find vertexId where tag1.col1 > 10"

Example2:
index key on Tag1-col1, tag2-col2:
**"_i" + Tag1 + Col1 + Tag2 + Col2 + value1 + value2 + vertexId**
To support "find vertexId where tag1.col1 > 10 && tag2.col2 > 20"

[...]

PR #1360, opened 2019-11-30
Index of edges and vertices in storage layer

bright-starry-sky:
Support for edges and vertices insert、delete and update


bright-starry-sky:
Jenkins go

src/common/base/NebulaKeyUtils.cpp
```diff
@@ -8,6 +8,8 @@
 
 namespace nebula {
 
+const std::string kIndexPrefix = "_i_";           // NOLINT
```
dangleptr: Do we still need it?  The first byte could be used as the prefix.

src/common/base/NebulaKeyUtils.cpp
```diff
@@ -88,6 +90,69 @@ std::string NebulaKeyUtils::kvKey(PartitionID partId, const folly::StringPiece&
...
+        raw.append(reinterpret_cast<const char*>(&len), sizeof(int32_t));
+    }
+    return raw;
+}
+
+// static
+std::string NebulaKeyUtils::vertexIndexKey(PartitionID partId, IndexID indexId, VertexID vId,
+                                           const IndexValues& values) {
+    auto raw = indexRaw(values);
+    int32_t item = (partId << 8) | static_cast<uint32_t>(NebulaKeyType::kData);
```
dangleptr: NebulaKeyType::kIndex ?

src/common/base/NebulaKeyUtils.cpp
```diff
@@ -88,6 +90,69 @@ std::string NebulaKeyUtils::kvKey(PartitionID partId, const folly::StringPiece&
...
+    std::string key;
+    key.reserve(kIndexPrefix.size() + kVertexIndexLen + raw.size());
+    key.append(reinterpret_cast<const char*>(&item), sizeof(int32_t))
+       .append(kIndexPrefix.data(), kIndexPrefix.size())
+       .append(reinterpret_cast<const char*>(&indexId), sizeof(IndexID))
+       .append(raw.data(), raw.size())
```
dangleptr: Could we avoid the double copy for "raw"?

src/interface/storage.thrift
```diff
@@ -208,15 +218,17 @@ struct EdgeKeyRequest {
 }
 
 struct DeleteVertexRequest {
-    1: common.GraphSpaceID space_id,
-    2: common.PartitionID  part_id,
-    3: common.VertexID     vid;
+    1: common.GraphSpaceID      space_id,
+    2: common.PartitionID       part_id,
+    3: common.VertexID          vid;
+    4: optional list<IndexItem> indexes,
```
dangleptr: Why we need indexes here?

src/interface/storage.thrift
```diff
@@ -199,6 +207,8 @@ struct AddEdgesRequest {
     2: map<common.PartitionID, list<Edge>>(cpp.template = "std::unordered_map") parts,
     // If true, it equals an upsert operation.
     3: bool overwritable,
+    // index item
+    4: optional list<IndexItem> indexes,
```
dangleptr: Why we need the indexes here?  
Storaged could read the information from metad.  We don't need to pass the information every request.

src/interface/storage.thrift
```diff
@@ -208,15 +218,17 @@ struct EdgeKeyRequest {
...
+    2: common.PartitionID       part_id,
+    3: common.VertexID          vid;
+    4: optional list<IndexItem> indexes,
 }
 
 struct DeleteEdgesRequest {
     1: common.GraphSpaceID space_id,
     // partId => edgeKeys
     2: map<common.PartitionID, list<EdgeKey>>(cpp.template = "std::unordered_map") parts,
+    3: optional list<IndexItem> indexes,
```
dangleptr: ditto

src/interface/storage.thrift
```diff
@@ -282,23 +294,25 @@ struct UpdateItem {
...
-    6: list<binary> return_columns,
-    7: bool insertable,
+    1: common.GraphSpaceID      space_id,
+    2: common.VertexID          vertex_id,
+    3: common.PartitionID       part_id,
+    4: binary                   filter,
+    5: list<UpdateItem>         update_items,
+    6: list<binary>             return_columns,
+    7: bool                     insertable,
+    8: optional list<IndexItem> indexes,
```
dangleptr: ditto

src/interface/storage.thrift
```diff
@@ -282,23 +294,25 @@ struct UpdateItem {
...
-    6: list<binary> return_columns,
-    7: bool insertable,
+    1: common.GraphSpaceID      space_id,
+    2: EdgeKey                  edge_key,
+    3: common.PartitionID       part_id,
+    4: binary                   filter,
+    5: list<UpdateItem>         update_items,
+    6: list<binary>             return_columns,
+    7: bool                     insertable,
+    8: optional list<IndexItem> indexes,
```
dangleptr: ditto

src/storage/AddEdgesProcessor.cpp
```diff
@@ -13,28 +13,116 @@ namespace nebula {
 namespace storage {
 
 void AddEdgesProcessor::process(const cpp2::AddEdgesRequest& req) {
```
dangleptr: asyncAtomicOp will hurt the performance. 
If there is no index,  we should use put interface directly.

src/common/base/NebulaKeyUtils.cpp
bright-starry-sky: > NebulaKeyType::kIndex ?

It's really all set，great！

src/common/base/NebulaKeyUtils.cpp
bright-starry-sky: > Do we still need it? The first byte could be used as the prefix.

make sense !  Will using kIndex for prefix.

src/interface/storage.thrift
bright-starry-sky: > Why we need the indexes here?
> Storaged could read the information from metad. We don't need to pass the information every request.

Aha, Index meta wasn't available then , let me improve it now.

src/storage/test/IndexTest.cpp
```diff
@@ -0,0 +1,674 @@
...
+        for (auto indexId = 3001; indexId < 3010; indexId++) {
+            auto prefix = NebulaKeyUtils::indexPrefix(partId, indexId);
+            std::unique_ptr<kvstore::KVIterator> iter;
+            EXPECT_EQ(kvstore::ResultCode::SUCCEEDED, kv->prefix(0, partId, prefix, &iter));
+            int32_t rowCount = 0;
+            while (iter->valid()) {
+                rowCount++;
+                iter->next();
+            }
+            EXPECT_EQ(10, rowCount);
```
critical27: Why we have 10 lines of index with only 9 lines of data?

bright-starry-sky:
Addressed all @dangleptr comments since 2019-12-15 10:15:00:
1:  Using kIndex instead of kData.
2: Removed 'kIndexPrefix' from index key.
3: Removed 'IndexItem' from all thrift requests.
4: Avoid the use of `asyncAtomicOp` if no index exists.
5: Avoid the double copy for `NebulaKeyUtils::vertexIndexKey` and `NebulaKeyUtils::edgeIndexKey`.

src/common/base/NebulaKeyUtils.cpp
```diff
@@ -88,6 +88,59 @@ std::string NebulaKeyUtils::kvKey(PartitionID partId, const folly::StringPiece&
...
+std::string NebulaKeyUtils::edgeIndexKey(PartitionID partId, IndexID indexId,
+                                         VertexID srcId, EdgeRanking rank,
+                                         VertexID dstId, const IndexValues& values) {
+    int32_t item = (partId << 8) | static_cast<uint32_t>(NebulaKeyType::kIndex);
+    std::string key;
+    key.reserve(256);
+    key.append(reinterpret_cast<const char*>(&item), sizeof(int32_t))
+       .append(reinterpret_cast<const char*>(&indexId), sizeof(IndexID));
+    indexRaw(values, key);
+    key.append(reinterpret_cast<const char*>(&srcId), sizeof(VertexID))
```
critical27: Why we don't encode edgeType here?

critical27: I am sorry for my poor understanding
1. Why we don't need to encode edge type here?
2. Why we encode the `edge key` in index key instead of index value?
3. I guess we can scan all index of first property, am I right?

bright-starry-sky: > I am sorry for my poor understanding
> 
> 1. Why we don't need to encode edge type here?
IndexId can  identify the `edgetype` or `tagId`
> 2. Why we encode the `edge key` in index key instead of index value?
we only need scan key by prefix or range, so we need encode the values into key , I don't know if my answer is clear. 
> 3. I guess we can only scan all index of first property in IndexValues, am I right?
Yes，you are right, if is a range query ,only scan first property. if is a equivalent query, can be first N properties.

src/storage/AddEdgesProcessor.cpp
```diff
@@ -13,27 +13,133 @@ namespace nebula {
...
+        batchHolder->put(std::move(key), std::move(prop));
+        for (auto& index : indexes_) {
+            /**
+             * skip the in-edge.
+             **/
+            if (index.get_schema() == type) {
+                cpp2::EdgeKey ek(apache::thrift::FragileConstructor::FRAGILE,
+                        srcId, type, rank, dstId);
+                auto ekVal = NebulaKeyUtils::edgePrefix(partId, srcId, type, rank, dstId);
+                newIndexes[ekVal] = std::make_pair(edge.get_props(), std::make_pair(ek, index));
```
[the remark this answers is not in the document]
bright-starry-sky: > Why we encode the edgeKey both in key and value of `newIndexes`? I guess you mean to prevent from different edge of same property value? Maybe `map<propertry, vector<edgekey>>` is enough?

good point ！There are two points here：
1: As your said,  different edge of same property value.
2: Same edge but different version number in a request.

src/common/base/NebulaKeyUtils.cpp
critical27: Maybe the second one need a little more hint.

src/storage/AddEdgesProcessor.cpp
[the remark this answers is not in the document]
critical27: Still a little confused, I have not see the place we use the key of `newIndexes` (aka `ekVal`). Besides, `ek` and `ekVal` is almost the same.

As the second case `Same edge but different version number in a request`, we didn't encode any version here. I see you insert a new index in `newIndex` and discards all index of the edges we insert in `obsoleteIndex`.

src/storage/test/IndexTest.cpp
```diff
@@ -0,0 +1,670 @@
...
```
critical27: Why we will have one more index row?

src/common/base/NebulaKeyUtils.cpp
bright-starry-sky: > Maybe the second one need a little more hint.

for second one, the index key format is :
`partId + indexId + propValues + srcId + rank + dstId . `
the propValues format is : 
`edgeType.col1 + edgeType.col2 + degeType.col3+... `. the column value is encoded(int and timestamp is big-endain).

src/storage/AddEdgesProcessor.cpp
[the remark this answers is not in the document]
bright-starry-sky: > Still a little confused, I have not see the place we use the key of `newIndexes` (aka `ekVal`). Besides, `ek` and `ekVal` is almost the same.
> 
> As the second case `Same edge but different version number in a request`, we didn't encode any version here. I see you insert a new index in `newIndex` and discards all index of the edges we insert in `obsoleteIndex`.

For example : 
insert below edges in one request:
kv (part1_src1_edgeType1_rank1_dst1 , v1)
kv (part1_src1_edgeType1_rank1_dst1 , v2)
kv (part1_src1_edgeType1_rank1_dst1 , v3)
kv (part1_src1_edgeType1_rank1_dst1 , v4)

Ultimately, only one record is stored in newIndexes ：kv (part1_src1_edgeType1_rank1_dst1 , v4) . It's just what I need.Answer your question，the `ekVal` only is a unique key.

src/storage/test/IndexTest.cpp
bright-starry-sky: Aha, Let's do the math：
**3part * 10vertex * 9tag = 270(row)**
```
for (auto partId = 0; partId < 3; partId++) {
        for (auto vertexId = 10 * partId; vertexId < 10 * (partId + 1); vertexId++) {
            EXPECT_EQ(9, rowCount);
        }
    }
```

**3part * 9index(same with tag num) * 10 vertex = 270(row)**
```
    for (auto partId = 0; partId < 3; partId++) {
        for (auto indexId = 3001; indexId < 3010; indexId++) {
            EXPECT_EQ(10, rowCount); <<-- this is means 10 vertices.
        }
    }
```

src/storage/AddEdgesProcessor.cpp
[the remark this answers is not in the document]
critical27: 👍

src/storage/test/IndexTest.cpp
critical27: My bad, got it.

src/storage/test/DeleteVertexTest.cpp
```diff
@@ -20,9 +20,10 @@ namespace storage {
 TEST(DeleteVertexTest, SimpleTest) {
     fs::TempDir rootPath("/tmp/DeleteVertexTest.XXXXXX");
     std::unique_ptr<kvstore::KVStore> kv(TestUtils::initKV(rootPath.path()));
+    auto schemaMan = TestUtils::mockSchemaMan();
```
darionyaphet: Do we need `schema manager` at here ?

bright-starry-sky: Yes, we need check the existing index via schemaMan.

src/storage/AddVerticesProcessor.cpp
```diff
@@ -22,30 +22,133 @@ void AddVerticesProcessor::process(const cpp2::AddVerticesRequest& req) {
     version = folly::Endian::big(version);
 
     const auto& partVertices = req.get_parts();
-    auto spaceId = req.get_space_id();
+    spaceId_ = req.get_space_id();
     callingNum_ = partVertices.size();
+    auto iRet = schemaMan_->getTagIndexes(spaceId_);
```
darionyaphet: I have added an index manager maybe you can use it to replace `schemaMan_`. I will release it later.

bright-starry-sky: That's cool !!

bright-starry-sky:
Resolve the conflict

src/storage/AddEdgesProcessor.cpp
```diff
@@ -13,27 +13,133 @@ namespace nebula {
...
+                data.emplace_back(std::move(key), std::move(edge.get_props()));
+            });
+            doPut(spaceId_, partId, std::move(data));
         });
-        doPut(spaceId, partId, std::move(data));
+    } else {
+        std::for_each(req.parts.begin(), req.parts.end(), [&](auto& partEdges){
+            auto partId = partEdges.first;
+            const auto &edges = partEdges.second;
+            auto atomic = [&]() -> std::string {
```
critical27: Maybe you should capture by value here, for example partId could be released.

src/common/base/NebulaKeyUtils.cpp
```diff
@@ -88,6 +88,59 @@ std::string NebulaKeyUtils::kvKey(PartitionID partId, const folly::StringPiece&
     return key;
 }
 
+// static
+void NebulaKeyUtils::indexRaw(const IndexValues &values, std::string& raw) {
+    std::vector<int32_t> colsLen;
```
wadeliuyi: colsLen init with size first better

src/storage/AddEdgesProcessor.cpp
```diff
@@ -13,27 +13,133 @@ namespace nebula {
...
-        std::vector<kvstore::KV> data;
-        std::for_each(partEdges.second.begin(), partEdges.second.end(), [&](auto& edge){
-            VLOG(4) << "PartitionID: " << partId << ", VertexID: " << edge.key.src
-                    << ", EdgeType: " << edge.key.edge_type << ", EdgeRanking: " << edge.key.ranking
-                    << ", VertexID: " << edge.key.dst << ", EdgeVersion: " << version;
-            auto key = NebulaKeyUtils::edgeKey(partId, edge.key.src, edge.key.edge_type,
-                                               edge.key.ranking, edge.key.dst, version);
-            data.emplace_back(std::move(key), std::move(edge.get_props()));
+    if (indexes_.empty()) {
+        std::for_each(req.parts.begin(), req.parts.end(), [&](auto& partEdges){
```
wadeliuyi: no need capture all local variables

bright-starry-sky: your means `[&]` ?

src/storage/AddEdgesProcessor.cpp
bright-starry-sky: Sorry, I don't quite follow you. your means [&] ?

src/storage/BaseProcessor.inl
```diff
@@ -83,5 +83,54 @@ void BaseProcessor<RESP>::doRemoveRange(GraphSpaceID spaceId,
...
+                                          const nebula::cpp2::ColumnDef& col) {
+    auto res = RowReader::getPropByName(reader, col.get_name());
+    if (!ok(res)) {
+        LOG(ERROR) << "Skip bad column prop " << col.get_name();
+        return "";
+    }
+    auto&& v = value(std::move(res));
+    switch (col.get_type().get_type()) {
+        case nebula::cpp2::SupportedType::BOOL: {
+            return folly::to<std::string>(boost::get<bool >(v));
```
dangleptr: folly::to is not reasonable for encode bool

src/storage/BaseProcessor.inl
```diff
@@ -83,5 +83,54 @@ void BaseProcessor<RESP>::doRemoveRange(GraphSpaceID spaceId,
...
+        }
+        case nebula::cpp2::SupportedType::INT:
+        case nebula::cpp2::SupportedType::TIMESTAMP: {
+            std::string raw;
+            raw.reserve(sizeof(int64_t));
+            auto val = folly::Endian::big(boost::get<int64_t>(v));
```
dangleptr: Take care of negative numbers.

src/storage/BaseProcessor.inl
```diff
@@ -83,5 +83,54 @@ void BaseProcessor<RESP>::doRemoveRange(GraphSpaceID spaceId,
...
+            raw.append(reinterpret_cast<const char*>(&val), sizeof(int64_t));
+            return raw;
+        }
+        case nebula::cpp2::SupportedType::FLOAT:
+        case nebula::cpp2::SupportedType::DOUBLE: {
+            return folly::to<std::string>(boost::get<double>(v));
```
dangleptr: Don't use folly::to

src/interface/storage.thrift
```diff
@@ -163,6 +163,12 @@ struct Edge {
     2: binary props,
 }
 
+struct IndexItem {
+     1: required common.IndexID          index_id,
+     2: required i32                     schema,
```
dangleptr: Why "schema" type is i32? 

It seems the name should be tagIdOrEdgeType.

src/storage/mutate/AddEdgesProcessor.cpp
```diff
@@ -13,27 +13,133 @@ namespace nebula {
...
+    std::unique_ptr<kvstore::BatchHolder> batchHolder = std::make_unique<kvstore::BatchHolder>();
+    std::map<std::string, std::pair<std::string,
+             std::pair<cpp2::EdgeKey, cpp2::IndexItem>>> newIndexes;
+    std::for_each(edges.begin(), edges.end(), [&](auto& edge){
+        auto prop = edge.get_props();
+        auto type = edge.key.edge_type;
+        auto srcId = edge.key.src;
+        auto rank = edge.key.ranking;
+        auto dstId = edge.key.dst;
+        VLOG(4) << "PartitionID: " << partId << ", VertexID: " << srcId
```
dangleptr: The highest vlog level is 3 in storage layer.

src/storage/mutate/AddEdgesProcessor.cpp
```diff
@@ -13,27 +13,133 @@ namespace nebula {
...
+                << ", EdgeType: " << type << ", EdgeRanking: " << rank
+                << ", VertexID: " << dstId << ", EdgeVersion: " << version;
+        auto key = NebulaKeyUtils::edgeKey(partId, srcId, type, rank, dstId, version);
+        batchHolder->put(std::move(key), std::move(prop));
+        for (auto& index : indexes_) {
+            /**
+             * skip the in-edge.
+             **/
+            if (index.get_schema() == type) {
+                cpp2::EdgeKey ek(apache::thrift::FragileConstructor::FRAGILE,
```
dangleptr: Don't use "apache::thrift::FragileConstructor::FRAGILE" any more. It has been deleted in the latest fbthrift.

src/storage/mutate/AddEdgesProcessor.cpp
```diff
@@ -13,27 +13,133 @@ namespace nebula {
...
+            };
+            this->kvstore_->asyncAtomicOp(spaceId_, partId, atomic, callback);
+        });
+    }
+}
+
+std::string AddEdgesProcessor::addEdges(int64_t version, PartitionID partId,
+                                        const std::vector<cpp2::Edge>& edges) {
+    std::unique_ptr<kvstore::BatchHolder> batchHolder = std::make_unique<kvstore::BatchHolder>();
+    std::map<std::string, std::pair<std::string,
```
dangleptr: Add more comments for this structure.  
It confused me.

src/storage/mutate/AddEdgesProcessor.h
```diff
@@ -28,6 +29,19 @@ class AddEdgesProcessor : public BaseProcessor<cpp2::ExecResponse> {
...
             : BaseProcessor<cpp2::ExecResponse>(kvstore, schemaMan, stats) {}
+
+    std::string addEdges(int64_t version, PartitionID partId,
+                         const std::vector<cpp2::Edge>& edges);
+
+    std::string obsoleteIndex(PartitionID partId,
+                              const std::pair<cpp2::EdgeKey, cpp2::IndexItem>& pair);
+
+    std::string newIndex(PartitionID partId,
+            const std::pair<std::string, std::pair<cpp2::EdgeKey, cpp2::IndexItem>>& index);
```
dangleptr: Alignment.

src/storage/mutate/AddEdgesProcessor.cpp
```diff
@@ -13,27 +13,133 @@ namespace nebula {
...
+        auto result = kvstore_->get(spaceId_, partId, indexKey, &val);
+        if (result == kvstore::ResultCode::SUCCEEDED) {
+            return indexKey;
+        }
+    }
+    return "";
+}
+
+std::string AddEdgesProcessor::newIndex(PartitionID partId,
+        const std::pair<std::string, std::pair<cpp2::EdgeKey, cpp2::IndexItem>>& index) {
```
dangleptr: Alignment

src/storage/mutate/AddEdgesProcessor.cpp
```diff
@@ -13,27 +13,133 @@ namespace nebula {
...
+    std::for_each(edges.begin(), edges.end(), [&](auto& edge){
+        auto prop = edge.get_props();
+        auto type = edge.key.edge_type;
+        auto srcId = edge.key.src;
+        auto rank = edge.key.ranking;
+        auto dstId = edge.key.dst;
+        VLOG(4) << "PartitionID: " << partId << ", VertexID: " << srcId
+                << ", EdgeType: " << type << ", EdgeRanking: " << rank
+                << ", VertexID: " << dstId << ", EdgeVersion: " << version;
+        auto key = NebulaKeyUtils::edgeKey(partId, srcId, type, rank, dstId, version);
```
dangleptr: When inserting an edge,   the operations looks like:
1.  Get the original value for the edgeKey.
2.  Delete the old index for the related indexId
3.  Insert the new index and data.

src/storage/mutate/AddEdgesProcessor.cpp
bright-starry-sky: Good point ! acquire knowledge from this.

src/interface/storage.thrift
dangleptr: A better name will be helpful when reading your code.  
So we'd better give a more meaningful name here.  
How about "tagOrEdge", short and meaningful.

bright-starry-sky: > A better name will be helpful when reading your code.
> So we'd better give a more meaningful name here.
> How about "tagOrEdge", short and meaningful.

I agree with `tagOrEdge`

bright-starry-sky:
Addressed all comments since 2019-12-26 23:00:00
Also , encode `int64_t` and `double` for string order completed.

bright-starry-sky:
Jenkins go

src/storage/mutate/AddEdgesProcessor.cpp
```diff
@@ -13,27 +13,163 @@ namespace nebula {
...
     // Switch version to big-endian, make sure the key is in ordered.
     version = folly::Endian::big(version);
 
     callingNum_ = req.parts.size();
+    auto iRet = schemaMan_->getEdgeIndexes(spaceId_);
+    if (iRet.ok()) {
+        for (auto& index : iRet.value()) {
+            indexes_.emplace_back(index);
+        }
+    }
```
dangleptr: What if !iRet.ok(),  return error directly.

src/storage/mutate/AddEdgesProcessor.cpp
```diff
@@ -13,27 +13,163 @@ namespace nebula {
...
+         * step 1 , Delete old version index if exists.
+         */
+        auto oi = obsoleteIndex(partId, index.first, index.second.second);
+        if (!oi.empty()) {
+            batchHolder->remove(std::move(oi));
+        }
+        /*
+         * step 2 , Insert new edge data
+         */
+        auto key = index.first;
```
dangleptr: auto&

src/storage/mutate/AddEdgesProcessor.cpp
```diff
@@ -13,27 +13,163 @@ namespace nebula {
...
+        auto prop = index.second.first;
```
dangleptr: auto&

src/storage/mutate/AddEdgesProcessor.cpp
```diff
@@ -13,27 +13,163 @@ namespace nebula {
...
+        const auto& cols = index.get_cols();
+        auto values = collectIndexValues(reader.get(), cols);
+        auto indexKey = NebulaKeyUtils::edgeIndexKey(partId,
+                                                     index.get_index_id(),
+                                                     NebulaKeyUtils::getSrcId(rawKey),
+                                                     NebulaKeyUtils::getRank(rawKey),
+                                                     NebulaKeyUtils::getDstId(rawKey),
+                                                     values);
+        std::string val;
+        auto result = kvstore_->get(spaceId_, partId, indexKey, &val);
```
dangleptr: What's the purpose about this get.

bright-starry-sky: > What's the purpose about this get.

First, we need to check if there is an older version of this edge. If it exists, then index row corresponding to old version data must also exist. So this `get` is redundant. you are right. good point.

src/storage/mutate/AddEdgesProcessor.cpp
bright-starry-sky: This schemaMan_ is going to be replaced by indexMan_ .
Currently, return` !iRet.ok()` means that the index does not exist

src/storage/BaseProcessor.inl
```diff
@@ -83,5 +83,54 @@ void BaseProcessor<RESP>::doRemoveRange(GraphSpaceID spaceId,
...
+        default:
+            LOG(ERROR) << "Unknown type: "
+                       << static_cast<int32_t>(col.get_type().get_type());
+    }
+    return "";
+}
+
+template <typename RESP>
+IndexValues BaseProcessor<RESP>::collectIndexValues(RowReader* reader,
+    const std::vector<nebula::cpp2::ColumnDef>& cols) {
```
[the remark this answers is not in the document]
bright-starry-sky: > Alignment

I was stumped 🤔, This parameter is too long. What should I do？

src/storage/BaseProcessor.inl
```diff
@@ -83,5 +83,54 @@ void BaseProcessor<RESP>::doRemoveRange(GraphSpaceID spaceId,
         });
 }
 
+template <typename RESP>
+std::string BaseProcessor<RESP>::indexStr(RowReader* reader,
```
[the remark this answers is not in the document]
bright-starry-sky: Of course, I'll do it.

src/storage/mutate/AddEdgesProcessor.cpp
bright-starry-sky: > auto&

Aha, because the argument of method `batchHolder->put` is an rvalue reference ，Here I use std::move to cast the reference.  So `auto&` doesn't work.

src/storage/mutate/DeleteVertexProcessor.cpp
```diff
@@ -17,54 +17,131 @@ void DeleteVertexProcessor::process(const cpp2::DeleteVertexRequest& req) {
...
+                                                          spaceId,
+                                                          tagId);
+                const auto& cols = index.get_cols();
+                auto values = collectIndexValues(reader.get(), cols);
+                auto indexKey = NebulaKeyUtils::vertexIndexKey(partId,
+                                                               indexId,
+                                                               vId,
+                                                               values);
+                std::string val;
+                auto result = kvstore_->get(spaceId, partId, indexKey, &val);
```
dangleptr: We don't need the 'get' operation.

src/storage/mutate/DeleteVertexProcessor.cpp
```diff
@@ -17,54 +17,131 @@ void DeleteVertexProcessor::process(const cpp2::DeleteVertexRequest& req) {
...
-            auto addr = value(std::move(addrRet));
-            leader.set_ip(addr.first);
-            leader.set_port(addr.second);
-            thriftResult.set_leader(std::move(leader));
+    /**
+     * There may be different versions of the same vertex,
+     * so delete duplicate index keys is required.
+     **/
+    if (!indexes.empty()) {
+        sort(indexes.begin(), indexes.end());
```
dangleptr: We should skip the duplicate versions key when iterating.

src/storage/mutate/UpdateEdgeProcessor.cpp
```diff
@@ -260,11 +262,50 @@ std::string UpdateEdgeProcessor::updateAndWriteBack() {
...
-    data.emplace_back(key_, updater_->encode());
-    auto log = kvstore::encodeMultiValues(kvstore::OP_MULTI_PUT, data);
-    return log;
+    std::unique_ptr<kvstore::BatchHolder> batchHolder = std::make_unique<kvstore::BatchHolder>();
+    auto nVal = updater_->encode();
+    batchHolder->put(std::move(key_), std::move(nVal));
+    if (!indexes_.empty()) {
+        for (auto& index : indexes_) {
+            auto indexId = index.get_index_id();
+            if (index.get_tagOrEdge() == edgeKey.edge_type) {
```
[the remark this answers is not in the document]
bright-starry-sky: There is no possible, Because ID is unique whether it's TagId or EdgeType

src/storage/mutate/DeleteVertexProcessor.cpp
bright-starry-sky: Make sense!

bright-starry-sky:
Jenkins go

bright-starry-sky:
Jenkins go

dutor:
Jenkins go

src/common/base/NebulaKeyUtils.h
```diff
@@ -192,6 +214,68 @@ class NebulaKeyUtils final {
...
+     * for example as below :
+     *    9223372036854775807     -> "\377\377\377\377\377\377\377\377"
+     *    1                       -> "\200\000\000\000\000\000\000\001"
+     *    0                       -> "\200\000\000\000\000\000\000\000"
+     *    -1                      -> "\177\377\377\377\377\377\377\377"
+     *    -9223372036854775808    -> "\000\000\000\000\000\000\000\000"
+     */
+
+    static std::string encodeInt64(int64_t v) {
+        v ^= folly::to<int64_t>(1) << 63;
```
critical27: How about define a constexpr?

bright-starry-sky: A good idea, Do you mean for method `encodeInt64` or constant `folly::to<int64_t>(1) << 63`?

src/storage/mutate/AddEdgesProcessor.cpp
```diff
@@ -13,27 +13,156 @@ namespace nebula {
...
+        auto dstId = edge.key.dst;
+        VLOG(3) << "PartitionID: " << partId << ", VertexID: " << srcId
+                << ", EdgeType: " << type << ", EdgeRanking: " << rank
+                << ", VertexID: " << dstId << ", EdgeVersion: " << version;
+        auto key = NebulaKeyUtils::edgeKey(partId, srcId, type, rank, dstId, version);
+        newIEdges[key] = std::move(prop);
     });
+    for (auto& e : newIEdges) {
+        for (auto& index : indexes_) {
+            if (NebulaKeyUtils::getEdgeType(e.first) == index.get_tagOrEdge()) {
```
dangleptr: 1.  It seems that we don't need the if condition.  
2. We'd better get the original value in the first step before processing all indexes_

bright-starry-sky:
Performance optimization for index write . 
The performance report please refer to #1580

src/storage/mutate/AddEdgesProcessor.cpp
```diff
@@ -13,27 +13,157 @@ namespace nebula {
...
+                << ", EdgeType: " << type << ", EdgeRanking: " << rank
+                << ", VertexID: " << dstId << ", EdgeVersion: " << version;
+        auto key = NebulaKeyUtils::edgeKey(partId, srcId, type, rank, dstId, version);
+        newEdges[key] = std::move(prop);
     });
+    for (auto& e : newEdges) {
+        std::string val;
+        std::unique_ptr<RowReader> nReader;
+        auto edgeType = NebulaKeyUtils::getEdgeType(e.first);
+        for (auto& index : indexes_) {
```
dangleptr: Preprocess the indexes_,   for each edgeType,  we just need to get the original value one time.

src/storage/StorageFlags.cpp
```diff
@@ -16,3 +16,5 @@ DEFINE_int32(waiting_catch_up_interval_in_secs, 30,
 DEFINE_int32(waiting_new_leader_retry_times, 30, "retry times when waiting for catching up data");
 DEFINE_int32(waiting_new_leader_interval_in_secs, 5,
              "interval between two requests for catching up state");
+DEFINE_bool(ignore_index_check_pre_insert, false,
```
dangleptr: Well done.

src/interface/common.thrift
```diff
@@ -100,4 +102,10 @@ struct Pair {
     2: string value,
 }
 
+struct IndexItem {
+     1: required IndexID          index_id,
+     2: required i32              tagOrEdge,
```
dangleptr: By the way,   why we need tagOrEdge here?  index_id is not enough?

And where did we use the IndexItem?

src/common/base/NebulaKeyUtils.cpp
```diff
@@ -86,6 +86,59 @@ std::string NebulaKeyUtils::kvKey(PartitionID partId, const folly::StringPiece&
...
+    for (auto& col : values) {
+        if (col.first == cpp2::SupportedType::STRING) {
+            colsLen.emplace_back(col.second.size());
+        }
+        raw.append(col.second.data(), col.second.size());
+    }
+    for (auto len : colsLen) {
```
darionyaphet: why append column size at the end of encoded-value?

src/storage/mutate/AddEdgesProcessor.cpp
bright-starry-sky: Yes, now is.

src/interface/common.thrift
bright-starry-sky: Different index id may correspond to the same tagID.
IndexItem will be used for storage layer and graph lookup .

bright-starry-sky: > By the way, why we need tagOrEdge here? index_id is not enough?
> 
> And where did we use the IndexItem?

I need a detailed explanation：
When addVertices or addEdges start executing, maybe have multiple kinds of tags or edgeType in the same request. And a tag or edgeType may have multiple indexes.Therefore, I need to know the relationship between specific tags and indexes.

Also, IndexItem will be used for each vertex or edge write, for example insert|delete|update ，the IndexId and tagOrEdge is required.

src/common/base/NebulaKeyUtils.cpp
bright-starry-sky: Used for variable-length column type ， for example `string`  .
if an index row likes string('abc') + string('abc') , when the where expression is col1 == 'abca', the current row should be skip.

src/interface/common.thrift
dangleptr: Make sense.

src/storage/mutate/AddEdgesProcessor.cpp
dangleptr: Got it

src/storage/mutate/UpdateVertexProcessor.cpp
```diff
@@ -259,13 +261,41 @@ std::string UpdateVertexProcessor::updateAndWriteBack() {
...
+        if (!indexes_.empty()) {
+            for (auto &index : indexes_) {
+                if (index.get_tagOrEdge() == u.first) {
+                    auto prop = u.second->updater->encode();
+                    auto reader = RowReader::getTagPropReader(this->schemaMan_,
+                                                              std::move(prop),
+                                                              spaceId_,
+                                                              u.first);
+                    const auto &cols = index.get_cols();
+                    auto values = collectIndexValues(std::move(reader).get(), cols);
```
dangleptr: why std::move(reader) ?

src/storage/mutate/UpdateVertexProcessor.cpp
```diff
@@ -259,13 +261,41 @@ std::string UpdateVertexProcessor::updateAndWriteBack() {
...
+                    auto iKey = NebulaKeyUtils::vertexIndexKey(partId, index.index_id, vId, values);
+                    batchHolder->put(std::move(iKey), "");
+                    if (!(u.second->kv.second.empty())) {
+                        auto oReader = RowReader::getTagPropReader(this->schemaMan_,
+                                                                   u.second->kv.second,
+                                                                   spaceId_,
+                                                                   u.first);
+                        const auto &oCols = index.get_cols();
+                        auto oValues = collectIndexValues(std::move(oReader).get(), oCols);
```
dangleptr: ditto

src/storage/mutate/UpdateVertexProcessor.cpp
bright-starry-sky: Good point! It got me thinking about performance optimization. I've solved the performance problem for delete|update vertices or edges.

bright-starry-sky:
1,Performance optimization. 2,Resolve the conflict. 3,Improve test cases

src/storage/mutate/UpdateVertexProcessor.cpp
bright-starry-sky: > why std::move(reader) ?

Good point! It got me thinking about performance optimization. I've solved the performance problem for delete|update vertices or edges.

jude-zhu:
close #458

PR #1369, opened 2019-12-03
rename package of java native client

critical27:


src/tools/native-client/src/main/java/com/vesoft/nebula/NativeClientResourceLoader.java
```diff
@@ -4,7 +4,7 @@
  * attached with Common Clause Condition 1.0, found in the LICENSES directory.
  */
 
-package com.vesoft.client;
+package com.vesoft.nebula;
```
darionyaphet: `com.vesoft.nebula.native` is better ?

critical27: Yeah, I will change to com.vesoft.nebula.codec

PR #1370, opened 2019-12-03
rename package name of native client

critical27:


critical27:
Jenkins go

src/tools/native-client/pom.xml
```diff
@@ -72,14 +72,12 @@
                             <goal>javah</goal>
                         </goals>
                         <configuration>
-                            <javahOutputDirectory>${basedir}/src/main/cpp
-                            </javahOutputDirectory>
+                            <javahOutputDirectory>${basedir}/src/main/cpp</javahOutputDirectory>
                             <javahOutputFileName>
-                                com_vesoft_client_NativeClient.h
+                                com_vesoft_nebula_codec.h
```
darionyaphet: the javah output file name should be `package name_file name` ?

critical27: Ok

src/tools/native-client/pom.xml
```diff
@@ -151,6 +151,61 @@
                     <argLine>-Djava.library.path=_build/</argLine>
                 </configuration>
             </plugin>
+            <plugin>
```
darionyaphet: would you publish natice codec to maven center ?

src/tools/native-client/pom.xml
```diff
@@ -151,6 +151,61 @@
...
+                        </goals>
+                    </execution>
+                </executions>
+            </plugin>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-javadoc-plugin</artifactId>
+                <version>3.1.1</version>
+                <configuration>
+                    <excludePackageNames>com.facebook.thrift:com.facebook.thrift.*</excludePackageNames>
```
darionyaphet: should not exclude thrift

src/tools/native-client/pom.xml
critical27: Yeah, then we can use in nebula-java

src/tools/native-client/pom.xml
critical27: Make sense.

src/tools/native-client/src/main/java/com/vesoft/client/NativeClientResourceLoader.java
```diff
@@ -1,47 +0,0 @@
-/* Copyright (c) 2018 vesoft inc. All rights reserved.
- *
```
darionyaphet: why remove `NativeClientResourceLoader`, maybe use `NativeClientResourceLoader.resourceLoader();` could load `.so`

critical27: We need to use it in nebula-java, but not here? We just need to pack the `so` in jar in codec.

PR #1371, opened 2019-12-04
Fix graphd crash problem caused by SetExecutor.

CPWstatic:
Fix 2 problem in this pr:
1. possible crash problem in SetExecutor.
2. memory leak in ColumnSpecification.


CPWstatic:
Jenkins go!

PR #1378, opened 2019-12-04
Enhance balancer

dangleptr:
To make it more robust, we add check step for each balance task.  It gives a chance to repair the inconsistent state between storage and meta. 

Of course,  during the development, we fix serval bugs by the way.  :(

Now the balancer looks good during my test, especially under heavy load.


dangleptr:
Jenkins go

src/kvstore/raftex/RaftPart.cpp
```diff
@@ -1743,6 +1744,24 @@ bool RaftPart::linkCurrentWAL(const char* newPath) {
     return wal_->linkCurrentWAL(newPath);
 }
 
+void RaftPart::checkAndResetPeers(const std::vector<HostAddr>& peers) {
+    std::lock_guard<std::mutex> lck(raftLock_);
+    // To avoid the iterator invalid, we use another container for it.
+    decltype(hosts_) hosts = hosts_;
```

src/meta/processors/admin/AdminClient.cpp
```diff
@@ -499,6 +560,8 @@ folly::Future<Status> AdminClient::getLeaderDist(HostLeaderMap* result) {
         }
 
         p.setValue(Status::OK());
+    }).thenError([p = std::move(promise)] (auto&& e) mutable {
```
critical27: My bad - -b

src/storage/AdminProcessor.h
```diff
@@ -41,14 +41,48 @@ class TransLeaderProcessor : public BaseProcessor<cpp2::AdminExecResp> {
...
+                    while (retry-- > 0) {
+                        auto leaderRet = kvstore_->partLeader(spaceId, partId);
+                        if (!ok(leaderRet)) {
+                            this->pushResultCode(to(error(leaderRet)), partId);
+                            onFinished();
+                            return;
+                        }
+                        auto leader = value(std::move(leaderRet));
+                        auto* store = static_cast<kvstore::NebulaStore*>(kvstore_);
+                        if (leader != HostAddr(0, 0) && leader != store->address()) {
```
critical27: Why not check leader is same with target leader?

src/meta/processors/admin/BalancePlan.cpp
```diff
@@ -84,13 +84,19 @@ void BalancePlan::invoke() {
...
                     }
+                    stopped = stopped_;
                 }
                 if (finished) {
                     CHECK_EQ(j, this->buckets_[i].size() - 1);
                     saveInStore(true);
                     onFinished_();
                 } else if (j + 1 < this->buckets_[i].size()) {
                     auto& task = this->tasks_[this->buckets_[i][j + 1]];
+                    if (tasks_[taskIndex].spaceId_ == task.spaceId_
```
critical27: Is this why we sometimes recover, and still have failed tasks?

src/kvstore/Part.h
```diff
@@ -76,6 +77,11 @@ class Part : public raftex::RaftPart {
     void reset() {
         LOG(INFO) << idStr_ << "Clean up all wals";
         wal()->reset();
+        ResultCode res = engine_->remove(NebulaKeyUtils::systemCommitKey(partId_));
+        if (res != ResultCode::SUCCEEDED) {
+            LOG(WARNING) << idStr_ << "Remove the committedLogId failed, error "
+                         << static_cast<int32_t>(res);
```
darionyaphet: maybe we need a method use to transfer error code to message

src/storage/AdminProcessor.h
dangleptr: Actually, we don't need to do it right now. Because the only thing we care about is the leader is not current host.

src/meta/processors/admin/BalancePlan.cpp
dangleptr: The logic here is to avoid messing up peers if the previous task failed for the same part.

PR #1379, opened 2019-12-04
Added "if not exists" option to create space/tag/edge.

### LINKED ISSUE #1366: Add IF NOT EXISTS for DDL statements
```ngql
CREATE SPACE IF NOT EXISTS test(partition_num=128, replica_factor=1);
CREATE TAG IF NOT EXISTS student(name string, age int);
CREATE EDGE IF NOT EXISTS like(likeness int);
```

monadbobo:
This option allows you to check if the space/tag/edge
that you create already exists in the database.

example:

create space if not exists testspace
create tag if not exists testtag()
create edge if not exists edge1()

close #1366

### DIFF
--- a/src/graph/test/SchemaTest.cpp
+++ b/src/graph/test/SchemaTest.cpp
@@ -138,7 +138,7 @@ TEST_F(SchemaTest, metaCommunication) {
         ASSERT_EQ(cpp2::ErrorCode::SUCCEEDED, code);
     }
     {
-        std::string query = "CREATE SPACE if not exists default_space";
+        std::string query = "CREATE SPACE IF NOT EXISTS default_space";
         cpp2::ExecutionResponse resp;
         auto code = client->execute(query, resp);
         ASSERT_EQ(cpp2::ErrorCode::SUCCEEDED, code);
@@ -180,7 +180,7 @@ TEST_F(SchemaTest, metaCommunication) {
     }
     {
         cpp2::ExecutionResponse resp;
-        std::string query = "CREATE TAG if not exists tag1()";
+        std::string query = "CREATE TAG IF NOT EXISTS tag1()";
         auto code = client->execute(query, resp);
         ASSERT_EQ(cpp2::ErrorCode::SUCCEEDED, code);
     }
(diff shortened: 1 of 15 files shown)


monadbobo:
close #1366

src/graph/test/SchemaTest.cpp
```diff
@@ -137,6 +137,12 @@ TEST_F(SchemaTest, metaCommunication) {
         auto code = client->execute(query, resp);
         ASSERT_EQ(cpp2::ErrorCode::SUCCEEDED, code);
     }
+    {
+        std::string query = "CREATE SPACE if not exists default_space";
```
yixinglu: make `if not exists` uppercase. 😃

src/graph/test/SchemaTest.cpp
```diff
@@ -172,6 +178,12 @@ TEST_F(SchemaTest, metaCommunication) {
         auto code = client->execute(query, resp);
         ASSERT_EQ(cpp2::ErrorCode::SUCCEEDED, code);
     }
+    {
+        cpp2::ExecutionResponse resp;
+        std::string query = "CREATE TAG if not exists tag1()";
```
yixinglu: ditto

src/graph/test/SchemaTest.cpp
```diff
@@ -376,6 +388,12 @@ TEST_F(SchemaTest, metaCommunication) {
         auto code = client->execute(query, resp);
         ASSERT_EQ(cpp2::ErrorCode::SUCCEEDED, code);
     }
+    {
+        cpp2::ExecutionResponse resp;
+        std::string query = "CREATE EDGE if not exists edge1()";
```
yixinglu: ditto

src/meta/processors/partsMan/CreateSpaceProcessor.cpp
```diff
@@ -18,6 +18,12 @@ void CreateSpaceProcessor::process(const cpp2::CreateSpaceReq& req) {
     auto properties = req.get_properties();
     auto spaceRet = getSpaceId(properties.get_space_name());
     if (spaceRet.ok()) {
+        if (req.get_if_not_exists()) {
+            resp_.set_id(to(spaceRet.value(), EntryType::SPACE));
+            resp_.set_code(cpp2::ErrorCode::SUCCEEDED);
+            onFinished();
+            return;
+        }
```
yixinglu: I think you only need to handle the `resp.set_code()` logic according to `req.get_if_not_exists()`.  And the rest is common.  For example:

```cpp
if (req.get_if_not_exists()) {
  resp.set_code(SUCCEED);
} else {
  LOG(ERROR) << "";
  resp.set_code(EXISTED);
}
resp.set_id();
onFinished();
return;
```

laura-ding:
In the case of if not exist, it is simply to judge whether space name, tag name or edge name exist, without checking whether the attributes are consistent. We need to judge the attributes, the attributes are the same then we can return success, if not, return error.

PR #1381, opened 2019-12-05
add scan edge/vertex in storage

critical27:
Support scan edge/vertex in storage, later for spark job. 

We can retrieve data of a part limited by single vertexId and time range. We could enhance it by adding complicated filter and group of vertex later.


critical27:
Jenkins go

src/storage/ScanVertexProcessor.cpp
```diff
@@ -0,0 +1,107 @@
...
+        }
+    } else {
+        start = *req.get_cursor();
+    }
+
+    if (req.get_vertex_id() != nullptr) {
+        auto vId = *req.get_vertex_id();
+        prefix = NebulaKeyUtils::vertexPrefix(partId, vId);
+    } else {
+        prefix = NebulaKeyUtils::prefix(partId);
```
wadeliuyi: the prefix default value is NebulaKeyUtils::prefix(partId); so here no need else

src/storage/ScanEdgeProcessor.cpp
```diff
@@ -0,0 +1,110 @@
...
+        }
+    } else {
+        start = *req.get_cursor();
+    }
+
+    if (req.get_vertex_id() != nullptr) {
+        auto vId = *req.get_vertex_id();
+        prefix = NebulaKeyUtils::edgePrefix(partId, vId);
+    } else {
+        prefix = NebulaKeyUtils::prefix(partId);
```
wadeliuyi: same with default value

src/storage/ScanEdgeProcessor.cpp
```diff
@@ -0,0 +1,110 @@
...
+                continue;
+            }
+        }
+
+        auto value = iter->val();
+        cpp2::ScanEdge data;
+        data.set_type(edgeType);
+        data.set_key(key.str());
+        data.set_value(value.str());
+        edgeData.emplace_back(data);
```
wadeliuyi: std::move(data)?

src/kvstore/plugins/hbase/HBaseStore.cpp
```diff
@@ -199,6 +199,27 @@ ResultCode HBaseStore::prefix(GraphSpaceID spaceId,
...
+                                       PartitionID  partId,
+                                       const std::string& start,
+                                       const std::string& prefix,
+                                       std::unique_ptr<KVIterator>* storageIter) {
+    UNUSED(partId);
+    auto tableName = this->spaceIdToTableName(spaceId);
+    std::string startRowKey, endRowKey;
+    startRowKey = this->getRowKey(start);
+    endRowKey.reserve(kMaxRowKeyLength);
+    for (size_t n = 0; n < kMaxRowKeyLength - prefix.size(); n++) {
```
wadeliuyi: size_t is unsigned, so take care about that if prefix.size() is bigger than kMaxRowKeyLength, is the compiler will convert the result (kMaxRowKeyLength - prefix.size()) to unsigned?

src/storage/ScanVertexProcessor.cpp
critical27: Good suggestion

src/storage/ScanEdgeProcessor.cpp
critical27: Good catch

src/kvstore/plugins/hbase/HBaseStore.cpp
critical27: Both `kMaxRowKeyLength` and `prefix.size()` is size_t, and the caller need make sure prefix is legal.

wadeliuyi: even so, I think double check is needed here.

critical27:
> Thanks, `scanVertex` is totally meet my requirements.

I didn't implement the interface in C++ client, perhaps you need to add one

src/interface/storage.thrift
```diff
@@ -310,6 +310,62 @@ struct UpdateEdgeRequest {
...
+struct ScanTag {
+    1: common.TagID     tagId,
+    2: binary           key,
+    3: binary           value,                  // decode according to vertex_schema.
+}
+
+struct ScanVertexResponse {
+    1: required ResponseCommon result,
+    2: optional map<common.TagID, common.Schema>(cpp.template = "std::unordered_map")    vertex_schema,
+    3: optional list<ScanTag> tag_data,
```
darionyaphet: maybe should named `ScanVertex` ?

critical27: Make sense, we need to update the corresponding code in java - -

src/meta/NebulaSchemaProvider.cpp
```diff
@@ -97,6 +97,20 @@ const nebula::cpp2::SchemaProp NebulaSchemaProvider::getProp() const {
...
+nebula::cpp2::Schema NebulaSchemaProvider::toSchema() const {
+    nebula::cpp2::Schema schema;
+    std::vector<nebula::cpp2::ColumnDef> columns;
+    for (auto& field : fields_) {
+        nebula::cpp2::ColumnDef column;
+        column.set_name(field->getName());
+        column.set_type(field->getType());
+        columns.emplace_back(column);
+    }
+    schema.set_columns(columns);
```
laura-ding: std::move(columns)

src/kvstore/RocksEngine.h
```diff
@@ -52,6 +52,40 @@ class RocksRangeIter : public KVIterator {
     rocksdb::Slice end_;
 };
 
+class RocksRangeWithPrefixIter : public KVIterator {
```
laura-ding: The destructor is not a virtual function, add final

src/storage/BaseProcessor.inl
```diff
@@ -83,5 +83,29 @@ void BaseProcessor<RESP>::doRemoveRange(GraphSpaceID spaceId,
...
+kvstore::ResultCode BaseProcessor<RESP>::doPrefix(GraphSpaceID spaceId,
+                                                  PartitionID partId,
+                                                  std::string prefix,
+                                                  std::unique_ptr<kvstore::KVIterator>* iter) {
+    return kvstore_->prefix(spaceId, partId, prefix, iter);
+}
+
+template<typename RESP>
+kvstore::ResultCode BaseProcessor<RESP>::doRangeWithPrefix(
+        GraphSpaceID spaceId, PartitionID partId, std::string start, std::string prefix,
```
laura-ding: const std::string &start, const std::string &prefix,

src/meta/NebulaSchemaProvider.cpp
critical27: Good catch

src/storage/BaseProcessor.inl
critical27: It seems most interface in this class expect caller to use right value reference.

src/kvstore/RocksEngine.h
dangleptr: Why we need it?

src/storage/query/ScanEdgeProcessor.cpp
```diff
@@ -0,0 +1,107 @@
...
+
+    for (; iter->valid() && rowCount < rowLimit && blockSize < FLAGS_max_scan_block_size;
+         iter->next()) {
+        auto key = iter->key();
+        if (!NebulaKeyUtils::isEdge(key)) {
+            continue;
+        }
+
+        EdgeType edgeType = NebulaKeyUtils::getEdgeType(key);
+        if (edgeType < 0) {
```
dangleptr: Why skip the reverse edge?

src/common/base/NebulaKeyUtils.h
```diff
@@ -93,6 +93,12 @@ class NebulaKeyUtils final {
         return !(tagId & tagMask);
     }
 
+    static TagID getVertexId(const folly::StringPiece& rawKey) {
```
dangleptr: Edge Or Vertex ?

src/interface/storage.thrift
```diff
@@ -310,6 +310,62 @@ struct UpdateEdgeRequest {
...
+
+struct ScanEdgeResponse {
+    1: required ResponseCommon result,
+    2: optional map<common.EdgeType, common.Schema>(cpp.template = "std::unordered_map")    edge_schema,
+    3: optional list<ScanEdge> edge_data,
+    4: bool has_next,
+    5: binary next_cursor,                      // next start key of scan
+}
+
+struct ScanVertexRequest {
```
dangleptr: Do we really need the scan interface for vertices?

src/interface/storage.thrift
```diff
@@ -310,6 +310,62 @@ struct UpdateEdgeRequest {
     7: bool insertable,
 }
 
+struct ScanEdgeRequest {
+    1: common.GraphSpaceID space_id,
+    2: common.PartitionID part_id,
+    // start key of this block
+    3: optional binary cursor,
+    // vertex ids used as prefix
+    4: optional common.VertexID vertex_id,
```
dangleptr: What's the purpose here?

src/interface/storage.thrift
```diff
@@ -310,6 +310,62 @@ struct UpdateEdgeRequest {
...
+    // max row count of edge in this response
+    5: i32 row_limit,
```
dangleptr: row_limit is not accurate here.  "Limit" is enough.

src/interface/storage.thrift
```diff
@@ -310,6 +310,62 @@ struct UpdateEdgeRequest {
...
+    6: i64 start_time,
+    7: i64 end_time,
+    8: i32 max_versions,                        // support later
+}
+
+struct ScanEdge {
+    1: common.EdgeType type,
+    2: binary          key,
```
dangleptr: What's the key format?

src/interface/storage.thrift
```diff
@@ -310,6 +310,62 @@ struct UpdateEdgeRequest {
...
+    3: binary          value,                   // decode according to edge_schema.
```
dangleptr: Do you return all columns inside?

src/dataman/ResultSchemaProvider.h
```diff
@@ -52,10 +52,14 @@ class ResultSchemaProvider : public meta::SchemaProviderIf {
     std::shared_ptr<const meta::SchemaProviderIf::Field> field(
         const folly::StringPiece name) const override;
 
+    nebula::cpp2::Schema toSchema() const override;
+
 protected:
     SchemaVer schemaVer_{0};
 
     ColumnDefs columns_;
+    nebula::cpp2::SchemaProp schemaProp_;
```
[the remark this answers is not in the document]
dangleptr: We don't need it in ResultSchemaProvider.

src/interface/storage.thrift
dangleptr: Why not use getNeighbors

src/interface/storage.thrift
dangleptr: Return the key structure is a better way.

src/kvstore/RocksEngine.h
dangleptr: It seems RangeIter is enough for your case.

critical27: No, because we also need to keep a prefix of partId.
This is RangeIter valid function, if we use this, we will scan data of other partition.
```
bool valid() const override {
    return !!iter_ && iter_->Valid() && (iter_->key().compare(end_) < 0);
}
```

src/dataman/ResultSchemaProvider.h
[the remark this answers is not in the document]
critical27: Good point, my bad

dutor:
Jenkins go

src/interface/storage.thrift
```diff
@@ -310,6 +310,61 @@ struct UpdateEdgeRequest {
     7: bool insertable,
 }
 
+struct ScanEdgeRequest {
+    1: common.GraphSpaceID space_id,
+    2: common.PartitionID part_id,
+    // start key of this block
+    3: optional binary cursor,
+    // If specified, only return specified columns, if not, no columns are returned
+    4: map<common.EdgeType, list<PropDef>> (cpp.template = "std::unordered_map") return_columns,
```
dangleptr: Why not use list directly?

src/interface/storage.thrift
```diff
@@ -310,6 +310,61 @@ struct UpdateEdgeRequest {
...
+    // max row count of edge in this response
+    6: i32 limit,
+    7: i64 start_time,
+    8: i64 end_time,
+}
+
+struct ScanEdgeResponse {
+    1: required ResponseCommon result,
+    2: map<common.EdgeType, common.Schema> (cpp.template = "std::unordered_map") edge_schema,
+    3: map<common.VertexID, list<ScanEdge>> (cpp.template = "std::unordered_map") edge_data,
```
dangleptr: Why not return list directly

src/interface/storage.thrift
```diff
@@ -310,6 +310,61 @@ struct UpdateEdgeRequest {
...
+    3: binary value, // decode according to edge_schema.
+}
+
+struct ScanVertexRequest {
+    1: common.GraphSpaceID space_id,
+    2: common.PartitionID part_id,
+    // start key of this block
+    3: optional binary cursor,
+    // If specified, only return specified columns, if not, no columns are returned
+    4: map<common.TagID, list<PropDef>> (cpp.template = "std::unordered_map") return_columns,
```
dangleptr: Ditto

src/interface/storage.thrift
```diff
@@ -310,6 +310,61 @@ struct UpdateEdgeRequest {
...
+
+struct ScanVertex {
+    1: common.TagID     tagId,
+    2: binary           value,                  // decode according to vertex_schema.
+}
+
+struct ScanVertexResponse {
+    1: required ResponseCommon result,
+    2: map<common.TagID, common.Schema> (cpp.template = "std::unordered_map") vertex_schema,
+    3: map<common.VertexID, list<ScanVertex>> (cpp.template = "std::unordered_map") vertex_data,
```
dangleptr: Ditto

src/interface/storage.thrift
```diff
@@ -310,6 +310,61 @@ struct UpdateEdgeRequest {
...
+    4: map<common.TagID, list<PropDef>> (cpp.template = "std::unordered_map") return_columns,
+    5: bool all_columns,
+    // max row count of tag in this response
+    6: i32 limit,
+    7: i64 start_time,
+    8: i64 end_time,
+}
+
+struct ScanVertex {
+    1: common.TagID     tagId,
```
dangleptr: Why not include VertexID directly in this structure.

src/kvstore/RocksEngine.h
dangleptr: Make sense. Could we extend prefixIterator for it.  Not reimplement a new one.

src/storage/query/ScanEdgeProcessor.cpp
dangleptr: We should not has this assumption.

src/interface/storage.thrift
critical27: In go we need to collect properties of vertex/edge to generate a table with same columns. But in scan I suppose we only to collect some properties among different edge types, we don't care about what the schema looks like of other edge types, the rows in the table may have different rows. I think the map is better here.

src/interface/storage.thrift
critical27: In the commits before, I return a list directly, we can save some space with a map. WDYT?

src/kvstore/RocksEngine.h
critical27: Good suggestion

src/storage/query/ScanEdgeProcessor.cpp
critical27: Let me ask for the how angel need for this in business process.

critical27: I have asked several business partners, return all the out edges is totally ok, maybe we can add a option, WDYT

src/kvstore/RocksEngine.h
```diff
@@ -81,12 +80,29 @@ class RocksPrefixIter : public KVIterator {
...
 
+class RocksRangeWithPrefixIter : public RocksPrefixIter {
+public:
+    RocksRangeWithPrefixIter(rocksdb::Iterator* iter, rocksdb::Slice start, rocksdb::Slice prefix)
+        : RocksPrefixIter(iter, prefix)
+        , start_(start) {}
+
+    ~RocksRangeWithPrefixIter() = default;
+
+    bool valid() const override {
```
dangleptr: Why override the method?  It looks the same.

src/storage/query/ScanEdgeProcessor.cpp
dangleptr: Sounds good to me.

src/storage/query/ScanEdgeProcessor.cpp
```diff
@@ -0,0 +1,173 @@
...
+        EdgeType edgeType = edgeIter.first;
+        std::vector<PropContext> propContexts;
+        auto schema = this->schemaMan_->getEdgeSchema(spaceId_, edgeType);
+        if (!schema) {
+            return cpp2::ErrorCode::E_EDGE_NOT_FOUND;
+        }
+
+        if (returnAllColumns_) {
+            // return all columns
+            edgeContexts_.emplace(edgeType, std::move(propContexts));
```
dangleptr: Do we really need to set edgeContexts_ and edgeSchema_ when returning all columns

src/kvstore/RocksEngine.h
critical27: Fixed

src/storage/query/ScanEdgeProcessor.cpp
critical27: edgeContexts_ and edgeSchema_ are used to store related edgeType, so we can check if we need to scan this edge

critical27:
action rerun

src/kvstore/NebulaStore.cpp
```diff
@@ -407,6 +407,21 @@ ResultCode NebulaStore::prefix(GraphSpaceID spaceId,
...
+                                        const std::string& start,
+                                        const std::string& prefix,
+                                        std::unique_ptr<KVIterator>* iter) {
+    auto ret = engine(spaceId, partId);
+    if (!ok(ret)) {
+        return error(ret);
+    }
+    auto* e = nebula::value(ret);
+    return e->rangeWithPrefix(start, prefix, iter);
+}
```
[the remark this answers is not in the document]
critical27: Because we don't scan the realtime data, for example we scan the data of yesterday, so it doesn't matter is leader or not.

src/storage/query/ScanEdgeProcessor.cpp
```diff
@@ -0,0 +1,173 @@
...
+    if (retCode != cpp2::ErrorCode::SUCCEEDED) {
+        this->pushResultCode(retCode, partId_);
+        this->onFinished();
+        return;
+    }
+
+    std::string start;
+    std::string prefix = NebulaKeyUtils::prefix(partId_);
+    if (req.get_cursor() == nullptr || req.get_cursor()->empty()) {
+        start = NebulaKeyUtils::prefix(partId_);
```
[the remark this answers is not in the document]
critical27: Sure.

src/storage/test/ScanEdgeTest.cpp
```diff
@@ -0,0 +1,448 @@
...
+    auto* processor = ScanEdgeProcessor::instance(kv.get(), schemaMan.get(), nullptr);
+    auto f = processor->getFuture();
+    processor->process(req);
+    auto resp = std::move(f).get();
+
+    LOG(INFO) << "Check the results...";
+    std::string cursor = "";
+    int32_t rowCount = 0, expectRowCount = 100;
+    checkResponse(partId, resp, cursor, rowCount, expectRowCount, 0, false, true);
+    EXPECT_EQ(rowCount, 100);
```
panda-sheep: For edges, when no output column attributes are specified. What is output?  
Does it output the following four fields：_src, _rank,  _dst,  _type?

critical27: We only return the key, aka, src, type and dst (we don't return rank for now)

src/common/base/NebulaKeyUtils.h
```diff
@@ -185,6 +185,12 @@ class NebulaKeyUtils final {
         return readInt<EdgeRanking>(rawKey.data() + offset, sizeof(EdgeRanking));
     }
 
+    static int64_t getVersion(const folly::StringPiece& rawKey) {
```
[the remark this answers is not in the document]
critical27: We use int64_t as timestamp for now, maybe we can unify them with a macro later in another PR, WDYT?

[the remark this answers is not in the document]
panda-sheep: Sorry, I just found out that the macro has already existed. 
Using `SchemaVer`

src/interface/storage.thrift
```diff
@@ -310,6 +310,63 @@ struct UpdateEdgeRequest {
...
+    2: common.PartitionID part_id,
+    // start key of this block
+    3: optional binary cursor,
+    // If specified, only return specified columns, if not, no columns are returned
+    4: map<common.TagID, list<PropDef>> (cpp.template = "std::unordered_map") return_columns,
+    5: bool all_columns,
+    // max row count of tag in this response
+    6: i32 limit,
+    7: i64 start_time,
+    8: i64 end_time,
```
CPWstatic: Must be given?

critical27: In java client, it has a default value from 0 to Long.MAX_VALUE.

src/storage/test/ScanEdgeTest.cpp
```diff
@@ -0,0 +1,448 @@
...
+    PartitionID partId = 1;
+    auto req = buildRequest(partId, "", rowLimit, true);
+    int32_t batchCount = 0;
+
+    while (true) {
+        auto* processor = ScanEdgeProcessor::instance(kv.get(), schemaMan.get(), nullptr);
+        auto f = processor->getFuture();
+        processor->process(req);
+        auto resp = std::move(f).get();
+        ++batchCount;
```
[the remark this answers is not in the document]
critical27: Never mind, we have count the total rows returned

PR #1384, opened 2019-12-06
fix test case for ubuntu docker

bright-starry-sky:
Temporary fix test case for ubuntu docker. The mv command file is in the /bin directory on ubuntu docker. 
I will add a renameDir method to FileUtils at later.

### DIFF
--- a/src/kvstore/test/NebulaStoreTest.cpp
+++ b/src/kvstore/test/NebulaStoreTest.cpp
@@ -760,13 +760,13 @@ TEST(NebulaStoreTest, ThreeCopiesCheckpointTest) {
         fs::FileUtils::remove(folly::stringPrintf("%s/data", rm.data()).c_str(), true);
         fs::FileUtils::remove(folly::stringPrintf("%s/wal", rm.data()).c_str(), true);
         std::string mv = folly::stringPrintf(
-                "/usr/bin/mv %s/disk%d/nebula/0/checkpoints/snapshot/data %s/disk%d/nebula/0/data",
+                "mv %s/disk%d/nebula/0/checkpoints/snapshot/data %s/disk%d/nebula/0/data",
                 rootPath.path(), i , rootPath.path(), i);
         sleep(1);
         auto ret = system(mv.c_str());
         ASSERT_EQ(0, ret);
         mv = folly::stringPrintf(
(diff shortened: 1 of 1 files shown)


src/kvstore/test/NebulaStoreTest.cpp
```diff
@@ -760,13 +760,13 @@ TEST(NebulaStoreTest, ThreeCopiesCheckpointTest) {
         fs::FileUtils::remove(folly::stringPrintf("%s/data", rm.data()).c_str(), true);
         fs::FileUtils::remove(folly::stringPrintf("%s/wal", rm.data()).c_str(), true);
         std::string mv = folly::stringPrintf(
-                "/usr/bin/mv %s/disk%d/nebula/0/checkpoints/snapshot/data %s/disk%d/nebula/0/data",
+                "mv %s/disk%d/nebula/0/checkpoints/snapshot/data %s/disk%d/nebula/0/data",
                 rootPath.path(), i , rootPath.path(), i);
         sleep(1);
         auto ret = system(mv.c_str());
```
Shylock-Hg: Consider the function `rename` in stdio.h?

bright-starry-sky: `std::rename` is for files only, not directories.
Actually, I'm more optimistic with `std::filesystem::rename` . However, the current environment does not yet allow for the use of `std::filesystem::rename`, which will take some time.

Shylock-Hg: The `rename` could apply to folder according to the man page. And I tried in nebula-dev containers.

bright-starry-sky: > The `rename` could apply to folder according to the man page. And I tried in nebula-dev containers.

Aha，Thanks for your verification ! Would you like to provide the PR？

Shylock-Hg: > > The `rename` could apply to folder according to the man page. And I tried in nebula-dev containers.
> 
> Aha，Thanks for your verification ! Would you like to provide the PR？

Ok.

bright-starry-sky: Close this one. And PR #1387 will  fix this bug.

PR #1391, opened 2019-12-06
fix unstable problem for link Wal file

bright-starry-sky:
Improve unstable test case snapshot_command_test .

### DIFF
--- a/src/meta/processors/admin/AdminClient.cpp
+++ b/src/meta/processors/admin/AdminClient.cpp
@@ -512,9 +512,14 @@ folly::Future<Status> AdminClient::createSnapshot(GraphSpaceID spaceId, const st
 
     folly::Promise<Status> pro;
     auto f = pro.getFuture();
+
+    /**
+     * Don't need retry.
+     * Because existing checkpoint directories leads to fail again.
+     **/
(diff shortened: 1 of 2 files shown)


src/meta/processors/admin/AdminClient.cpp
```diff
@@ -512,9 +512,14 @@ folly::Future<Status> AdminClient::createSnapshot(GraphSpaceID spaceId, const st
...
     auto f = pro.getFuture();
+
+    /**
+     * Don't need retry.
+     * Because existing checkpoint directories leads to fail again.
+     **/
     getResponse(allHosts, 0, std::move(req), [] (auto client, auto request) {
         return client->future_createCheckpoint(request);
-    }, 0, std::move(pro), 1 /*The snapshot operation only needs to be retried twice*/);
+    }, 0, std::move(pro), 0);
```
dangleptr: What if there are some connection failure?

bright-starry-sky: An error returned, and rerun the create command by console. Because some nodes may succeed, the checkpoint directory will exist, retry will not work.

dangleptr: make sense.

PR #1394, opened 2019-12-08
Remove unnecessary null pointer checks

### LINKED ISSUE #1392: Remove unnecessary null pointer checks
[An extra null pointer check is not needed](https://isocpp.org/wiki/faq/freestore-mgmt#delete-handles-null "Do I need to check for null before delete p?") in [the destructor and the function “parse” for the class “GQLParser”](https://github.com/vesoft-inc/nebula/blob/70ecf18d303dd5cbd0fefd305e4b182a7cd0b68e/src/parser/GQLParser.h#L36 "Update candidate").

Reminiscent:
fix #1392


Reminiscent:
Besides, should we remove [this](https://github.com/vesoft-inc/nebula/blob/70ecf18d303dd5cbd0fefd305e4b182a7cd0b68e/src/parser/GQLParser.h#L54-L57) null pointer checks? Like
```
scanner_.flushBuffer();
delete sentences_;
sentences_ = nullptr;
return Status::SyntaxError(error_);
```

Shylock-Hg:
Good job, I think it's safe to remove the checking before delete since c++ 14.

Reminiscent:
I don't know how to find all the same patterns in the code. I use the comment ```grep -rn '\<delete\>' *``` and check them one by one. I find the patterns only use here. Could you help me find more? Or this is the only place to use the null pointer check? @Shylock-Hg  PTAL

elfring:
>Could you help me find more?

How do you think about to [use a development tool like “clang-tidy” for corresponding source code adjustments](https://clang.llvm.org/extra/clang-tidy/checks/readability-delete-null-pointer.html "Deletion of unnecessary null pointer checks")?

-----

Would you like to wrap any pointers with the class template “[std::unique_ptr](https://en.wikipedia.org/wiki/Smart_pointer#unique_ptr "Description for the usage of smart pointers")”?

-----

Can the pull request subject be improved another bit?

dutor:
This change is OK for me.

But, I don't think it would be better to do so. After all, not everyone knows that fact that `delete` will check `NULL`ness for us, for whom we have to pay effort to explain on this, which is not worthy.

@Reminiscent @Shylock-Hg @elfring @laura-ding

dutor:
I will merge this PR this time. Please don't make such change any more.

dutor:
Jenkins go

dutor:
Jenkins, go

PR #1397, opened 2019-12-09
[doc]Manual en readme

wilsonyou:
minor updates


Amber1990Zhang:
Why are there so many commits? I don't understand, are they from you, and I noticed there is only one file changed, I'm confused.

wilsonyou:
> Why are there so many commits? I don't understand, are they from you, and I noticed there is only one file changed, I'm confused.

Only the last commit maters. Please ignore the previous ones.

dangleptr:
Jenkins go

PR #1399, opened 2019-12-09
Add timeout for storage/meta clients and add some error logs

dangleptr:
It is reported by our users.  

After network broken down,  the storaged can't connect to metad some times,   meanwhile the graphd can't connect to storaged either.  Currently,  the client will wait for the response without timeout,  it is unreasonable.   
"timeout" will be helpful to recover from broken connections.

The error logs here is to help our users figure out the problems if failed.

### DIFF
--- a/src/common/filter/Expressions.cpp
+++ b/src/common/filter/Expressions.cpp
@@ -674,6 +674,7 @@ OptVariantType UUIDExpression::eval() const {
                    << ", part " << rc.get_part_id() << ", str id " << toString();
         return OptVariantType(Status::Error("Get UUID Failed"));
      }
+     VLOG(3) << "Get UUID from " << *field_ << " to " << v.get_id();
      return v.get_id();
 }
 
--- a/src/common/filter/Expressions.cpp
+++ b/src/common/filter/Expressions.cpp
@@ -664,11 +664,17 @@ OptVariantType UUIDExpression::eval() const {
      auto client = context_->storageClient();
      auto space = context_->space();
      auto uuidResult = client->getUUID(space, *field_).get();
-     if (!uuidResult.ok() ||
-         !uuidResult.value().get_result().get_failed_codes().empty()) {
-         return OptVariantType(Status::Error("Get UUID Failed"));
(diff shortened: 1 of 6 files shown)


zlcook:
Can you log the mapping of uuid-value and value ?

dangleptr:
> Can you log the mapping of uuid-value and value ?

Good point.

src/common/filter/Expressions.cpp
```diff
@@ -664,11 +664,17 @@ OptVariantType UUIDExpression::eval() const {
...
+        return OptVariantType(Status::Error("Get UUID Failed"));
      }
-     return uuidResult.value().get_id();
+     auto v = std::move(uuidResult).value();
+     for (auto& rc : v.get_result().get_failed_codes()) {
+        LOG(ERROR) << "Get UUID failed, error " << static_cast<int32_t>(rc.get_code())
+                   << ", part " << rc.get_part_id() << ", str id " << toString();
+        return OptVariantType(Status::Error("Get UUID Failed"));
+     }
+     return v.get_id();
```
zlcook: How about vlog the mapping of uuid and value for success ?

zlcook:
> > Can you log the mapping of uuid-value and value ?
> 
> Good point. I have added some error logs for uuid request. It will output the string id.
> If you want to know the the partId for the string id, you could calc it by yourself with this command.
> 
> ```
> YIELD hash("string id") % 1024 + 1
> ```

Very good

Amber1990Zhang:
> > Can you log the mapping of uuid-value and value ?
> 
> Good point. I have added some error logs for uuid request. It will output the string id.
> If you want to know the the partId for the string id, you could calc it by yourself with this command.
> 
> ```
> YIELD hash("string id") % 1024 + 1
> ```

Sorry I have a small question here, is `1024` the default partition number? If so, it should be `100` now, shouldn't it?

src/common/filter/Expressions.cpp
dangleptr: Sounds good

zlcook:
LGTM

PR #1402, opened 2019-12-09
Refactor JNI library

dangleptr:
The old jni lib depends on fbthrift,  it is unfriendly when our users want to compile it locally. 

Now, the new JNI library only depends on STL and glog.  
Our users could build it more easily by themself. 

For datamanlite,  it is a simpler codec library than dataman.  Although there are some duplicate codes currently,  it will be cleaned up after @sherman-the-tank refactor it. 

Of course, we fix serval bugs during the development.  :(


src/jni/src/datamanlite/RowReader.h
```diff
@@ -0,0 +1,283 @@
...
+    int32_t headerLen_ = 0;
+    int32_t numBytesForOffset_ = 0;
+    // Block offet value is composed by two integers. The first one is
+    // the block offset, the second one is the largest index being visited
+    // in the block. This index is zero-based
+    mutable std::vector<std::pair<int64_t, uint8_t>> blockOffsets_;
+    mutable std::vector<int64_t> offsets_;
+
+private:
+    static int32_t getSchemaVer(Slice row);
```
critical27: Do we need to unify the type of schema version? The `SchemaVer` used in thrift is int64_t

src/jni/java/pom.xml
```diff
@@ -0,0 +1,213 @@
...
+ attached with Common Clause Condition 1.0, found in the LICENSES directory.
+-->
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+
+    <groupId>com.vesoft</groupId>
+    <artifactId>nebula-utils</artifactId>
+    <version>1.0.0-beta</version>
```
darionyaphet: maybe it's `1.0.0-RC2`

critical27:
There is other problems, when I add jni as dependency in java client, the so is not loaded. Never mind, we can merge this first, fix it later.

src/jni/src/datamanlite/RowReader.h
dangleptr: Good point

src/jni/java/pom.xml
dangleptr: Good point.

src/jni/src/datamanlite/NebulaSchemaProvider.h
```diff
@@ -0,0 +1,86 @@
...
+
+        const char* getName() const override {
+            return name_.c_str();
+        }
+
+        const ValueType& getType() const override {
+            return type_;
+        }
+
+        bool isValid() const override {
```
darionyaphet: If there is nothing to check, could we remove `isValid` ?

darionyaphet: If there is nothing to check, could we remove `isValid` ?

src/jni/src/datamanlite/RowReader.h
```diff
@@ -0,0 +1,283 @@
...
+
+public:
+    class Iterator;
+
+    class Cell final {
+        friend class Iterator;
+    public:
+        template<typename T>
+        typename std::enable_if<std::is_integral<T>::value, ResultType>::type
+            getInt(T& v) const noexcept;
```
darionyaphet: please alignment

src/jni/src/datamanlite/RowReader.h
```diff
@@ -0,0 +1,283 @@
...
+    // Skip to the {index}Th field
+    // The method retuns the offset of the field
+    // It returns a negative number when the data corrupts
+    int64_t skipToField(int64_t index) const noexcept;
+
+    // The following methods all return the number of bytes read
+    // A negative number will be returned if an error occurs
+    template<typename T>
+    typename std::enable_if<std::is_integral<T>::value, int32_t>::type
+        readInteger(int64_t offset, T& v) const noexcept;
```
darionyaphet: please alignment

src/jni/src/datamanlite/RowReader.h
```diff
@@ -0,0 +1,283 @@
...
+    int32_t readDouble(int64_t offset, double& v) const noexcept;
+    int32_t readString(int64_t offset, Slice& v) const noexcept;
+    int32_t readInt64(int64_t offset, int64_t& v) const noexcept;
+    int32_t readVid(int64_t offset, int64_t& v) const noexcept;
+
+    // Following methods assume the parameters index and offset are valid
+    // When succeeded, offset will advance
+    template<typename T>
+    typename std::enable_if<std::is_integral<T>::value, ResultType>::type
+        getInt(int64_t index, int64_t& offset, T& v) const noexcept;
```
darionyaphet: alignment

src/jni/src/datamanlite/RowReader.cpp
```diff
@@ -0,0 +1,595 @@
...
+            int64_t v;
+            int32_t len = readInteger(offset, v);
+            if (len <= 0) {
+                return static_cast<int64_t>(ResultType::E_DATA_INVALID);
+            }
+            offset += len;
+            break;
+        }
+        case ValueType::FLOAT: {
+            // Eight bytes
```
darionyaphet: sizeof(float) is 4 ?

src/jni/java/src/main/java/com/vesoft/nebula/NebulaCodec.java
```diff
@@ -0,0 +1,104 @@
...
+import java.util.Objects;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class NebulaCodec {
+    private static final Logger LOGGER = LoggerFactory.getLogger(NebulaCodec.class.getName());
+
+    static {
+        System.loadLibrary("nebula_codec");
```
critical27: We'd better catch exception/error here, when java client import this as dependency, will throw exception and error.
```
        try {
            System.loadLibrary("nebula_codec");
        } catch (Exception e) {
            LOGGER.error(e.getMessage(), e);
        } catch (Error e) {
            LOGGER.error(e.getMessage(), e);
        }
```

src/jni/src/datamanlite/NebulaSchemaProvider.h
dangleptr: It is copied from dataman.  It will be refactored after dataman2

dangleptr: ditto

src/jni/src/datamanlite/RowReader.h
dangleptr: good catch

src/jni/src/datamanlite/RowReader.h
dangleptr: good catch

src/jni/src/datamanlite/RowReader.cpp
dangleptr: good catch.

src/jni/java/src/main/java/com/vesoft/nebula/NebulaCodec.java
dangleptr: Good point

Issue #1364, opened 2019-12-02
go from vid over *  reversely leads graphd server to crash


jievince:

jievince:
logs/nebula-graphd.ERROR
```
Log file created at: 2019/12/02 16:16:12
Running on machine: localhost.localdomain
Log line format: [IWEF]mmdd hh:mm:ss.uuuuuu threadid file:line] msg
F1202 16:16:12.468133 39624 GoExecutor.cpp:1316] Check failed: iter != edges_.end()
```

Amber1990Zhang:
go from vid over edge1, edge2
go form vid over *
both works. So I guess it's the `REVERSELY` that matters.

jievince:
In my test, ` go from hash("叶莉") over relation reversely` works well, while `go from hash("叶莉") over reversely` will lead graphd server to crash.

Amber1990Zhang:
![image](https://user-images.githubusercontent.com/42762957/69945319-4642a480-1524-11ea-9e6f-74a12fd1313d.png)

PR #1403, opened 2019-12-09
 Fixed graphd crash

monadbobo:
Fixed crash caused when not specifying edge type (over *) and setting reversely.


src/graph/test/GoTest.cpp
```diff
@@ -917,6 +917,25 @@ TEST_F(GoTest, ReverselyOneStep) {
         };
         ASSERT_TRUE(verifyResult(resp, expected));
     }
+    {
+        cpp2::ExecutionResponse resp;
+        auto query = "GO FROM hash('Tim Duncan') OVER * REVERSELY "
+                     "YIELD like._src";
```
CPWstatic: Would you please add an UT with multi-edges?  I am not quite sure if this test case could cover it.

src/graph/test/GoTest.cpp
```diff
@@ -380,7 +380,90 @@ TEST_F(GoTest, MULTI_EDGES) {
         };
         ASSERT_TRUE(verifyResult(resp, expected));
     }
-
+    {
+        cpp2::ExecutionResponse resp;
```
laura-ding: Could you add a test to cover reversely multi-edge and the result has different props on the different edges?

src/graph/test/GoTest.cpp
```diff
@@ -380,7 +380,117 @@ TEST_F(GoTest, MULTI_EDGES) {
...
+        ASSERT_TRUE(verifyResult(resp, expected));
+    }
+    {
+        cpp2::ExecutionResponse resp;
+        auto *fmt = "GO FROM %ld OVER * REVERSELY yield like.likeness, teammate.start_year, "
+                    "serve.start_year, $$.player.name";
+        auto &player = players_["Manu Ginobili"];
+        auto query = folly::stringPrintf(fmt, player.vid());
+        auto code = client_->execute(query, resp);
+        ASSERT_NE(cpp2::ErrorCode::SUCCEEDED, code);
```
CPWstatic: Why are we not check the result here? We could also yield source vertex props.

jude-zhu:
close #1364

PR #1409, opened 2019-12-10
Support show with default info

darionyaphet:
```
create TAG tag_with_default(name string default "darion", age int default 18)
Execution succeeded (Time spent: 1840/2313 us)

create EDGE like_with_default(likeness int default 80)
Execution succeeded (Time spent: 1560/2033 us)

show create tag tag_with_default;
==================================
| Tag              | Create Tag   |
==================================
| tag_with_default | CREATE TAG tag_with_default (
  name string default darion,
  age int default 18
) ttl_duration = 0, ttl_col = "" |
-----------------------------------------------
Got 1 rows (Time spent: 1116/1653 us)

show create edge like_with_default;
==================================
| Edge              | Create Edge |
==================================
| like_with_default | CREATE EDGE like_with_default (
  likeness int default 80
) ttl_duration = 0, ttl_col = "" |
-----------------------------------------------
Got 1 rows (Time spent: 1105/1656 us)
```


Amber1990Zhang:
Is `ttl` supported now?

src/graph/ShowExecutor.cpp
```diff
@@ -460,11 +460,34 @@ void ShowExecutor::showCreateTag() {
...
+                    case nebula::cpp2::SupportedType::STRING:
+                        defaultValue = folly::to<std::string>(value->get_string_value());
+                        break;
+                    default:
+                        LOG(ERROR) << "Unsupported type";
+                        return;
+                }
+                buf += " default ";
+                buf += defaultValue;
+            }
```
liuyu85cn: why not extract the two changes in file to one function ?

darionyaphet: It looks like won't be reused anywhere else and it's really simple :)

liuyu85cn: all right

PR #1411, opened 2019-12-10
local conf mode

### LINKED ISSUE #1375: Takes too much hard disk space
image

partition_num=100
replica_factor=1

critical27:
1. For test, we can use local conf instead of use config in meta server.
2. Support force update configs if user knows exactly what he is doing, so we can update some immutable configs. 

e.g. `update configs storage:custom_filter_interval_secs=36000 force`
Even though `custom_filter_interval_secs` is a IMMUTABLE config by default, we can still update it with force flag.

close #1375


critical27:
Jenkins go

dangleptr:
Please update the docs @Amber1990Zhang @critical27

PR #1427, opened 2019-12-11
Add some error logs and give conf templates for production cluster

dangleptr:
1. Configurations on production cluster are different with test cluster. So we give the templates for production cluster. Our users should use these templates as their default configurations when doing load test. 
2. Add more error logs on storaged


dangleptr:
Jenkins go

yixinglu:
Do these **production** configure files need to be copied to `/usr/local/nebula/etc` when we make nebula rpm/deb packages?

src/kvstore/Part.cpp
```diff
@@ -441,6 +423,23 @@ bool Part::preProcessLog(LogID logId,
     return true;
 }
 
+ResultCode Part::toResultCode(raftex::AppendLogResult res) {
```
Shylock-Hg: Could make static.

etc/nebula-graphd.conf.production
```diff
@@ -0,0 +1,54 @@
...
+# Whether to run as a daemon process
+--daemonize=true
+# The file to host the process id
+--pid_file=pids/nebula-graphd.pid
+
+########## logging ##########
+# The directory to host logging files, which must already exists
+--log_dir=logs
+# Log level, 0, 1, 2, 3 for INFO, WARNING, ERROR, FATAL respectively
+--minloglevel=0
```
Shylock-Hg: Maybe not enable INFO level.

src/kvstore/Part.cpp
dangleptr: It can't be static, because we need idStr_.

etc/nebula-graphd.conf.production
dangleptr: On production cluster,  the default log level is INFO.

dangleptr:
> Do these **production** configure files need to be copied to `/usr/local/nebula/etc` when we make nebula rpm/deb packages?

Good point.  WDYT?  @yixinglu

src/kvstore/Part.cpp
```diff
@@ -441,6 +423,23 @@ bool Part::preProcessLog(LogID logId,
...
+    switch (res) {
+        case raftex::AppendLogResult::SUCCEEDED:
+            return ResultCode::SUCCEEDED;
+        case raftex::AppendLogResult::E_NOT_A_LEADER:
+            return ResultCode::ERR_LEADER_CHANGED;
+        case raftex::AppendLogResult::E_WRITE_BLOCKING:
+            return ResultCode::ERR_WRITE_BLOCK_ERROR;
+        default:
```
critical27: Maybe we can assign AppendLogResult::E_BUFFER_OVERFLOW to another ResultCode?

dangleptr: There are already too many codes inside ResultCode and cpp2::ErrorCode. 
We could check the internal consensus error code with this log.

dangleptr:
> > Do these **production** configure files need to be copied to `/usr/local/nebula/etc` when we make nebula rpm/deb packages?
> 
> Good point. WDYT? @yixinglu

Let's open another pr if needed in the future.

PR #1432, opened 2019-12-12
Fix typo problem

dangleptr:


laura-ding:
@dangleptr The change has been included in #1403

PR #1437, opened 2019-12-12
index scan

bright-starry-sky:


src/common/base/NebulaKeyUtils.cpp
```diff
@@ -88,6 +90,73 @@ std::string NebulaKeyUtils::kvKey(PartitionID partId, const folly::StringPiece&
...
+// static
+std::string NebulaKeyUtils::edgeIndexKey(PartitionID partId, IndexID indexId,
+                                         VertexID srcId, EdgeRanking rank,
+                                         VertexID dstId, const IndexValues& values) {
+    auto raw = indexRaw(values);
+    int32_t item = (partId << 8) | static_cast<uint32_t>(NebulaKeyType::kData);
+    std::string key;
+    key.reserve(kIndexPrefix.size() + kEdgeIndexLen + raw.size());
+    key.append(reinterpret_cast<const char*>(&item), sizeof(int32_t))
+            .append(kIndexPrefix.data(), kIndexPrefix.size())
```
darionyaphet: please alignment thanks

bright-starry-sky:
1: Using kIndex instead of kData. 
2: Removed kIndexPrefix from index key. 
3: Removed IndexItem from all thrift requests. 
4: Moved hint from storage.thrift to common.thrift. 
5:fix alignment problem.

bright-starry-sky:
1, Resolve conflict. 
2, Directory structure refactoring

bright-starry-sky:
Please hold this pr. A code synchronization is required with #1360

bright-starry-sky:
1: Resolve the conflict
2: Reconcile pr #1360
3: Disable the property data get when no columns need to be returned.

src/interface/common.thrift
```diff
@@ -100,4 +102,22 @@ struct Pair {
...
+struct IndexItem {
+    1: required IndexID          index_id,
+    2: required i32              tagOrEdge,
+    3: required list<ColumnDef>  cols,
+}
+
+struct IndexHintItem {
+    1: binary                first_str,
+    2: binary                second_str,
+    3: SupportedType         type,
```
bright-starry-sky: It seems that `type` can be delete. let me check.

bright-starry-sky:
Remove `type` from `IndexHintItem`
Ready to review

bright-starry-sky:
1, Resolved conflict.
2,Improved error handling.

dangleptr:
Let's go on after #1360 merged.

src/common/base/NebulaKeyUtils.h
```diff
@@ -276,6 +276,27 @@ class NebulaKeyUtils final {
         return raw;
     }
 
+    static VertexID getIndexVertexID(const folly::StringPiece& rawKey) {
+         auto offset = rawKey.size() - sizeof(VertexID);
```
dangleptr: You'd better check the key.size()

src/common/base/NebulaKeyUtils.h
```diff
@@ -276,6 +276,27 @@ class NebulaKeyUtils final {
...
+         return *reinterpret_cast<const VertexID*>(rawKey.data() + offset);
+     }
+
+    static VertexID getIndexSrcId(const folly::StringPiece& rawKey) {
+        auto offset = rawKey.size() -
```
dangleptr: ditto

src/common/base/NebulaKeyUtils.h
```diff
@@ -276,6 +276,27 @@ class NebulaKeyUtils final {
...
+                      sizeof(VertexID) * 2 - sizeof(EdgeRanking);
+        return readInt<VertexID>(rawKey.data() + offset, sizeof(VertexID));
+    }
+
+    static VertexID getIndexDstId(const folly::StringPiece& rawKey) {
+        auto offset = rawKey.size() - sizeof(VertexID);
```
dangleptr: ditto

src/common/base/NebulaKeyUtils.h
```diff
@@ -276,6 +276,27 @@ class NebulaKeyUtils final {
...
+        return readInt<VertexID>(rawKey.data() + offset, sizeof(VertexID));
+    }
+
+    static EdgeRanking getIndexRank(const folly::StringPiece& rawKey) {
+        auto offset = rawKey.size() - sizeof(VertexID) - sizeof(EdgeRanking);
```
dangleptr: ditto

src/interface/common.thrift
```diff
@@ -103,9 +103,20 @@ struct Pair {
...
 struct IndexItem {
-     1: required IndexID          index_id,
-     2: required i32              tagOrEdge,
-     3: required list<ColumnDef>  cols,
+    1: required IndexID          index_id,
+    2: required i32              tagOrEdge,
+    3: required list<ColumnDef>  cols,
+}
+
+struct IndexHintItem {
```
dangleptr: Could you add some comments for this structure.

src/interface/common.thrift
```diff
@@ -103,9 +103,20 @@ struct Pair {
...
+    1: binary                first_str,
+    2: binary                second_str,
+}
+
+struct IndexHint {
```
dangleptr: ditto

src/interface/storage.thrift
```diff
@@ -188,6 +208,13 @@ struct EdgePropRequest {
     5: list<PropDef> return_columns,
 }
 
+struct IndexScanRequest {
+    1: common.GraphSpaceID       space_id,
+    2: list<common.PartitionID>  parts,
```
dangleptr: Do we really need this field?

src/interface/storage.thrift
```diff
@@ -188,6 +208,13 @@ struct EdgePropRequest {
...
+    3: common.IndexHint          hint,
+    4: list<string>              return_columns,
```
dangleptr: Each element is the column name?

src/interface/storage.thrift
```diff
@@ -412,4 +439,8 @@ service StorageService {
     ExecResponse      removeRange(1: RemoveRangeRequest req);
 
     GetUUIDResp getUUID(1: GetUUIDReq req);
+
+    // Interfaces for edge and vertex index scan
+    ScanVertexResponse scanVertexIndex(1: IndexScanRequest req);
```
dangleptr: LookUpVertexIndex

src/interface/storage.thrift
```diff
@@ -412,4 +439,8 @@ service StorageService {
...
+    ScanEdgeResponse   scanEdgeIndex(1: IndexScanRequest req);
```
dangleptr: Ditto

src/interface/storage.thrift
```diff
@@ -188,6 +208,13 @@ struct EdgePropRequest {
     5: list<PropDef> return_columns,
 }
 
+struct IndexScanRequest {
```
dangleptr: LookUpIndexRequest

src/interface/storage.thrift
```diff
@@ -118,6 +126,18 @@ struct QueryResponse {
     4: optional list<VertexData> vertices,
 }
 
+struct ScanVertexResponse {
```
dangleptr: LookUpVertexIndexResp

src/interface/storage.thrift
```diff
@@ -118,6 +126,18 @@ struct QueryResponse {
...
+    1: required ResponseCommon             result,
+    2: optional common.Schema              schema,
+    3: optional list<VertexIndexData>      rows,
+}
+
+struct ScanEdgeResponse {
```
dangleptr: LookUpEdgeIndexResp

src/storage/client/StorageClient.inl
```diff
@@ -234,6 +234,87 @@ folly::Future<StatusOr<Response>> StorageClient::getResponse(
     return f;
 }
 
+
+template<class Request, class RemoteFunc, class Response>
+folly::SemiFuture<StorageRpcResponse<Response>> StorageClient::getResponse(
```
dangleptr: Why not reuse current 'getResponse' method ?

src/storage/index/ScanEdgeIndexProcessor.cpp
```diff
@@ -0,0 +1,168 @@
...
+namespace nebula {
+namespace storage {
+
+void ScanEdgeIndexProcessor::process(const cpp2::IndexScanRequest& req) {
+    spaceId_ = req.get_space_id();
+    hint_ = req.get_hint();
+    auto indexId = hint_.index_id;
+    auto iRet = schemaMan_->getEdgeIndex(spaceId_, indexId);
+    if (!iRet.ok()) {
+        putResultCodes(cpp2::ErrorCode::E_LOAD_META_FAILED, req.get_parts());
```
dangleptr: Maybe INDEX_NOT_FOUND is a better choice

src/storage/index/ScanEdgeIndexProcessor.cpp
```diff
@@ -0,0 +1,168 @@
...
+
+    for (const auto& col : index_.get_cols()) {
+        if (col.get_type().get_type() == nebula::cpp2::SupportedType::STRING) {
+            vlColNum_ += 1;
+        }
+    }
+
+    std::vector<folly::Future<std::pair<PartitionID, kvstore::ResultCode>>> results;
+    for (auto& part : req.get_parts()) {
+        results.emplace_back(asyncProcess(part, indexId, edgeType));
```
dangleptr: I think search all parts one by one is enough here.

src/interface/storage.thrift
bright-starry-sky: Yes , each element is a column name. the column list corresponds to yield clause.

src/storage/index/LookUpIndexBaseProcessor.inl
```diff
@@ -0,0 +1,201 @@
...
+template<typename REQ, typename RESP>
+cpp2::ErrorCode ScanIndexBaseProcessor<REQ, RESP>::buildIndexHint() {
+    // TODO(sky) : Currently only supported :
+    //             1) Range query for the first column
+    //             2) Equivalent query for all columns
+    if (hint_.is_range) {
+        if (hint_.hint_items.size() != 1) {
+            return cpp2::ErrorCode::E_INVALID_INDEX_HINT;
+        }
+        range_ = std::make_pair(hint_.hint_items[0].get_first_str(),
```
critical27: When we can only scan index within range [start, end], user specify it or graph?

src/storage/index/LookUpIndexBaseProcessor.inl
```diff
@@ -0,0 +1,201 @@
...
+                                hint_.hint_items[0].get_second_str());
+    } else {
+        prefix_.reserve(128);
+        for (auto& hint : hint_.hint_items) {
+                prefix_.append(hint.get_first_str());
```
critical27: Who take care of the index hint order? I guess graph check the what kind of index we have, and if thery are the first N columns?

src/storage/index/LookUpEdgeIndexProcessor.cpp
```diff
@@ -0,0 +1,168 @@
...
+    if (hint_.is_range) {
+        std::unique_ptr<kvstore::KVIterator> iter;
+        auto begin = prefix + range_.first;
+        auto end = prefix + range_.second;
+        auto ret = kvstore_->range(spaceId_,
+                                   part,
+                                   begin,
+                                   end,
+                                   &iter);
+        if (ret == kvstore::SUCCEEDED) {
```
critical27: Similar to the code block when we use prefix, maybe we can unify them to base.

bright-starry-sky:
Please hold this pr review, I'm refactoring the code.
I'll push down the optimization of the index scan to the storage layer.

src/interface/common.thrift
```diff
@@ -103,9 +103,20 @@ struct Pair {
...
-     1: required IndexID          index_id,
-     2: required i32              tagOrEdge,
-     3: required list<ColumnDef>  cols,
+    1: required IndexID          index_id,
+    2: required i32              tagOrEdge,
+    3: required list<ColumnDef>  cols,
+}
+
+struct IndexHintItem {
+    1: binary                first_str,
```
darionyaphet: what is the first and second string’s meaning ？

bright-starry-sky: > what is the first and second string’s meaning ？

 Aha , I'm refactoring the index scan logic now. Most of the existing code will be discarded. 
By the way, the first and second string is a data scan range-pair,  I'll use the native expression instead of it in the storage layer.

bright-starry-sky: > what is the first and second string’s meaning ？


what is the first and second string’s meaning ？

The index scan filter logic push down to storage layer.

src/storage/client/StorageClient.h
```diff
@@ -222,7 +237,8 @@ class StorageClient {
     folly::SemiFuture<StorageRpcResponse<Response>> collectResponse(
         folly::EventBase* evb,
         std::unordered_map<HostAddr, Request> requests,
-        RemoteFunc&& remoteFunc);
+        RemoteFunc&& remoteFunc,
+        GetIdFunc f);
```
dangleptr: GetPartIDFunc maybe a better name. 
And could we give it a default value?

src/storage/index/IndexPolicyMaker.h
```diff
@@ -0,0 +1,117 @@
...
+     */
+    PREFIX_SCAN        = 0x02,
+    /**
+     * accurate scan, Fit all filter conditions. The results are accurate,
+     * so no more filter is needed.
+     */
+    ACCURATE_SCAN      = 0x03,
+};
+
+using OperatorList = std::vector<std::pair<std::pair<std::string, VariantType>, bool>>;
```
dangleptr: Add comments for this structure.

src/storage/index/IndexPolicyMaker.h
```diff
@@ -0,0 +1,117 @@
...
+     * Scan policy optimized,  PolicyScanType will be valid here.
+     */
+    OPTIMIZED_POLICY   = 0x02,
+};
+
+enum class PolicyScanType : uint8_t {
+    /**
+     * No index column match. seek scan and using expression condition filter.
+     */
+    SEEK_SCAN          = 0x01,
```
dangleptr: INDEX_SCAN maybe a better name.

src/storage/index/IndexPolicyMaker.h
```diff
@@ -0,0 +1,117 @@
...
+    /**
+     * Few index column match, index scan in subset.
+     * And using expression condition filter for remaining index column.s
+     */
+    PREFIX_SCAN        = 0x02,
```
dangleptr: PREFIX_SCAN and ACCURATE_SCAN are the same things.  Unify them.

src/storage/index/IndexPolicyMaker.h
```diff
@@ -0,0 +1,117 @@
...
+#define STORAGE_INDEXPOLICYMAKER_H
+#include "base/Base.h"
+#include "meta/SchemaManager.h"
+#include "storage/CommonUtils.h"
+#include "storage/index/IndexOptimizer.h"
+
+namespace nebula {
+namespace storage {
+
+enum class PolicyType : uint8_t {
```
dangleptr: I am confused about this enum.  Why we need it.

src/storage/index/IndexPolicyMaker.h
bright-starry-sky: They are different, `ACCURATE_SCAN` does not need to use expression filtering ,  PREFIX_SCAN and SEEK_SCAN needed. I will unify PREFIX_SCAN and SEEK_SCAN . 
WDYT?

src/storage/index/IndexPolicyMaker.h
bright-starry-sky: When we traverse the expression tree, we terminate the traversal when  `OR` Operator occur.  if so , the  `PolicyType = SIMPLE_POLICY`
Also, when create scan policy, if the `PolicyType = SIMPLE_POLICY` ,  The logic of policy creation will be skipped。

bright-starry-sky: resolved.

src/storage/index/IndexExecutor.inl
```diff
@@ -0,0 +1,329 @@
...
+    return ret;
+}
+
+template<typename RESP>
+kvstore::ResultCode IndexExecutor<RESP>::getVertexRow(PartitionID partId,
+                                                      const folly::StringPiece& key,
+                                                      cpp2::VertexIndexData* data) {
+    auto vId = NebulaKeyUtils::getIndexVertexID(key);
+    data->set_vertex_id(vId);
+    if (schema_ == nullptr) {
```
darionyaphet: `schema_ == null` Is that possible?

src/storage/index/IndexExecutor.inl
```diff
@@ -0,0 +1,329 @@
...
+    auto src = NebulaKeyUtils::getIndexSrcId(key);
+    auto rank = NebulaKeyUtils::getIndexRank(key);
+    auto dst = NebulaKeyUtils::getIndexDstId(key);
+    cpp2::EdgeKey edge;
+    edge.set_src(src);
+    edge.set_edge_type(tagOrEdge_);
+    edge.set_ranking(rank);
+    edge.set_dst(dst);
+    data->set_key(edge);
+    if (schema_ == nullptr) {
```
darionyaphet: similar with the above

src/storage/index/IndexExecutor.inl
bright-starry-sky: > `schema_ == null` Is that possible?

Emm… the schema_ is a shared_ptr as below : 
std::shared_ptr<SchemaWriter>          schema_{nullptr};

src/storage/index/IndexPolicyMaker.cpp
```diff
@@ -0,0 +1,149 @@
...
+    prefix_.reserve(256);
+    decltype(operatorList_.size()) hintNum = 0;
+    bool hasStr = false;
+    for (auto& col : index_->get_fields()) {
+        auto it = std::find_if(operatorList_.begin(), operatorList_.end(),
+                               [&col] (const auto& tup) {
+                                   return col.get_name() == std::get<0>(tup);;
+                               });
+        if (it != operatorList_.end()) {
+            if (std::get<2>(*it) == RelationalExpression::Operator::EQ) {
```
dangleptr: We don't handle ">, <, >=, <=" ?

bright-starry-sky:
> Totally, the pr looks good to me now.
> Do you want to add more policies in the future?
> Especially for the case "A > 10" or "10 < A < 20"

Of course, in the future, we'll add different policies, for example ：
tag (col1 , col2 , col3) 
index (col1, col2)
where col1 > 10 and col1 < 20  
| where (col2 < 10 or col2 > 20) and col1 == 5
| col1 > 10 and col2 > 10
| Etc.

src/storage/index/IndexExecutor.inl
darionyaphet: when `checkReturnColumns` the schema have constructed.

BTW is the schema is null why return `SUCCEEDED`?

src/storage/index/IndexPolicyMaker.cpp
bright-starry-sky: > We don't handle ">, <, >=, <=" ?

Yes , currently use expression to handle such operations  ">, <, >=, <=" .
 In fact, the best way is to handle with it in a subset of range.

src/storage/index/IndexPolicyMaker.cpp
```diff
@@ -0,0 +1,149 @@
...
+                prop = *aExpr->prop();
+            } else {
+                traversalExpression(left);
+                traversalExpression(right);
+                break;
+            }
+            operatorList_.emplace_back(std::make_tuple(std::move(prop), std::move(v), rExpr->op()));
+            break;
+        }
+        case nebula::Expression::kFunctionCall : {
```
darionyaphet: `kUUID` maybe same with `kFunctionCall `

src/storage/index/IndexExecutor.inl
bright-starry-sky: > when `checkReturnColumns` the schema have constructed.
> 
> BTW is the schema is null why return `SUCCEEDED`?

Aha, The result set such as `tag_index(VertexID , col1, col2...) `or `edge_index(srcVID, dstVID, Ranking, col1 , col2...)` , if the yield clause is empty , so we only need return (VertexID) or (srcVID, dstVID, Ranking) . So we don't have to handle with the schema, so return SUCCEEDED direct.

src/storage/index/IndexPolicyMaker.cpp
```diff
@@ -0,0 +1,149 @@
...
+                hasStr = (v.which() == VAR_STR);
+                prefix_.append(NebulaKeyUtils::encodeVariant(v));
+            } else {
+                break;
+            }
+        } else {
+            break;
+        }
+    }
+    if (hintNum == operatorList_.size() && !hasStr) {
```
darionyaphet: why could not contain String ?

src/storage/index/IndexPolicyMaker.cpp
bright-starry-sky: Good point, In the future, we need add the logic when uuid supprot where clause in the future.

src/storage/index/IndexPolicyMaker.cpp
bright-starry-sky: > why could not contain String ?

Because string is a variable-length field, We need to filter this field with expression. for example : 
index (c1, c2)
row1 ("ABC", "ABC")
row2("AB", "CABC")
where c1 == "ABC"    

the two rows has same prefix string "ABCABC" , we need using expression's eval() method to get the correct row row1.

src/storage/index/IndexExecutor.h
```diff
@@ -0,0 +1,102 @@
...
+     *         1, Trigger policy generator
+     *         2, Build prefix string for first n columns of index.
+     *         3, Collect information needed for index scanning,
+     **/
+    cpp2::ErrorCode buildExecutionPlan(const std::string& filter);
+
+    /**
+     * Details Scan index part as one by one.
+     **/
+    kvstore::ResultCode executeExecutionPlan(PartitionID part);
```
darionyaphet: Sorry to say, the execution plan was representative the behavior of `Query Engine` do you mind use another name ?

src/storage/index/IndexPolicyMaker.cpp
dangleptr: Yes,  full scan the whole index is too costly.

src/storage/index/IndexPolicyMaker.cpp
darionyaphet: Got~

src/storage/index/IndexExecutor.h
bright-starry-sky: > Sorry to say, the execution plan was representative the behavior of `Query Engine` do you mind use another name ?

Sounds reasonable，any Suggestions?

dangleptr: I think its fine here.

src/storage/index/LookUpEdgeIndexProcessor.cpp
```diff
@@ -0,0 +1,69 @@
...
+namespace nebula {
+namespace storage {
+
+void LookUpEdgeIndexProcessor::process(const cpp2::LookUpIndexRequest& req) {
+    /**
+     * step 1 : prepare index meta and structure of return columns.
+     */
+    auto ret = prepareRequest(req);
+    if (ret != cpp2::ErrorCode::SUCCEEDED) {
+        putResultCodes(ret, req.get_parts());
```
dangleptr: onFinished

src/storage/index/LookUpEdgeIndexProcessor.cpp
```diff
@@ -0,0 +1,69 @@
...
+        return;
+    }
+
+    /**
+     * step 2 : build execution plan
+     */
+    ret = buildExecutionPlan(req.get_filter());
+    if (ret != cpp2::ErrorCode::SUCCEEDED) {
+        putResultCodes(ret, req.get_parts());
```
dangleptr: ditto

src/storage/index/LookUpVertexIndexProcessor.cpp
```diff
@@ -0,0 +1,69 @@
...
+namespace nebula {
+namespace storage {
+
+void LookUpVertexIndexProcessor::process(const cpp2::LookUpIndexRequest& req) {
+    /**
+     * step 1 : prepare index meta and structure of return columns.
+     */
+    auto ret = prepareRequest(req);
+    if (ret != cpp2::ErrorCode::SUCCEEDED) {
+        putResultCodes(ret, req.get_parts());
```
dangleptr: ditto

src/storage/index/LookUpVertexIndexProcessor.cpp
```diff
@@ -0,0 +1,69 @@
...
+        return;
+    }
+
+    /**
+     * step 2 : build execution plan
+     */
+    ret = buildExecutionPlan(req.get_filter());
+    if (ret != cpp2::ErrorCode::SUCCEEDED) {
+        putResultCodes(ret, req.get_parts());
```
dangleptr: ditto

src/storage/index/LookUpEdgeIndexProcessor.cpp
darionyaphet: When `putResultCodes` is called, onFinished() is included in this function.

src/storage/test/IndexScanTest.cpp
```diff
@@ -0,0 +1,1487 @@
+/* Copyright (c) 2018 vesoft inc. All rights reserved.
```
darionyaphet: 2020

src/storage/index/IndexExecutor.inl
```diff
@@ -0,0 +1,329 @@
...
+           rowNum_ < FLAGS_max_rows_returned_per_lookup) {
+        auto key = iter->key();
+        /**
+         * Need to filter result with expression if is not accurate scan.
+         */
+        if (isIndexScan_ && !conditionsCheck(key)) {
+            iter->next();
+            continue;
+        }
+        ret = getDataRow(part, key);
```
dangleptr: We'd better find all valid edges/vertices firstly,  then get the properties from then.

src/storage/client/StorageClient.h
```diff
@@ -306,6 +322,28 @@ class StorageClient {
...
+            return Status::Error("Space not found, spaceid: %d", spaceId);
+        }
+
+        auto parts = status.value();
+        for (auto partId = 1; partId <= parts; partId++) {
+            auto metaStatus = getPartMeta(spaceId, partId);
+            if (!metaStatus.ok()) {
+                return metaStatus.status();
+            }
+            auto partMeta = metaStatus.value();
```
laura-ding: std::move(metaStatus).value()

src/storage/index/IndexExecutor.inl
```diff
@@ -0,0 +1,329 @@
...
+template <typename RESP>
+cpp2::ErrorCode IndexExecutor<RESP>::checkReturnColumns(const std::vector<std::string> &cols) {
+    int32_t index = 0;
+    if (!cols.empty()) {
+        schema_ = std::make_shared<SchemaWriter>();
+        auto schema = isEdgeIndex_ ?
+                      schemaMan_->getEdgeSchema(spaceId_, tagOrEdge_) :
+                      schemaMan_->getTagSchema(spaceId_, tagOrEdge_);
+        if (!schema) {
+            VLOG(3) << "Can't find schema, spaceId "
```
laura-ding: why use `VLOG`

src/storage/index/IndexPolicyMaker.h
```diff
@@ -0,0 +1,80 @@
...
+
+protected:
+    meta::SchemaManager*                     schemaMan_{nullptr};
+    meta::IndexManager*                      indexMan_{nullptr};
+    std::unique_ptr<ExpressionContext>       expCtx_{nullptr};
+    std::unique_ptr<Expression>              exp_{nullptr};
+    std::string                              prefix_;
+    std::shared_ptr<nebula::cpp2::IndexItem> index_{nullptr};
+    bool                                     optimizedPolicy_{true};
+    bool                                     isIndexScan_{true};
```
dangleptr: maybe use a better name here.  For example,  skipFilter_ ?

src/storage/index/IndexExecutor.inl
```diff
@@ -0,0 +1,329 @@
...
+                                                  tagOrEdge_);
+        auto row = getRowFromReader(reader.get());
+        data->set_props(std::move(row));
+        if (FLAGS_enable_vertex_cache && vertexCache_ != nullptr) {
+            vertexCache_->insert(std::make_pair(vId, tagOrEdge_),
+                                 iter->val().str(), partId);
+            VLOG(3) << "Insert cache for vId " << vId << ", tagId " << tagOrEdge_;
+        }
+    } else {
+        VLOG(3) << "Missed partId " << partId << ", vId " << vId << ", tagId " << tagOrEdge_;
```
laura-ding: Why error log use VLOG(3)

src/storage/index/IndexExecutor.inl
```diff
@@ -0,0 +1,329 @@
...
+    };
+    return exprEval(getters);
+}
+
+template<typename RESP>
+OptVariantType IndexExecutor<RESP>::decodeValue(const folly::StringPiece& key,
+                                                const folly::StringPiece& prop) {
+    using nebula::cpp2::SupportedType;
+    auto type = indexCols_[prop.str()];
+    return NebulaKeyUtils::decodeVariant(getIndexVal(key, prop), type);
```
laura-ding: `decodeVariant ` return `VariantType`, but here return `OptVariantType`, I think `decodeVariant ` need handle some error.

src/storage/index/IndexExecutor.inl
```diff
@@ -0,0 +1,329 @@
...
+            }
+            default:
+                len = 0;
+        }
+        if (col.get_name() == prop.str()) {
+            break;
+        }
+        offset += len;
+    }
+    return key.subpiece(offset, len);
```
laura-ding: Need to check the` key.len` with `offset + len`, if not equal, need return error

bright-starry-sky: These are never going to be equal.

src/common/base/NebulaKeyUtils.h
```diff
@@ -262,16 +262,30 @@ class NebulaKeyUtils final {
...
+
     static std::string encodeDouble(double v) {
         if (v < 0) {
-            v = -(std::numeric_limits<double>::max() + v);
+            /**
+             *   TODO : now, the -(std::numeric_limits<double>::min())
+             *   have a problem of precision overflow. current return value is -nan.
+             */
+            auto i = *reinterpret_cast<const int64_t*>(&v);
+            i = -(std::numeric_limits<int64_t >::max() + i);
```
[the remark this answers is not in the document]
bright-starry-sky: Sharp eyes 👍

bright-starry-sky:
Addressed comments.

src/storage/index/IndexExecutor.inl
```diff
@@ -0,0 +1,333 @@
...
+    if (ret != nebula::kvstore::SUCCEEDED) {
+        return ret;
+    }
+    while (iter->valid() &&
+           rowNum_ < FLAGS_max_rows_returned_per_lookup) {
+        auto key = iter->key();
+        /**
+         * Need to filter result with expression if is not accurate scan.
+         */
+        if (skipFilter_ && !conditionsCheck(key)) {
```
dangleptr: skipFilter_ || !conditionsCheck(key) ?

bright-starry-sky: I think the best is (skipFilter_ && !conditionsCheck(key)) , because we can step out the if block direct when skipFilter == false .

dangleptr: I am confused. "skipFilter_" means skip the whole expression logic.

bright-starry-sky: Your are right, I've changed the variable name from `skipFilter_` to `requiredFilter_`.

PR #1443, opened 2019-12-13
Optimize go for dst id with N steps

dangleptr:
If opened,  the latency for each rpc request will be dumped. 

This pr is just for debug now.

### DIFF
--- a/src/graph/GoExecutor.cpp
+++ b/src/graph/GoExecutor.cpp
@@ -12,6 +12,8 @@
 #include "dataman/ResultSchemaProvider.h"
 
 
+DEFINE_bool(trace_go, false, "Whether to dump the detail trace log from one go request");
+
 namespace nebula {
 namespace graph {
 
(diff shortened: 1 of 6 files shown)


src/graph/GoExecutor.cpp
```diff
@@ -453,6 +455,15 @@ void GoExecutor::stepOut() {
                            << "error code: " << static_cast<int>(error.second);
             }
         }
+        if (FLAGS_trace_go) {
+            LOG(INFO) << "Step:" << curStep_
+                      << " finished, total request vertices " << starts_.size();
+            auto& hostLatency = result.hostLatency();
+            for (size_t i = 0; i < hostLatency.size(); i++) {
+                LOG(INFO) << hostLatency[i].first << ", time cost " << hostLatency[i].second
```
bright-starry-sky: Is it possible to launch multiple storaged on one host？Would it be better to print the port number as well？

dangleptr: HostAddr includes ip and port.

dangleptr:
Will open it later.

PR #1448, opened 2019-12-13
Business key2 UUID

### LINKED ISSUE #1447: Support uuid(bussiness_key) in spark
1. Support using `uuid` expression in vertex & edge  id generation, when business PK field in hive is of non-integer type.
2. Support  string literal prefix to discriminate different hive tables, whose PK fields could have duplicate value.

Consider this:
hive table1 

PK1 | prop1
---|---
abc@abc.com|value1

hive table2 

PK2| prop1 
---|--- 
abc@abc.com|value1

1. PK columns could be non-integer type, we could use `uuid("abc@abc.com")` to generate integer type id 
2. PK1 & PK2 may have duplicated value, When generating id, we use `uuid("table2"+"abc@abc.com")` to keep it unique.

spacewalkman:
close #1447

### DIFF
--- a/src/tools/spark-sstfile-generator/pom.xml
+++ b/src/tools/spark-sstfile-generator/pom.xml
@@ -24,8 +24,6 @@
         <config.version>1.4.0</config.version>
         <nebula.version>1.0.0-rc2</nebula.version>
         <s2.version>1.0.0</s2.version>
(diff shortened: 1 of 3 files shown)


dangleptr:
What's your thoughts about this pr?   @darionyaphet 
Do we need it in 2.0.  Let me close it in 1.x

PR #1450, opened 2019-12-15
Add timezone set/get and datetime function

### LINKED ISSUE #1289: Support date and time functions

laura-ding:
close #1289

- Add interface to set system timezone and get system timezone
- Add time function
- Modify timestamp display

### DIFF
--- a/src/common/filter/FunctionManager.cpp
+++ b/src/common/filter/FunctionManager.cpp
@@ -8,6 +8,8 @@
 #include "filter/FunctionManager.h"
 #include "time/WallClock.h"
 #include "filter/geo/GeoFilter.h"
+#include "TimeFunction.h"
+#include "TimeType.h"
 
 namespace nebula {
 
@@ -24,7 +26,7 @@ FunctionManager::FunctionManager() {
         auto &attr = functions_["abs"];
         attr.minArity_ = 1;
         attr.maxArity_ = 1;
-        attr.body_ = [] (const auto &args) {
+        attr.body_ = [] (const auto &args, const auto*) {
             return std::abs(Expression::asDouble(args[0]));
         };
     }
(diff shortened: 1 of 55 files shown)


src/common/filter/FunctionManager.cpp
```diff
@@ -272,7 +274,7 @@ FunctionManager::FunctionManager() {
         auto &attr = functions_["length"];
         attr.minArity_ = 1;
         attr.maxArity_ = 1;
-        attr.body_ = [] (const auto &args) {
+        attr.body_ = [] (const auto &args, const auto*) {
```
darionyaphet: why change to pointer？Does the reference have some problem ?

src/common/filter/TimeFunction.h
```diff
@@ -0,0 +1,516 @@
...
+#include "Expressions.h"
+
+/**
+ * Time functions can calculate from 1400 years, and the ones before 1400 years are not supported
+ */
+
+namespace nebula {
+class TimeCommon final {
+public:
+    TimeCommon() = default;
```
darionyaphet: maybe `TimeCommon` don't need to create instance? so we could use delete ?

src/common/filter/test/CMakeLists.txt
```diff
@@ -26,6 +26,14 @@ nebula_add_test(
     LIBRARIES gtest gtest_main ${THRIFT_LIBRARIES} wangle
 )
 
+nebula_add_test(
+    NAME timefunction_test
```
darionyaphet: time_function_test

src/common/filter/FunctionManager.cpp
laura-ding: It is timezone，mostly function doesn't need it.

src/common/filter/TimeType.h
```diff
@@ -0,0 +1,65 @@
...
+    ErrorType = -1,
+    NullType = 0,
+    DataType = 1,
+    DataTimeType = 2,
+    TimeType = 3,
+    TimestampType = 4,
+};
+
+struct Interval {
+    int16_t  year;
```
darionyaphet: the type should be uint

PR #1451, opened 2019-12-15
Support stats of all  executor in graphd

laura-ding:


src/graph/Executor.h
```diff
@@ -28,8 +28,11 @@ namespace graph {
 
 class Executor : public cpp::NonCopyable, public cpp::NonMovable {
 public:
-    explicit Executor(ExecutionContext *ectx) {
+    explicit Executor(ExecutionContext *ectx, const std::string statsName) {
```
yixinglu:
```cpp
const std::string &statsName = ""
```

src/graph/PipeExecutor.cpp
```diff
@@ -11,7 +11,7 @@ namespace nebula {
 namespace graph {
 
 PipeExecutor::PipeExecutor(Sentence *sentence,
-                           ExecutionContext *ectx) : TraverseExecutor(ectx) {
+                           ExecutionContext *ectx) : TraverseExecutor(ectx, "") {
```
yixinglu: Use default parameter: `TraverseExecutor(ectx)`

src/graph/SequentialExecutor.cpp
```diff
@@ -14,7 +14,7 @@ namespace nebula {
 namespace graph {
 
 SequentialExecutor::SequentialExecutor(SequentialSentences *sentences,
-                                       ExecutionContext *ectx) : Executor(ectx) {
+                                       ExecutionContext *ectx) : Executor(ectx, "") {
```
yixinglu: ditto

yixinglu: the default empty string also can be removed here.

Shylock-Hg:
Add lock require for #1444

Issue #1421, opened 2019-12-11
The default value for a property does not work when inserting a vertex


wilsonyou:
I created a tag with a default value specified for a property.

When I inserted a new vertex for the tag without providing the property value, the following error occurred:

> (user@127.0.0.1) [(none)]> use nba
> Execution succeeded (Time spent: 786/1236 us)
> 
> (user@127.0.0.1) [nba]> create tag test5(age int DEFAULT 10);
> Execution succeeded (Time spent: 2005359/2005767 us)
> 
> (user@127.0.0.1) [nba]> insert vertex test5(age) values 100:();
> E1211 03:08:55.570917     1 GraphClient.cpp:85] Thrift rpc call failed: Channel got EOF. Check for server hitting connection limit, server connection idle timeout, and server crashes.
> [ERROR (-3)]:
> (user@127.0.0.1) [nba]>
>

Amber1990Zhang:
I reproduced your issue. 
```(user@127.0.0.1) [liverpool]> insert vertex test5(age) values 101:();
E1211 06:10:51.522341     1 GraphClient.cpp:85] Thrift rpc call failed: Channel got EOF. Check for server hitting connection limit, server connection idle timeout, and server crashes.
[ERROR (-3)]:
```

zhaizhifen:
(user@127.0.0.1) [nba]> 
(user@127.0.0.1) [nba]> insert vertex player(name,age) values 100:("Tim Duncan", 42)
[ERROR (-8)]: Not exist default value
(user@127.0.0.1) [nba]> 
(user@127.0.0.1) [nba]> 
(user@127.0.0.1) [nba]> desc tag player;
==================
| Field | Type   |
==================
| name  | string |
------------------
| aget  | int    |
------------------
Got 2 rows (Time spent: 1135/2201 us)

(user@127.0.0.1) [nba]> insert vertex player(name,age) values 100:("Tim Duncan", 42)
[ERROR (-8)]: Not exist default value
(user@127.0.0.1) [nba]> 
(user@127.0.0.1) [nba]>

laura-ding:
@zhaizhifen, your schema `player`’s column name is `aget` not `age`.

laura-ding:
Fixed, closed

PR #1452, opened 2019-12-15
Fix default value bug

### LINKED ISSUE #1368: It need to judge the default value is the same as the given type when create schema
image

laura-ding:
colse #1421, close #1368


src/graph/InsertVertexExecutor.cpp
```diff
@@ -48,14 +48,14 @@ Status InsertVertexExecutor::check() {
         auto *tagName = item->tagName();
         auto tagStatus = ectx()->schemaManager()->toTagID(spaceId_, *tagName);
         if (!tagStatus.ok()) {
-            LOG(ERROR) << "No schema found for " << tagName;
+            LOG(ERROR) << "No schema found for " << *tagName;
```
darionyaphet: May bad orz

src/graph/test/SchemaTest.cpp
```diff
@@ -46,6 +46,36 @@ TEST_F(SchemaTest, TestComment) {
     }
 }
 
+TEST_F(SchemaTest, TestDefaultValue) {
+    auto client = gEnv->getClient();
```
darionyaphet: have a test case about default value ?

laura-ding: data_test has include it

src/graph/test/SchemaTest.cpp
```diff
@@ -46,6 +46,36 @@ TEST_F(SchemaTest, TestComment) {
...
+        std::string cmd = "CREATE TAG default_tag(name string, age int DEFAULT \"10\")";
+        auto code = client->execute(cmd, resp);
+        ASSERT_EQ(cpp2::ErrorCode::E_EXECUTION_ERROR, code);
+    }
+    {
+        cpp2::ExecutionResponse resp;
+        std::string cmd = "CREATE TAG default_tag(name string  DEFAULT \"\", "
+                          "age int DEFAULT \"10\")";
+        auto code = client->execute(cmd, resp);
+        ASSERT_EQ(cpp2::ErrorCode::E_EXECUTION_ERROR, code);
```
darionyaphet: duplicate with previous

PR #1470, opened 2019-12-18
fix dead lock in uuid

critical27:
My bad, we should not write in synchronous way, which would cause all worker thread blocked.


zhangguoqing:
LGTM!

PR #1471, opened 2019-12-18
Optimize go N steps over edge

dangleptr:
Optimize go executor.
The pr improve the perf for "go 2 steps over edge" about 20x with about 1M vertex ids returned.
Of course,  it will improve the perf for other queries, but i have't test it yet. 

Main changes:
1.  Modify the structure storaged returned,  separate vertexId from props to avoid "getPropByName" called when getting dstId.  (This is a normal request for go)

2.  Return directly if current executor is the rightest.  It avoid one extra encode/decode procedure which cost lots of time if we have many rows returned.

3. Pre calculate the yield columns types to avoid calculate them for every edge.


src/graph/GoExecutor.cpp
```diff
@@ -1010,100 +1103,68 @@ bool GoExecutor::processFinalResult(RpcResponse &rpcResp, Callback cb) const {
...
-                        EdgeType type;
-                        auto found = expCtx_->getEdgeType(edgeName, type);
-                        if (!found) {
-                            return Status::Error(
-                                    "Get edge type for `%s' failed in getters.", edgeName.c_str());
+                for (auto& edge : edata.edges) {
+                    auto dstId = edge.get_dst();
+                    // Check if duplicate
+                    if (distinct_) {
+                        auto ret = uniqResult->emplace(std::make_pair(dstId, edge.get_props()));
```
CPWstatic: It seems not correct here, if we do calculations with the results, the final result could be duplicated.

dangleptr: It should be the same with current logic. 
Could you give me a case for it?

dangleptr: I got your point.  You are right,  we should do distinct operation on final result.

src/graph/GoExecutor.cpp
```diff
@@ -535,44 +548,42 @@ void GoExecutor::maybeFinishExecution(RpcResponse &&rpcResp) {
...
+                return std::make_pair(s.first, std::make_shared<ResultSchemaProvider>(s.second));
+            });
         }
-    }
+        for (auto& vdata : *vertices) {
+            for (auto& edata : vdata.edge_data) {
+                for (auto& edge : edata.get_edges()) {
+                    auto dst = edge.get_dst();
+                    std::unique_ptr<RowReader> reader;
+                    if (!schemas.empty()) {
```
wadeliuyi: maybe can move this check out of the loop

src/graph/GoExecutor.cpp
```diff
@@ -535,44 +548,42 @@ void GoExecutor::maybeFinishExecution(RpcResponse &&rpcResp) {
...
-                    edgeKeys.back().set_ranking(rank);
-                    edgeKeys.back().set_edge_type(type);
-                    ++iter;
-                }
-            }
+        if (eschema != nullptr) {
+            std::transform(eschema->cbegin(),
+                           eschema->cend(),
+                           std::inserter(schemas, schemas.begin()), [] (auto &s) {
+                return std::make_pair(s.first, std::make_shared<ResultSchemaProvider>(s.second));
```
wadeliuyi: Indent

src/graph/GoExecutor.cpp
```diff
@@ -535,44 +548,42 @@ void GoExecutor::maybeFinishExecution(RpcResponse &&rpcResp) {
...
-    }
+        for (auto& vdata : *vertices) {
+            for (auto& edata : vdata.edge_data) {
+                for (auto& edge : edata.get_edges()) {
+                    auto dst = edge.get_dst();
+                    std::unique_ptr<RowReader> reader;
+                    if (!schemas.empty()) {
+                        auto it = schemas.find(edata.type);
+                        DCHECK(it != schemas.end());
+                        reader = RowReader::getRowReader(edge.props, it->second);
```
wadeliuyi: the reader just use once in the loop, can we reload the RowReader::getRowReader and just return a static reader, so it will not malloc and free reader in every loop.

src/graph/GoExecutor.cpp
```diff
@@ -723,85 +708,207 @@ void GoExecutor::finishExecution(RpcResponse &&rpcResp) {
...
-    }
 
     if (onResult_) {
+        std::unique_ptr<InterimResult> outputs;
+        if (!setupInterimResult(std::move(rpcResp), outputs)) {
+            return;
+        }
         onResult_(std::move(outputs));
     } else {
+        auto start = time::WallClock::fastNowInMicroSec();
```
wadeliuyi: check the flag FLAGS_trace_go first?

src/graph/GoExecutor.cpp
```diff
@@ -723,85 +708,207 @@ void GoExecutor::finishExecution(RpcResponse &&rpcResp) {
...
+            && (yields_.empty()
+                || (yields_.size() == 1 && yields_[0]->expr()->isEdgeDstIdExpression()));
+}
+
+StatusOr<std::vector<cpp2::RowValue>> GoExecutor::toThriftResponse(RpcResponse&& rpcResp) {
+    std::vector<cpp2::RowValue> rows;
+    int64_t totalRows = 0;
+    for (auto& resp : rpcResp.responses()) {
+        if (resp.get_total_edges() != nullptr) {
+            totalRows += *resp.get_total_edges();
```
wadeliuyi: maybe *(resp.get_total_edges()) more readable

src/graph/GoExecutor.cpp
```diff
@@ -723,85 +708,207 @@ void GoExecutor::finishExecution(RpcResponse &&rpcResp) {
...
+                        std::vector<cpp2::ColumnValue> row;
+                        row.emplace_back();
+                        row.back().set_id(edge.get_dst());
+                        rows.emplace_back();
+                        rows.back().set_columns(std::move(row));
+                    }  // for edges
+                }  // for edata
+            }   // for `vdata'
+        }   // for `resp'
+        return rows;
```
wadeliuyi: maybe define this function in a sub function is more readable, else the function is too long

src/graph/GoExecutor.cpp
```diff
@@ -1104,44 +1169,78 @@ bool GoExecutor::processFinalResult(RpcResponse &rpcResp, Callback cb) const {
...
+                    getters.getInputProp = [&srcId,
+                                            this] (const std::string &prop) {
+                        return getPropFromInterim(srcId, prop);
                     };
 
+                    std::unique_ptr<RowReader> reader;
+                    if (!edgeSchema.empty()) {
+                        auto it = edgeSchema.find(edgeType);
+                        DCHECK(it != edgeSchema.end());
+                        reader = RowReader::getRowReader(edge.props, it->second);
```
wadeliuyi: because the reader depend edgeSchema, if the edgeSchema is empty, the reader is nullptr, then CHECK(reader != nullptr); will crash , so if can we return directly when edgeSchema is empty?

src/graph/FindPathExecutor.cpp
```diff
@@ -598,6 +599,13 @@ FindPathExecutor::getStepOutProps(bool reversely) {
     }
     std::vector<storage::cpp2::PropDef> props;
     for (auto &e : *edges) {
+        {
+            storage::cpp2::PropDef pd;
+            pd.owner = storage::cpp2::PropOwner::EDGE;
+            pd.name = "_dst";
+            pd.id.set_edge_type(e);
+            props.emplace_back(std::move(pd));
+        }
```
CPWstatic: If `_dst` would return by default, then this property is no more required in prop list.

src/graph/GoExecutor.cpp
```diff
@@ -642,17 +653,10 @@ void GoExecutor::onVertexProps(RpcResponse &&rpcResp) {
...
     auto spaceId = ectx()->rctx()->session()->space();
-    auto &resp = rpcResp.responses();
-    auto *edgeSchema = resp[0].get_edge_schema();
-    if (edgeSchema == nullptr) {
-        return names;
-    }
-
-    for (auto &schema : *edgeSchema) {
-        auto edgeType = schema.first;
+    for (auto edgeType : edgeTypes_) {
```
CPWstatic: As i known, this function is used for supporting `over *`. According to the current implementation, i don't think it is appropriate to write it this way. In most cases you would not get all edge-type records. So this modification would make the output a wide-table which is sparse.

src/graph/GoExecutor.cpp
```diff
@@ -1363,5 +1461,73 @@ SupportedType GoExecutor::getPropTypeFromInterim(const std::string &prop) const
     return index_->getColumnType(prop);
 }
 
+nebula::cpp2::SupportedType GoExecutor::calciteExprType(Expression* exp) const {
```
CPWstatic: Emm, what is `calcite` means?

src/storage/query/QueryBaseProcessor.inl
```diff
@@ -471,9 +478,19 @@ kvstore::ResultCode QueryBaseProcessor<REQ, RESP>::collectEdgeProps(
...
                 };
-                getters.getEdgeRank = [&] () -> VariantType {
+                getters.getEdgeRank = [&rank] () -> VariantType {
                     return rank;
                 };
+                getters.getEdgeDstId = [this,
+                                        &edgeType,
+                                        &dstId] (const std::string& edgeName) -> OptVariantType {
+                    auto edgeStatus = this->schemaMan_->toEdgeType(spaceId_, edgeName);
+                    CHECK(edgeStatus.ok());
```
CPWstatic: We could return error here.

src/interface/storage.thrift
```diff
@@ -88,9 +88,29 @@ struct ResultCode {
...
+struct IdAndProp {
+    1: common.VertexID dst,
+    2: binary props,
+}
+
 struct EdgeData {
-    1: common.EdgeType type,
-    2: binary          data,   // decode according to edge_schema.
+    1: common.EdgeType   type,
+    2: binary            data,   // decode according to edge_schema.
```
CPWstatic: Is this property still in use?

jude-zhu:
close #1604

src/graph/GoExecutor.cpp
dangleptr: Sorry,  what i want is "calculate"

src/graph/FindPathExecutor.cpp
```diff
@@ -580,8 +582,7 @@ Status FindPathExecutor::doFilter(
                         boost::get<int64_t>(temps[1]),
```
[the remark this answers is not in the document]
dangleptr: Good point

src/graph/GoExecutor.cpp
dangleptr: Good point.

src/graph/GoExecutor.cpp
```diff
@@ -535,44 +548,42 @@ void GoExecutor::maybeFinishExecution(RpcResponse &&rpcResp) {
...
+                for (auto& edge : edata.get_edges()) {
+                    auto dst = edge.get_dst();
+                    std::unique_ptr<RowReader> reader;
+                    if (!schemas.empty()) {
+                        auto it = schemas.find(edata.type);
+                        DCHECK(it != schemas.end());
+                        reader = RowReader::getRowReader(edge.props, it->second);
+                        DCHECK(reader != nullptr);
+                        EdgeRanking rank;
+                        auto rc = reader->getVid(_RANK, rank);
```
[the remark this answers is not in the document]
dangleptr: Good point.  Actually,  getVid is not correct,  because we use varint in RowWriter::writeInt

src/graph/GoExecutor.cpp
dangleptr: Good point.

src/graph/GoExecutor.cpp
dangleptr: Actually,  the logic is the same.  Further more,  if you extract edgeNames from edgeSchema in response, it is unordered.

src/graph/GoExecutor.cpp
dangleptr: Good point.

src/graph/GoExecutor.cpp
dangleptr: If getAliasProp invoked, it means the reader should not be NULL.  Otherwise it is a bug.

src/interface/storage.thrift
dangleptr: Good catch.

src/graph/FindPathExecutor.cpp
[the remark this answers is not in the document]
dangleptr: Currently,  on storaged,  we does not handle the case  if "_dst" not passed,

src/graph/GoExecutor.cpp
CPWstatic: Got it.

src/graph/GoExecutor.cpp
```diff
@@ -1408,5 +1468,73 @@ SupportedType GoExecutor::getPropTypeFromInterim(const std::string &prop) const
     return index_->getColumnType(prop);
 }
 
+nebula::cpp2::SupportedType GoExecutor::calculateExprType(Expression* exp) const {
```
laura-ding: If the schema cache of graphd and the storaged are different, maybe has problem.

dangleptr: Yes,  we use the schema in graphd. 
And the problem also existed in current master branch, because different storaged maybe have different versions schema when updating it.

src/graph/GoExecutor.cpp
```diff
@@ -743,85 +729,170 @@ void GoExecutor::finishExecution(RpcResponse &&rpcResp) {
...
-                doError(std::move(ret).status());
-                return;
-            }
+        auto ret = toThriftResponse(std::forward<RpcResponse>(rpcResp));
+        if (FLAGS_trace_go) {
+            LOG(INFO) << "Process the resp from storaged, total time "
+                      << time::WallClock::fastNowInMicroSec() - start << "us";
+        }
+        if (!ret.ok()) {
+            LOG(ERROR) << "Get rows failed: " << ret.status();
```
laura-ding: add `doError()`

PR #1473, opened 2019-12-18
Helm chart for nebula

knightXun:
This pr is to provide a helm chart for nebula. With this chart, we can install nebula in few steps.


knightXun:
@yixinglu this can not be merged, still has some problem

kubernetes/helm/templates/deployment.yaml
```diff
@@ -0,0 +1,76 @@
...
+            cpu: {{ .Values.resources.graphd.limits.cpu | quote }}
+            memory: {{ .Values.resources.graphd.limits.memory | quote }}
+        ports:
+        - name: thrift
+          containerPort: {{ .Values.port.graphd.thriftPort }}
+        - name: http
+          containerPort: {{ .Values.port.graphd.httpPort }}
+        - name: http2
+          containerPort: {{ .Values.port.graphd.http2Port }}
+        command: ["/usr/local/nebula/bin/nebula-graphd", "--flagfile=/usr/local/nebula/etc/nebula-graphd.conf"]
```
yixinglu: Please  refer this entrypoint: https://github.com/vesoft-inc/nebula/blob/3208cabc9320674e04c283e9028c3a4e4b499cde/docker/Dockerfile.graphd#L21

kubernetes/helm/templates/configmap.yaml
```diff
@@ -0,0 +1,165 @@
...
+    ########## basics ##########
+    # Whether to run as a daemon process  
+    --daemonize=false
+    # The file to host the process id
+    --pid_file=pids/nebula-graphd.pid
+    ########## logging ##########
+    # The directory to host logging files, which must already exists
+    --log_dir=logs
+    # Log level, 0, 1, 2, 3 for INFO, WARNING, ERROR, FATAL respectively
+    --minloglevel=2
```
yixinglu: Should let user specify the level of log

knightXun: good point

PR #1481, opened 2019-12-19
add uuid cache

critical27:
Add uuid cache.

We first check if we cached. If not we need to create one in a CAS operation. If CAS succeeded, we can return the result, if failed, return errorcode.

### DIFF
--- a/src/interface/storage.thrift
+++ b/src/interface/storage.thrift
@@ -25,6 +25,7 @@ enum ErrorCode {
     E_PART_NOT_FOUND = -14,
     E_KEY_NOT_FOUND = -15,
     E_CONSENSUS_ERROR = -16,
+    E_ATOMIC_OP_FAIL = -17,
 
     // meta failures
     E_EDGE_PROP_NOT_FOUND = -21,
--- a/src/kvstore/Common.h
+++ b/src/kvstore/Common.h
@@ -26,6 +26,7 @@ enum ResultCode {
     ERR_UNSUPPORTED         = -8,
     ERR_CHECKPOINT_ERROR    = -9,
     ERR_WRITE_BLOCK_ERROR   = -10,
+    ERR_ATOMIC_OP_FAILURE   = -11,
     ERR_UNKNOWN             = -100,
 };
 
--- a/src/kvstore/Part.cpp
+++ b/src/kvstore/Part.cpp
@@ -431,6 +431,8 @@ ResultCode Part::toResultCode(raftex::AppendLogResult res) {
             return ResultCode::ERR_LEADER_CHANGED;
         case raftex::AppendLogResult::E_WRITE_BLOCKING:
             return ResultCode::ERR_WRITE_BLOCK_ERROR;
+        case raftex::AppendLogResult::E_ATOMIC_OP_FAILURE:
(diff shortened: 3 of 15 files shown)


critical27:
> Besides uuid cache, for inserting vertex, maybe we could insert UUID directly if the UUID expression used.
> 
> WDYT?

I don't quite get it, for now if we inserting vertex with uuid, need to acquire a uuid at first, if getUUID ok, then it can insert vertex later

critical27:
> Besides uuid cache, for inserting vertex, maybe we could insert UUID directly if the UUID expression used.
> 
> WDYT?

I don't quite get it, if we insert a vertex with uuid, graph need to send two request in sequence, we acquire a uuid at first, if ok, then we can insert vertex.

dangleptr:
> > Besides uuid cache, for inserting vertex, maybe we could insert UUID directly if the UUID expression used.
> > WDYT?
> 
> I don't quite get it, if we insert a vertex with uuid, graph need to send two request in sequence, we acquire a uuid at first, if ok, then we can insert vertex.

Currently,  it is an CAS operation.  Actually, we could insert the uuid directly when inserting vertex.

critical27:
> > > Besides uuid cache, for inserting vertex, maybe we could insert UUID directly if the UUID expression used.
> > > WDYT?
> > 
> > 
> > I don't quite get it, if we insert a vertex with uuid, graph need to send two request in sequence, we acquire a uuid at first, if ok, then we can insert vertex.
> 
> Currently, it is an CAS operation. Actually, we could insert the uuid directly when inserting vertex.

But it seems the leader of uuid and leader of vId maybe not the same one? I think add a field to indicate if its vId is generated by uuid when insert a vertex, WDYT?

dangleptr:
> > > > Besides uuid cache, for inserting vertex, maybe we could insert UUID directly if the UUID expression used.
> > > > WDYT?
> > > 
> > > 
> > > I don't quite get it, if we insert a vertex with uuid, graph need to send two request in sequence, we acquire a uuid at first, if ok, then we can insert vertex.
> > 
> > 
> > Currently, it is an CAS operation. Actually, we could insert the uuid directly when inserting vertex.
> 
> But it seems the leader of uuid and leader of vId maybe not the same one? How about add a field to indicate if vId is generated by uuid when insert a vertex, WDYT?

In InsertVertexExecutor,  if we hit UUID expression,  we could insert the uuid directly.

critical27:
> > > > > Besides uuid cache, for inserting vertex, maybe we could insert UUID directly if the UUID expression used.
> > > > > WDYT?
> > > > 
> > > > 
> > > > I don't quite get it, if we insert a vertex with uuid, graph need to send two request in sequence, we acquire a uuid at first, if ok, then we can insert vertex.
> > > 
> > > 
> > > Currently, it is an CAS operation. Actually, we could insert the uuid directly when inserting vertex.
> > 
> > 
> > But it seems the leader of uuid and leader of vId maybe not the same one? How about add a field to indicate if vId is generated by uuid when insert a vertex, WDYT?
> 
> In InsertVertexExecutor, if we hit UUID expression, we could insert the uuid directly.

Good point, but the cache only works for insert? go, fetch could use it as well.

dangleptr:
Let's close it. UUID will be abandoned in 2.0.

Issue #1091, opened 2019-10-16
Make the backend service HTTP APIs confirm to RESTful specification


yixinglu:
for example: fetch and update flags of storage:

```
# get all flags of storage service
GET https://ip:port/storage/flags

# get flag with name filter
GET https://ip:port/storage/flags?name=num_workers

# update flags of storage service
POST https://ip:port/storage/flags?num_workers=4&&load_data_interval_secs=10
```

doublex:
+1
`/src/client` has tons of dependencies.
A restful client interface would help a lot.

PR #1482, opened 2019-12-19
Refactor web service implementation

yixinglu:
This is a subtask of #1091 

In this PR, I add a `Router` class to manage all routes registered in web service. Now we can get path parameters from request url.

In addition, I remove the web service singleton instance since we don't use it in multi-thread scenario.

At last, I also cleanup some `base/Base.h` usage in some headers, and replace it with class forward declaration.


yixinglu:
Jenkins go

yixinglu:
Jenkins go

src/webservice/Router.cpp
```diff
@@ -0,0 +1,123 @@
...
+    this->pattern_ = std::make_unique<std::regex>(ss.str());
+}
+
+bool Route::matches(proxygen::HTTPMethod method, const std::string &path) const {
+    if (method_ != method) {
+        return false;
+    }
+    std::string p = path;
+    if (p.size() > 1 && p.back() == '/') {
+        p = path.substr(0, path.length() - 1);
```
laura-ding: p.pop_back();

src/webservice/Router.h
```diff
@@ -0,0 +1,110 @@
...
+    }
+
+    Route *next() const {
+        return next_;
+    }
+
+    bool matches(proxygen::HTTPMethod method, const std::string &path) const;
+
+    Route &handler(ReqHandlerGenerator generator) {
+        this->generator_ = generator;
```
laura-ding: Is here support to modify handler multi times? If not, need to judge the reentrant

src/webservice/Router.cpp
```diff
@@ -0,0 +1,123 @@
...
+        }
+    }
+    return new NotFoundHandler();
+}
+
+void Router::append(Route *route) {
+    if (head_ != nullptr) {
+        // Add current route to tail
+        auto iter = head_;
+        while (iter->next() != nullptr) {
```
laura-ding: Here begins from the head every time. what about adding a tail ptr?

yixinglu: You are right. I consider that there are not too many routes for iterating, so I don't have a pointer to save the tail.  But I can fix it.

src/webservice/Router.cpp
yixinglu: good catch. thx.

src/webservice/Router.h
yixinglu: I want to manage all handlers chain later, so here this generator will be replaced by the latest. I can add some comments and do some checks for that.

yixinglu:
I have restored code formats, please help me to review this PR,  @laura-ding  @dutor  @monadbobo  @dangleptr

src/storage/test/StorageHttpStatusHandlerTest.cpp
```diff
@@ -51,14 +57,14 @@ TEST(StoragehHttpStatusHandlerTest, StorageStatusTest) {
...
     }
-    {
-        auto url = "";
-        auto request = folly::stringPrintf("http://%s:%d%s", FLAGS_ws_ip.c_str(),
-                                           FLAGS_ws_http_port, url);
-        auto resp = http::HttpClient::get(request);
-        ASSERT_TRUE(resp.ok());
-        ASSERT_EQ("status=running\n", resp.value());
-    }
+    // {
```
dangleptr: Why remove the UT?

src/common/network/TypesDef.h
```diff
@@ -0,0 +1,18 @@
...
+#pragma once
+
+#include <utility>
+
+#include "common/base/ThriftTypes.h"
+
+namespace nebula {
+
+// Host address type and utility functions
+using HostAddr = std::pair<IPv4, Port>;
```
dangleptr: Why not define it in Base.h?

src/webservice/Router.cpp
```diff
@@ -0,0 +1,125 @@
...
+        // exact path
+        if (path.back() == '/') {
+            this->path_ = path.substr(0, path.length() - 1);
+        } else {
+            this->path_ = path;
+        }
+    }
+}
+
+std::unique_ptr<std::regex> Route::reToken_ =
```
laura-ding: add `const`

yixinglu: this is static variables

src/graph/test/GraphHttpHandlerTest.cpp
```diff
@@ -52,14 +59,14 @@ TEST_F(GraphHttpHandlerTest, GraphStatusTest) {
...
     }
-    {
-        auto url = "";
-        auto request = folly::stringPrintf("http://%s:%d%s", FLAGS_ws_ip.c_str(),
-                                           FLAGS_ws_http_port, url);
-        auto resp = http::HttpClient::get(request);
-        ASSERT_TRUE(resp.ok());
-        ASSERT_EQ("status=running\n", resp.value());
-    }
+    // {
```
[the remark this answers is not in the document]
yixinglu: Next subtask to fix it.

src/common/network/TypesDef.h
yixinglu: Because some headers will only use this type. If defined in base.h, there are too many needless codes to  include for this type usage.

src/webservice/Router.cpp
yixinglu: @laura-ding  @dangleptr  I have fixed all suggest issues. please review again!

src/webservice/Router.cpp
```diff
@@ -0,0 +1,128 @@
...
+    while (head_ != nullptr) {
+        auto next = head_->next();
+        delete head_;
+        head_ = next;
+    }
+}
+
+Route &Route::handler(ReqHandlerGenerator generator) {
+    CHECK(generators_.empty()) << "Only allowed to register handler once for a route";
+    generators_.push_back(generator);
```
dangleptr: If generators_ has only one element,  why use vector?

src/webservice/Router.cpp
```diff
@@ -0,0 +1,128 @@
...
+}
+
+proxygen::RequestHandler *Route::generateHandler(const std::string &path) const {
+    if (!pattern_) {
+        return generators_.back()({});
+    }
+    std::smatch m;
+    std::regex_search(path, m, *pattern_);
+    PathParams params;
+    CHECK_EQ(groups_.size(), m.size() - 1UL) << "groups is not equal to matches";
```
dangleptr: Why use CHECK_EQ?  Is it normal for mismatch?

src/webservice/Router.cpp
yixinglu: In the beginning, I want to allow to register multiple handlers for one route path. But now I fix it for better code reading.

src/webservice/Router.cpp
yixinglu: Because the regex pattern is generated according to the route path registered by user, something must be wrong when its matched path parameters size is not equal to the user's route path if a url is matched by this pattern.

src/webservice/WebService.h
```diff
@@ -7,49 +7,57 @@
...
-    static std::unique_ptr<proxygen::HTTPServer> server_;
-    static std::unique_ptr<thread::NamedThread> wsThread_;
+    // Check whether this web service is stopped
+    bool stopped() const {
+        return stopped_;
+    }
 
-    static HandlerGen handlerGenMap_;
+private:
+    bool started_{false};
```
dutor: For such a scenario, we typically use either an atomic boolean or a volatile one.

BTW, is it really necessary to use two flags?

src/webservice/Router.h
```diff
@@ -0,0 +1,109 @@
...
+class HTTPMessage;
+}   // namespace proxygen
+
+namespace nebula {
+namespace web {
+
+using PathParams = std::unordered_map<std::string, std::string>;
+using ReqHandlerGenerator = std::function<proxygen::RequestHandler *(PathParams &&)>;
+
+class Route final {
```
dutor: I wonder if it's applicable to make `Route` a subclass of `Router`.

src/storage/CMakeLists.txt
```diff
@@ -28,6 +28,8 @@ nebula_add_library(
     http/StorageHttpAdminHandler.cpp
 )
 
+add_dependencies(storage_http_handler ws_common_obj)
```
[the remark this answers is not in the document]
yixinglu: The storage_http_handler library uses some static function of ws_common_obj and gcc will complain some errors in link step.

src/webservice/Router.h
yixinglu: In my opinion, Router is a manager for all routes.

src/webservice/WebService.h
yixinglu: OK, stopped member is not necessary now, I have removed it. Since the web service only used in one thread, so I don't prevent the accession to the variable from other threads.

PR #1487, opened 2019-12-20
fix insert a extra vertex of vid=0

critical27:
Whenever insert a vertex, it will automatically try to insert a vertex with vId = 0


bright-starry-sky:
LGTM, well down!

PR #1488, opened 2019-12-20
show the port on nebula console prompt

zhangguoqing:
Sometimes we maybe run multiple **graphd** instances on the same IP(machine) but different ports, so the PR will show the port on nebula console prompt.
```
bin/nebula --addr=127.0.0.1 --port=3699 --u=user --p=password
(user@127.0.0.1:3699) [(none)]> 

bin/nebula --addr=127.0.0.1 --port=12345--u=user --p=password
(user@127.0.0.1:12345) [(none)]> 
```


laura-ding:
jenkins go

laura-ding:
jenkins go

PR #1489, opened 2019-12-20
Only update mutable configs, remove force flag in config manager

critical27:
Follow-up of #1485. Remove some unused code. Only update mutable configs.


src/interface/meta.thrift
```diff
@@ -608,7 +608,6 @@ struct GetConfigResp {
 
 struct SetConfigReq {
     1: ConfigItem           item,
-    2: bool                 force,
```
dangleptr: We'd better don't remove the filed in thrift structure.  I

critical27: Make sense.

PR #1492, opened 2019-12-21
Changed format of latency in nebula console

doJCHbest:


laura-ding:
jenkins go

laura-ding:
jenkins go

Issue #1184, opened 2019-10-31
Nebula should remove the unreasonable exit


monadbobo:
Nebula should remove the unreasonable exit(LOG(FATAL)/CHECK).

Amber1990Zhang:
Absolutely.

jude-zhu:
#1495

PR #1495, opened 2019-12-23
Remove CHECK and FATAL in graphd

laura-ding:


src/graph/YieldExecutor.cpp
```diff
@@ -446,8 +446,9 @@ void YieldExecutor::feedResult(std::unique_ptr<InterimResult> result) {
 }
 
 void YieldExecutor::setupResponse(cpp2::ExecutionResponse &resp) {
-    CHECK(resp_ != nullptr);
-    resp = std::move(*resp_);
+    if (resp_ != nullptr) {
```
monadbobo: Here maybe an empty resp should be set to return to the caller, similar to the GoExecutor :: setupResponse method？

laura-ding: `GoExecutor::setupResponse ` create an empty object is redundant, wdyt?

src/graph/AggregateFunction.h
```diff
@@ -71,7 +71,8 @@ class GroupByHash {
                 return 0;
             }
             default: {
-                LOG(FATAL) << "Untreated value type: " << static_cast<int32_t>(col.getType());
+                LOG(ERROR) << "Untreated value type: " << static_cast<int32_t>(col.getType());
+                return 0;
```
whitewum: return 0 or other error code?

jude-zhu:
close #1495

jude-zhu:
#1184

PR #1497, opened 2019-12-23
enhance data type in storage test cases

darionyaphet:


src/storage/test/TestUtils.h
```diff
@@ -117,13 +115,40 @@ class TestUtils {
         return vertices;
     }
 
+    static std::vector<cpp2::Edge>
+    setupEdges(PartitionID part, VertexID srcFrom,
+               VertexID srcTo, std::string tmp = "%d_%ld") {
```
dutor: How about rename `tmp` to `fmt` or some other meaningful name?

dutor:
Please resolve the conflicts.

src/storage/test/TestUtils.h
darionyaphet: yep ,`fmt` is a good choice.

Issue #1501, opened 2019-12-23
When graphd fail to resolve dns,It do not exit


knightXun:
When graphd init, it will try to resolve metad dns.If dns name fail to resolve, Graphd will not exit. I want graphd restart and retry resolve metad dns name.

knightXun:
/assign knightXun 
@monadbobo @dutor

laura-ding:
Adjust the order in which parameters are initialized and move it to the front of `start WebService`

Shylock-Hg:
> When graphd init, it will try to resolve metad dns.If dns name fail to resolve, Graphd will not exit. I want graphd restart and retry resolve metad dns name.

Seems graphd will exit if resolve failed.

laura-ding:
@Shylock-Hg exit without stop WebService

PR #1507, opened 2019-12-24
making graphd exit and retry dns resolve when resolve failed

knightXun:
When graphd fail to resolve metad dns, it will hung there, not exit. In normal case, it should exit. 
this pr make graphd exit, this will make graphd run correctly in kubernetes. 
related issus:  #1501


Shylock-Hg:
We'd better using SCOPE_EXIT to avoid duplicate error.

dutor:
This is issue has been addressed in #1513 , so close it.

PR #1512, opened 2019-12-25
Enhance Spark Writer

darionyaphet:
1. Currently we are using schema name for the config's key. But actually some user will loading data into the same schema from different data source. So I have modified the config's structure to support it.

2. Support error log save into separated HDFS file and reload. 

3. Support rate limit in spark writer


src/tools/spark-sstfile-generator/src/main/scala/com/vesoft/nebula/tools/generator/v2/SparkClientGenerator.scala
```diff
@@ -512,49 +614,68 @@ object SparkClientGenerator {
...
+                          override def onSuccess(result: Optional[Integer]): Unit = {
+                            batchSuccess.add(1)
+                          }
 
-                        override def onFailure(t: Throwable): Unit = {
-                          if (batchFailure.value > DEFAULT_ERROR_TIMES) {
-                            throw new TooManyErrorException("too many error")
+                          override def onFailure(t: Throwable): Unit = {
+                            if (batchFailure.value > DEFAULT_ERROR_TIMES) {
+                              throw new TooManyErrorException("too many error")
```
Amber1990Zhang: too many errors

darionyaphet:
TODO : I will update the document later :D

src/jni/java/pom.xml
```diff
@@ -12,7 +12,7 @@
 
     <groupId>com.vesoft</groupId>
     <artifactId>nebula-utils</artifactId>
-    <version>1.0.0-rc2</version>
+    <version>1.0.0-rc3</version>
```
critical27: Have we uploaded rc3 version jni? rc2 is ok I guess

darionyaphet: I understand that RC3 will release later so I update this version number.
Actually, when RC2 release we should update this version.

critical27: Ok, remember to upload it to mvn

PR #1523, opened 2019-12-26
Make compact operations parallel

zhangguoqing:
Compaction is executed in multiple threads, that is, all rocksdb engines
are compact at the same time when a compact command is received.


zhangguoqing:
jenkins go

PR #1527, opened 2019-12-26
对cpp的更改

kangshaojie:

### DIFF
--- a/build_dep.sh
+++ b/build_dep.sh
@@ -71,7 +71,8 @@ function aptget_install {
         python \
         maven \
         openjdk-8-jdk \
-        unzip
+        unzip \
+	curl ####
 
     installPackage $1
     addAlias $1
--- a/src/common/time/test/DurationTest.cpp
+++ b/src/common/time/test/DurationTest.cpp
@@ -3,7 +3,6 @@
  * This source code is licensed under Apache 2.0 License,
  * attached with Common Clause Condition 1.0, found in the LICENSES directory.
  */
-
 #include "base/Base.h"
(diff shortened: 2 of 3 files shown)


build_dep.sh
```diff
@@ -71,7 +71,8 @@ function aptget_install {
         python \
         maven \
         openjdk-8-jdk \
-        unzip
+        unzip \
+	curl ####
```
whitewum: alignment

src/common/time/test/DurationTest.cpp
```diff
@@ -46,7 +45,7 @@ int main(int argc, char** argv) {
     testing::InitGoogleTest(&argc, argv);
     folly::init(&argc, &argv, true);
     google::SetStderrLogging(google::INFO);
-
+    std::cout << "test\n";
```
whitewum: no cout
delete this line
checkout this file.

src/daemons/GraphDaemon.cpp
```diff
@@ -179,18 +179,23 @@ Status setupSignalHandler() {
...
-    switch (sig) {
-        case SIGINT:
-        case SIGTERM:
-            FLOG_INFO("Signal %d(%s) received, stopping this server", sig, ::strsignal(sig));
-            gServer->stop();
-            break;
-        default:
-            FLOG_ERROR("Signal %d(%s) received but ignored", sig, ::strsignal(sig));
-    }
+    // switch (sig) {
```
whitewum: There are many sigs. So  switch case is better.
git checkout this file.

PR #1529, opened 2019-12-26
install curl in build_dep.sh

NekoKedama:
To add curl in dep in order to make http_test work properly

To install curl while users don't have it 

NO(I THINK)

I uninstall the curl and do build_dep again,it shows the curl will be installed.Then I run http_test and it runs properly.


build_dep.sh
```diff
@@ -35,7 +35,8 @@ function yum_install {
         readline-devel \
         maven \
         java-1.8.0-openjdk \
-        unzip
+        unzip \
+	curl
```
whitewum: make alignment

build_dep.sh
```diff
@@ -71,7 +72,8 @@ function aptget_install {
         python \
         maven \
         openjdk-8-jdk \
-        unzip
+        unzip \
+	curl
```
whitewum: ditto

laura-ding:
jenkins go

dutor:
Jenkins go

PR #1534, opened 2019-12-27
Update CmdProcessor.cpp

999wwx:
use function to realize time_transformation


whitewum:
Dup with #1492.
Anyway, thank you for your passion.

PR #1537, opened 2019-12-27
Make internal error more clear and readable.

CPWstatic:
We have been troubled by `Internal error` for a long time. So this pr is about to make the error msg much more detailed.


CPWstatic:
Jenkins go!

src/graph/AlterTagExecutor.cpp
```diff
@@ -52,16 +52,19 @@ void AlterTagExecutor::execute() {
...
         doFinish(Executor::ProcessControl::kNext);
     };
 
     auto error = [this] (auto &&e) {
-        LOG(ERROR) << "Exception caught: " << e.what();
-        doError(Status::Error("Internal error"));
+        auto msg = folly::stringPrintf("Alter tag `%s' exception: %s.",
+                sentence_->name()->c_str(), e.what().c_str());
+        LOG(ERROR) << msg;
+        doError(Status::Error(msg));
```
laura-ding: std::move(msg), under ditto

src/graph/ConfigExecutor.cpp
```diff
@@ -204,9 +207,11 @@ void ConfigExecutor::getVariables() {
         doFinish(Executor::ProcessControl::kNext);
     };
 
-    auto error = [this] (auto &&e) {
-        LOG(ERROR) << "Exception caught: " << e.what();
-        doError(Status::Error(folly::stringPrintf("Internal error : %s", e.what().c_str())));
+    auto error = [this, name] (auto &&e) {
+        auto msg = folly::stringPrintf("Get config `%s' exception: %s",
```
zhangguoqing: auto msg = folly::stringPrintf("**Get config `%s' exception**: %s", name.c_str(), e.what().c_str());

- [ ]  `%s'  
- [ ]   \`%s\`  
- [x]   '%s'

src/graph/ConfigExecutor.cpp
```diff
@@ -144,20 +144,22 @@ void ConfigExecutor::setVariables() {
     auto future = ectx()->gflagsManager()->setConfig(module, name, type, value);
     auto *runner = ectx()->rctx()->runner();
 
-    auto cb = [this] (auto && resp) {
+    auto cb = [this, name] (auto && resp) {
```
dutor: Please note that `name` here will be copied. But I think it's OK for this scenario.

CPWstatic: The name itself is allocated in stack, when this cb is called, the context has been changed, so it must be copied.

Issue #1425, opened 2019-12-11
Add HTTP API to provide build information.


Shylock-Hg:
**Is your feature request related to a problem? Please describe.**
Get build information from HTTP API such as :
```bash
curl localhost:13000/info
# Nebula-meta
# GIT-SHA: xxxxx
# BUILT DATE: xxxx
# BUILT TYPE: Release
...
```

**Describe the solution you'd like**
Generate the info when building and compile into object.(such as generate sources file)
Or refactor the versionString().
**Describe alternatives you've considered**
Assets installed.

**Additional context**

PR #1540, opened 2019-12-27
Support update heartbeat_interval_secs dynamically

dangleptr:
After #1236 merged,  we can't change the schema through update the load_data_interval_secs dynamically.   

It is unfriendly for our users because the schema changes take effect at least heartbeat_interval_secs seconds. 

Now,  we could support to change "heartbeat_interval_secs" on console like before.


zhangguoqing:
LGTM!

jude-zhu:
close #1425

PR #1541, opened 2019-12-27
Add average error

johnson-623:


whitewum:
Notice that usleep is not accuracy.

PR #1548, opened 2019-12-30
encode leader info from heartbeat, update show hosts/parts

critical27:
Follow-up of #1236, just encode all leader info into a new table `leaders`, and when we show hosts/parts, we can use this table instead of contacting with storage service.

After this merged, we can use it to preheat storage client.

Known issues: because we judge whether a host is active by heartbeat, if we kill a storage node, it would seems we have `more` leader than we should have. See examples below: `heartbeat interval` is 3s, `expired_threshold_sec` is 30s, total leader count should be 100.

1. Normal state
image

2. Now we kill node 27006, we will get extra 35 leaders, which is the leader count of 27006 before, however the first nodes leader count sum is 100.
image

3. When pass `expired_threshold_sec`, the 27006 is regarded as `offline`, everything looks good.
image


dangleptr:
Well done.  
For the problem you mentioned,  maybe we could filter the host which lastHeartBeatTime expired with heartbeat_interval as the threshold

WDYT?

critical27:
> Well done.
> For the problem you mentioned, maybe we could filter the host which lastHeartBeatTime expired with heartbeat_interval as the threshold when showing hosts.
> 
> WDYT?

It that a little bit vulnerable to network? If we lost one heartbeat, leader count would be zero. Let me have a try.

src/meta/processors/partsMan/ListHostsProcessor.cpp
```diff
@@ -56,43 +55,52 @@ StatusOr<std::vector<cpp2::HostItem>> ListHostsProcessor::allHostsWithStatus(
...
         iter->next();
     }
 
     std::unordered_map<HostAddr,
                        std::unordered_map<std::string, std::vector<PartitionID>>> allParts;
-    for (const auto& spaceId : spaces) {
+    for (const auto& spaceId : spaceIds_) {
         // get space name by space id
-        auto spaceName = spaceIdNameMap[spaceId];
+        auto spaceName = spaceIdNameMap_[spaceId];
```
wadeliuyi: const auto &spaceName?

src/meta/processors/partsMan/ListPartsProcessor.cpp
```diff
@@ -91,34 +96,28 @@ ListPartsProcessor::getAllParts() {
...
-                } else if (showAllParts_ || std::find(partIds_.begin(), partIds_.end(), partId) ==
-                                            partIds_.end()) {
-                    LOG(ERROR) << "Maybe not get the leader of partition " << partId;
+    auto activeHosts = ActiveHostsMan::getActiveHosts(kvstore_);
+    while (iter->valid()) {
+        auto host = MetaServiceUtils::parseLeaderKey(iter->key());
+        if (std::find(activeHosts.begin(), activeHosts.end(),
+                      HostAddr(host.ip, host.port)) != activeHosts.end()) {
+            LeaderParts leaderParts = MetaServiceUtils::parseLeaderVal(iter->val());
+            auto partIds = leaderParts[spaceId_];
```
wadeliuyi: const auto &partIds?

src/meta/processors/partsMan/ListPartsProcessor.cpp
```diff
@@ -91,34 +96,28 @@ ListPartsProcessor::getAllParts() {
...
-                    it->set_leader(this->toThriftHost(leader));
-                } else if (showAllParts_ || std::find(partIds_.begin(), partIds_.end(), partId) ==
-                                            partIds_.end()) {
-                    LOG(ERROR) << "Maybe not get the leader of partition " << partId;
+    auto activeHosts = ActiveHostsMan::getActiveHosts(kvstore_);
+    while (iter->valid()) {
+        auto host = MetaServiceUtils::parseLeaderKey(iter->key());
+        if (std::find(activeHosts.begin(), activeHosts.end(),
+                      HostAddr(host.ip, host.port)) != activeHosts.end()) {
+            LeaderParts leaderParts = MetaServiceUtils::parseLeaderVal(iter->val());
```
wadeliuyi: each loop will create a template map, and will destruct after the loop, maybe here can define a function named parseLeaderValBySpaceID(iter->val(), spaceId_) to just return the leaders of current space.

src/meta/processors/admin/HBProcessor.cpp
```diff
@@ -37,25 +37,32 @@ void HBProcessor::process(const cpp2::HBReq& req) {
...
-        ret = ActiveHostsMan::updateHostInfo(kvstore_, host, info);
         if (req.__isset.leader_partIds) {
-            const auto& leaderIds = req.get_leader_partIds();
-            UNUSED(leaderIds);
-            // TODO(zhangguoqing) put the leader parts information into kvstore
+            ret = ActiveHostsMan::updateHostInfo(kvstore_, host, info,
+                                                 req.get_leader_partIds());
+            updateLeader = true;
+            lastUpdateTime = time::WallClock::fastNowInMilliSec();
+            LastUpdateTimeMan::update(kvstore_, lastUpdateTime);
```
dangleptr: It seems that we don't need to update the lastUpdateTime in HeartBeat processor.

src/meta/processors/partsMan/ListHostsProcessor.cpp
critical27: I remeber that operator[] return T&?

src/meta/processors/partsMan/ListPartsProcessor.cpp
critical27: Good suggestion

src/meta/processors/admin/HBProcessor.cpp
critical27: Indeed, fixed

dutor:
Jenkins, go

src/meta/processors/admin/HBProcessor.cpp
```diff
@@ -40,20 +40,20 @@ void HBProcessor::process(const cpp2::HBReq& req) {
...
         HostInfo info(time::WallClock::fastNowInMilliSec());
-        ret = ActiveHostsMan::updateHostInfo(kvstore_, host, info);
         if (req.__isset.leader_partIds) {
-            const auto& leaderIds = req.get_leader_partIds();
-            UNUSED(leaderIds);
-            // TODO(zhangguoqing) put the leader parts information into kvstore
+            ret = ActiveHostsMan::updateHostInfo(kvstore_, host, info,
+                                                 req.get_leader_partIds());
+        } else {
+            ret = ActiveHostsMan::updateHostInfo(kvstore_, host, info);
```
dangleptr:
```
ret = ActiveHostsMan::updateHostInfo(
    kvstore_, 
    host, 
    info,  
    req.__isset.leader_partIds ? req.get_leader_partIds() : nullptr);
```

PR #1553, opened 2019-12-30
Fixed use of precompiled header

dutor:
Currently, PCH does not take effect when building in the out-of-source way, since the _.gch_ file, by design, must be at the same directory with the original header. This PR resolves this problem by generating the _.gch_ file in the source tree, i.e. `${CMAKE_CURRENT_SOURCE_DIR}`.

At the mean time, the former implementation of _FindPCHSupport_ does not handle the compile options properly, so that the options used for PCH generation and source compilation are different, which is erroneous. This PR also addresses this issue.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -203,7 +203,6 @@ add_compile_options(-Wall)
 add_compile_options(-Werror)
 add_compile_options(-Wunused-parameter)
 add_compile_options(-Wshadow)
-add_compile_options(-fpch-preprocess)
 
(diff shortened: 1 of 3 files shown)


dutor:
Jenkins, go!

dutor:
Jenkins, go!

dutor:
Jenkins, go!

dutor:
Jenkins, go!

dutor:
jenkins, go!

dutor:
jenkins go

dutor:
Jenkins go

dutor:
Jenkins, go!

PR #1559, opened 2019-12-31
add metrics gapped by 5s 

knightXun:
I add some metrics gapped by 5s, So prometheus can collect metrics in every 5s.


src/common/stats/StatsManager.cpp
```diff
@@ -86,7 +87,7 @@ int32_t StatsManager::registerHisto(folly::StringPiece counterName,
                 bucketSize,
                 min,
                 max,
-                StatsType(60, {seconds(60), seconds(600), seconds(3600)}))));
```
laura-ding: Why change `numBuckets` to `720`, `numBuckets` is each level

knightXun: i thought `numBuckets` means time gap, `60` means statsmanager scrape metrics every minute. If we want to scape metrics every 5s, It needs 60 * 12 buckets.

knightXun:
ping @laura-ding

laura-ding:
jenkins go

knightXun:
jenkins go

knightXun:
ping @dutor

Shylock-Hg:
The duration record the metrics don't matter to prometheus scrape cycle.The folly statistics already handle time, so the metrics expose to pometheus or other metrics collector don't need to concern `time`.

knightXun:
@Shylock-Hg folly store last minute metrics. For example, Promethues will get some result in 1s and 59s. The time gap is too large. It is better shrink one minute to 5 seconds. Prometheus can scrape metrics every 5 seconds.

PR #1564, opened 2019-12-31
Added the Nebula test tool.

monadbobo:
Added the Nebula test tool.

### DIFF
--- a/tests/common/nebula_test_suite.py
+++ b/tests/common/nebula_test_suite.py
@@ -130,6 +130,13 @@ class NebulaTestSuite(object):
                 if col.get_timestamp() != expect:
                     return False, msg
             return True, ''
+
+        if col.getType() == ttypes.ColumnValue.PATH:
+            msg = self.check_format_str.format(col.get_path(), expect)
+            if col.get_path() != expect:
+                return False, msg
+            return True, ''
+
         return False, 'ERROR: Type unsupported'
 
     @classmethod
--- a/tests/common/nebula_test_suite.py
+++ b/tests/common/nebula_test_suite.py
@@ -4,13 +4,15 @@
 #
 # This source code is licensed under Apache 2.0 License,
 # attached with Common Clause Condition 1.0, found in the LICENSES directory.
-import pytest
-import sys
 import math
-from graph import ttypes
-from nebula.ConnectionPool import ConnectionPool
-from nebula.Client import GraphClient, AuthException, ExecutionException
+import sys
 from typing import Pattern
+
+import pytest
+from graph import ttypes
+from nebula.Client import AuthException, ExecutionException, GraphClient
+from nebula.ConnectionPool import ConnectionPool
+
 
 class NebulaTestSuite(object):
     @classmethod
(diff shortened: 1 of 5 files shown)


tests/query/test_simple_query.py
```diff
@@ -0,0 +1,79 @@
...
+
+
+class TestNebula(NebulaTestSuite):
+    @classmethod
+    def prepare(self):
+        resp = self.client.execute('CREATE SPACE IF NOT EXISTS space1')
+        self.check_resp_successed(resp)
+
+    def test_schema(self):
+        resp = self.client.execute('USE space1')
```
yixinglu: I think we could provide a `execute` function to avoid calling `self.client` each time.

tests/common/nebula_test_suite.py
```diff
@@ -0,0 +1,159 @@
...
+
+        if col.getType() == ttypes.ColumnValue.TIMESTAMP:
+            msg = self.check_format_str.format(col.get_timestamp(), expect)
+            if isinstance(expect, Pattern):
+                if not expect.match(str(col.get_timestamp())):
+                    return False, msg
+            else:
+                if col.get_timestamp() != expect:
+                    return False, msg
+            return True, ''
```
laura-ding: add `Path` type

tests/common/nebula_test_suite.py
```diff
@@ -0,0 +1,159 @@
...
+                if ok:
+                    break
+            assert ok, msg
+            expect.remove(exp)
+
+    @classmethod
+    def check_result(self, rows, expect):
+        msg = 'len(rows)[%d] != len(expect)[%d]' % (len(rows), len(expect))
+        assert len(rows) == len(expect), msg
+        for row, i in zip(rows, range(0, len(expect))):
```
laura-ding: need to support  `result is the same, but the order of the rows is not the same`

monadbobo: we using "search_result" check this condition.

laura-ding: good

tests/query/test_simple_query.py
```diff
@@ -0,0 +1,110 @@
...
+
+    def test_path(self):
+        time.sleep(3)
+        resp = self.execute('USE space1')
+        assert resp.error_code == 0
+
+        resp = self.execute_query('FIND SHORTEST PATH FROM 1 to 3 OVER *')
+        assert resp.error_code == 0
+
+        path = ttypes.Path()
```
laura-ding: I  think check `PATH` like this,  is hard to write expected.

tests/common/nebula_test_suite.py
```diff
@@ -0,0 +1,186 @@
...
+            for col, j in zip(row.columns, range(0, len(expect[i]))):
+                ok, msg = self.check_value(col, expect[i][j])
+                assert ok, msg
+
+    @classmethod
+    def check_path_result(self, rows, expect):
+        for row, i in zip(rows, range(0, len(expect))):
```
laura-ding: check the length of `row` and `expect`

PR #1567, opened 2020-01-01
Added support to disable -Werror

dutor:
This is for end users those who want to build from source, allowing them to disable `-Werror` without tweaking the code.


dutor:
> Em... Do you want to be compatible for different gcc versions?

What do you mean by _compatible_? I can't see there is anything about compiler itself.

dangleptr:
> > Em... Do you want to be compatible for different gcc versions?
> 
> What do you mean by _compatible_? I can't see there is anything about compiler itself.

Is there any warnings now?  Only for some different compilers,  some warnings maybe dumped.

dutor:
> Em... Do you want to be compatible for different gcc versions?

I think I got you now.

There are chances that we cannot cover all situations for environments of users. For example, users may use the latest compiler which can trigger more warnings, and we might not catch up immediately.

dangleptr:
> > Em... Do you want to be compatible for different gcc versions?
> 
> I think I got you now.
> 
> There are chances that we cannot cover all situations for environments of users. For example, users may use the latest compiler which can trigger more warnings, and we might not catch up immediately.

Nice.

PR #1568, opened 2020-01-01
Link against libgcc statically

dutor:
Exception handling related code reside in libgcc(libgcc_eh, to be specific). There are chances to crash during exception handling if different libgcc used at compile and run time. The practice is to link libgcc and libstdc++ in the same way.


dutor:
Jenkins go

dutor:
jenkins go

dutor:
jenkins go

dutor:
jenkins go

PR #1569, opened 2020-01-02
Ignore some build target

darionyaphet:


.gitignore
```diff
@@ -28,6 +28,8 @@ Testing/
 target/
 cluster.id
 pids/
+bin/
```
dutor: We strongly discourage to build the project directly in the source tree.

darionyaphet: Sounds good. we should make nebula under `build` directory. closed

PR #1572, opened 2020-01-02
Support assign set sentence to a variable.

CPWstatic:


CPWstatic:
Jenkins go!

src/parser/parser.yy
```diff
@@ -1169,6 +1169,9 @@ assignment_sentence
     : VARIABLE ASSIGN piped_sentence {
         $$ = new AssignmentSentence($1, $3);
     }
+    | VARIABLE ASSIGN set_sentence {
```
dangleptr: piped_sentence is not one kind of set_sentence?

CPWstatic: No, it isn't.

dangleptr: IMO we should do that.  @dutor WDYT?

CPWstatic: i'm ok with your opinion.

PR #1576, opened 2020-01-02
Update meta client interface and add skipConfig flag

dangleptr:
There are too many arguments for meta client now.  Some arguments are not required for each caller. 
Now we compact them into one structure "MetaClientOptions".  Meanwhile,  we add a new option to skip all config manager logic in it.  It is useful for many tools. 

By the way,  there are some problems for current heart beat logic after #1236  merged in. This pr has fixed it.


src/meta/client/MetaClient.cpp
```diff
@@ -79,9 +73,11 @@ bool MetaClient::isMetadReady() {
 }
 
 bool MetaClient::waitForMetadReady(int count, int retryIntervalSecs) {
-    std::string gflagsJsonPath;
-    GflagsManager::getGflagsModule(gflagsModule_);
-    gflagsDeclared_ = GflagsManager::declareGflags(gflagsModule_);
+    if (!options_.skipConfig_) {
```
critical27: 👍

src/storage/StorageServer.cpp
```diff
@@ -89,11 +90,14 @@ bool StorageServer::start() {
     workers_->start();
 
     // Meta client
+    meta::MetaClientOptions options;
+    options.localHost_ = localHost_;
+    options.inStoraged_ = true;
+    options.serviceName_ = "";
```
darionyaphet: why the service name is empty ?

dangleptr: Ha,  it is compatible with current logic.  I did not handled it in this pr.

src/tools/storage-perf/StoragePerfTool.cpp
```diff
@@ -74,7 +74,10 @@ class Perf {
             threads.emplace_back(std::move(t));
         }
         threadPool_ = std::make_shared<folly::IOThreadPoolExecutor>(FLAGS_io_threads);
-        mClient_ = std::make_unique<meta::MetaClient>(threadPool_, metaAddrsRet.value());
+
+        meta::MetaClientOptions options;
+        options.skipConfig_ = true;
+        mClient_ = std::make_unique<meta::MetaClient>(threadPool_, metaAddrsRet.value(), options);
```
CPWstatic: Would you please update the code in tools/db-dump/DbDumper.cpp ? It also initiates a meta client.

PR #1579, opened 2020-01-02
Fixed linking error with gold

dutor:
When using _gold_ as the linker, building fails with undefined referenced symbols. As for the the reason underhood, I will dig it later.


dutor:
I thinks the error is due to the linker option `--gc-sections`, which is enabled with `ld.bfd` while disabled with `ld.gold`.

To be specific, with `--gc-sections`, some unused functions are eliminated during the link stage(this is one of several ways I reduce the build size), as well as their dependencies. At the meantime, `--gc-sections` is disabled when `gold` linker is used, since it's buggy when they are used together(something tedious on use of systemtap by folly, forget it).

dangleptr:
Awesome

PR #1583, opened 2020-01-03
Invalid leader when LEADER_CHANGED with leader = (0, 0)

critical27:
We should randomly pick a host to send request if LEADER_CHANGED is returned with leader = (0, 0), both in MetaClient and StorageClient.


zhangguoqing:
LGTM!

PR #1584, opened 2020-01-03
support TTL in storage

panda-sheep:
support TTL in storage

### DIFF
--- a/src/common/base/Cord.cpp
+++ b/src/common/base/Cord.cpp
@@ -112,9 +112,8 @@ size_t Cord::appendTo(std::string& str) const {
     }
 
     // Last block
-    if (tail_) {
-        str.append(tail_, blockPt_);
-    }
+    str.append(tail_, blockPt_);
+
     return len_;
 }
 
--- a/src/common/base/Cord.cpp
+++ b/src/common/base/Cord.cpp
@@ -112,8 +112,9 @@ size_t Cord::appendTo(std::string& str) const {
     }
 
     // Last block
-    str.append(tail_, blockPt_);
-
+    if (tail_) {
+        str.append(tail_, blockPt_);
+    }
     return len_;
 }
 
--- a/src/dataman/ResultSchemaProvider.cpp
+++ b/src/dataman/ResultSchemaProvider.cpp
@@ -56,8 +56,7 @@ std::string ResultSchemaProvider::ResultSchemaField::getDefaultValue() const {
  *
  **********************************/
 ResultSchemaProvider::ResultSchemaProvider(Schema schema)
-        : columns_(std::move(schema.get_columns())),
-          schemaProp_(std::move(schema.get_schema_prop())) {
+        : columns_(std::move(schema.get_columns())) {
     for (int64_t i = 0; i < static_cast<int64_t>(columns_.size()); i++) {
         const std::string& name = columns_[i].get_name();
         nameIndex_.emplace(std::make_pair(SpookyHashV2::Hash64(name.data(), name.size(), 0), i));
@@ -128,10 +127,6 @@ nebula::cpp2::Schema ResultSchemaProvider::toSchema() const {
     nebula::cpp2::Schema schema;
     schema.set_columns(columns_);
     return schema;
-}
-
-const cpp2::SchemaProp ResultSchemaProvider::getProp() const {
-    return schemaProp_;
 }
 
 }  // namespace nebula
--- a/src/dataman/ResultSchemaProvider.cpp
+++ b/src/dataman/ResultSchemaProvider.cpp
@@ -56,7 +56,8 @@ std::string ResultSchemaProvider::ResultSchemaField::getDefaultValue() const {
  *
  **********************************/
 ResultSchemaProvider::ResultSchemaProvider(Schema schema)
-        : columns_(std::move(schema.get_columns())) {
+        : columns_(std::move(schema.get_columns())),
+          schemaProp_(std::move(schema.get_schema_prop())) {
     for (int64_t i = 0; i < static_cast<int64_t>(columns_.size()); i++) {
         const std::string& name = columns_[i].get_name();
         nameIndex_.emplace(std::make_pair(SpookyHashV2::Hash64(name.data(), name.size(), 0), i));
@@ -127,6 +128,10 @@ nebula::cpp2::Schema ResultSchemaProvider::toSchema() const {
     nebula::cpp2::Schema schema;
     schema.set_columns(columns_);
     return schema;
+}
+
+const cpp2::SchemaProp ResultSchemaProvider::getProp() const {
+    return schemaProp_;
 }
 
 }  // namespace nebula
--- a/src/meta/SchemaProviderIf.h
+++ b/src/meta/SchemaProviderIf.h
@@ -103,6 +103,8 @@ public:
 
     virtual nebula::cpp2::Schema toSchema() const = 0;
 
+    virtual const nebula::cpp2::SchemaProp getProp() const = 0;
+
     /******************************************
      *
      * Iterator implementation
--- a/src/meta/SchemaProviderIf.h
+++ b/src/meta/SchemaProviderIf.h
@@ -103,8 +103,6 @@ public:
 
     virtual nebula::cpp2::Schema toSchema() const = 0;
 
-    virtual const nebula::cpp2::SchemaProp getProp() const = 0;
-
     /******************************************
      *
      * Iterator implementation
--- a/src/storage/CommonUtils.cpp
+++ b/src/storage/CommonUtils.cpp
@@ -1,0 +1,53 @@
+/* Copyright (c) 2020 vesoft inc. All rights reserved.
+ *
+ * This source code is licensed under Apache 2.0 License,
+ * attached with Common Clause Condition 1.0, found in the LICENSES directory.
+ */
+
+#include "base/Base.h"
+#include "storage/CommonUtils.h"
+
+namespace nebula {
+namespace storage {
+
+bool checkDataExpiredForTTL(const meta::SchemaProviderIf* schema,
+                            RowReader* reader,
+                            const std::string& ttlCol,
+                            int64_t ttlDuration) {
+    auto now = time::WallClock::fastNowInSec();
+    const auto& ftype = schema->getFieldType(ttlCol);
+
+    int64_t v = 0;
+    switch (ftype.type) {
+        case nebula::cpp2::SupportedType::TIMESTAMP:
+        case nebula::cpp2::SupportedType::INT: {
+            auto ret = reader->getInt<int64_t>(ttlCol, v);
+            if (ret != ResultType::SUCCEEDED) {
+                // Reading wrong data should not be deleted
+                return false;
+            }
+            break;
+        }
+        case nebula::cpp2::SupportedType::VID: {
+            auto ret = reader->getVid(ttlCol, v);
+            if (ret != ResultType::SUCCEEDED) {
+                // Reading wrong data should not be deleted
+                return false;
+            }
+            break;
+        }
+        default: {
+            VLOG(1) << "Unsupport TTL column type";
+            return false;
+        }
+    }
+
+    if (now > (v + ttlDuration)) {
+        return true;
+    }
+    return false;
+}
+
+
+}  // namespace storage
+}  // namespace nebula
--- a/src/storage/CompactionFilter.h
+++ b/src/storage/CompactionFilter.h
@@ -9,6 +9,7 @@
 
 #include "base/Base.h"
 #include "base/NebulaKeyUtils.h"
+#include "dataman/RowReader.h"
 #include "kvstore/CompactionFilter.h"
 
 DEFINE_bool(storage_kv_mode, false, "True for kv mode");
@@ -35,7 +36,7 @@ public:
             if (!schemaValid(spaceId, key)) {
                 return true;
             }
-            if (!ttlValid(spaceId, val)) {
+            if (!ttlValid(spaceId, key, val)) {
                 VLOG(3) << "TTL invalid for key " << key;
                 return true;
             }
@@ -76,8 +77,85 @@ public:
         return true;
     }
 
-    bool ttlValid(GraphSpaceID, const folly::StringPiece&) const {
-        // TODO(heng): Implement ttl filter after ttl schema pr merged in
+    bool ttlValid(GraphSpaceID spaceId, const folly::StringPiece& key,
+                  const folly::StringPiece& val) const {
+        if (NebulaKeyUtils::isVertex(key)) {
+            auto tagId = NebulaKeyUtils::getTagId(key);
+            auto schema = this->schemaMan_->getTagSchema(spaceId, tagId);
+            if (!schema) {
+                VLOG(3) << "Space " << spaceId << ", Tag " << tagId << " invalid";
+                return false;
+            }
+            auto reader = nebula::RowReader::getTagPropReader(schemaMan_, val, spaceId, tagId);
+            return checkDataTtlValid(schema.get(), reader.get());
+        } else if (NebulaKeyUtils::isEdge(key)) {
+            auto edgeType = NebulaKeyUtils::getEdgeType(key);
+            if (edgeType < 0) {
+                // TODO(panda) reverse edge supports ttl compaction
+                return true;
+            }
+            auto schema = this->schemaMan_->getEdgeSchema(spaceId, edgeType);
+            if (!schema) {
+                VLOG(3) << "Space " << spaceId << ", EdgeType " << edgeType << " invalid";
+                return false;
+            }
+            auto reader = nebula::RowReader::getEdgePropReader(schemaMan_, val, spaceId, edgeType);
+            return checkDataTtlValid(schema.get(), reader.get());
+        }
+        return true;
+    }
+
+    bool checkDataTtlValid(const meta::SchemaProviderIf* schema, nebula::RowReader* reader) const {
+        nebula::cpp2::SchemaProp schemaProp = schema->getProp();
+        int ttlDuration = 0;
+        if (schemaProp.get_ttl_duration()) {
+            ttlDuration = *schemaProp.get_ttl_duration();
+        }
+        std::string ttlCol;
+        if (schemaProp.get_ttl_col()) {
+            ttlCol = *schemaProp.get_ttl_col();
+        }
+
+        // Only support the specified ttl_col mode
+        // Not specifying or non-positive ttl_duration behaves like ttl_duration = infinity
+        if (ttlCol.empty() ||  ttlDuration <= 0) {
+            return true;
+        }
+
+        auto now = time::WallClock::slowNowInSec();
+        const auto& ftype = schema->getFieldType(ttlCol);
+
+        int64_t v = 0;
+        switch (ftype.type) {
+            case nebula::cpp2::SupportedType::TIMESTAMP:
+            case nebula::cpp2::SupportedType::INT: {
+                auto ret = reader->getInt<int64_t>(ttlCol, v);
+                if (ret != ResultType::SUCCEEDED) {
+                    VLOG(3) << "Data read error";
+                    // Reading wrong data should not be deleted
+                    return true;
+                }
+                break;
+            }
+            case nebula::cpp2::SupportedType::VID: {
+                auto ret = reader->getVid(ttlCol, v);
+                if (ret == ResultType::SUCCEEDED) {
+                    VLOG(3) << "Data read error";
+                    // Reading wrong data should not be deleted
+                    return true;
+                }
+                break;
+            }
+            default: {
+                // Reading wrong data should not be deleted
+                VLOG(1) << "Unsupport TTL column type";
+                return true;
+            }
+        }
+
+        if (now > (v + ttlDuration)) {
+            return false;
+        }
         return true;
     }
 
--- a/src/storage/CompactionFilter.h
+++ b/src/storage/CompactionFilter.h
@@ -12,6 +12,7 @@
 #include "dataman/RowReader.h"
 #include "meta/NebulaSchemaProvider.h"
 #include "kvstore/CompactionFilter.h"
+#include "storage/CommonUtils.h"
 
 DEFINE_bool(storage_kv_mode, false, "True for kv mode");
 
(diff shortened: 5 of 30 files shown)


panda-sheep:
Jenkins go!

panda-sheep:
> We should also provide a mechanism to delete the expired dat

Yeah, when compaction ，delete the expired dat.  Not implement int this pr. 
This pr  can only  read  in ttl.

dangleptr:
> > We should also provide a mechanism to delete the expired dat
> 
> Yeah, when compaction ，delete the expired dat. Not implement int this pr.
> This pr can only read in ttl.

We'd better do it in this pr.  Without it,  it is not complete.

src/dataman/ResultSchemaProvider.cpp
```diff
@@ -129,5 +130,9 @@ nebula::cpp2::Schema ResultSchemaProvider::toSchema() const {
     return schema;
 }
 
+const cpp2::SchemaProp ResultSchemaProvider::getProp() const {
```
dangleptr: For ResultSchemaProvider,  we don't need schemaProp_

src/meta/SchemaProviderIf.h
```diff
@@ -103,6 +103,8 @@ class SchemaProviderIf {
 
     virtual nebula::cpp2::Schema toSchema() const = 0;
 
+    virtual const nebula::cpp2::SchemaProp getProp() const = 0;
```
dangleptr: We don't need the function here.

src/storage/query/QueryBaseProcessor.inl
```diff
@@ -170,6 +171,53 @@ cpp2::ErrorCode QueryBaseProcessor<REQ, RESP>::checkAndBuildContexts(const REQ&
     return cpp2::ErrorCode::SUCCEEDED;
 }
 
+template<typename REQ, typename RESP>
+bool QueryBaseProcessor<REQ, RESP>::checkDataExpiredForTTL(RowReader* reader) {
+    auto schema = reader->getSchema();
```
dangleptr: It is not correct.  You should use schemaMan_ directly.

src/storage/query/QueryBaseProcessor.inl
```diff
@@ -170,6 +171,53 @@ cpp2::ErrorCode QueryBaseProcessor<REQ, RESP>::checkAndBuildContexts(const REQ&
...
+
+    auto now = time::WallClock::slowNowInSec();
+    const auto& ftype = schema->getFieldType(ttlCol);
+
+    int64_t v = 0;
+    switch (ftype.type) {
+        case nebula::cpp2::SupportedType::TIMESTAMP:
+        case nebula::cpp2::SupportedType::INT: {
+            auto ret = reader->getInt<int64_t>(ttlCol, v);
+            CHECK(ret == ResultType::SUCCEEDED);
```
dangleptr: Don't use CHECK in our code.

src/dataman/ResultSchemaProvider.cpp
panda-sheep: Yeah.
Logically， don't need schemaProp_ in ResultSchemaProvider
Only add  schemaProp_  to NebulaSchemaProvider.
  
Here add  schemaProp_  in  ResultSchemaProvider  only  for unit test.
For example，  `genEdgeSchemaProvider `and `genTagSchemaProvider `in TestUtils.h in storage   use ResultSchemaProvider to  create schema.

src/meta/SchemaProviderIf.h
panda-sheep: ditto

src/storage/query/QueryBaseProcessor.inl
panda-sheep: Good Catch！

src/storage/query/QueryBaseProcessor.inl
panda-sheep: good catch, fixed

panda-sheep:
> > > We should also provide a mechanism to delete the expired dat
> > 
> > 
> > Yeah, when compaction ，delete the expired dat. Not implement int this pr.
> > This pr can only read in ttl.
> 
> We'd better do it in this pr. Without it, it is not complete.

ok，has already done

src/dataman/ResultSchemaProvider.cpp
panda-sheep: According to offline communication,   use  NebulaSchemaProvider  to replace ResultSchemaProvider in storage UT.

src/common/base/Cord.cpp
```diff
@@ -112,8 +112,9 @@ size_t Cord::appendTo(std::string& str) const {
     }
 
     // Last block
-    str.append(tail_, blockPt_);
-
+    if (tail_) {
```
dangleptr: the tail_ maybe nullptr?

src/storage/query/QueryBaseProcessor.inl
```diff
@@ -170,6 +172,65 @@ cpp2::ErrorCode QueryBaseProcessor<REQ, RESP>::checkAndBuildContexts(const REQ&
...
+    std::string ttlCol;
+    if (schemaProp.get_ttl_col()) {
+        ttlCol = *schemaProp.get_ttl_col();
+    }
+
+    // Only support the specified ttl_col mode
+    // Not specifying or non-positive ttl_duration behaves like ttl_duration = infinity
+    if (ttlCol.empty() ||  ttlDuration <= 0) {
+        return false;
+    }
```
dangleptr: Maybe we could extract the ttl col and ttlDuration in checkAndBuildContext.

src/storage/query/QueryBaseProcessor.inl
```diff
@@ -522,9 +597,16 @@ kvstore::ResultCode QueryBaseProcessor<REQ, RESP>::collectEdgeProps(
                     continue;
                 }
             }
+
+            // Check if the schema has TTL
+            if (!checkDataExpiredForTTL(schema.get(), reader.get())) {
```
dangleptr: I suggest we check ttl before checking the expression logic.

src/storage/query/QueryBaseProcessor.inl
```diff
@@ -522,9 +597,16 @@ kvstore::ResultCode QueryBaseProcessor<REQ, RESP>::collectEdgeProps(
...
+                proc(reader.get(), key, props);
```
dangleptr: If the ttl exprired,  we could "continue" the loop.

src/storage/CompactionFilter.h
```diff
@@ -76,8 +78,90 @@ class StorageCompactionFilter final : public kvstore::KVFilter {
...
+            if (!schema) {
+                VLOG(3) << "Space " << spaceId << ", Tag " << tagId << " invalid";
+                return false;
+            }
+            auto reader = nebula::RowReader::getTagPropReader(schemaMan_, val, spaceId, tagId);
+            return checkDataTtlValid(schema.get(), reader.get());
+        } else if (NebulaKeyUtils::isEdge(key)) {
+            auto edgeType = NebulaKeyUtils::getEdgeType(key);
+            if (edgeType < 0) {
+                // TODO(panda) reverse edge supports ttl compaction
```
dangleptr: After #1717 merged in,  we could support it now.

src/storage/query/QueryBaseProcessor.inl
panda-sheep: Good catch!

src/storage/query/QueryBaseProcessor.inl
panda-sheep: Good!

src/common/base/Cord.cpp
dangleptr: I am confused in which case the tail_ will be nullptr.

src/storage/CompactionFilter.h
```diff
@@ -76,8 +78,87 @@ class StorageCompactionFilter final : public kvstore::KVFilter {
...
+            ttlCol = *schemaProp.get_ttl_col();
+        }
+
+        // Only support the specified ttl_col mode
+        // Not specifying or non-positive ttl_duration behaves like ttl_duration = infinity
+        if (ttlCol.empty() ||  ttlDuration <= 0) {
+            return true;
+        }
+
+        auto now = time::WallClock::slowNowInSec();
```
dangleptr: fastNowInSec

src/storage/CompactionFilter.h
```diff
@@ -76,8 +78,87 @@ class StorageCompactionFilter final : public kvstore::KVFilter {
...
+        const auto& ftype = nschema->getFieldType(ttlCol);
```
dangleptr: Add TODO comments,  we should optimize the method in the future.

src/storage/query/QueryBaseProcessor.inl
```diff
@@ -98,6 +100,8 @@ cpp2::ErrorCode QueryBaseProcessor<REQ, RESP>::checkAndBuildContexts(const REQ&
                 } else {
                     tagContexts_[it->second].props_.emplace_back(std::move(prop));
                 }
+
+                buildTagOrEdgeTTLInfo(schema.get(), true, tagId, 0);
```
dangleptr: Maybe should put the method after the for loop.  We don't need to do it for every props.

src/storage/CompactionFilter.h
panda-sheep: ok

src/storage/query/QueryBaseProcessor.inl
panda-sheep: merge  `buildTagOrEdgeTTLInfo `and  build `RespSchema `  to  `buildTTLInfoAndRespSchema`

src/storage/CompactionFilter.h
panda-sheep: fix

src/common/base/Cord.cpp
panda-sheep: Haha, I finally found out why.
When the pr created,  rows 101 and 102  in this file did not exist
Reference # 950
now this case not exitsbin. 

👍👍👍

src/storage/query/QueryBaseProcessor.inl
```diff
@@ -163,10 +165,85 @@ cpp2::ErrorCode QueryBaseProcessor<REQ, RESP>::checkAndBuildContexts(const REQ&
...
+    if (edgeFound != edgeTTLInfo_.end()) {
+        return edgeFound->second;
+    }
+    return std::make_pair(std::string(""), 0);
+}
+
+template<typename REQ, typename RESP>
+bool QueryBaseProcessor<REQ, RESP>::checkDataExpiredForTTL(const meta::SchemaProviderIf* schema,
+                                                           RowReader* reader,
+                                                           std::string ttlCol,
```
dangleptr: const std::string&

src/storage/query/QueryBaseProcessor.inl
```diff
@@ -163,10 +165,85 @@ cpp2::ErrorCode QueryBaseProcessor<REQ, RESP>::checkAndBuildContexts(const REQ&
...
+std::pair<std::string, int64_t>  QueryBaseProcessor<REQ, RESP>::getEdgeTTLInfo(EdgeType edgeType) {
+    auto edgeFound = edgeTTLInfo_.find(edgeType);
+    if (edgeFound != edgeTTLInfo_.end()) {
+        return edgeFound->second;
+    }
+    return std::make_pair(std::string(""), 0);
+}
+
+template<typename REQ, typename RESP>
+bool QueryBaseProcessor<REQ, RESP>::checkDataExpiredForTTL(const meta::SchemaProviderIf* schema,
```
dangleptr: Could you unify the methods with checkDataTtlValid inside compacionFilter

src/storage/query/QueryBaseProcessor.inl
```diff
@@ -456,6 +564,19 @@ kvstore::ResultCode QueryBaseProcessor<REQ, RESP>::collectEdgeProps(
                                                   val,
                                                   spaceId_,
                                                   std::abs(edgeType));
+
+            // Check if the schema has TTL
+            if (edgeHasTTL(edgeType)) {
+                std::pair<std::string, int64_t> edgeTTL = getEdgeTTLInfo(edgeType);
```
dangleptr: Move getEdgeTTLInfo outside of the for loop.

src/storage/query/QueryBaseProcessor.inl
```diff
@@ -517,9 +638,13 @@ kvstore::ResultCode QueryBaseProcessor<REQ, RESP>::collectEdgeProps(
                     continue;
                 }
             }
+
+            proc(reader.get(), key, props);
```
dangleptr: what's the purpose about the changes?

src/storage/query/QueryBaseProcessor.inl
```diff
@@ -163,10 +165,85 @@ cpp2::ErrorCode QueryBaseProcessor<REQ, RESP>::checkAndBuildContexts(const REQ&
...
+bool QueryBaseProcessor<REQ, RESP>::edgeHasTTL(EdgeType edgeType) {
+    auto edgeFound = edgeTTLInfo_.find(edgeType);
+    if (edgeFound != edgeTTLInfo_.end()) {
+        return true;
+    }
+    return false;
+}
+
+template<typename REQ, typename RESP>
+std::pair<std::string, int64_t>  QueryBaseProcessor<REQ, RESP>::getTagTTLInfo(TagID tagId) {
```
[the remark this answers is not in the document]
panda-sheep: Worshipping, I got new skills. 👍👍👍

src/storage/query/QueryBaseProcessor.inl
```diff
@@ -163,10 +165,85 @@ cpp2::ErrorCode QueryBaseProcessor<REQ, RESP>::checkAndBuildContexts(const REQ&
...
+    auto tagFound = tagTTLInfo_.find(tagId);
+    if (tagFound != tagTTLInfo_.end()) {
+        return tagFound->second;
+    }
+    return std::make_pair(std::string(""), 0);
+}
+
+template<typename REQ, typename RESP>
+std::pair<std::string, int64_t>  QueryBaseProcessor<REQ, RESP>::getEdgeTTLInfo(EdgeType edgeType) {
```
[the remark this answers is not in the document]
panda-sheep: 👍👍👍

src/storage/query/QueryBaseProcessor.inl
panda-sheep: fix

src/storage/query/QueryBaseProcessor.inl
panda-sheep: fix

src/storage/query/QueryBaseProcessor.inl
panda-sheep: Here is a check of the value of each row

src/storage/query/QueryBaseProcessor.inl
panda-sheep: when data expired， Do not do the following
```
proc(reader.get(), key, props);
 ++cnt;
```

src/storage/CompactionFilter.h
```diff
@@ -76,11 +79,57 @@ class StorageCompactionFilter final : public kvstore::KVFilter {
...
+            auto reader = nebula::RowReader::getEdgePropReader(schemaMan_, val,
+                                                               spaceId, std::abs(edgeType));
+            return checkDataTtlValid(schema.get(), reader.get());
+        }
         return true;
     }
 
+    // TODO(panda) Optimize the method in the future
+    bool checkDataTtlValid(const meta::SchemaProviderIf* schema, nebula::RowReader* reader) const {
+        const meta::NebulaSchemaProvider* nschema =
```
bright-starry-sky: `const auto*` is better?

src/storage/CompactionFilter.h
```diff
@@ -76,11 +79,57 @@ class StorageCompactionFilter final : public kvstore::KVFilter {
...
+            dynamic_cast<const meta::NebulaSchemaProvider*>(schema);
+        if (nschema == NULL) {
+            return true;
+        }
+        const nebula::cpp2::SchemaProp schemaProp = nschema->getProp();
```
bright-starry-sky: const auto is better?

src/storage/CompactionFilter.h
```diff
@@ -76,11 +79,57 @@ class StorageCompactionFilter final : public kvstore::KVFilter {
...
         return true;
     }
 
+    // TODO(panda) Optimize the method in the future
+    bool checkDataTtlValid(const meta::SchemaProviderIf* schema, nebula::RowReader* reader) const {
+        const meta::NebulaSchemaProvider* nschema =
+            dynamic_cast<const meta::NebulaSchemaProvider*>(schema);
+        if (nschema == NULL) {
+            return true;
+        }
```
bright-starry-sky: A question，Is the redundant check ？

src/storage/CommonUtils.cpp
```diff
@@ -0,0 +1,53 @@
...
+                // Reading wrong data should not be deleted
+                return false;
+            }
+            break;
+        }
+        default: {
+            VLOG(1) << "Unsupport TTL column type";
+            return false;
+        }
+    }
```
bright-starry-sky: Why do need to handling VID?

panda-sheep: When the column type is VID, it can be used as the ttl_col column.

src/storage/CompactionFilter.h
panda-sheep: In some test case, it may be const meta::ResultSchemaProvider*

src/storage/CompactionFilter.h
panda-sheep: Yeah，good catch

src/storage/CompactionFilter.h
panda-sheep: Yeah,  `const auto*` is better

src/storage/query/QueryBaseProcessor.inl
dangleptr: As your logic,  we don't need the changes.

src/storage/query/QueryBaseProcessor.inl
```diff
@@ -456,6 +513,16 @@ kvstore::ResultCode QueryBaseProcessor<REQ, RESP>::collectEdgeProps(
                                                   val,
                                                   spaceId_,
                                                   std::abs(edgeType));
+            // Check if the schema has TTL
+            auto retTTL = getEdgeTTLInfo(edgeType);
```
dangleptr: Move it outside of the loop

src/storage/query/QueryBaseProcessor.inl
```diff
@@ -382,14 +406,33 @@ kvstore::ResultCode QueryBaseProcessor<REQ, RESP>::collectVertexProps(
...
         auto result = vertexCache_->get(std::make_pair(vId, tagId), partId);
         if (result.ok()) {
             auto v = std::move(result).value();
             auto reader = RowReader::getTagPropReader(this->schemaMan_, v, spaceId_, tagId);
-            this->collectProps(reader.get(), "", props, fcontext, collector);
-            VLOG(3) << "Hit cache for vId " << vId << ", tagId " << tagId;
-            return kvstore::ResultCode::SUCCEEDED;
+
+            // Check if the schema has TTL
+             auto retTtlOpt = getTagTTLInfo(tagId);
```
dangleptr:
```
        auto retTtlOpt = getTagTTLInfo(tagId);
        if (retTtlOpt.hasValue()) {
            auto ttlValue = retTtlOpt.value();
            if (checkDataExpiredForTTL(schema.get(),
                                       reader.get(),
                                       ttlValue.first,
                                       ttlValue.second)) {
                return ret;
            }
        }
```

src/common/base/Cord.cpp
panda-sheep: fix

src/storage/query/QueryBaseProcessor.inl
panda-sheep: I originally wanted to do this, but then the log information would be incomplete.
But this code is more concise, let me do it first

src/storage/query/QueryBaseProcessor.inl
panda-sheep: The comment is incomplete here， I will  fix
Here check if each row of data is expired,  so get the value of each row of data to check。
So it does not move it outside of the loop

src/storage/query/QueryBaseProcessor.inl
panda-sheep: If  no change， for example only one row data has ttl，and ttl data expired，
shouldn't do the following, but do
```
proc(reader.get(), key, props);
 ++cnt;
```

src/storage/query/QueryBaseProcessor.inl
dangleptr: why call "getEdgeTTLInfo" for every edge?  The same edgeType has the same TTL Info.

src/storage/query/QueryBaseProcessor.inl
dangleptr: As your logic, if ttl is expired,  it will continue the loop,  right?

src/storage/query/QueryBaseProcessor.inl
panda-sheep: You are right.
My mistake.

👍👍👍,

panda-sheep: I mistakenly thought  move "checkDataExpiredForTTL" . My mistake.

panda-sheep:
Have fix comments，please continue.

src/storage/query/QueryBaseProcessor.inl
dangleptr: So the original logic is right. we dont need the changes.

src/storage/query/QueryEdgePropsProcessor.cpp
```diff
@@ -27,16 +27,32 @@ kvstore::ResultCode QueryEdgePropsProcessor::collectEdgesProps(
...
         auto reader = RowReader::getEdgePropReader(schemaMan_,
                                                    iter->val(),
                                                    spaceId_,
-                                                   edgeKey.edge_type);
-        this->collectProps(reader.get(), iter->key(), props, nullptr, &collector);
-        rsWriter.addRow(writer);
+                                                   std::abs(edgeKey.edge_type));
+
+        // Check if ttl data expired
+        auto retTTLOpt = getEdgeTTLInfo(edgeKey.edge_type);
```
dangleptr: We don't need to do it for every EdgeKey

panda-sheep: Yeah.  Here is in `if`  case , which is actually only executed once.
I will do it first.

src/storage/query/QueryEdgePropsProcessor.cpp
```diff
@@ -27,17 +27,30 @@ kvstore::ResultCode QueryEdgePropsProcessor::collectEdgesProps(
...
+                                                   std::abs(edgeKey.edge_type));
+
+        // Check if ttl data expired
+        if (retTTLOpt.has_value()) {
+            auto ttlValue = retTTLOpt.value();
+            if (checkDataExpiredForTTL(schema.get(),
+                                       reader.get(),
+                                       ttlValue.first,
+                                       ttlValue.second)) {
+                iter->next();
```
dangleptr: why next when expired.

panda-sheep: Because, line 54  `iter->next();`   I don’t think it ’s useful. I just wrote it for consistency.

panda-sheep: I removed them all

src/storage/CompactionFilter.h
```diff
@@ -76,11 +79,56 @@ class StorageCompactionFilter final : public kvstore::KVFilter {
...
-    bool ttlValid(GraphSpaceID, const folly::StringPiece&) const {
-        // TODO(heng): Implement ttl filter after ttl schema pr merged in
+    bool ttlValid(GraphSpaceID spaceId, const folly::StringPiece& key,
+                  const folly::StringPiece& val) const {
+        if (NebulaKeyUtils::isVertex(key)) {
+            auto tagId = NebulaKeyUtils::getTagId(key);
+            auto schema = this->schemaMan_->getTagSchema(spaceId, tagId);
+            if (!schema) {
+                VLOG(3) << "Space " << spaceId << ", Tag " << tagId << " invalid";
+                return false;
```
bright-starry-sky: Why return false ? means this row was expired?

panda-sheep: When there is no schema, the data will not expire

bright-starry-sky: > When there is no schema, the data will not expire

Got , The purpose of TTL is to remove expired data. schema not found means the row is invalid data, maybe the schema has been removed.

panda-sheep: Yeah，compaction remove data when  `ttlValid `   function  return false.
In function `checkDataTtlValid` to check if ttl  is valid.

Anyway, ttl expired data and no valid schema data will remove

PR #1593, opened 2020-01-06
Fix three copies under different states

dangleptr:
Allow the three copies under different states.  Meanwhile,  avoid busy loop when sending snapshot if related follower is empty. 

For example,   the three copies with no WALs.
A  committedLogId 10,   term 10
B  committedLogId  9,    term 10
C  committedLogId  11000,   term  9

In this case,  A will be elected as the leader.  
But because C committedLogId is larger than A logIdSent,  so the C will return LOG_STALE, meanwhile A will return success and lastLogSent_ will be updated to be 11000.

Then the HB will never be sent to the C because the so large lastLogSent_ and C will fall in leader election infinity loop.  Very sad. 

To recover from the case,  we force the C to cleanup its data if it found the leader's term is larger but logId is smaller.  Then it will pull snapshot from leader.


src/kvstore/raftex/Host.cpp
```diff
@@ -277,6 +291,14 @@ void Host::appendLogsInternal(folly::EventBase* eb,
                         cpp2::AppendLogResponse r;
                         r.set_error_code(res);
                         self->setResponse(r);
+                    } else if (self->lastLogIdSent_ == resp.get_last_log_id()) {
```
critical27: 👍, this block and code block line 229, will prevent from sending logs that follower have received. But still confused about why the leader always trying to seek log 1?

src/kvstore/raftex/Host.cpp
```diff
@@ -190,19 +192,24 @@ void Host::appendLogsInternal(folly::EventBase* eb,
             {
                 std::lock_guard<std::mutex> g(self->lock_);
                 self->setResponse(r);
-                self->lastLogIdSent_ = self->logIdToSend_;
+                self->lastLogIdSent_ = self->logIdToSend_ - 1;
```
darionyaphet: Not quite understanding why `logIdToSend_ - 1`

dangleptr: It is used for avoid infinite loop when retrying for errors.  For more detail,  let's talk about it offline,  or you could ask for @critical27

PR #1601, opened 2020-01-07
suppress BalanceIntegrationTest timeout in clang with asan on

critical27:
Thanks for @dutor's help, it would time out because main process would lock in time, while forked sub process lock again, and then timeout.

### DIFF
--- a/src/common/time/detail/TscHelper.cpp
+++ b/src/common/time/detail/TscHelper.cpp
@@ -66,7 +66,7 @@ volatile std::atomic<double> ticksPerMSecFactor{0.0};
 volatile std::atomic<double> ticksPerUSecFactor{[] {
     launchTickTockThread();
     // re-launch tick-tock thread after forking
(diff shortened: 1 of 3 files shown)


critical27:
jenkins go

dutor:
Thanks for the big efforts on this.

But, I think we cannot resolve such an issue by simply remove the `pthread_atfork` call, without which our daemons will take the risk of losing time accuracy in front of time drifting, which is why the so-called _tick-tock_ thread exists.

For the sake of communication cost, would you mind handing this issue over to me?

critical27:
> Thanks for the big efforts on this.
> 
> But, I think we cannot resolve such an issue by simply remove the `pthread_atfork` call, without which our daemons will take the risk of losing time accuracy in front of time drifting, which is why the so-called _tick-tock_ thread exists.
> 
> For the sake of communication cost, would you mind handing this issue over to me?

Sure, it works for clang with asan now, orz.

PR #1603, opened 2020-01-08
Fixed deadlock with Clang/ASan

dutor:
This PR is mainly to address a deadlock issue illustrated in the screenshot below.
508F1508-16FF-4A20-A897-EB86990CC3DD

The root cause is very tedious to explain. Please contact me offline if interested.


dutor:
Disscussed with @dangleptr offline, the former fix is too hacky to understand. Update accordingly.

dutor:
Now, along with #1602 , Clang/ASan builds OK for now.

PR #1606, opened 2020-01-08
Return default value when prop is not find in old version tag/edge.

### LINKED ISSUE #1154: Storaged aborts when ALTER TAG to add props
image
image

### LINKED ISSUE #1185: [bug] FETCH PROP ON failed when tag schema changed
I add a new property to a tag only to find the old vertex property can't be fetched.
This is the old vertex before tag schema changed:
image
image
Then I changed the tag schema 
image
FETCH crashed
image
**QUESYION:** `FETCH` returns different results.

CPWstatic:
close #1185 , close #1154

For now, the storage would be crashed when we want a new property from the latest version tag/edge. The reason is that the data was stored with old tag/edge version, while the storage can't figure out what value type the property is.

### DIFF
--- a/src/storage/Collector.h
+++ b/src/storage/Collector.h
@@ -39,28 +39,65 @@ public:
                 : writer_(writer) {}
 
     void collectVid(int64_t v, const PropContext& prop) override {
-        UNUSED(prop);
+        switch (prop.type_.type) {
+            case nebula::cpp2::SupportedType::VID:
+                break;
+            default:
+                // Give a default value
+                // TODO: Should give null
+                v = 0;
+        }
         (*writer_) << RowWriter::ColType(nebula::cpp2::SupportedType::VID) << v;
     }
 
     void collectBool(bool v, const PropContext& prop) override {
-        UNUSED(prop);
+        switch (prop.type_.type) {
+            case nebula::cpp2::SupportedType::BOOL:
+                break;
+            default:
+                // Give a default value
+                // TODO: Should give null
+                v = false;
+        }
         (*writer_) << RowWriter::ColType(nebula::cpp2::SupportedType::BOOL) << v;
     }
 
     void collectInt64(int64_t v, const PropContext& prop) override {
-        UNUSED(prop);
+        switch (prop.type_.type) {
+            case nebula::cpp2::SupportedType::INT:
+            case nebula::cpp2::SupportedType::TIMESTAMP:
+                break;
+            default:
+                // Give a default value
+                // TODO: Should give null
+                v = 0;
+        }
         (*writer_) << RowWriter::ColType(nebula::cpp2::SupportedType::INT) << v;
     }
 
     void collectDouble(double v, const PropContext& prop) override {
-        UNUSED(prop);
+        switch (prop.type_.type) {
+            case nebula::cpp2::SupportedType::DOUBLE:
+            case nebula::cpp2::SupportedType::FLOAT:
+                break;
+            default:
+                // Give a default value
+                // TODO: Should give null
+                v = 0;
+        }
         (*writer_) << RowWriter::ColType(nebula::cpp2::SupportedType::DOUBLE) << v;
     }
 
     void collectString(const std::string& v, const PropContext& prop) override {
-        UNUSED(prop);
-        (*writer_) << RowWriter::ColType(nebula::cpp2::SupportedType::STRING) << v;
+        switch (prop.type_.type) {
+            case nebula::cpp2::SupportedType::STRING:
+                (*writer_) << RowWriter::ColType(nebula::cpp2::SupportedType::STRING) << v;
+                break;
+            default:
+                // Give a default value
+                // TODO: Should give null
+                (*writer_) << RowWriter::ColType(nebula::cpp2::SupportedType::STRING) << "";
+        }
     }
 
     template<typename V>
--- a/src/storage/Collector.h
+++ b/src/storage/Collector.h
@@ -39,65 +39,28 @@ public:
                 : writer_(writer) {}
 
     void collectVid(int64_t v, const PropContext& prop) override {
-        switch (prop.type_.type) {
-            case nebula::cpp2::SupportedType::VID:
-                break;
-            default:
-                // Give a default value
-                // TODO: Should give null
-                v = 0;
-        }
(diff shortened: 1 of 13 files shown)


dangleptr:
Where did it crashed?

CPWstatic:
> Where did it crashed?

LOG(FATAL)

CPWstatic:
Jenkins go!

src/storage/Collector.h
```diff
@@ -39,28 +39,65 @@ class PropsCollector : public Collector {
                 : writer_(writer) {}
 
     void collectVid(int64_t v, const PropContext& prop) override {
-        UNUSED(prop);
+        switch (prop.type_.type) {
```
dangleptr: Why not reuse RowWriter to do type converting?

src/storage/query/QueryBaseProcessor.inl
```diff
@@ -335,12 +335,23 @@ void QueryBaseProcessor<REQ, RESP>::collectProps(RowReader* reader,
...
             auto res = RowReader::getPropByName(reader, name);
             if (!ok(res)) {
-                VLOG(1) << "Skip the bad value for prop " << name;
-                continue;
+                VLOG(1) << "Bad value for prop " << name;
+                // TODO: Should return NULL
+                auto defaultVal = RowReader::getDefaultProp(prop.type_.type);
+                if (!defaultVal.ok()) {
+                    // Should never reach here.
+                    LOG(FATAL) << "Get default value failed for " << name;
```
laura-ding: if someone adds new value type and doesn't modify the `getDefaultProp`, here will crash.

src/storage/test/AdHocSchemaManager.h
```diff
@@ -54,6 +56,7 @@ class AdHocSchemaManager final : public nebula::meta::SchemaManager {
 
     StatusOr<std::string> toTagName(GraphSpaceID, TagID) override {
         LOG(FATAL) << "Unimplemented";
```
laura-ding: modify `FATAL` to `ERROR`

src/storage/test/AdHocSchemaManager.h
```diff
@@ -72,6 +75,7 @@ class AdHocSchemaManager final : public nebula::meta::SchemaManager {
 
     StatusOr<std::vector<std::string>> getAllEdge(GraphSpaceID) override {
         LOG(FATAL) << "Unimplemented";
```
laura-ding: ditto

src/storage/query/QueryBaseProcessor.inl
CPWstatic: We want it crash exactly. Someone add new type, must add the corresponding test.

laura-ding: It might  be discovered by user

src/storage/query/QueryBaseProcessor.inl
```diff
@@ -556,6 +565,55 @@ std::vector<Bucket> QueryBaseProcessor<REQ, RESP>::genBuckets(
...
+                    vertexSchemaResp_.emplace(tc.tagId_, std::move(respTag));
+                }
+            }
+        }
+    }
+
+    if (!edgeContexts_.empty()) {
+        for (const auto& ec : edgeContexts_) {
+            nebula::cpp2::Schema respEdge;
+            RowSetWriter rsWriter;
```
dangleptr: We don't need it anymore.

src/storage/query/QueryBoundProcessor.cpp
```diff
@@ -113,50 +121,13 @@ kvstore::ResultCode QueryBoundProcessor::processVertex(PartitionID partId, Verte
...
-                    edgeSchema.emplace(ec.first, std::move(respEdge));
-                }
-            }
-        }
-
-        if (!edgeSchema.empty()) {
-            resp_.set_edge_schema(std::move(edgeSchema));
-        }
+    if (!edgeSchemaResp_.empty()) {
+            resp_.set_edge_schema(std::move(edgeSchemaResp_));
```
dangleptr: ditto

src/storage/query/QueryBoundProcessor.cpp
```diff
@@ -113,50 +121,13 @@ kvstore::ResultCode QueryBoundProcessor::processVertex(PartitionID partId, Verte
...
-                if (it == vertexSchema.end()) {
-                    vertexSchema.emplace(tc.tagId_, respTag);
-                }
-            }
-        }
-        if (!vertexSchema.empty()) {
-            resp_.set_vertex_schema(std::move(vertexSchema));
-        }
+    if (!vertexSchemaResp_.empty()) {
+            resp_.set_vertex_schema(std::move(vertexSchemaResp_));
```
dangleptr: indent?

src/storage/query/QueryEdgePropsProcessor.cpp
```diff
@@ -26,9 +26,15 @@ kvstore::ResultCode QueryEdgePropsProcessor::collectEdgesProps(
     if (ret != kvstore::ResultCode::SUCCEEDED) {
         return ret;
     }
+
+    auto schema = edgeSchema_.find(edgeKey.edge_type);
```
dangleptr: For QueryEdgeProcessor,  it only support one edgeType one request.  So you could pass the related edge schema directly into collectEdgesProps

CPWstatic: Got it.

src/storage/query/QueryEdgePropsProcessor.cpp
```diff
@@ -67,17 +69,19 @@ void QueryEdgePropsProcessor::doProcess(const cpp2::EdgePropRequest& req) {
...
     }
 
-    int32_t edgeSize = std::accumulate(edgeContexts_.cbegin(), edgeContexts_.cend(), 0,
-                                       [](int ac, auto& ec) { return ac + ec.second.size(); });
-
-    int32_t returnColumnsNum = req.get_return_columns().size() + edgeSize;
+    auto schema = edgeSchema_.find(req.edge_type);
+    if (schema == edgeSchema_.end()) {
+        LOG(ERROR) << "Not found the edge type: " << req.edge_type;
+        this->onFinished();
```
dangleptr: We should return immediately after calling ”this->onFinished“

PR #1613, opened 2020-01-09
Separate read/write thread pool on storaged

dangleptr:
So the slow read request will not disrupt the normal hear beat logic.

### DIFF
--- a/src/storage/StorageServiceHandler.cpp
+++ b/src/storage/StorageServiceHandler.cpp
@@ -32,6 +32,7 @@
 
 DEFINE_int32(vertex_cache_num, 16 * 1000 * 1000, "Total keys inside the cache");
 DEFINE_int32(vertex_cache_bucket_exp, 4, "Total buckets number is 1 << cache_bucket_exp");
+DEFINE_int32(reader_handlers, 32, "Total reader handlers");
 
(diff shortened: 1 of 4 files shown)


src/storage/StorageServiceHandler.cpp
```diff
@@ -61,14 +62,17 @@ StorageServiceHandler::future_getProps(const cpp2::VertexPropRequest& req) {
...
 }
 
 folly::Future<cpp2::EdgePropResponse>
 StorageServiceHandler::future_getEdgeProps(const cpp2::EdgePropRequest& req) {
-    auto* processor = QueryEdgePropsProcessor::instance(kvstore_, schemaMan_, &edgePropsQpsStat_);
+    auto* processor = QueryEdgePropsProcessor::instance(kvstore_,
+                                                        schemaMan_,
+                                                        &edgePropsQpsStat_,
+                                                        readerPool_.get());
     RETURN_FUTURE(processor);
```
panda-sheep: Should the following  `future_getEdgeKeys `and `future_getUUID ` functions also be processed?

dangleptr: Good point.

PR #1618, opened 2020-01-10
Add VertexID, SrcID, DstID, Rank in FETCH query result

laura-ding:


src/graph/test/DataTest.cpp
```diff
@@ -257,14 +257,14 @@ TEST_F(DataTest, InsertTest) {
...
+        std::vector<std::tuple<int64_t, std::string, int64_t>> expected = {
+                {std::hash<std::string>()("Bob"), "Bob", 9},
         };
         ASSERT_TRUE(verifyResult(resp, expected));
         cmd = "FETCH PROP ON student hash(\"Bob\")";
         code = client_->execute(cmd, resp);
-        std::vector<std::tuple<std::string, int64_t>> expected2 = {
-                {"four", 20191106001},
+        std::vector<std::tuple<int64_t, std::string, int64_t>> expected2 = {
+                {std::hash<std::string>()("Bob"), "four", 20191106001},
```
yixinglu: hash("Bob") ? Isn't it `hash("four")` ?

laura-ding: The first column is the vertexID

src/graph/FetchEdgesExecutor.cpp
```diff
@@ -49,6 +49,11 @@ Status FetchEdgesExecutor::prepareClauses() {
         if (!status.ok()) {
             break;
         }
+
+        // Add SrcID, DstID, Rank before prepareYield()
+        returnColNames_.emplace_back("SrcID");
+        returnColNames_.emplace_back("DstID");
+        returnColNames_.emplace_back("Rank");
```
CPWstatic: I suggest we use the reserved property name, `_src`, `_dst`, `_rank`

src/graph/FetchVerticesExecutor.cpp
```diff
@@ -57,6 +57,9 @@ Status FetchVerticesExecutor::prepareClauses() {
             LOG(ERROR) << "Prepare vertex id failed: " << status;
             break;
         }
+
+        // Add VertexID before prepareYield()
+        returnColNames_.emplace_back("VertexID");
```
CPWstatic: `_id`

src/graph/FetchEdgesExecutor.cpp
laura-ding: `_src` and ` _dst` and `_rank` are internal names

laura-ding:
@Amber1990Zhang here need to update docs about `FETCH`
image
image
Add some columns in result about `FETCH`

src/graph/FetchEdgesExecutor.cpp
CPWstatic: It is, but we would better not rename for users. And users could also use the internal names in the nGQL. By the way, the storage has considered this situation, and return these value by default, please check it.

CPWstatic: Emm, `_src`, `_dst`, `_rank`, they are not internal names, i think. They are just reserved prop name which could be used to get the `key` of a vertex/edge. For example, people could use `<edge>._src` to get a source vertex id in an edge.

PR #1620, opened 2020-01-10
[doc]fix typo

Amber1990Zhang:


docs/manual-EN/1.overview/3.design-and-architecture/1.design-and-architecture.md
```diff
@@ -89,7 +89,7 @@ The binary of the meta service is **nebula-metad**. Here is the list of its main
   - TTL (time-to-live) management, supporting automatic data deletion and space reclamation.
 The meta service is stateful, and just like the storage service, it persists data to a key-value store.
```
wilsonyou: It uses Shell pipe as an alternative, which indicates that the output in the former query acts as the input in the latter one. Key features of nGQL are as follows:

PR #1625, opened 2020-01-14
Fix external llvm symbolizer error for clang asan test

yixinglu:
This PR depends on vesoft-inc/nebula-dev-docker#28

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -47,4 +47,4 @@ jobs:
         run: cd build && make -j $(nproc)
       - name: CTest with single thread
         timeout-minutes: 40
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yaml
```diff
@@ -47,4 +47,4 @@ jobs:
         run: cd build && make -j $(nproc)
       - name: CTest with single thread
         timeout-minutes: 40
-        run: cd build && ASAN_SYMBOLIZER_PATH=/usr/bin/llvm-symbolizer-8 ASAN_OPTIONS=fast_unwind_on_malloc=0 ctest --timeout 300 --output-on-failure
```
Shylock-Hg: `ASAN_SYMBOLIZER_PATH=/usr/bin/llvm-symbolizer-8`, the variable useless ?

yixinglu: refer to https://stackoverflow.com/questions/38079761/why-does-asan-symbolizer-path-no-longer-work-with-version-adorned-binaries

PR #1626, opened 2020-01-14
fix crash in storage with high payload

critical27:
Introduced in #1614, which will cause storage crashed because of `CHECK(replicatingLogs_);`.

Thread 1 could cas `replicatingLogs_` from false to true with `logsLock_`, if thread 2 set `replicatingLogs_` right after cas, it would crash.

### DIFF
--- a/src/kvstore/raftex/RaftPart.cpp
+++ b/src/kvstore/raftex/RaftPart.cpp
@@ -896,11 +896,11 @@ void RaftPart::processAppendLogResponses(
         }
         // Step 5: Check whether need to continue
         // the log replication
+        std::lock_guard<std::mutex> lck(logsLock_);
         CHECK(replicatingLogs_);
         // Continue to process the original AppendLogsIterator if necessary
         iter.resume();
         if (iter.empty()) {
-            std::lock_guard<std::mutex> lck(logsLock_);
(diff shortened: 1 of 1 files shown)


dangleptr:
I wan't to know how could it happens.  If the replicatingLogs_ is true,  before being reset to false,  it has no chance to send logs.

src/kvstore/raftex/RaftPart.cpp
```diff
@@ -896,11 +896,11 @@ void RaftPart::processAppendLogResponses(
         }
         // Step 5: Check whether need to continue
         // the log replication
+        std::lock_guard<std::mutex> lck(logsLock_);
```
dangleptr: Don't do that.

src/kvstore/raftex/RaftPart.cpp
```diff
@@ -896,11 +896,11 @@ void RaftPart::processAppendLogResponses(
...
         CHECK(replicatingLogs_);
         // Continue to process the original AppendLogsIterator if necessary
         iter.resume();
         if (iter.empty()) {
-            std::lock_guard<std::mutex> lck(logsLock_);
             VLOG(2) << idStr_ << "logs size " << logs_.size();
```
dangleptr: move the reset replicatingLogs_  logic after constructing iter

Issue #735, opened 2019-08-07
Preheat the storage client


dangleptr:
Currently,  storage client will pick up one host randomly to send request,  and after a while,  all leader addresses will be saved in storage client.   During the time,  almost all requests failed.  

To solve the problem,  we could use one preheat request to get all leaders from server-side before storage client is ready to work.

sherman-the-tank:
Good idea 👍

whitewum:
can you provide a command such as `check ok` to test if the cluster is available (read and write some tmp keys)? -- instead of `show hosts` or `show parts`, and check by `human eyes`-- It sucks.

( return not available is better than sth like internal error, or fetch error)

PR #1627, opened 2020-01-14
simple preheat for storage client in graph service

critical27:
Since we have many poc are going online, one problem is that first query of every partition probably would fail because of StorageClient has no leader info.

This PR is a simple version of preheat after #1236 and #1548 merged, when MetaClient init, it would try to get leader distribution from meta server, and storage client in graph could use it directly.

For simplicity, I only load leader distribution only once instead of update it periodically.  Which bring a problem, **you have to start meta and storage service first, after a few seconds, then you can start graph services**. 

PS: Just like problem above, after `balance leader`, the leader info probably is wrong. If you want to preheat again, you need to restart graph.

@Amber1990Zhang, in doc, we could advice user start meta and storage first, and start graph later.


dangleptr:
>>> Which bring a problem, you have to start meta and storage service first, after a few seconds, then you can start graph services.

We could handle the case by lazy init the leadersMap when accessing it.

dangleptr:
>>> PS: Just like problem above, after balance leader, the leader info probably is wrong. If you want to preheat again, you need to restart graph.

I think we don't need to repreheat after the leader balanced.

critical27:
> > > > Which bring a problem, you have to start meta and storage service first, after a few seconds, then you can start graph services.
> 
> We could handle the case by lazy init the leadersMap when accessing it.

Good suggestion

src/storage/client/StorageClient.cpp
```diff
@@ -502,6 +567,11 @@ folly::SemiFuture<StorageRpcResponse<storage::cpp2::GeneralResponse>>
 StorageClient::get(GraphSpaceID space,
                    const std::vector<std::string>& keys,
                    folly::EventBase* evb) {
+    if (!loadLeaderBefore_ && !loadLeader()) {
```
dangleptr: Could we move loadLeader inside this->leader() function.

critical27: fixed

dangleptr:
> Please note that: Since there involves synchronous RPC calls, there might be performance issues when bursting requests arrives(especially async ones) if `loadLeader` has not been invoked.

Yes, we should handle this problem.

critical27:
> > Please note that: Since there involves synchronous RPC calls, there might be performance issues when bursting requests arrives(especially async ones) if `loadLeader` has not been invoked.
> 
> Yes, we should handle this problem.

I've considered this problem before, how about double check the loadLeaderBefore_ in loadLeader, WDYT?

src/meta/client/MetaClient.cpp
```diff
@@ -1779,5 +1779,38 @@ Status MetaClient::refreshCache() {
...
+
+    LeaderMap leaderMap;
+    auto hostItems = std::move(ret).value();
+    for (auto& item : hostItems) {
+        auto hostAddr = HostAddr(item.hostAddr.ip, item.hostAddr.port);
+        for (auto& spaceEntry : item.get_leader_parts()) {
+            auto spaceName = spaceEntry.first;
+            auto status = getSpaceIdByNameFromCache(spaceName);
+            if (!status.ok()) {
+                continue;
```
dangleptr: Why continue here?

critical27: Cause `show hosts` return map of <space name, leader>, we have to convert to spaceId here. We just can ignore the if the space exists in `show hosts` but not in `show spaces`, actually they are both from meta, the `continue` rarely can hit.

dangleptr: Fine.

jude-zhu:
relate to #735

PR #1635, opened 2020-01-16
Fix thrift structure initialization

darionyaphet:


dutor:
If I were right, this fix is originated from the fact that `FRAGILE` construction is no longer supported in newer fbthrift, right?

dutor:
BTW. please resolve the conflicts first.

PR #1636, opened 2020-01-16
Make flags/stats/status HTTP API confirm to RESTful specification

yixinglu:
This PR is subtask of #1091  and depends on PR #1482

* Change `update` related http method to `PUT` and allow to update multiple flags in one time
* Rename `GET` API's filter condition like: `GET graph/flags?names=minloglevel`
* Update all related documents


src/storage/test/StorageHttpAdminHandlerTest.cpp
```diff
@@ -54,56 +54,28 @@ class StorageHttpAdminHandlerTestEnv : public ::testing::Environment {
...
-        auto resp = http::HttpClient::get(request);
-        ASSERT_TRUE(resp.ok());
-        ASSERT_EQ("ok", resp.value());
-    }
+static void checkInvalidRequest(const std::string& url, const std::string& errMsg) {
+    ASSERT_EQ(0, request(url).find(errMsg));
+}
+
+TEST(StoragehHttpAdminHandlerTest, TestInvalidRequests) {
+    checkInvalidRequest("/admin", "Space should not be empty");
```
Amber1990Zhang: should >- must

dangleptr:
Address it if needed in 2.0. So close it now.

PR #1646, opened 2020-01-17
Add UBSAN for clang compilation (github workflow).

monadbobo:


dutor:
I close this PR since UBSAN is too memory consuming, as mentioned in #1691 .

PR #1665, opened 2020-01-18
Disabled strict aliasing by default

dutor:
By disabling strict aliasing via `-fno-strict-aliasing`, it might have some negligible impacts on performance while avoiding relevant bugs which is hard to hunt, which is worthy for us.


bright-starry-sky:
The line 61 : `add_compile_options(-fno-strict-aliasing)` should be delete.

PR #1673, opened 2020-01-19
Fix the separator in spark doc

darionyaphet:


darionyaphet:
> Thanks for taking care of this, I have a pr #1648 fixed this too.

Currently we can't support `-` in the schema name so I have replace it with `_`

PR #1677, opened 2020-01-19
Keep relevant targets even if testing disabled

dutor:
When `ENABLE_TESTING` is set to `OFF`, this PR excludes the _**test**_ subdirectories out of the default build, while still keep the targets. So that we can build specific target explicitly. Meanwhile, targets will get built automatically if needed.


dutor:
> LGTM. This will speed up packaging and docker images building.

Not exactly, normally it's the same behavior as before.

PR #1681, opened 2020-01-19
[hot fix] pass value in lambda in asyncAtomicOp

critical27:
See [failed ut](https://github.com/vesoft-inc/nebula/pull/1677/checks?check_run_id=397469707).

Failed because we pass value by reference in async operation, some of them may be released before we use them.

PS: Since we have merged index, we need to add some UT no matter have index or not. (e.g. DeleteVertexTest only cover delete with index)


bright-starry-sky:
It's my fault. Thanks for your excellent solution.

PR #1689, opened 2020-01-20
Upgrade nebula-dev docker image for sanitize test

yixinglu:
Disable conversion null warning for generated source file


yixinglu:
@dutor Please rename required check again for newer compiler version

PR #1695, opened 2020-01-21
Check the index have existed although the name not the same

darionyaphet:


src/meta/processors/indexMan/CreateEdgeIndexProcessor.cpp
```diff
@@ -44,6 +44,37 @@ void CreateEdgeIndexProcessor::process(const cpp2::CreateEdgeIndexReq& req) {
...
+            if (i == fieldNames.size() - 1) {
+                LOG(ERROR) << "Edge Index " << indexName << " have existed";
+                resp_.set_code(cpp2::ErrorCode::E_EXISTED);
+                onFinished();
+                return;
+            }
+        }
+        checkIter->next();
+    }
+
```
bright-starry-sky: A question， Do we need to keep naming unique between tag index and edge index like tag and edge naming?

darionyaphet: Actually, I don't know why the index name is duplicated.

src/meta/processors/indexMan/CreateEdgeIndexProcessor.cpp
```diff
@@ -44,6 +44,37 @@ void CreateEdgeIndexProcessor::process(const cpp2::CreateEdgeIndexReq& req) {
...
+    while (checkIter->valid()) {
+        auto val = checkIter->val();
+        auto item = MetaServiceUtils::parseIndex(val);
+        if (item.get_schema_id().getType() != nebula::cpp2::SchemaID::Type::edge_type) {
+            continue;
+        }
+
+        for (size_t i = 0; i < fieldNames.size(); i++) {
+            if (fieldNames[i] != item.get_fields()[i].get_name()) {
```
dangleptr: Do you ensure the item have the same numbers fields as the filedNames?

darionyaphet: In fact, when the index's field is disaccord, they are not the same.

dangleptr: So maybe access invalid index on "item.get_fields()[i]"

src/meta/processors/indexMan/CreateTagIndexProcessor.cpp
```diff
@@ -43,6 +43,37 @@ void CreateTagIndexProcessor::process(const cpp2::CreateTagIndexReq& req) {
...
+
+            if (i == fieldNames.size() - 1) {
+                LOG(ERROR) << "Tag Index " << indexName << " have existed";
+                resp_.set_code(cpp2::ErrorCode::E_EXISTED);
+                onFinished();
+                return;
+            }
+        }
+        checkIter->next();
+    }
```
bright-starry-sky: For example , exist index1 (col1, col2), then we want create index2(col1, col2, col4), is it work well?

darionyaphet: Sure. 

In the test case `single_index` create on `name` and `multi_index` indexed on `name` and `email`.  It's work.

src/meta/processors/indexMan/CreateEdgeIndexProcessor.cpp
darionyaphet: Add a checker, if the size of current index is longer than the previous one , it will skip.

src/meta/processors/indexMan/CreateTagIndexProcessor.cpp
```diff
@@ -43,6 +43,38 @@ void CreateTagIndexProcessor::process(const cpp2::CreateTagIndexReq& req) {
...
```
bright-starry-sky: If index1 on tag1(col1, col2) exist,  create index on tag2(col1, col2)  will not be allowed. right?

bright-starry-sky: Could we unify the code block? I think the tag and edge will be able to share.

darionyaphet: Seems you should skip edge or tag by item.get_schema_id().getType()

bright-starry-sky: > Seems you should skip edge or tag by item.get_schema_id().getType()

There seems is a logic error, this check you should use tagId instead of id_type. 
you could add a test case : 
create tag1(col1, col2);
create tag2(col1, col2);
create index1 on tag1(col1,col2);
create index2 on tag2(col1, col2) --> here expect success . actual return an error E_EXISTED.

src/meta/processors/indexMan/CreateTagIndexProcessor.cpp
```diff
@@ -43,6 +43,38 @@ void CreateTagIndexProcessor::process(const cpp2::CreateTagIndexReq& req) {
         return;
     }
 
+    auto prefix = MetaServiceUtils::indexPrefix(space);
```
dangleptr: Could we unify the code.

src/meta/processors/indexMan/CreateTagIndexProcessor.cpp
darionyaphet: This compare just for the same schema

src/meta/processors/BaseProcessor.inl
```diff
@@ -454,5 +454,21 @@ BaseProcessor<RESP>::indexCheck(const std::vector<nebula::cpp2::IndexItem>& item
     return cpp2::ErrorCode::SUCCEEDED;
 }
 
+template<typename RESP>
```
Shylock-Hg, line 457: I think better put this function to `processor/indexMan/` folder, because it's not the concept of `BaseProcessor`.

Shylock-Hg: Please check `item.get_fields().szie() >= fields.size()`.

src/meta/processors/BaseProcessor.inl
```diff
@@ -454,5 +454,21 @@ BaseProcessor<RESP>::indexCheck(const std::vector<nebula::cpp2::IndexItem>& item
...
+        }
+
+        if (i == fields.size() - 1) {
+            LOG(ERROR) << "Index " << item.get_index_name() << " have existed";
+            return true;
+        }
+    }
+    return false;
+}
+
```
bright-starry-sky, line 472: Can we handle the error as below ? 
tag (col1, col2 ,col3)
create index on tag (col1, col2, col3)
create index2 on tag (col1, col2, col3, col1)

darionyaphet: Good Point We should check whether inculcate the duplicate column.

dangleptr:
It is a bug.  We should address it in next version.   @darionyaphet Could you rebase it?

src/meta/processors/BaseProcessor.inl
darionyaphet: before we check the index have checked the fields size.

PR #1696, opened 2020-01-21
Fixed session timeout

dutor:
Session should not have been expired if `FLAGS_session_idle_timeout_secs` is zero


Amber1990Zhang:
doc done.

PR #1697, opened 2020-01-21
Delete duplicate IndexID

bright-starry-sky:
IndexID is re-defined in file src/interface/common.thrift.


bright-starry-sky:
This is a duplicate pr, trace the issue using #1708 .

PR #1701, opened 2020-01-22
cmake 3.16 + sha256sum

goranc:
-- update to new version of cmake 3.16.3
-- change validity check to sha256sum where we can use cmake-3.16.3-SHA-256.txt file to get original sha256 hash from repository


dangleptr:
I think a better way is to change it locally,  we don't need to submit the changes to the repo. 
Thanks @goranc

Issue #465, opened 2019-05-29
Support query with index.


dangleptr:
Subtask of #71 

We should support query some vertexes and edges props with index on storage

sherman-the-tank:
Do you mean ```FIND``` statement?

PR #1705, opened 2020-01-31
Lookup sentence

bright-starry-sky:
This PR based on PR #1437.


bright-starry-sky:
Ready to review.  Thanks.

src/parser/scanner.lex
```diff
@@ -184,6 +185,7 @@ IP_OCTET                    ([0-9]|[1-9][0-9]|1[0-9][0-9]|2[0-4][0-9]|25[0-5])
 {WHEN}                      { return TokenType::KW_WHEN; }
 {DELETE}                    { return TokenType::KW_DELETE; }
 {FIND}                      { return TokenType::KW_FIND; }
```
darionyaphet: why don't remove `FIND` ?

bright-starry-sky: > why don't remove `FIND` ?

Find path using it.

src/graph/LookupExecutor.cpp
```diff
@@ -0,0 +1,496 @@
...
+        tagOrEdge_ = tagResult.value();
+        return Status::OK();
+    }
+    auto edgeResult = mc->getEdgeTypeByNameFromCache(spaceId_, *from_);
+    if (edgeResult.ok()) {
+        tagOrEdge_ = edgeResult.value();
+        isEdge_ = true;
+        return Status::OK();
+    }
+    return Status::Error("Tag or Edge was not found : %s", from_->c_str());
```
laura-ding: Add print error log, under ditto

src/graph/LookupExecutor.cpp
```diff
@@ -0,0 +1,496 @@
...
+    return status;
+}
+
+Status LookupExecutor::prepareYield() {
+    auto *clause = sentence_->yieldClause();
+    if (clause != nullptr) {
+        auto schema = (isEdge_) ? ectx_->schemaManager()->getEdgeSchema(spaceId_, tagOrEdge_)
+                                : ectx_->schemaManager()->getTagSchema(spaceId_, tagOrEdge_);
+        if (schema == nullptr) {
+            return Status::Error("No tag schema found : %s", from_->c_str());
```
laura-ding: No tag or edge schema found

src/graph/LookupExecutor.cpp
```diff
@@ -0,0 +1,496 @@
...
+}
+
+void LookupExecutor::setOutputYields(SchemaWriter *outputSchema,
+                                     const std::shared_ptr<ResultSchemaProvider>& schema) {
+    if (schema == nullptr) {
+        return;
+    }
+    for (auto& col : returnCols_) {
+        auto index = schema->getFieldIndex(col);
+        outputSchema->appendCol(col, schema->getFieldType(index).type);
```
laura-ding: there has `const ValueType& getFieldType(const Slice& name)`

bright-starry-sky: Good point !

src/graph/test/LookupTest.cpp
```diff
@@ -0,0 +1,180 @@
...
+        ASSERT_TRUE(verifyResult(resp, expected));
+    }
+    /**
+     * No matching index was found
+     */
+    {
+        cpp2::ExecutionResponse resp;
+        auto query = "LOOKUP ON lookup_edge_2 WHERE lookup_edge_2.col1 == \"col2_200\"";
+        auto code = client_->execute(query, resp);
+        ASSERT_EQ(cpp2::ErrorCode::E_EXECUTION_ERROR, code);
```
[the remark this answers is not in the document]
bright-starry-sky: here means no matching index was found.

src/graph/test/LookupTestBase.h
```diff
@@ -0,0 +1,192 @@
...
+    {
+        cpp2::ExecutionResponse resp;
+        std::string cmd = "CREATE EDGE lookup_edge_2(col1 string,"
+                          "col2 int, col3 double, col4 bool)";
+        auto code = client_->execute(cmd, resp);
+        if (cpp2::ErrorCode::SUCCEEDED != code) {
+            return TestError() << "Do cmd:" << cmd << " failed";
+        }
+    }
+    sleep(FLAGS_heartbeat_interval_secs + 3);
```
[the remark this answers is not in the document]
bright-starry-sky: for schemaManger and IndexManger.

src/graph/LookupExecutor.cpp
```diff
@@ -0,0 +1,499 @@
...
+        LOG(ERROR) << "Preparing failed: " << status;
+        return status;
+    }
+    return status;
+}
+
+Status LookupExecutor::prepareFrom() {
+    from_ = sentence_->from();
+    auto *mc = ectx()->getMetaClient();
+    auto tagResult = mc->getTagIDByNameFromCache(spaceId_, *from_);
```
critical27: Do we have forbidden same name in tag and edge?

src/graph/LookupExecutor.cpp
```diff
@@ -0,0 +1,499 @@
...
+                break;
+            }
+            ++hintCount;
+        }
+        /**
+         * return the index of max hint count.
+         */
+        indexHint[hintCount] = indexId;
+    }
+    index_ = indexHint.rbegin()->second;
```
critical27: Why we can return the indexId with most hintCount? Like the example above: `index1 (col4, col3, col2, col1)` and `index2 (col1, col2, col3, col4)`, I suppose the hintCount of both index are 3?

src/graph/LookupExecutor.cpp
```diff
@@ -0,0 +1,499 @@
...
+        if (outputSchema == nullptr) {
+            if (!yields_.empty()) {
+                schema = std::make_shared<ResultSchemaProvider>(*(resp.get_schema()));
+            }
+            outputSchema = std::make_shared<SchemaWriter>();
+            outputSchema->appendCol("VertexID", nebula::cpp2::SupportedType::VID);
+            setOutputYields(outputSchema.get(), schema);
+            rsWriter = std::make_unique<RowSetWriter>(outputSchema);
+        }
+        for (const auto& data : *resp.get_rows()) {
```
critical27: Perhaps you have missed the vid column when output. See example below, if type this more times, the console would crash.
image

src/graph/LookupExecutor.cpp
```diff
@@ -0,0 +1,499 @@
...
+void LookupExecutor::processVertexResult(VertexRpcResponse &&result) {
+    auto all = result.responses();
+    std::shared_ptr<SchemaWriter> outputSchema;
+    std::unique_ptr<RowSetWriter> rsWriter;
+    std::shared_ptr<ResultSchemaProvider> schema = nullptr;
+    for (auto &resp : all) {
+        if (!resp.__isset.rows || resp.get_rows() == nullptr || resp.rows.empty()) {
+            continue;
+        }
+        if (outputSchema == nullptr) {
```
critical27: When no yield, the vertex didn't show in the output in console.
image

src/graph/LookupExecutor.cpp
```diff
@@ -0,0 +1,499 @@
...
+        isEdge_ = true;
+        return Status::OK();
+    }
+    LOG(ERROR) << "Tag or Edge was not found : " << from_->c_str();
+    return Status::Error("Tag or Edge was not found : %s", from_->c_str());
+}
+
+Status LookupExecutor::prepareWhere() {
+    auto *clause = sentence_->whereClause();
+    whereWrapper_ = std::make_unique<WhereWrapper>(clause);
```
dangleptr: We don't need WhereWrapper here,  all filters should be pushed down to the storaged.

src/graph/LookupExecutor.cpp
```diff
@@ -0,0 +1,499 @@
...
+        auto status = yieldClauseWrapper_->prepare(inputs_.get(), varHolder, yields_);
+        if (!status.ok()) {
+            return status;
+        }
+        for (auto *col : yields_) {
+            if (!col->getFunName().empty()) {
+                return Status::SyntaxError("Not supported yet");
+            }
+            auto* prop = col->expr();
+            if (prop->kind() != Expression::kAliasProp) {
```
dangleptr: Maybe it is too strict.

src/graph/LookupExecutor.cpp
```diff
@@ -0,0 +1,499 @@
...
+                auto* aExpr = dynamic_cast<const AliasPropertyExpression*>(left);
+                prop = *aExpr->prop();
+                filters_.emplace_back(std::make_pair(prop, rExpr->op()));
+            } else if (left->kind() == nebula::Expression::kPrimary
+                       && right->kind() == nebula::Expression::kAliasProp) {
+                auto* aExpr = dynamic_cast<const AliasPropertyExpression*>(right);
+                prop = *aExpr->prop();
+                filters_.emplace_back(std::make_pair(prop, rExpr->op()));
+            } else {
+                return Status::SyntaxError("Unsupported syntax ：%s", rExpr->toString().c_str());
```
dangleptr: It is not correct.  For example,  a > 1 && (b > 1 && c > 1) ,  it could use the index (a, b, c)

src/graph/LookupExecutor.cpp
```diff
@@ -0,0 +1,499 @@
...
+    return Status::OK();
+}
+
+Status LookupExecutor::checkFilter() {
+    auto status = traversalExpr(whereWrapper_->getFilter());
+    return status;
+}
+
+Status LookupExecutor::chooseIndex() {
+    auto ret = (isEdge_) ? ectx()->getMetaClient()->getEdgeIndexesFromCache(spaceId_)
```
dangleptr: Put it in prepareClause.

src/graph/LookupExecutor.cpp
```diff
@@ -0,0 +1,499 @@
...
+                                    : mc->getTagIndexFromCache(spaceId_, indexId);
+        if (!itemStatus.ok()) {
+            return itemStatus.status();
+        }
+        const auto& index = itemStatus.value();
+        int32_t hintCount = 0;
+        for (const auto& field : index->get_fields()) {
+            auto it = std::find_if(filters_.begin(), filters_.end(),
+                                   [field](const auto &rel) {
+                                       return field.get_name() == rel.first &&
```
dangleptr: Only EQ could use the index?

src/graph/LookupExecutor.cpp
```diff
@@ -0,0 +1,499 @@
...
+    if (skipOptimize_) {
+        index_ = ids[0];
+        return status;
+    }
+    auto *mc = ectx()->getMetaClient();
+
+    /**
+     * std::map<column_hint_count, IndexID>
+     */
+    std::map<int32_t, IndexID> indexHint;
```
dangleptr: I think the logic is not correct

src/graph/LookupExecutor.cpp
```diff
@@ -0,0 +1,499 @@
...
+    if (rsWriter != nullptr) {
+        outputs->setInterim(std::move(rsWriter));
+    }
+
+    if (onResult_) {
+        onResult_(std::move(outputs));
+    } else {
+        resp_ = std::make_unique<cpp2::ExecutionResponse>();
+        auto colNames = outputs->getColNames();
+        resp_->set_column_names(std::move(colNames));
```
dangleptr: If no onResult_,  we could skip the extra encode/decode,  you could recheck the logic in GoExecutor.

src/graph/LookupExecutor.cpp
bright-starry-sky: The expr should be : 
LogicalExpr_root(RelationalExpr_1(a > 1)  AND LogicalExpr_1(RelationalExpr_2(b>1) AND RelationalExpr_3( c > 1))) . So this is a iteration and this index can be used.

src/graph/LookupExecutor.cpp
bright-starry-sky: > Only EQ could use the index?

No , The purpose of this method is to choose an optimal index , for example : 
Suppose the where clause is `col1 == 1 and col2 == 2 and col3 == 3`,
Indexes are all correct and available as below, here can be choose best one as index2 : 
```
index1 (col4, col3, col2, col1)
index2 (col1, col2, col3, col4)
index3 (col1, colA, col2, col3)
```

src/graph/LookupExecutor.cpp
bright-starry-sky: Fiexed

src/graph/LookupExecutor.cpp
bright-starry-sky: > Perhaps you have missed the vid column when output. See example below, if type this more times, the console would crash.
> image

fixed

src/graph/LookupExecutor.cpp
dangleptr: why you think index2 is the best,  the policy should be based on the data size for each index.

src/graph/LookupExecutor.cpp
dangleptr: As your logic,  because the root expr is LogicalExpr(AND),  so it will fall in the second switch,  and the left expr is RelationExpr and right is LogicExpr,  so SyntaxError will be returned. right?

src/graph/LookupExecutor.cpp
bright-starry-sky: > why you think index2 is the best, the policy should be based on the data size for each index.

index1's prefix : ""
index2's prefix : "col1+col2+col3"
index3.s prefix : "col1"

So the index2 is the best index.

src/graph/LookupExecutor.cpp
bright-starry-sky: > As your logic, because the root expr is LogicalExpr(AND), so it will fall in the second switch, and the left expr is RelationExpr and right is LogicExpr, so SyntaxError will be returned. right?

No ， This is an iterative function, if the root expr is LogicalExpr(AND), it should be call traversalExpr(left) and traversalExpr(right) again.

src/graph/LookupExecutor.cpp
bright-starry-sky: The purpose of `chooseIndex` is to select an optimal index, It required some conditions after the expression has been analyzed. I think it is a part of optimize, so I think it should be right here.

bright-starry-sky:
Addressed dangleptr's comments

src/graph/LookupExecutor.cpp
bright-starry-sky: > Do we have forbidden same name in tag and edge?

Yes, the tag name and edge names are unique. so  it's ok.

bright-starry-sky:
Add expression check to avoid memory leak

src/graph/test/LookupTestBase.h
```diff
@@ -0,0 +1,192 @@
...
+        cpp2::ExecutionResponse resp;
+        std::string cmd = "CREATE TAG lookup_tag_1(col1 string, col2 string, col3 string)";
+        auto code = client_->execute(cmd, resp);
+        if (cpp2::ErrorCode::SUCCEEDED != code) {
+            return TestError() << "Do cmd:" << cmd << " failed";
+        }
+    }
+    {
+        cpp2::ExecutionResponse resp;
+        std::string cmd = "CREATE TAG lookup_tag_2(col1 string, col2 int, col3 double, col4 bool)";
```
Amber1990Zhang: I think `double` or `float` is not supported.

src/graph/LookupExecutor.cpp
```diff
@@ -0,0 +1,555 @@
...
+            auto* left = rExpr->left();
+            auto* right = rExpr->right();
+            /**
+             *  TODO (sky) : Support WHERE ：tag1.col2 != tag1.col3
+             *  Handler error for FuncExpr or  ArithmeticExpr contains
+             *  AliasPropExpr , for example :
+             *  WHERE lookup_tag_2.col2 > (lookup_tag_2.col3 - 100)
+             */
+            if (left->kind() == nebula::Expression::kAliasProp &&
+                right->kind() == nebula::Expression::kAliasProp) {
```
dangleptr: In your case in comments,   the right expression  kind is not nebula::Expression::kAliasProp

src/graph/LookupExecutor.cpp
```diff
@@ -0,0 +1,555 @@
...
+    return findOptimalIndex(validIndexes);
+}
+
+/**
+ * Choose the index that filter best matches, for example :
+ * the filter is col1 == 1 and col2 == 2 and col3 == 3,
+ * the indexes are index1 (col4, col3, col2, col1)
+ *                 index2 (col1, col2, col3, col4)
+ *                 index3 (col1, colA, col2, col3)
+ * as above, these indexes all are valid, but the index2 should be choose.
```
dangleptr: In your case,  for the expression " col1 == 1 and col2 == 2 and col3 == 3",   only index2 is valid. 
index1 and index3 can't be used.

src/graph/LookupExecutor.cpp
dangleptr: I mean put the codes from line 198 to 208 into prepare step.

src/graph/LookupExecutor.cpp
```diff
@@ -0,0 +1,555 @@
...
+        resp_ = std::make_unique<cpp2::ExecutionResponse>();
+    }
+    doFinish(Executor::ProcessControl::kNext);
+}
+
+void LookupExecutor::finishExecution(std::unique_ptr<RowSetWriter> rsWriter) {
+    auto resultColNames = getResultColumnNames();
+    auto outputs = std::make_unique<InterimResult>(std::move(resultColNames));
+    if (rsWriter != nullptr) {
+        outputs->setInterim(std::move(rsWriter));
```
dangleptr: Please recheck the code in GoExecutor,  if there is no onResult_,  we could return directly without using RowSetWriter.

src/storage/index/IndexPolicyMaker.cpp
```diff
@@ -98,30 +109,27 @@ cpp2::ErrorCode IndexPolicyMaker::traversalExpression(const Expression *expr) {
...
                     return cpp2::ErrorCode::E_INVALID_FILTER;
                 }
                 v = value.value();
                 auto* aExpr = dynamic_cast<const AliasPropertyExpression*>(right);
                 prop = *aExpr->prop();
             } else {
-                traversalExpression(left);
-                traversalExpression(right);
-                break;
+                optimizedPolicy_ = false;
```
dangleptr: Why exit?

src/graph/LookupExecutor.cpp
```diff
@@ -0,0 +1,707 @@
...
+     * where where clause is below :
+     * col1 > 1 and col2 > 1 --> index1 and index2 are valid.
+     * col1 > 1 --> index1 is valid.
+     * col2 > 1 --> index2 is valid.
+     * col3 > 1 --> index3 is valid.
+     */
+    for (auto& index : indexes_) {
+        bool matching = true;
+        size_t filterNum = 1;
+        for (const auto& field : index->get_fields()) {
```
dangleptr: I think the logic still has some problems. 
As the examples in your comments,  for "col1 > 1",  index1 should be valid. 
But as you logic here, index1 is invalid.  Right?

src/graph/LookupExecutor.cpp
```diff
@@ -0,0 +1,707 @@
...
+
+    if (indexes.empty()) {
+        return Status::IndexNotFound();
+    }
+
+    /**
+     * step 2 , if have multiple valid indexes, get the best one.
+     * for example : if where clause is :
+     * col1 > 1 and col2 > 1 --> index1 and index2 are valid. get one of these at random.
+     * col1 > 1 and col2 == 1 --> only index2 is valid. because col2 have a equivalent value.
```
dangleptr: I think both index1 and index2 are valid too in this case.

src/graph/LookupExecutor.cpp
bright-starry-sky: Emm.. index1 really valid when col1 > 1,  and by this example only index1 is valid. 
here I using index's columns to check the index validity. Check field matching in order of columns, the loop will break when not matching column found.

src/graph/LookupExecutor.cpp
bright-starry-sky: > I think both index1 and index2 are valid too in this case.

Sure, the both indexes are valid, but I need choose a best one, because only need one index in storage layer.

dangleptr: Update the comments.

src/graph/LookupExecutor.cpp
dangleptr: make sense

src/graph/LookupExecutor.cpp
dangleptr: We'd better remove the if condition.

src/storage/index/IndexPolicyMaker.cpp
dangleptr: Index (col1, col2)
For the expression "co1 > 1 and col1 < 10 and col2 > 5",   the index can't be used?

bright-starry-sky: This index can be used. This is a standard logical expression. 
for example in my test :
```
CREATE EDGE lookup_edge_1(col1 int, col2 int, col3 int);
CREATE EDGE INDEX e_index_1 ON lookup_edge_1(col1, col2, col3);
LOOKUP ON lookup_edge_1 WHERE lookup_edge_1.col1 >1 and lookup_edge_1.col1 < 3 and lookup_edge_1.col2 > 1;
Execution succeeded (Time spent: 2.461/4.992 ms)
```

dangleptr: Got it

src/graph/LookupExecutor.cpp
```diff
@@ -0,0 +1,704 @@
...
+    std::vector<std::string> result;
+    result.reserve(yields_.size() + ((isEdge_) ? 3 : 1));
+    if (isEdge_) {
+        result.emplace_back("SrcVID");
+        result.emplace_back("DstVID");
+        result.emplace_back("Ranking");
+    } else {
+        result.emplace_back("VertexID");
+    }
+    result.reserve(yields_.size());
```
panda-sheep: line 393 has reserved

bright-starry-sky: Sharp eyes

src/graph/LookupExecutor.cpp
```diff
@@ -0,0 +1,704 @@
...
+    Status status = Status::OK();
+    spaceId_ = ectx()->rctx()->session()->space();
+    expCtx_ = std::make_unique<ExpressionContext>();
+    expCtx_->setStorageClient(ectx()->getStorageClient());
+    expCtx_->setSpace(spaceId_);
+    do {
+        status = checkIfGraphSpaceChosen();
+        if (!status.ok()) {
+            break;
+        }
```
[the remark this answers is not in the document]
bright-starry-sky: good suggestion.

bright-starry-sky:
Addressed panda-sheep's comment.

jude-zhu:
close #465

PR #1712, opened 2020-02-03
limit PARTITION_NUM/REPLICA_FACTOR

knightXun:
we should limit the range of `PARTITION_NUM` and `REPLICA_FACTOR`. The data type of the two values is int64, The value could be very large.

### DIFF
--- a/src/graph/CreateSpaceExecutor.cpp
+++ b/src/graph/CreateSpaceExecutor.cpp
@@ -23,13 +23,17 @@ Status CreateSpaceExecutor::prepare() {
             case SpaceOptItem::PARTITION_NUM:
                 partNum_ = item->get_partition_num();
                 if (partNum_ <= 0) {
-                    return Status::Error("Partition_num value should be greater than zero");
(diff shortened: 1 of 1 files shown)


src/graph/CreateSpaceExecutor.cpp
```diff
@@ -23,13 +23,17 @@ Status CreateSpaceExecutor::prepare() {
             case SpaceOptItem::PARTITION_NUM:
                 partNum_ = item->get_partition_num();
                 if (partNum_ <= 0) {
-                    return Status::Error("Partition_num value should be greater than zero");
+                    return Status::Error("partition_num value should be greater than zero");
+                } else if (partNum_ >=100000) {
```
dangleptr: If you really want to limit the max numbers,  use one gflag to control it.  And the logic should be implemented in metad.

knightXun:
close this, we need better design

src/graph/CreateSpaceExecutor.cpp
knightXun: ok i understand

PR #1717, opened 2020-02-05
Support properties on in-edge

dangleptr:
To improve the efficient, we store properties on in-edge too.  

Main changes:
1.  Take properties when inserting reverse edges.
2.  Remove all special handling code for reversely traverse in GoExecutor.  Now,  for reversely traverse, we handle it as a special edgeType(which is negative.)
3.  Remove all checks on storage-side.

Notes:
I haven't take care of updating logic,  because i think it needs a big refactor.  I will do it later.

### DIFF
--- a/src/graph/GoExecutor.cpp
+++ b/src/graph/GoExecutor.cpp
@@ -253,10 +253,6 @@ Status GoExecutor::prepareOver() {
 
     isReversely_ = clause->isReversely();
 
-    if (isReversely()) {
-        edgeHolder_ = std::make_unique<EdgeHolder>();
-    }
-
     auto edges = clause->edges();
     for (auto e : edges) {
         if (e->isOverAll()) {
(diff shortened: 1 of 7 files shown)


src/graph/GoExecutor.cpp
```diff
@@ -1136,6 +996,7 @@ bool GoExecutor::processFinalResult(RpcResponse &rpcResp, Callback cb) const {
                         }
                         return isReversely() ? srcId : dstId;
```
CPWstatic: This definition actually made users confused, as they talk about this in issue#1600. And it makes conflict with the definition of `GO BIDIRECT`. And i have change it in pr#1694. I will change the implementation of #1694 after this pr being merged.

dangleptr:
OK...  I will post another pr to make update logic correctly.

src/graph/GoExecutor.cpp
dangleptr: That's good.

PR #1722, opened 2020-02-06
[doc] limit statement updated

wilsonyou:


docs/manual-CN/2.query-language/2.functions-and-operators/limit-syntax.md
```diff
@@ -24,12 +24,12 @@ LIMIT [<offset_value>,] <number_rows>
 例如：
 
 ```ngql
-nebula> GO FROM 105 OVER follow YIELD $$.player.name AS Friend, $$.player.age AS Age, follow._dst AS Follow | LIMIT 2
-================================
+nebula> GO FROM 105 OVER follow YIELD $$.player.name AS Friend, $$.player.age AS Age, follow._dst AS Follow | LIMIT 2;
```
laura-ding: It is better to use `ORDER BY` before `LIMIT`, the result will be unique

wilsonyou: Sure.

PR #1724, opened 2020-02-06
Fix update logic after store properties on in-edges.

dangleptr:
Now after updating the out-edges successfully,  we update in-edges.

### DIFF
--- a/src/graph/UpdateEdgeExecutor.cpp
+++ b/src/graph/UpdateEdgeExecutor.cpp
@@ -144,7 +144,7 @@ std::vector<std::string> UpdateEdgeExecutor::getReturnColumns() {
 }
 
 
-void UpdateEdgeExecutor::finishExecution(storage::cpp2::UpdateResponse &&rpcResp) {
+void UpdateEdgeExecutor::toResponse(storage::cpp2::UpdateResponse &&rpcResp) {
     resp_ = std::make_unique<cpp2::ExecutionResponse>();
     std::vector<std::string> columnNames;
     columnNames.reserve(yields_.size());
@@ -193,9 +193,7 @@ void UpdateEdgeExecutor::finishExecution(storage::cpp2::UpdateResponse &&rpcResp
         rows.back().set_columns(std::move(row));
     }
     resp_->set_rows(std::move(rows));
-    doFinish(Executor::ProcessControl::kNext);
 }
-
 
 void UpdateEdgeExecutor::setupResponse(cpp2::ExecutionResponse &resp) {
     if (resp_ == nullptr) {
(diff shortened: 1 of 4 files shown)


src/graph/UpdateEdgeExecutor.cpp
```diff
@@ -279,11 +247,11 @@ void UpdateEdgeExecutor::execute() {
...
         }
-        if (insertable_ && rpcResp.get_upsert()) {
-            // TODO(zhangguoqing) Making the reverse edge of insertion is transactional
-            this->insertReverselyEdge(std::move(rpcResp));
+        if (reversely) {
+            doFinish(Executor::ProcessControl::kNext);
         } else {
-            this->finishExecution(std::move(rpcResp));
+            this->toResponse(std::move(rpcResp));
+            this->updateEdge(true);
```
bright-starry-sky: 👍

src/graph/UpdateEdgeExecutor.cpp
```diff
@@ -190,68 +190,36 @@ void UpdateEdgeExecutor::finishExecution(storage::cpp2::UpdateResponse &&rpcResp
...
     resp = std::move(*resp_);
 }
 
-
-void UpdateEdgeExecutor::insertReverselyEdge(storage::cpp2::UpdateResponse &&rpcResp) {
-    std::vector<storage::cpp2::Edge> edges;
-    storage::cpp2::Edge reverselyEdge;
-    reverselyEdge.key.set_src(edge_.dst);
-    reverselyEdge.key.set_dst(edge_.src);
-    reverselyEdge.key.set_ranking(edge_.ranking);
```
laura-ding: why not set ranking

dangleptr: The edge_ is reused here.  The ranking has been set in the previous step.

src/graph/UpdateEdgeExecutor.cpp
```diff
@@ -190,68 +190,36 @@ void UpdateEdgeExecutor::finishExecution(storage::cpp2::UpdateResponse &&rpcResp
...
+    std::string filterStr = "";
+    std::vector<std::string> returns;
+    if (reversely) {
+        auto src = edge_.src;
+        auto dst = edge_.dst;
+        edge_.set_src(dst);
+        edge_.set_dst(src);
+        edge_.set_edge_type(-edge_.edge_type);
+    } else {
+        filterStr = filter_ ? Expression::encode(filter_) : "";
```
darionyaphet: So the filter just processing when the edge is not reverse ?

dangleptr: When updating the reverse edges, it means the out edges have been updated succeeded. So we don't need to check the filter again.

src/storage/mutate/UpdateEdgeProcessor.cpp
```diff
@@ -268,13 +270,13 @@ std::string UpdateEdgeProcessor::updateAndWriteBack(PartitionID partId,
...
                 if (!val_.empty()) {
                     if (rReader == nullptr) {
                         rReader = RowReader::getEdgePropReader(this->schemaMan_,
                                                                val_,
                                                                spaceId_,
-                                                               edgeKey.edge_type);
+                                                               std::abs(edgeKey.edge_type));
                     }
                     auto rValues = collectIndexValues(rReader.get(),
                                                       index->get_fields());
```
bright-starry-sky: Let me handle the in-edge prop of index with insert edge together.
Could you help change back line 273, 279 and 295?

dangleptr: Let me check the edge_type, if it is negative, just skip directly.  You could handle it later.

bright-starry-sky: > Let me check the edge_type, if it is negative, just skip directly. You could handle it later.

excellent!

bright-starry-sky:
Well done!

Issue #1609, opened 2020-01-09
optimize "go ... reversely yield edge._src" performance


jude-zhu:
as titled

Issue #1600, opened 2020-01-07
go...reversely, "$$" returns the src of the edge while edgetype._dst returns the dst of the edge


jude-zhu:
the topology of the data is:

```
202 -follow-> 200 -follow-> 201
```

```
> go from 200 over follow reversely yield $$.student.name
===================
| $$.student.name |
===================
| 202             |
-------------------

go from 200 over follow reversely yield follow._src
===============
| follow._src |
===============
| 202         |
---------------

go from 200 over follow reversely yield follow._dst
===============
| follow._dst |
===============
| 200         |
---------------
```
 `$$` refers to the target of the edge, however, `$$.student.name` gets the same result as ` `follow._src` instead of `follow._dst`.

tong-hao:
@jude-zhu   Assign to me

Issue #1358, opened 2019-11-29
change reverse query to return the source vertex of the edge


jude-zhu:
vertexA -> [relationship] -> vertexB

Without specifying the return props, `Go from vertexB over relationship reversely;` returns the dst of the edge, which is vertexB. Should we return the src of the edge for the reverse query?

![1](https://user-images.githubusercontent.com/51590253/69854832-08941080-12c5-11ea-841a-f4893990cf08.jpg)

ffisherr:
@jude-zhu, can i help with this issue? Seems pretty easy. I'm new to this project, so, what should i start with?

Shylock-Hg:
> @jude-zhu, can i help with this issue? Seems pretty easy. I'm new to this project, so, what should i start with?

Welcome of course. You can solve any issue you want if nobody assigned.
I think you should read the nGQL syntax from the `docs` first.

PR #1740, opened 2020-02-07
Bidirectional traverse.

### LINKED ISSUE #1570: 'go' graph traversal supports both direction
graph traversal supports both direction，as in cypher:

```
match n1-[r:TYPE]-n2 -- Direction: BOTH (a bidirectional edge)
```

CPWstatic:
Close #1570 #1600 #1609 #1358

I will close #1694, if this is an accepted solution.


src/parser/Clauses.h
```diff
@@ -219,10 +218,16 @@ class OverEdges final {
 
 class OverClause final : public Clause {
 public:
-    explicit OverClause(OverEdges *edges, bool isReversely = false) {
+    enum Direction : uint8_t {
```
monadbobo: enum class was more better.

src/storage/query/QueryBaseProcessor.inl
```diff
@@ -137,6 +137,15 @@ cpp2::ErrorCode QueryBaseProcessor<REQ, RESP>::checkAndBuildContexts(const REQ&
...
+                auto it2 = edgeProps.find(edgeType);
+                if (it2 == edgeProps.end()) {
+                    std::vector<PropContext> v{std::move(prop)};
+                    edgeProps.emplace(edgeType, std::move(v));
+                    break;
+                }
+                it2->second.emplace_back(std::move(prop));
+                break;
+                /*
                 auto it2 = edgeContexts_.find(edgeType);
```
Shylock-Hg: Remove the comment code?

src/graph/GoExecutor.cpp
```diff
@@ -229,13 +229,18 @@ Status GoExecutor::prepareOverAll() {
         }
 
         auto v = edgeStatus.value();
-        if (isReversely()) {
+        if (direction_ == OverClause::Direction::kForward) {
```
monadbobo: It would be better to replace "if" with a "switch"

src/graph/GoExecutor.cpp
```diff
@@ -267,17 +272,23 @@ Status GoExecutor::prepareOver() {
...
```
monadbobo: This code is exactly the same as above, so it would be better to use a function to represent it.

src/storage/query/QueryBaseProcessor.inl
```diff
@@ -101,13 +102,13 @@ cpp2::ErrorCode QueryBaseProcessor<REQ, RESP>::checkAndBuildContexts(const REQ&
                 break;
             }
             case cpp2::PropOwner::EDGE: {
-                EdgeType edgeType = col.id.get_edge_type();
-                auto edgeName = this->schemaMan_->toEdgeName(spaceId_, std::abs(edgeType));
+                EdgeType edgeType = std::abs(col.id.get_edge_type());
```
dangleptr: Return positive edgeType for reverse edge?  I think it is not correct. 
Currently,  we take reverse edge as a special edge type.  So when you want to traverse the reverse edge,  it should return the negative edge type.

dangleptr: But i notice that edgeSchemaResp_ only contains abs(edgeType)

CPWstatic: The schema always to be same for both direction. But data is different.

dangleptr: Don't do that.  For reverse edge, we should return negative edgeType.  So QE could distinguish which one is reversely.

src/graph/GoExecutor.cpp
```diff
@@ -990,13 +1020,12 @@ bool GoExecutor::processFinalResult(RpcResponse &rpcResp, Callback cb) const {
                                         "Get edge type for `%s' failed in getters.",
                                         edgeName.c_str());
                             }
-                            if (type != edgeType) {
+                            if (type != std::abs(edgeType)) {
                                 return 0L;
                             }
                         }
-                        return isReversely() ? srcId : dstId;
+                        return dstId;
```
dangleptr: @whitewum Please note the changes here.

src/graph/GoExecutor.cpp
```diff
@@ -1074,16 +1102,49 @@ bool GoExecutor::processFinalResult(RpcResponse &rpcResp, Callback cb) const {
                             return Status::Error(
                                     "Get edge type for `%s' failed in getters.", edgeName.c_str());
                         }
-                        if (edgeType != type) {
-                            auto sit = edgeSchema.find(type);
-                            if (sit == edgeSchema.end()) {
-                                LOG(ERROR) << "Can't find schema for " << edgeName;
-                                return Status::Error("get schema failed");
+                        if (std::abs(edgeType) != type) {
+                            switch (direction_) {
```
dangleptr: If seems that we don't need the switch case.  The follows code could be enough
```
auto sit = edgeSchema.find(direction_ == OverClause::Direction::kBackward ? -type : type);
if (sit == edgeSchema.end()) {
     LOG(ERROR) << "Can't find schema for " << edgeName;
     return Status::Error("get schema failed");
}
return RowReader::getDefaultProp(sit->second.get(), prop);
```

src/dataman/RowReader.h
```diff
@@ -91,7 +91,12 @@ class RowReader {
     static StatusOr<VariantType> getDefaultProp(const meta::SchemaProviderIf* schema,
                                                 const std::string& prop) {
         auto& vType = schema->getFieldType(prop);
-        return getDefaultProp(vType.type);
+        auto defaultVal = getDefaultProp(vType.type);
+        if (!defaultVal.ok()) {
+            LOG(ERROR) << "Get default value for `" << prop << "' failed: " << defaultVal.status();
```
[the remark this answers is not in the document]
CPWstatic: We could return `defaultVal` directly whether ok or not.

src/graph/GoExecutor.cpp
```diff
@@ -442,10 +453,13 @@ void GoExecutor::stepOut() {
     }
     auto returns = status.value();
     std::string filterPushdown = "";
-    if (FLAGS_filter_pushdown && isFinalStep() && !isReversely()) {
+    if (FLAGS_filter_pushdown && isFinalStep()
+            && direction_ == OverClause::Direction::kForward) {
```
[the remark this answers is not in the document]
CPWstatic: There is a new pr #1785 to do this, based on this pr.

src/dataman/RowReader.h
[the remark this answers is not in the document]
dangleptr: If is not OK,  what's the default value?  Is there any value inside it?

src/graph/GoExecutor.cpp
whitewum: can anybody tell me:

E.g.,
```
Match (n)-[xx]->(m)
Return id(m)
```
is 
```
GO FROM n OVER xx YIELD xx._dst
```

```
Match (m)<-[xx]-(n)
Return id(n)
```
is it 
```
GO FROM m REVERSELY OVER xx YIELD xx._dst or xx._src
```

I'm confused several times, and can't get it anymore.

dangleptr: e.g.2
GO FROM m REVERSELY OVER xx YIELD xx._dst 

e.g.3
GO FROM n BIDIRECT OVER xx YIELD xx._dst


You can think the "_src" represents myself,   "_dst" means the other point.

PR #1741, opened 2020-02-08
Remove discard spark doc and upgrade java client's version

darionyaphet:


critical27:
Do we need to update pom since client has been updated?

PR #1742, opened 2020-02-10
[doc] yield statement updated

wilsonyou:


docs/manual-EN/2.query-language/4.statement-syntax/2.data-query-and-manipulation-statements/yield-syntax.md
```diff
@@ -42,56 +45,57 @@ YIELD
...
+---------------------
+| 110               |
+---------------------
 ```
 
-### As Separate Statement
+### As Stand-alone Statement
 
-`YIELD` statement can be used independently to retrieve computation results without reference to any graph. You can use `AS` to rename an alias.
+`YIELD` statement can be used independently to retrieve computation results without reference to any graph. You can use `AS` to rename it an alias.
```
Amber1990Zhang: You can use `AS` to rename it an alias. >- You can use `AS` to rename it as an alias.
This sentence is weird to me but I'm not sure if I fixed it.

docs/manual-EN/2.query-language/4.statement-syntax/2.data-query-and-manipulation-statements/yield-syntax.md
```diff
@@ -42,56 +45,57 @@ YIELD
...
+| 102               |
+---------------------
+| 104               |
+---------------------
+| 110               |
+---------------------
 ```
 
-### As Separate Statement
+### As Stand-alone Statement
```
Amber1990Zhang: Stand-Alone

wilsonyou: alone should be alone

PR #1746, opened 2020-02-10
Reservoir sampling.

CPWstatic:
This implementation is a kind of cutoff strategy.


src/storage/query/QueryBaseProcessor.inl
```diff
@@ -524,7 +543,125 @@ kvstore::ResultCode QueryBaseProcessor<REQ, RESP>::collectEdgeProps(
...
+template<typename REQ, typename RESP>
+void QueryBaseProcessor<REQ, RESP>::samplingCollectEdgeProps(
+                                               kvstore::KVIterator* iter,
+                                               VertexID vId,
+                                               EdgeType edgeType,
+                                               const std::vector<PropContext>& props,
+                                               FilterContext* fcontext,
+                                               EdgeProcessor proc) {
+    auto sampler = std::make_unique<ReservoirSampling>();
+    std::vector<std::pair<std::string, std::string>> kvs;
```
monadbobo: maybe std::array more better?

src/storage/query/QueryBaseProcessor.inl
```diff
@@ -524,7 +543,125 @@ kvstore::ResultCode QueryBaseProcessor<REQ, RESP>::collectEdgeProps(
...
+            kvs[index] = std::make_pair(key.str(), val.str());
+            ++cnt;
+        }
+    }
+
+    if (cnt > FLAGS_max_edge_returned_per_vertex) {
+        cnt = FLAGS_max_edge_returned_per_vertex;
+    }
+
+    for (auto i = 0; i < cnt; ++i) {
```
monadbobo: Most of the code in this loop is the same as in collectEdgeProps, so it is best to abstract a common method.

src/storage/query/QueryBaseProcessor.h
```diff
@@ -34,6 +37,32 @@ struct Bucket {
...
+            if (index < FLAGS_max_edge_returned_per_vertex) {
+                return index;
+            } else {
+                return -1;
+            }
+        }
+    }
+
+private:
+    boost::random::mt19937  rng_;
```
dutor: Why not use PRNG from the STL?
Why not use a 64bit generator?
IMO. `dist` should be reused.

CPWstatic: The distribution changes as the sample space grows, it can not be reused.

src/storage/query/QueryBaseProcessor.inl
CPWstatic: This is a big loop, is it a good idea to make it a function call?

dangleptr:
What's the purpose about this pr?  What's the problem it want to solve?

src/storage/query/QueryBaseProcessor.h
```diff
@@ -34,6 +37,32 @@ struct Bucket {
 
 using OneVertexResp = std::tuple<PartitionID, VertexID, kvstore::ResultCode>;
 
+class ReservoirSampling final {
+public:
+    ReservoirSampling() {
+        rng_.seed(nebula::time::WallClock::fastNowInMicroSec());
+    }
+
+    int64_t sampling() {
```
dangleptr: I suggest pass the samples total number N in ctor,  put the final results vector in this class. 
And as a common class,  it should be a template class.  The class looks like:

```
template<class T>
class ReservoirSampling final {
public:
    ReservoirSampling(int32_t N) {
        rng_.seed(nebula::time::WallClock::fastNowInMicroSec());
        num_ = N;   
    }

    int64_t sampling(T&&  element) {
           ....
    }

    
private:
       std::vector<T>   samples_;
       int32_t num_;
       ....
```



At last, add a UT for ReservoirSampling,  add we'd better move it into common dir.

src/storage/query/QueryBaseProcessor.h
```diff
@@ -34,6 +37,32 @@ struct Bucket {
...
+        ++cnt_;
+        if (cnt_ < FLAGS_max_edge_returned_per_vertex) {
+            return cnt_;
+        } else {
+            std::uniform_int_distribution<> dist(0, cnt_);
+            int64_t index = dist(rng_);
```
dangleptr: You could use folly::random directly here.

CPWstatic: I had read it before, its a simple wrapper of PRNG, the code is almost the same as here. So it's not necessary i think.

src/storage/query/QueryBaseProcessor.h
CPWstatic: I'll do it.

critical27:
I guess this PR is for random walk? Clean code!

PR #1753, opened 2020-02-11
[doc] property_reference updated

wilsonyou:


docs/manual-CN/2.query-language/3.language-structure/property-reference.md
```diff
@@ -23,7 +23,7 @@ $$.tag_name.prop_name
 ### 示例
 
 ```ngql
-nebula> GO FROM 1 OVER e1 YIELD $^.start.name AS startName, $$.end.Age AS endAge
+nebula> GO FROM 100 OVER follow YIELD $^.player.name AS StartName, $$.player.age AS EndAge;
```
whitewum: StartName is an alias. why capital inital letter?

PR #1756, opened 2020-02-12
lease read, check has committed logs in this term, modify collectN eval

critical27:
Some changes during jepsen test
1. Add leader lease check, if leader has not send message successfully in a lease, reject read request.
2. Add a `ready` flag, when a leader is elected, accept read request after the heartbeat.
3. Also collect the result when `eval` is false in `collectN`, so we can handle the resp in `processElectionResponses` and `processAppendLogRequest`.


src/kvstore/raftex/RaftPart.cpp
```diff
@@ -1162,7 +1167,14 @@ bool RaftPart::leaderElection() {
                 }
             }
             weight_ = 1;
-            sendHeartbeat();
+            sendHeartbeat().then([this] (folly::Try<AppendLogResult>&& t) {
+                CHECK(!t.hasException());
+                auto result = std::move(t).value();
+                if (result == AppendLogResult::SUCCEEDED) {
+                    std::lock_guard<std::mutex> g(raftLock_);
```
dangleptr: Do we really need the lock here?

src/kvstore/raftex/RaftPart.cpp
```diff
@@ -1162,7 +1167,14 @@ bool RaftPart::leaderElection() {
                 }
             }
             weight_ = 1;
-            sendHeartbeat();
+            sendHeartbeat().then([this] (folly::Try<AppendLogResult>&& t) {
```
dangleptr: Could we unify the logic with leaseValid?

src/kvstore/raftex/RaftPart.cpp
```diff
@@ -1113,22 +1124,16 @@ bool RaftPart::leaderElection() {
...
             quorum_,
             // Result evaluator
-            [hosts, this](size_t idx, cpp2::AskForVoteResponse& resp) {
-                if (resp.get_error_code() == cpp2::ErrorCode::E_LOG_STALE) {
-                    LOG(INFO) << idStr_ << "My last log id is less than " << hosts[idx]->address()
-                              << ", double my election interval.";
-                    uint64_t curWeight = weight_.load();
-                    weight_.store(curWeight * 2);
-                }
+            [hosts] (size_t idx, cpp2::AskForVoteResponse& resp) {
```
dangleptr: After modify the collectN, Is there any difference with the original logic?

critical27: Actually no difference, but we can handle the corresponding ErrorCode, e.g. we can handle `E_EXCEPTION` to step down later.

src/kvstore/raftex/RaftPart.cpp
critical27: Good suggestion

src/kvstore/raftex/RaftPart.cpp
critical27: You are right, I suppose `atomic_bool` would be enough.

src/kvstore/raftex/RaftPart.cpp
```diff
@@ -34,6 +34,9 @@ DEFINE_int32(wal_buffer_size, 8 * 1024 * 1024, "Default wal buffer size");
 DEFINE_int32(wal_buffer_num, 2, "Default wal buffer number");
 DEFINE_bool(trace_raft, false, "Enable trace one raft request");
 
+DEFINE_double(lease_valid_deviation, 0.8, "");
```
darionyaphet: Would you support some  description ?

src/kvstore/raftex/RaftPart.cpp
```diff
@@ -962,7 +966,7 @@ bool RaftPart::needToSendHeartbeat() {
     std::lock_guard<std::mutex> g(raftLock_);
     return status_ == Status::RUNNING &&
            role_ == Role::LEADER &&
-           lastMsgSentDur_.elapsedInSec() >= FLAGS_raft_heartbeat_interval_secs * 2 / 5;
+           lastMsgSentDur_.elapsedInMSec() >= FLAGS_raft_heartbeat_interval_secs * 1000 * 2 / 5;
```
darionyaphet: Could using `FLAGS_raft_heartbeat_interval_ms` ?

src/kvstore/raftex/RaftPart.cpp
critical27: My bad.

src/kvstore/raftex/RaftPart.cpp
critical27: That would be better, I just want to reserve the name `raft_heartbeat_interval_secs`.

critical27:
I will add follower stickness later, it still has some problems to solve in ut.

src/kvstore/raftex/RaftPart.cpp
```diff
@@ -763,6 +767,7 @@ void RaftPart::replicateLogs(folly::EventBase* eb,
 
     VLOG(2) << idStr_ << "About to replicate logs to all peer hosts";
 
+    lastMsgSentDur_.reset();
```
dangleptr: Why move it here?  It is not the meaning of lastMsgSentDur_

critical27: I've renamed the original `lastMsgSentDur_` to `lastMsgAcceptedDur_`, `lastMsgSentDur_` is used to record how long the net io time

dangleptr: i think what you want is lastMsgSendingMs_ and lastMsgSentMs_

src/kvstore/raftex/RaftPart.cpp
dangleptr: before sendHeartBeat succeeded,  the lease should be invalid,  so we could unify it?

src/kvstore/raftex/RaftPart.cpp
critical27: Yeah, so I use `lastMsgAcceptedCostMs_` to store difference between lastMsgSendingMs_ and lastMsgSentMs_

dangleptr: That's ok

src/kvstore/raftex/RaftPart.cpp
dangleptr: make sense.

src/kvstore/raftex/RaftPart.cpp
```diff
@@ -1295,6 +1316,16 @@ void RaftPart::processAskForVoteRequest(
         return;
     }
 
+    auto candidate = HostAddr(req.get_candidate_ip(), req.get_candidate_port());
```
dangleptr: we should have the same check when processing append log response.

src/kvstore/raftex/RaftPart.cpp
critical27: Agreed, the lease should be invalid. I mean it seems we can not tell whether sendHeartBeat succeeded by `lastMsgAcceptedDur_`? 
So if a node restart very quickly and elected as leader, but it has not committed any logs in this term. So we need to reject read request before any log committed, because there could be some logs  of previous term which has not been committed. I suppose it is different from lease, in this case, `lastMsgAcceptedDur_` would be quite small, and `leaseValid` would return valid. WDYT?

dangleptr: > I suppose it is different from lease, in this case, lastMsgAcceptedDur_ would be quite small, and leaseValid would return valid. 

I am confused,  why lastMsgAcceptedDur_ would be quite small,  if the heartbeat not been accepted by the quorum followers,  it should be larger than the lease interval.

critical27: I mean if a node elected as leader, before the heartbeat was accepted by quorum, `lastMsgAcceptedDur_` could be less than lease interval

dangleptr: lastMsgAcceptedDur_  equals "now - lastMsgAcceptedTime"，  why it could be less than lease interval

critical27: When it start up or after reboot, `lastMsgAcceptedTime` could be quite close to now?

dangleptr: lastMsgAcceptedTime  should be inited to be zero after reboot.

critical27: You're right

src/kvstore/NebulaStore.cpp
```diff
@@ -803,7 +803,7 @@ int32_t NebulaStore::allLeader(std::unordered_map<GraphSpaceID,
 }
 
 bool NebulaStore::checkLeader(std::shared_ptr<Part> part) const {
-    return !FLAGS_check_leader || part->isLeader();
+    return !FLAGS_check_leader || (part->isLeader() && part->leaseValid());
```
darionyaphet: Check lease is valid 👍

src/kvstore/raftex/RaftPart.cpp
```diff
@@ -1843,6 +1866,17 @@ void RaftPart::checkAndResetPeers(const std::vector<HostAddr>& peers) {
...
 }
 
+bool RaftPart::leaseValid() {
+    if (!FLAGS_has_leader_lease) {
+        return true;
+    }
+    // When majority has accepted a log, leader obtains a lease which last for heartbeat.
+    // However, we need to take off the net io time. On the left side of the inequality is
+    // the time duration since last time leader send a log (the log has been accepted as well)
+    return time::WallClock::fastNowInMilliSec() - lastMsgAcceptedTime_
```
darionyaphet: miracle

PR #1760, opened 2020-02-12
Support more case about Raft

darionyaphet:


dangleptr:
Excellent. Very great work.  

We could think more about it,  for raft, we need a framework for testing.  Let's talk about it offline.

Of course, the UTs in this pr is very valuable.

dangleptr:
The cases are good.  The pr is ready now?

darionyaphet:
Currently it’s not ready and need more carding. So close it temporary.

PR #1769, opened 2020-02-13
Fix potential raft bug

dangleptr:
It happens in our users testing.

case 1:
Suppose we have three replicas A(leader), B, C, after A crashed,
B, C will begin the election. B win, and send hb, C has gap with B
and need the snapshot from B. Meanwhile C begin the election,
C will be Candidate, but because C is in WAITING_SNAPSHOT,
so prepareElectionRequest will return false and go on the election.
Because C is in Candidate, so it will reject the snapshot request from B.
Infinite loop begins.

So we need to back the follower to avoid the case.

case 2:
It maybe loss data during balance,  because kvstore::prefix maybe return leader-change, 
at this time,  an empty snapshot will be sent to related learner.

### DIFF
--- a/src/kvstore/raftex/RaftPart.cpp
+++ b/src/kvstore/raftex/RaftPart.cpp
@@ -1086,6 +1086,16 @@ bool RaftPart::leaderElection() {
     cpp2::AskForVoteRequest voteReq;
     decltype(hosts_) hosts;
     if (!prepareElectionRequest(voteReq, hosts)) {
+        // Suppose we have three replias A(leader), B, C, after A crashed,
+        // B, C will begin the election. B win, and send hb, C has gap with B
+        // and need the snapshot from B. Meanwhile C begin the lection,
+        // C will be Candidate, but becasue C is in WAITING_SNAPSHOT,
+        // so prepareElectionRequest will return false and go on the election.
+        // Becasue C is in Candidate, so it will reject the snapshot request from B.
+        // Infinite loop begins.
+        // So we neeed to back the follower to avoid the case.
+        std::lock_guard<std::mutex> g(raftLock_);
+        role_ = Role::FOLLOWER;
         return false;
     }
 
--- a/src/kvstore/raftex/RaftPart.cpp
+++ b/src/kvstore/raftex/RaftPart.cpp
@@ -1086,14 +1086,14 @@ bool RaftPart::leaderElection() {
     cpp2::AskForVoteRequest voteReq;
     decltype(hosts_) hosts;
     if (!prepareElectionRequest(voteReq, hosts)) {
-        // Suppose we have three replias A(leader), B, C, after A crashed,
+        // Suppose we have three replicas A(leader), B, C, after A crashed,
         // B, C will begin the election. B win, and send hb, C has gap with B
-        // and need the snapshot from B. Meanwhile C begin the lection,
-        // C will be Candidate, but becasue C is in WAITING_SNAPSHOT,
+        // and need the snapshot from B. Meanwhile C begin the election,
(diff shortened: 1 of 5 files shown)


src/kvstore/raftex/RaftPart.cpp
```diff
@@ -1086,6 +1086,16 @@ bool RaftPart::leaderElection() {
...
     decltype(hosts_) hosts;
     if (!prepareElectionRequest(voteReq, hosts)) {
+        // Suppose we have three replias A(leader), B, C, after A crashed,
+        // B, C will begin the election. B win, and send hb, C has gap with B
+        // and need the snapshot from B. Meanwhile C begin the lection,
+        // C will be Candidate, but becasue C is in WAITING_SNAPSHOT,
+        // so prepareElectionRequest will return false and go on the election.
+        // Becasue C is in Candidate, so it will reject the snapshot request from B.
+        // Infinite loop begins.
+        // So we neeed to back the follower to avoid the case.
```

src/kvstore/raftex/test/TestShard.h
```diff
@@ -174,21 +174,19 @@ class SnapshotManagerImpl : public SnapshotManager {
...
-                cb(std::move(data), totalCount, totalSize, false);
+                cb(std::move(data), totalCount, totalSize, SnapshotStatus::IN_PROGRESS);
                 data.clear();
             }
             auto encoded = encodeSnapshotRow(row.first, row.second);
             totalSize += encoded.size();
             totalCount++;
             data.emplace_back(std::move(encoded));
         }
-        if (data.size() > 0) {
```
darionyaphet: seems should not remove this check ?

dangleptr: We remove the if condition to ensure the callback function always being called in any case.

src/kvstore/raftex/RaftPart.cpp
```diff
@@ -1086,6 +1086,16 @@ bool RaftPart::leaderElection() {
...
+        std::lock_guard<std::mutex> g(raftLock_);
+        role_ = Role::FOLLOWER;
```
bright-starry-sky: got !

src/kvstore/raftex/RaftPart.cpp
```diff
@@ -1086,6 +1086,16 @@ bool RaftPart::leaderElection() {
     cpp2::AskForVoteRequest voteReq;
     decltype(hosts_) hosts;
     if (!prepareElectionRequest(voteReq, hosts)) {
+        // Suppose we have three replias A(leader), B, C, after A crashed,
+        // B, C will begin the election. B win, and send hb, C has gap with B
+        // and need the snapshot from B. Meanwhile C begin the lection,
```
liuyu85cn: typo

src/kvstore/raftex/SnapshotManager.cpp
```diff
@@ -42,7 +42,12 @@ folly::Future<Status> SnapshotManager::sendSnapshot(std::shared_ptr<RaftPart> pa
                                            std::vector<std::string>&& data,
                                            int64_t totalCount,
                                            int64_t totalSize,
-                                           bool finished) mutable {
+                                           SnapshotStatus status) mutable {
+            if (status == SnapshotStatus::FAILED) {
+                LOG(INFO) << part->idStr_ << "Snapshot send failed, the leader changed?";
```
liuyu85cn: if LOG(ERROR) better ?

dangleptr: ERROR means users should take care of it.   But here,  we don't need our users to do any special actions. Just dump the information is enough.

PR #1773, opened 2020-02-13
Fix get edge_type failed

laura-ding:


src/common/filter/Expressions.cpp
```diff
@@ -241,8 +241,7 @@ Status VariablePropertyExpression::prepare() {
 
 
 OptVariantType EdgeTypeExpression::eval(Getters &getters) const {
-    UNUSED(getters);
-    return *alias_;
+    return getters.getAliasProp(*alias_, *prop_);
```
bright-starry-sky: Be careful memory leaks. memory leaks occur when getters.getAliasProp == nullptr . right?

laura-ding: `getters ` needs to guarantee that getAliasProp is not null

src/graph/test/GoTest.cpp
```diff
@@ -406,15 +406,16 @@ TEST_P(GoTest, MULTI_EDGES) {
...
         auto &player = players_["Russell Westbrook"];
         auto query = folly::stringPrintf(fmt, player.vid());
         auto code = client_->execute(query, resp);
         ASSERT_EQ(cpp2::ErrorCode::SUCCEEDED, code);
-        std::vector<std::tuple<int64_t, int64_t>> expected = {
-            {0, players_["James Harden"].vid()},
-            {0, players_["Dejounte Murray"].vid()},
-            {0, players_["Paul George"].vid()},
+        std::vector<std::tuple<int64_t, int64_t, int64_t, int64_t>> expected = {
+            {0, players_["James Harden"].vid(), 0, -5},
```
CPWstatic: I remember back in the day that the `_type` would return a name which is a string. And i think it is not helpful to users if we return an integer. Maybe it is better to return the type name.

laura-ding: Good catch! but the type of propName is EdgeType, the type is **INT**,  there are many places will get it, if change the type of it, need to change the type in the schema, it may bring other problems.

src/graph/test/GoTest.cpp
```diff
@@ -832,15 +834,15 @@ TEST_P(GoTest, ReferenceVariableInYieldAndWhere) {
...
+        auto *fmt = "GO FROM %ld OVER serve YIELD $^.test";
         auto query = folly::stringPrintf(fmt, players_["Tim Duncan"].vid());
         auto code = client_->execute(query, resp);
         ASSERT_NE(cpp2::ErrorCode::SUCCEEDED, code);
     }
 
     {
         cpp2::ExecutionResponse resp;
-        auto *fmt = "GO FROM %ld OVER serve yield serve.test";
+        auto *fmt = "GO FROM %ld OVER serve YIELD serve.test";
```
Shylock-Hg: What's the purpose?

src/common/filter/Expressions.cpp
```diff
@@ -137,6 +137,9 @@ std::string AliasPropertyExpression::toString() const {
 }
 
 OptVariantType AliasPropertyExpression::eval(Getters &getters) const {
+    if (getters.getAliasProp == nullptr) {
+        return Status::Error("`getAliasProp' function is not implemented");
+    }
```
bright-starry-sky: Like this.

src/graph/test/GoTest.cpp
laura-ding: Uniform keyword uppercase

PR #1774, opened 2020-02-13
Check TTL before creating index

panda-sheep:
```
It mainly includes the following parts
1)    use ttl when create tag/edge,  drop ttl when alter tag/edge
       not use ttl when create tag/edge ,  add ttl when alter tag/edge
2)     tag/edge contains index,   not  used  ttl when alter tag/edge
3)     tag/edge uses ttl,   not  create index
4)     add the above unit tests

5)    change columns are not allowed when the column is ttl_col
```

### DIFF
--- a/src/meta/MetaServiceUtils.cpp
+++ b/src/meta/MetaServiceUtils.cpp
@@ -423,15 +423,15 @@ cpp2::ErrorCode MetaServiceUtils::alterColumnDefs(std::vector<nebula::cpp2::Colu
             for (auto it = cols.begin(); it != cols.end(); ++it) {
                 if (col.get_name() == it->get_name()) {
                     // Check if there is a TTL on the column to be deleted
-                    if (!prop.get_ttl_col() ||
-                        (prop.get_ttl_col() && (*prop.get_ttl_col() != col.get_name()))) {
-                        cols.erase(it);
-                        return cpp2::ErrorCode::SUCCEEDED;
-                    } else {
-                        LOG(ERROR) << "Column can't be dropped, a TTL attribute on it: "
-                                   << col.get_name();
-                        return cpp2::ErrorCode::E_NOT_DROP;
+                    // drop the column with ttl
+                    //if (!prop.get_ttl_col() ||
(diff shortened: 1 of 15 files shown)


src/meta/MetaServiceUtils.cpp
```diff
@@ -423,15 +423,15 @@ cpp2::ErrorCode MetaServiceUtils::alterColumnDefs(std::vector<nebula::cpp2::Colu
...
-                        (prop.get_ttl_col() && (*prop.get_ttl_col() != col.get_name()))) {
-                        cols.erase(it);
-                        return cpp2::ErrorCode::SUCCEEDED;
-                    } else {
-                        LOG(ERROR) << "Column can't be dropped, a TTL attribute on it: "
-                                   << col.get_name();
-                        return cpp2::ErrorCode::E_NOT_DROP;
+                    // drop the column with ttl
+                    // if (!prop.get_ttl_col() ||
+                    //    (prop.get_ttl_col() && (*prop.get_ttl_col() != col.get_name()))) {
```
bright-starry-sky: Extra comments.

panda-sheep: Yeah, good catch

src/meta/processors/BaseProcessor.h
```diff
@@ -226,12 +233,21 @@ class BaseProcessor {
...
+     * Fetch the latest version edge's schema.
+     */
+    StatusOr<nebula::cpp2::Schema>
+    getLatestEdgeSchema(GraphSpaceID spaceId, const EdgeType edgeType);
+
+    /**
+     * Check if edge has ttl
+     */
+    StatusOr<bool>
+    edgeHasTTL(GraphSpaceID spaceId, const EdgeType edgeType);
```
dangleptr: Do we need the method?

panda-sheep: This function is deprecated. I remember it has been removed.

panda-sheep:
The regression test issue has been fixed（#1779）, please ignore it here

PR #1424, opened 2019-12-11
Add Job manager

liuyu85cn:
# add job manager to support admin jobs.
### submit job [flush/compact]
        add a"flush" job in the chosen space, return a job id.
        this job will run in async.
        use "show job <id>" to check the status 
        there are two job type now (flush and compact)

### show jobs
        list all the not expired jobs*

### show job [id]
        show a job and all its tasks

### stop job [id]
        stop a job if not finished

### recover job
        after fail over, use this command to put all the "queue" back into queue again.

### Comments
* Job:
   a user input command. Send from graph to Meta

* Task.
   after a 'job' arrived to Meta. Meta will split the job to tasks, and send them to storage.
  
* expired jobs 
   default 1 week, controlled by meta FLAGs "job_expired_secs"

PR #1777, opened 2020-02-14
fix potential wal bug in ut

critical27:
https://github.com/vesoft-inc/nebula/pull/1424/checks?check_run_id=445093798

Let me try to clarify what happens:
1. A is leader of term 5, append some logs
2. A is killed and rebooted, lastLogId = 42
3. A is elected as leader of term 6.
image

However, A keeps saying `There is a gap in the log id. The last log id is 43, and the id being appended is 43`, ant test case failed.

The problem is that we only `stop` and `start` the part in test, so when we reboot a part, the `lastLogId` could be previous result when it was killed. 

So I reload the last log id in wal. I think we don't need to scan wal files here in test, since we don't actually kill it. In reality, whenever we kill a node and restart, we will always scan all files in constructor.

### DIFF
--- a/src/kvstore/raftex/RaftPart.cpp
+++ b/src/kvstore/raftex/RaftPart.cpp
@@ -235,8 +235,6 @@ RaftPart::RaftPart(ClusterID clusterId,
                                                                logClusterId,
                                                                log);
                                 });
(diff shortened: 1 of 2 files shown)


dangleptr:
Very good analyzation

PR #1779, opened 2020-02-14
fix unstable testcase index_scan_test

bright-starry-sky:
The test case `index_scan_test` often failing on CI system.
The failing reason is : all parts need to elect leaders before data insert. 
image

### DIFF
--- a/src/storage/test/IndexScanTest.cpp
+++ b/src/storage/test/IndexScanTest.cpp
@@ -16,6 +16,8 @@
 #include "dataman/RowSetReader.h"
 #include "dataman/RowReader.h"
 
+DECLARE_uint32(raft_heartbeat_interval_secs);
+
 namespace nebula {
 namespace storage {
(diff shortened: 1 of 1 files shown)


src/storage/test/IndexScanTest.cpp
```diff
@@ -371,6 +373,7 @@ static cpp2::LookUpEdgeIndexResp checkLookupEdgesString(const std::string& filte
     }
     std::unique_ptr<meta::IndexManager> indexMan(im);
     auto eindex = indexMan->getEdgeIndex(spaceId, type).value();
+    sleep(FLAGS_raft_heartbeat_interval_secs);
```
dangleptr: why sleep here?

bright-starry-sky: > why sleep here?

Execute the data insert immediately when the schema created just done. some times individual part have not ready yet , so using sleep waiting a heartbeat let to make partitioned folder and waiting part leader ready.

dangleptr: OK

PR #1782, opened 2020-02-16
charset compare and  display

### LINKED ISSUE #1400: Characters indent not correctly always.
**Describe the bug(__must be provided__)**
The table output not indent correctly if contains wide-size characters.
(user@127.0.0.1) [test]> fetch prop on building uuid('pin_测试') yield building.name
=================
| building.name |
=================
| 测试        |
-----------------
Got 1 rows (Time spent: 6454/7737 us)

**Your Environments(__must be provided__)**
  * OS: Linux hp-server 4.18.16-300.fc29.x86_64 #1 SMP Sat Oct 20 23:24:08 UTC 2018 x86_64 x86_64 x86_64 GNU/Linux

  * Compliler: 
g++ (GCC) 8.2.0
Copyright (C) 2018 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.

  * CPU: 
Architecture:        x86_64
CPU op-mode(s):      32-bit, 64-bit
Byte Order:          Little Endian
CPU(s):              40
On-line CPU(s) list: 0-39
Thread(s) per core:  2
Core(s) per socket:  10
Socket(s):           2
NUMA node(s):        2
Vendor ID:           GenuineIntel
CPU family:          6
Model:               62
Model name:          Intel(R) Xeon(R) CPU E5-2690 v2 @ 3.00GHz
Stepping:            4
CPU MHz:             1629.598
CPU max MHz:         3600.0000
CPU min MHz:         1200.0000
BogoMIPS:            5985.55
Virtualization:      VT-x
L1d cache:           32K
L1i cache:           32K
L2 cache:            256K
L3 cache:            25600K
NUMA node0 CPU(s):   0-9,20-29
NUMA node1 CPU(s):   10-19,30-39
Flags:               fpu vme de pse tsc msr pae mce cx8 apic sep mtrr pge mca cmov pat pse36 clflush dts acpi mmx fxsr sse sse2 ss ht tm pbe syscall nx pdpe1gb rdtscp lm constant_tsc arch_perfmon pebs bts rep_good nopl xtopology nonstop_tsc cpuid aperfmperf pni pclmulqdq dtes64 monitor ds_cpl vmx smx est tm2 ssse3 cx16 xtpr pdcm pcid dca sse4_1 sse4_2 x2apic popcnt tsc_deadline_timer aes xsave avx f16c rdrand lahf_lm cpuid_fault epb pti ssbd ibrs ibpb stibp tpr_shadow vnmi flexpriority ept vpid fsgsbase smep erms xsaveopt dtherm ida arat pln pts flush_l1d

Fetch the result contains wide-size characters.

**Expected behavior**
Indent correctly based on characters instead of bytes.

**Additional context**
Maybe we need support the inner encoding format for characters such as utf-8, ANSI etc.

### LINKED ISSUE #1666: Supports multiple character sets
1) Supports multiple character sets. By default, each character set supports collation in binary mode.
2) The default character set is utf8, and in nebula, utf8 is an alias of utf8mb4, so the default collation is utf8_bin
3) You can specify the character set and collation when creating a space. If it is not specified, the default character set and collation are used.

The following three create space statements are equivalent:

```
create space my_space (partition_num = 10, replica_factor = 1, charset = utf8, collate = utf8_bin);
create space my_space (partition_num = 10, replica_factor = 1, charset = utf8);
create space my_space (partition_num = 10, replica_factor = 1);
```

panda-sheep:
1.  order by and RelationalExpression evaluation use current space  collate to compare
2.  cache the charset information in meta client
3. Result display in console shoud be based on the number of bytes in the current space charset.(fix 
 #1400)

close #1666


panda-sheep:
There is a problem with the current regression test environment and the supported character set is incomplete. as shown below.
image
https://github.com/vesoft-inc/nebula-dev-docker/issues/33
we need to support the full character set

panda-sheep:
There is a problem with the current regression test environment and the supported character set is incomplete. 
https://github.com/vesoft-inc/nebula-dev-docker/issues/33
we need to support the full character set

src/common/charset/Charset.cpp
```diff
@@ -0,0 +1,178 @@
...
+        return true;
+    } else {
+        return false;
+    }
+}
+
+
+StatusOr<bool>
+CharsetInfo::nebulaStrCmpGT(const std::string& collateName, const char* p1, const char* p2) {
+    auto ret = nebulaStrCmpLE(collateName, p1, p2);
```
liuyu85cn: what if just return  nebulaStrCmpLE(collateName, p2, p1) ?

src/common/charset/Charset.cpp
```diff
@@ -0,0 +1,196 @@
...
+            len = 1;
+        }
+        length++;
+    }
+    return length;
+}
+
+
+StatusOr<int>
+CharsetInfo::nebulaStrCmp(const std::string& collateName, const char* p1, const char* p2) {
```
liuyu85cn: maybe check if p1 == nullptr ?

src/common/charset/Charset.cpp
```diff
@@ -0,0 +1,196 @@
...
+            len = 3;
+        } else if (byte >= 0xC0) {
+            len = 2;
+        } else {
+            len = 1;
+        }
+        length++;
+    }
+    return length;
+}
```
bright-starry-sky: A question, this method same with MultiByteToWideChar?

src/common/charset/Charset.cpp
bright-starry-sky: A question, why parameter was defined `const char*` not `const std::string& `?

src/common/charset/Charset.cpp
```diff
@@ -0,0 +1,196 @@
...
+    auto ret = nebulaStrCmp(collateName, p1, p2);
+    if (!ret.ok()) {
+        return ret.status();
+    }
+    auto val = ret.value();
+    if (val <= 0) {
+        return true;
+    } else {
+        return false;
+    }
```
bright-starry-sky: Is better return val <= 0？

src/common/charset/test/CharsetTest.cpp
```diff
@@ -0,0 +1,93 @@
+/* Copyright (c) 2020 vesoft inc. All rights reserved.
+ *
+ * This source code is licensed under Apache 2.0 License,
+ * attached with Common Clause Condition 1.0, found in the LICENSES directory.
+ */
+
+#include "base/Base.h"
```
bright-starry-sky: Redundant definition.

src/common/filter/Expressions.h
```diff
@@ -37,6 +37,8 @@ struct Getters {
 
 class ExpressionContext final {
 public:
+    explicit ExpressionContext(std::string spaceCollate) : spaceCollate_(spaceCollate) {}
```
bright-starry-sky: A suggestion, set a default value of param spaceCollate.

src/common/charset/Charset.cpp
```diff
@@ -0,0 +1,196 @@
...
+
+#include "charset/Charset.h"
+
+namespace nebula {
+
+std::unordered_set<std::string> CharsetInfo::supportCharset = {"utf8"};
+
+std::unordered_set<std::string> CharsetInfo::supportCollation = {"utf8_bin"};
+
+std::unordered_map<std::string, CharsetInfo::CharsetToCollation> CharsetInfo::charsetToCollation = {
```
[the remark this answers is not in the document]
panda-sheep: Yeah， I will add comments about this file.
this map key is charset name
               value is charset description  information.

src/common/charset/Charset.cpp
```diff
@@ -0,0 +1,196 @@
...
+    {"utf8_bin", "en_US.UTF-8"}
+};
+
+Status CharsetInfo::isSupportCharset(const std::string& charsetName) {
+    for (auto& sc : supportCharset) {
+        if (!sc.compare(charsetName)) {
+            return Status::OK();
+        }
+    }
+    return Status::Error("Charset `%s' not support", charsetName.c_str());
```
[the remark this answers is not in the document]
panda-sheep: Yeah.

src/common/charset/Charset.cpp
panda-sheep: This function counts how many characters are in the string according to the UTF8 character encoding

src/common/charset/Charset.cpp
panda-sheep: Yeahc， check it out more safely.
The result of the string c_str function will not be nullptr.

panda-sheep: Yeah,  good catch.

src/common/charset/Charset.cpp
```diff
@@ -0,0 +1,196 @@
...
+    auto ret = nebulaStrCmp(collateName, p1, p2);
+    if (!ret.ok()) {
+        return ret.status();
+    }
+    auto val = ret.value();
+    if (val < 0) {
+        return true;
+    } else {
+        return false;
+    }
```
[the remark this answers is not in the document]
panda-sheep: Good catch.

src/common/charset/Charset.cpp
panda-sheep: Yeah,  this is equivalent

src/common/charset/test/CharsetTest.cpp
panda-sheep: Yeah， good catch

src/common/filter/Expressions.h
panda-sheep: When default values are set, no parameters are passed during construction, which may result in different charset.

src/common/charset/Charset.cpp
```diff
@@ -0,0 +1,196 @@
...
+
+    // For character set is incomplete in locale
+    std::locale loc(iter->second);
+    if (!std::has_facet<std::ctype<char>>(loc)) {
+        return Status::Error("Locale: %s  environment is not supported", iter->second.c_str());
+    }
+    auto& f = std::use_facet<std::collate<char>>(loc);
+
+    std::string str1(p1), str2(p2);
```
liuyu85cn: how about also provide a 
CharsetInfo::nebulaStrCmp(const std::string& collateName, const std::string& p1, const std::string& p2);
overload ?

src/console/CmdProcessor.cpp
```diff
@@ -264,9 +285,17 @@ void CmdProcessor::printHeader(
 
     size_t idx = 0;
     for (auto& cname : (*resp.get_column_names())) {
-        std::string fmt = folly::stringPrintf(" %%-%lds |", widths[idx++]);
+        auto len = widths[idx++];
+        if (curSpaceCharset_.compare("utf8") == 0) {
+            auto byteLen = cname.size();
+            auto charLen = CharsetInfo::getUtf8Charlength(cname);
+            auto diff = (byteLen - charLen) / 2;
+            len += diff;
```
critical27: Why we add diff here but minus diff in `calColumnWidths`?

src/graph/ClientSession.h
```diff
@@ -31,15 +32,28 @@ class ClientSession final {
         return space_;
     }
 
-    void setSpace(const std::string &name, GraphSpaceID space) {
+    void setSpace(const std::string &name,
+                  GraphSpaceID space,
+                  const std::string &charsetName,
```
critical27: Maybe we could use `FLAGS_default_charset` as default value

src/graph/ShowExecutor.cpp
```diff
@@ -79,6 +81,12 @@ void ShowExecutor::execute() {
         case ShowSentence::ShowType::kShowSnapshots:
             showSnapshots();
             break;
+        case ShowSentence::ShowType::kShowCharset:
```
critical27: How about show them in `show spaces`?

src/common/charset/Charset.cpp
panda-sheep: Yeah，fixed

src/console/CmdProcessor.cpp
panda-sheep: One less byte to display when displaying a Chinese character

src/graph/ClientSession.h
panda-sheep: Good catch!

src/graph/ShowExecutor.cpp
panda-sheep: show spaces  not display charset,  only show create space XXX to display space charset.

panda-sheep:
> Some questions :
> 1, where is the conversion from the client character set to the server character set.
Now，default client charset and server charset are same，no conver.
> 2, If different Spaces defined different character sets, How to ensure the correct of data insert for different spaces?
 Not check the legality of the data,  the users are to  guarantee.
> 3, how to set the character collation in different user session?
charset collation does not set.  The use must be to guarantee  current operating environment charset、 space charset、 data encoding charset the same.

bright-starry-sky:
> Some questions :
> 1, where is the conversion from the client character set to the server character set.
> Now，default client charset and server charset are same，no conver.
> 2, If different Spaces defined different character sets, How to ensure the correct of data insert for different spaces?
> Not check the legality of the data, the users are to guarantee.
> 3, how to set the character collation in different user session?
> charset collation does not set. The use must be to guarantee current operating environment charset、 space charset、 data encoding charset the same.

got!

panda-sheep:
I  will format the code now.

dangleptr:
Rebase it please.

src/common/charset/Charset.cpp
```diff
@@ -59,4 +60,116 @@ StatusOr<std::string> CharsetInfo::getCharsetbyCollation(const std::string& coll
...
+    }
+
+    // For charset is incomplete in locale
+    std::locale loc(iter->second);
+    if (!std::has_facet<std::ctype<char>>(loc)) {
+        return Status::Error("Locale: `%s' environment is not supported", iter->second.c_str());
+    }
+    auto& f = std::use_facet<std::collate<char>>(loc);
+
+    std::string str1(p1), str2(p2);
```
dangleptr: Why copy it?

src/common/charset/Charset.h
```diff
@@ -86,6 +146,13 @@ class CharsetInfo final {
      * Description information of supported charsets
      */
     std::unordered_map<std::string, CharsetDesc> charsetDesc_;
+
+    /**
+     * Mapping collation to locale
+     */
+    std::unordered_map<std::string, std::string> collationToLocale_ = {
```
dangleptr: We should support utf8_general_ci  at least.

panda-sheep: My test results show that the Chinese sorting results are the same in both utf8_bin and utf8_general_ci. utf8_general_ci is just a sort that ignores case.
I personally feel that there is no use case for utf8_general_ci?

src/common/charset/Charset.cpp
panda-sheep: Good catch
Copying is worried that it will be modified during compare. In fact, It is not necessary to copy the string here.

src/common/charset/Charset.h
panda-sheep: Fixed，We currently support two charset:
```
utf8
utf8mb4
```
We have extended utf8 to handle 4-byte characters, so
utf8 and utfmb4 are equivalent

Four collations are supported:
```
utf8_bin
utf8_general_ci
utf8mb4_bin
utf8mb4_general_c
```i

panda-sheep:
pr is ok, request a review.

dangleptr:
We need to reimplement it in 2.0.  So close it now.

PR #1785, opened 2020-02-18
Push down filters in reversely or bidirectional traverse.

CPWstatic:


dangleptr:
This is a great improvement.  Do we want it in 1.x,  IMO we should accept it. 
@jude-zhu @whitewum @CPWstatic

src/storage/query/QueryBaseProcessor.inl
```diff
@@ -581,6 +580,42 @@ kvstore::ResultCode QueryBaseProcessor<REQ, RESP>::collectEdgeProps(
...
+                    }
+                    VLOG(1) << "Hit srcProp filter for tag " << tag << ", prop "
+                            << prop << ", value " << it->second;
+                    return it->second;
+                };
+                auto value = exp_->eval(getters);
+                if (value.ok() && !Expression::asBool(value.value())) {
+                    VLOG(1) << "Filter the edge "
+                            << vId << "-> " << dstId << "@" << rank << ":" << edgeType;
+                    continue;
```
critical27, line 616: if `value` is not ok, need to continue as well

PR #1786, opened 2020-02-18
fix some problems during balance

critical27:
Fix some problems during test:
1. Previously we update when BalanceProcessor has return, but the part alloc probably would be updated later, so some machine won't `loadData` in MetaClient. So I update meta when a BalanceTask has done `updateMeta`.
2. When we `balance remove xxx`, the src of BalanceTask could be alive, but previously we regard them as dead, which would cause some part was not removed.
3. We should forbid `balance leader` during `balance data`.
4. We don't need to `refreshCache` when add a part.

### DIFF
--- a/src/meta/processors/admin/BalanceTask.cpp
+++ b/src/meta/processors/admin/BalanceTask.cpp
@@ -152,14 +152,6 @@ void BalanceTask::invoke() {
                 if (!resp.ok()) {
                     LOG(INFO) << taskIdStr_ << "Update meta failed, status " << resp;
                     ret_ = Result::FAILED;
-                } else if (kv_ != nullptr) {
-                    if (LastUpdateTimeMan::update(kv_, time::WallClock::fastNowInMilliSec()) !=
-                            kvstore::ResultCode::SUCCEEDED) {
-                        LOG(INFO) << taskIdStr_ << "Update meta failed";
-                        ret_ = Result::FAILED;
-                    } else {
-                        status_ = Status::REMOVE_PART_ON_SRC;
-                    }
                 } else {
                     status_ = Status::REMOVE_PART_ON_SRC;
                 }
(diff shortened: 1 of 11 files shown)


src/meta/processors/admin/BalanceTask.cpp
```diff
@@ -152,6 +152,14 @@ void BalanceTask::invoke() {
                 if (!resp.ok()) {
                     LOG(INFO) << taskIdStr_ << "Update meta failed, status " << resp;
                     ret_ = Result::FAILED;
+                } else if (kv_ != nullptr) {
+                    if (LastUpdateTimeMan::update(kv_, time::WallClock::fastNowInMilliSec()) !=
```
dangleptr: Each task finished will result in the meta info updated?

critical27: Do you have better suggestion? How about when BalancePlan::invoke is done?

dangleptr: Sounds good.

src/meta/processors/partsMan/ListHostsProcessor.cpp
```diff
@@ -74,7 +74,7 @@ Status ListHostsProcessor::allHostsWithStatus() {
     }
 
     // get hosts which have send heartbeat recently
-    auto activeHosts = ActiveHostsMan::getActiveHosts(kvstore_, FLAGS_heartbeat_interval_secs + 1);
+    auto activeHosts = ActiveHostsMan::getActiveHosts(kvstore_, FLAGS_heartbeat_interval_secs * 2);
```
darionyaphet: Actually I'm not sure why heartbeat_interval_secs * 2 after talk with dooble offline, maybe more safe

dangleptr: Yes,  the purpose is to make the active hosts  returned more stable.

PR #1792, opened 2020-02-20
Remove ignore_index_check_pre_insert flag

darionyaphet:


src/storage/mutate/AddEdgesProcessor.cpp
```diff
@@ -88,17 +88,14 @@ std::string AddEdgesProcessor::addEdges(int64_t version, PartitionID partId,
         newEdges[key] = std::move(prop);
     });
     for (auto& e : newEdges) {
-        std::string val;
+        auto val = findObsoleteIndex(partId, e.first);
```
bright-starry-sky: Here is a performance bug.

src/storage/mutate/AddVerticesProcessor.cpp
```diff
@@ -104,18 +104,15 @@ std::string AddVerticesProcessor::addVertices(int64_t version, PartitionID partI
     });
 
     for (auto& v : newVertices) {
-        std::string val;
         std::unique_ptr<RowReader> nReader;
         auto tagId = NebulaKeyUtils::getTagId(v.first);
         auto vId = NebulaKeyUtils::getVertexId(v.first);
+        auto val = findObsoleteIndex(partId, vId, tagId);
```
bright-starry-sky: Here is a performance bug.

PR #1794, opened 2020-02-20
Return leader changed when meta has multiple replica

critical27:
During balance test with multiple replica meta, sometimes the `balance` only print previous balance id. At last I realize that we don't return `leader change` could cause this problem.

1. I move `cpp2::ErrorCode to(kvstore::ResultCode code)` to `MetaCommon` class, since not processor need them, e.g. (Balancer and BalancePlan)
2. Add a `handleErrorCode` just like what we did in storage, so each processor could call this method to set result code.
3. Reject `balance data` if meta is not leader. (I think we don't need to reject `balance leader`)
4. Fix we set `Result::SUCCEEDED` twice in `BalanceTask`


Shylock-Hg:
Close https://github.com/vesoft-inc/nebula/pull/1526

PR #1814, opened 2020-02-25
Job Manager judged some tasks failed. However they can run success manually by curl. 

liuyu85cn:
Job Manager judged some tasks as failed. However they can run success manually. 

I guess this may caused by different output of curl. 

print more info if this happen again. and why job manager think those tasks failed.


bright-starry-sky:
A little suggestion,  we could provide some test cases for job failing handler.
Of course, you can do it in other PR.

PR #1818, opened 2020-02-25
fix makedir failed when dir is SYM_LINK

### LINKED ISSUE #1802: can't recognize soft link directory
as /usr/local/nebula/ has very limited disk space, I try to make a soft link from other space (/ebs/) and make a soft link to /usr/local/nebula/data/. (I know how to write in storage.conf for multi-disk).

But the service can't start. I think it is due to some errors about soft link directory

### LINKED ISSUE #1812: data目录如果是软链接的将会无法创建默认的目录导致服务挂了
软链接正常，手动使用mkdir创建所需要的目录服务才能正常运行。

laura-ding:
close #1812 
close #1802

### DIFF
--- a/src/kvstore/RocksEngine.cpp
+++ b/src/kvstore/RocksEngine.cpp
@@ -99,7 +99,13 @@ RocksEngine::RocksEngine(GraphSpaceID spaceId,
         , dataPath_(folly::stringPrintf("%s/nebula/%d", dataPath.c_str(), spaceId)) {
     auto path = folly::stringPrintf("%s/data", dataPath_.c_str());
     if (FileUtils::fileType(path.c_str()) == FileType::NOTEXIST) {
-        FileUtils::makeDir(path);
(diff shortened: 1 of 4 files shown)


src/kvstore/RocksEngine.cpp
```diff
@@ -99,7 +99,13 @@ RocksEngine::RocksEngine(GraphSpaceID spaceId,
...
     if (FileUtils::fileType(path.c_str()) == FileType::NOTEXIST) {
-        FileUtils::makeDir(path);
+        if (!FileUtils::makeDir(path)) {
+            LOG(FATAL) << "makeDir " << path << " failed";
+        }
+    }
+
+    if (FileUtils::fileType(path.c_str()) != FileType::DIRECTORY) {
+        LOG(FATAL) << path << " is not directory";
     }
```
dutor: The result of `FileUtils::fileType` could be saved, instead of calling twice.

PR #1822, opened 2020-02-26
Stop scanning the data if sending snapshot failed

dangleptr:
1. If sending snapshot failed,  we should stop scanning the data,  otherwise it may send lots of large but useless requests to learner,  it is a risk to cause OOM.

2. Don't malloc memory each time for sending snapshot,  it hurt the performance of the normal service.  Of course,  there is an extra string-copy here,  but i think it is deserved


src/kvstore/raftex/RaftPart.cpp
```diff
@@ -271,13 +271,13 @@ void RaftPart::start(std::vector<HostAddr>&& peers, bool asLearner) {
...
     lastLogId_ = wal_->lastLogId();
     lastLogTerm_ = wal_->lastLogTerm();
+    term_ = proposedTerm_ = lastLogTerm_;
 
     // Set the quorum number
     quorum_ = (peers.size() + 1) / 2;
 
     auto logIdAndTerm = lastCommittedLogId();
     committedLogId_ = logIdAndTerm.first;
-    term_ = proposedTerm_ = logIdAndTerm.second;
```
critical27: Do we need to set lastLogTerm_ to committedLogTerm when `lastLogId_ < committedLogId_`, on line 286

src/kvstore/raftex/SnapshotManager.cpp
```diff
@@ -71,34 +71,34 @@ folly::Future<Status> SnapshotManager::sendSnapshot(std::shared_ptr<RaftPart> pa
...
                             p.setValue(Status::OK());
                         }
-                        return;
+                        return true;
                     }
                 } catch (const std::exception& e) {
                     LOG(ERROR) << part->idStr_ << "Send snapshot failed, exception " << e.what();
                     p.setValue(Status::Error("Send snapshot failed!"));
-                    return;
+                    return false;
```
critical27: Do we need to return false when we fail only once? Or you think the timeout is big enough?

src/kvstore/raftex/RaftPart.cpp
dangleptr: Good point.

src/kvstore/raftex/SnapshotManager.cpp
dangleptr: Make sense.  Maybe we should add a timeout control and retry when catching exceptions.

src/kvstore/NebulaStore.cpp
```diff
@@ -101,7 +101,14 @@ bool NebulaStore::init() {
                             enginePtr->removePart(partId);
                             continue;
                         } else {
-                             partIds.emplace_back(partId);
+                             auto it = std::find(partIds.begin(), partIds.end(), partId);
+                             if (it != partIds.end()) {
+                                LOG(INFO) << "Part " << partId
```
critical27: When could this happen?

dangleptr: I am not sure how it happens.  But i notice one part be opened twice on the same host in our user's logs.

PR #1823, opened 2020-02-26
Meta compaction filter

### LINKED ISSUE #285: Implement compaction filter for meta server.
For example,  after some space has been dropped,  we should delete all edges/tags under it in compaction.

critical27:
1. Add meta compaction filter, when we drop a space, we can filter the tag/edge/part/... of the space when we compact.
2. Use a MetaCFHelper to get all spaces when we begin to compact.
3. Add a MetaHttpAdminHandler to `compact` and `flush` manually.

close #285

### DIFF
--- a/src/daemons/MetaDaemon.cpp
+++ b/src/daemons/MetaDaemon.cpp
@@ -8,6 +8,7 @@
 #include "common/base/SignalHandler.h"
 #include <thrift/lib/cpp2/server/ThriftServer.h>
 #include "meta/MetaServiceHandler.h"
+#include "meta/MetaHttpAdminHandler.h"
 #include "meta/MetaHttpIngestHandler.h"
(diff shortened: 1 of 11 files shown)


critical27:
> A question, why do compaction in meta server? I think the meta data can be delete via each processor.

Yeah, you are right, but it seems we have not delete them, especially after `drop space`. This PR just provide another way, at least we don't have to delete them everywhere.

critical27:
> A question, why do compaction in meta server? I think the meta data can be delete via each processor.

Yeah, you are right, but it seems we have not delete them, especially after `drop space`. This PR just provide another way, at least we don't have to delete them everywhere.

dangleptr:
Let's reimplement it in 2.0 if needed.  So close this pr.

PR #1826, opened 2020-02-26
Fix meta would crash when balance in multiple relica

### LINKED ISSUE #1820: Failed to commit logs  导致nebula-storaged服务挂了
具体logs已经发给开发人员。

critical27:
1. As title, introduced in #1786, when recovery failed, the plan would be nullptr.
2. Add an error log when commit failed.

close #1820

### DIFF
--- a/src/meta/processors/admin/Balancer.h
+++ b/src/meta/processors/admin/Balancer.h
@@ -106,10 +106,6 @@ public:
 
     void finish() {
         CHECK(!lock_.try_lock());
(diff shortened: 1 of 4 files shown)


src/meta/processors/admin/Balancer.h
```diff
@@ -106,10 +106,6 @@ class Balancer {
 
     void finish() {
         CHECK(!lock_.try_lock());
```
bright-starry-sky: A question，Is the useful now of lock?

critical27: Yes, we need to make sure we hold the lock whenever we want to call `finish`

Issue #1831, opened 2020-02-26
Provide best conf practices for production environment and test environment 


jude-zhu:
As titled

PR #1833, opened 2020-02-26
Add config for AWS EC2 c5d.xlarge

whitewum:

I think an AWS EC2 config will be more comparable/illustrative for users to set conf file.

These are conf file examples. I've been asked many times of recommended configs.
AWS EC2 is typical. AWS EC2 xlarge is for minimal & 12xlarge is for standard usage.

No.

Not necessary.

### DIFF
--- a/etc/nebula-storage.conf.c5d.xlarge
+++ b/etc/nebula-storage.conf.c5d.xlarge
@@ -1,0 +1,60 @@
+########## bench ########### 
+# AWS EC2 c5d.xlarge
+# vCPU 4
+# memory 8 GiB
+# Disk 100GiB NVMe SSD
+
+########## basics ##########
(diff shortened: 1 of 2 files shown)


whitewum:
refer to #1831

etc/nebula-storage.conf.c5d.xlarge
```diff
@@ -1,3 +1,9 @@
+########## bench ###########
```
dangleptr, line 1: Why rename conf.production to conf.c5d.xlarge

darionyaphet:
Maybe we should have another config file to adapt EC2？

whitewum:
> Maybe we should have another config file to adapt EC2？

Agree. I'll close this one.

PR #1834, opened 2020-02-26
Check if props exists when go or fetch.

### LINKED ISSUE #1816: [bug] edge property should check schema first
```
GO FROM -44342802513 OVER i2i  YIELD i2i.view_cf
[ERROR (-8)]: Get neighbors failed
Tue Feb 25 15:11:54 2020
```
The property should be i2i.view_cf_item not view_cf. Should check schema before go to storage.

CPWstatic:
close #1816


src/graph/FetchEdgesExecutor.cpp
```diff
@@ -106,6 +112,32 @@ Status FetchEdgesExecutor::prepareEdgeKeys() {
...
+            return Status::SyntaxError(
+                "Near [%s.%s], edge should be declared in `ON' clause first.",
+                    pair.first.c_str(), pair.second.c_str());
+        }
+        auto propName = pair.second;
+        if (propName == _SRC || propName == _DST
+                || propName == _RANK || propName == _TYPE) {
+            continue;
+        }
+        auto es = ectx()->schemaManager()->getEdgeSchema(spaceId_, edgeType_);
```
laura-ding: use `labelSchema_`

src/graph/FetchVerticesExecutor.cpp
```diff
@@ -92,6 +97,27 @@ Status FetchVerticesExecutor::prepareVids() {
...
+Status FetchVerticesExecutor::checkTagProps() {
+    auto aliasProps = expCtx_->aliasProps();
+    for (auto &pair : aliasProps) {
+        if (pair.first != *labelName_) {
+            return Status::SyntaxError(
+                "Near [%s.%s], tag should be declared in `ON' clause first.",
+                    pair.first.c_str(), pair.second.c_str());
+        }
+
+        auto ts = ectx()->schemaManager()->getTagSchema(spaceId_, tagId_);
```
laura-ding: ditto

dangleptr:
what about "update", "delete" etc.

src/graph/FetchVerticesExecutor.cpp
```diff
@@ -92,6 +97,23 @@ Status FetchVerticesExecutor::prepareVids() {
     return Status::OK();
 }
 
+Status FetchVerticesExecutor::checkTagProps() {
+    auto aliasProps = expCtx_->aliasProps();
```
dangleptr: We'd better add a validate function inside Expression to do that.

CPWstatic: The `AliasProps` have different meanings in different statements. We don't know if it is a tag or edge., so we can't validate it inside the expression.

PR #1840, opened 2020-02-28
support return partial result in get of kv interfaces

critical27:
1. Add a return_partly in get, when it is set to true, we can get the keys which exist, and ignore those keys which are not existed.
2. To implement this, add a `tryGet` method in KVStore, which return each status of key.
3. Remove async thread in `GetProcessor`


src/kvstore/KVStore.h
```diff
@@ -83,6 +84,13 @@ class KVStore {
                                 const std::vector<std::string>& keys,
                                 std::vector<std::string>* values) = 0;
 
+    // Read multiple keys, if key[i] does not exist, the i-th value in return value
+    // would be Status::KeyNotFound
+    virtual ErrorOr<ResultCode, std::vector<Status>> tryGet(GraphSpaceID spaceId,
```
dangleptr: Why not reuse multiGet directly?

critical27: Because we need to return the Status, so we can judge whether the key exists. If we reuse `multiGet`, the value of key_not_exist would be empty, it is hard to tell whether the value is really empty or the key doesn't exist.

dangleptr: I mean modify the multiGet interface directly...

critical27: ok

src/kvstore/NebulaStore.cpp
```diff
@@ -359,10 +359,12 @@ ResultCode NebulaStore::get(GraphSpaceID spaceId,
...
-ResultCode NebulaStore::multiGet(GraphSpaceID spaceId,
-                                 PartitionID partId,
-                                 const std::vector<std::string>& keys,
-                                 std::vector<std::string>* values) {
+ErrorOr<ResultCode, std::vector<Status>> NebulaStore::multiGet(
+        GraphSpaceID spaceId,
+        PartitionID partId,
+        const std::vector<std::string>& keys,
+        std::vector<std::string>* values,
+        bool returnPartly) {
```
dangleptr: Do we really need the param here? 
If only partial results returned, we could return ResultCode::PARTIAL_RESULTS;

critical27: I think we'd better keep it, if we return `ResultCode::PARTIAL_RESULTS` we can't tell the difference between `key not exists` and `value is empty`,

critical27:
Hold on a little, I will update ut.

critical27:
Ready to merge

PR #1842, opened 2020-03-01
User manager for graph layer

bright-starry-sky:
User manager for graph layer.


src/graph/UserExecutor.cpp
```diff
@@ -0,0 +1,185 @@
...
+
+CreateUserExecutor::CreateUserExecutor(Sentence *sentence,
+                                       ExecutionContext *ectx)
+    : Executor(ectx, "create_user") {
+    sentence_ = static_cast<CreateUserSentence*>(sentence);
+}
+
+Status CreateUserExecutor::prepare() {
+    DCHECK(sentence_ != nullptr);
+    return Status::OK();
```
panda-sheep: Now not use `DCHECK`   in  code

src/graph/UserExecutor.cpp
```diff
@@ -0,0 +1,185 @@
...
+
+void DropUserExecutor::execute() {
+    auto *mc = ectx()->getMetaClient();
+    auto *account = sentence_->getAccount();
+    auto future = mc->dropUser(*account, sentence_->ifExists());
+    auto *runner = ectx()->rctx()->runner();
+    auto cb = [this] (auto &&resp) {
+        if (!resp.ok()) {
+            doError(Status::Error("Drop user failed: %s.",
+                                         resp.status().toString().c_str()));
```
panda-sheep: Aligned

src/graph/UserExecutor.cpp
```diff
@@ -0,0 +1,185 @@
...
+AlterUserExecutor::AlterUserExecutor(Sentence *sentence,
+                                     ExecutionContext *ectx)
+    : Executor(ectx, "alter_user") {
+    sentence_ = static_cast<AlterUserSentence*>(sentence);
+}
+
+Status AlterUserExecutor::prepare() {
+    DCHECK(sentence_ != nullptr);
+    Status status = Status::OK();
+    return status;
```
panda-sheep: `return  Status::OK();`  is better？

src/graph/ShowExecutor.cpp
```diff
@@ -967,6 +970,184 @@ void ShowExecutor::showCollation() {
...
+        }
+
+        resp_->set_rows(std::move(rows));
+        doFinish(Executor::ProcessControl::kNext);
+    };
+
+    auto error = [this] (auto &&e) {
+        LOG(ERROR) << "Exception caught: " << e.what();
+        doError(Status::Error(folly::stringPrintf("Show tags exception: %s",
+                                                  e.what().c_str())));
```
panda-sheep: ditto

src/graph/ShowExecutor.cpp
```diff
@@ -967,6 +970,184 @@ void ShowExecutor::showCollation() {
...
+        resp_ = std::make_unique<cpp2::ExecutionResponse>();
+        std::vector<cpp2::RowValue> rows;
+        std::vector<std::string> header{"Account", "IsLock", "Login Type",
+                                        "MAX_QUERIES_PER_HOUR",
+                                        "MAX_UPDATES_PER_HOUR",
+                                        "MAX_CONNECTIONS_PER_HOUR",
+                                        "MAX_USER_CONNECTIONS"};
+        resp_->set_column_names(std::move(header));
+        for (auto& item : value) {
+            auto user = item.second;
```
critical27: It seems we could unify code with `showUser`?

src/graph/UserExecutor.cpp
```diff
@@ -0,0 +1,185 @@
...
+
+Status DropUserExecutor::prepare() {
+    DCHECK(sentence_ != nullptr);
+    return Status::OK();
+}
+
+void DropUserExecutor::execute() {
+    auto *mc = ectx()->getMetaClient();
+    auto *account = sentence_->getAccount();
+    auto future = mc->dropUser(*account, sentence_->ifExists());
```
critical27: Do we need to check the user privilege? I guess only user like GOD can drop/alter user

src/graph/ShowExecutor.cpp
```diff
@@ -967,6 +970,184 @@ void ShowExecutor::showCollation() {
...
+                                        "MAX_CONNECTIONS_PER_HOUR",
+                                        "MAX_USER_CONNECTIONS"};
+        resp_->set_column_names(std::move(header));
+        std::vector<cpp2::ColumnValue> row;
+        row.resize(7);
+        row[0].set_str(user.get_account());
+        row[1].set_bool_val(user.__isset.is_lock ? *user.get_is_lock() : false);
+        if (user.__isset.login_type) {
+            row[2].set_str(*user.get_login_type() == nebula::cpp2::UserLoginType::PASSWORD
+                           ? "PASSWORD" : "LDAP");
```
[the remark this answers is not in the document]
bright-starry-sky: > When there are many types of login authentication in the future, it is best to use a function to handle it here，for example `login_type_to_string`?

good suggest. improved.

src/graph/ShowExecutor.cpp
```diff
@@ -967,6 +970,184 @@ void ShowExecutor::showCollation() {
...
+        rows.back().set_columns(std::move(row));
+
+        resp_->set_rows(std::move(rows));
+        doFinish(Executor::ProcessControl::kNext);
+    };
+
+    auto error = [this] (auto &&e) {
+        LOG(ERROR) << "Exception caught: " << e.what();
+        doError(Status::Error(folly::stringPrintf("Show tags exception: %s",
+                                                  e.what().c_str())));
```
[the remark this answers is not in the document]
bright-starry-sky: > The error message is incorrect.

good catch.

src/parser/parser.yy
```diff
@@ -1772,24 +1787,25 @@ with_user_opt_list
...
-    | KW_CREATE KW_USER name_label KW_WITH KW_PASSWORD STRING COMMA with_user_opt_list {
-        auto sentence = new CreateUserSentence($3, $6);
-        sentence->setOpts($8);
+    ;
+
+create_user_sentence
+    : KW_CREATE KW_USER opt_if_not_exists name_label KW_WITH user_login_type {
+        auto sentence = new CreateUserSentence($4, $3);
+        sentence->setLoginType($6);
         $$ = sentence;
```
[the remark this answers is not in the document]
bright-starry-sky: > I think when create user does not specify a login type, the default is KW_PASSWORD

emm . I think we'd also to set the "password" to default if as your said the default is KW_PASSWORD. so I think best way is do not set default KW_PASSWORD.

src/parser/UserSentences.h
```diff
@@ -13,32 +13,87 @@
...
+public:
+    enum LoginType : uint8_t {
+        PASSWORD,
+        LDAP
+    };
+
+    explicit UserLoginType(std::string *password) {
+        loginType_ = LoginType::PASSWORD;
+        password_.reset(password);
+    }
```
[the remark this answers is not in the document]
bright-starry-sky: > Can the password here be empty?

Password complexity checking is not yet available. allow "" password. but not allow null password.

src/graph/UserExecutor.cpp
bright-starry-sky: good point.

src/graph/ShowExecutor.cpp
bright-starry-sky: good idea. I've unified.

src/graph/UserExecutor.cpp
bright-starry-sky: > Do we need to check the user privilege? I guess only user like GOD can drop/alter user

Aha, Good question. Only GOD and ADMIN are allowed to create, delete and drop users. and GOD allows authorize for any space. ADMIN only allows authorize space within own permission.

And I will coding the privilege check function in next PR. the next PR is "Permission Manager"

ansarizafar:
Its better to use term Super Admin/User than GOD.

src/graph/UserExecutor.cpp
critical27: 👍

src/meta/MetaServiceUtils.cpp
```diff
@@ -516,48 +516,41 @@ std::string MetaServiceUtils::userKey(UserID userId) {
...
+    std::string val;
+    apache::thrift::CompactSerializer::serialize(userItem, &val);
     return val;
 }
 
-folly::StringPiece MetaServiceUtils::userItemVal(folly::StringPiece rawVal) {
-    auto offset = sizeof(int32_t) + *reinterpret_cast<const int32_t *>(rawVal.begin());
-    return rawVal.subpiece(offset, rawVal.size() - offset);
-}
+std::string MetaServiceUtils::replaceUserVal(const nebula::cpp2::UserItem& user,
```
liuyu85cn: how about change the parameter name like user=>newUser, val=>oldUserVal

src/interface/meta.thrift
```diff
@@ -64,21 +64,6 @@ enum AlterSchemaOp {
...
 
-/**
-** GOD is A global senior administrator.like root of Linux systems.
-** ADMIN is an administrator for a given Graph Space.
-** USER is a normal user for a given Graph Space. A User can access (read and write) the data in the Graph Space.
-** GUEST is a read-only role for a given Graph Space. A Guest cannot modify the data in the Graph Space.
-** Refer to header file src/graph/PermissionManager.h for details.
-**/
-
-enum RoleType {
```
liuyu85cn: why not just use those roles in common.thrift

src/meta/MetaServiceUtils.cpp
bright-starry-sky: > how about change the parameter name like user=>newUser, val=>oldUserVal

Aha, here I using thrift object direct. So I can modifying individual user property value when thrift val was set. And serialize it for storage. Thrift object cab be backward compatible.

src/interface/meta.thrift
bright-starry-sky: > why not just use those roles in common.thrift

Because using the RoleType in meta layer and graph layer.

src/graph/UserExecutor.cpp
```diff
@@ -0,0 +1,178 @@
...
+    auto future = mc->alterUser(userItem);
+    auto *runner = ectx()->rctx()->runner();
+    auto cb = [this] (auto &&resp) {
+        if (!resp.ok()) {
+            DCHECK(onError_);
+            onError_(resp.status());
+            return;
+        }
+
+        DCHECK(onFinish_);
```
panda-sheep: ditto

src/graph/UserExecutor.cpp
```diff
@@ -0,0 +1,178 @@
...
+    userItem.set_login_type(type);
+    if (type == nebula::cpp2::UserLoginType::PASSWORD) {
+        userItem.set_encoded_pwd(encryption::MD5Utils::md5Encode(*loginType->getPassword()));
+    }
+    UserUtils::resetUserItem(opts, userItem);
+    auto future = mc->createUser(userItem, sentence_->ifNotExists());
+    auto *runner = ectx()->rctx()->runner();
+    auto cb = [this] (auto &&resp) {
+        if (!resp.ok()) {
+            DCHECK(onError_);
```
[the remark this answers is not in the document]
bright-starry-sky: Good catch, I was changed.

panda-sheep:
Great job,  LGTM
Please resolve the conflict first

src/parser/scanner.lex
```diff
@@ -157,6 +153,15 @@ SNAPSHOT                    ([Ss][Nn][Aa][Pp][Ss][Hh][Oo][Tt])
 SNAPSHOTS                   ([Ss][Nn][Aa][Pp][Ss][Hh][Oo][Tt][Ss])
 FORCE                       ([Ff][Oo][Rr][Cc][Ee])
 BIDIRECT                    ([Bb][Ii][Dd][Ii][Rr][Ee][Cc][Tt])
+MAX_QUERIES_PER_HOUR        ([Mm][Aa][Xx][_][Qq][Uu][Ee][Rr][Ii][Ee][Ss][_][Pp][Ee][Rr][_][Hh][Oo][Uu][Rr])
```
dangleptr: Actually, we don't need them anymore.

src/parser/test/ScannerTest.cpp
```diff
@@ -429,6 +421,26 @@ TEST(Scanner, Basic) {
         CHECK_SEMANTIC_TYPE("SNAPSHOTS", TokenType::KW_SNAPSHOTS),
         CHECK_SEMANTIC_TYPE("Snapshots", TokenType::KW_SNAPSHOTS),
         CHECK_SEMANTIC_TYPE("snapshots", TokenType::KW_SNAPSHOTS),
+        CHECK_SEMANTIC_TYPE("MAX_QUERIES_PER_HOUR", TokenType::KW_MAX_QUERIES_PER_HOUR),
```
dangleptr: Ditto

src/interface/common.thrift
```diff
@@ -112,5 +112,50 @@ struct Pair {
...
+struct UserItem {
+    1: required string        account;
+    // Disable user if lock status is true.
+    2: optional bool          is_lock,
+    // user login type
+    3: optional UserLoginType login_type,
+    // encoded password
+    4: optional string        encoded_pwd,
+    // The number of queries an account can issue per hour
+    5: optional i32           max_queries_per_hour,
```
dangleptr: We don't need them any more before we don't implement user's isolation.

bright-starry-sky: > We don't need them any more before we don't implement user's isolation.

Good suggest ! 
Yes , They are useless here if we don't implement user's isolation.
I also hesitated to add these options at that time. Now, let me delete them.

src/interface/common.thrift
```diff
@@ -112,5 +112,50 @@ struct Pair {
...
+}
+
+struct RoleItem {
+    1: string        user,
+    2: string        space,
+    3: RoleType      role_type,
+}
+
+struct UserItem {
+    1: required string        account;
```
dangleptr: Do we have id for user?

src/interface/common.thrift
```diff
@@ -112,5 +112,50 @@ struct Pair {
...
+} (cpp.enum_strict)
+
+enum UserLoginType {
+   PASSWORD = 0x01,
+   LDAP     = 0x02,
+}
+
+struct RoleItem {
+    1: string        user,
+    2: string        space,
```
dangleptr: Use spaceId

src/meta/processors/usersMan/AuthenticationProcessor.cpp
```diff
@@ -112,27 +118,66 @@ void DropUserProcessor::process(const cpp2::DropUserReq& req) {
...
+    GraphSpaceID spaceId;
+    const auto& roleItem = req.get_role_item();
+    auto userRet = getUserId(roleItem.get_user());
+    if (!userRet.ok()) {
+        handleErrorCode(MetaCommon::to(userRet.status()));
+        onFinished();
+        return;
+    }
+    if (roleItem.get_role_type() == nebula::cpp2::RoleType::GOD) {
+        spaceId = kDefaultSpaceId;
```
dangleptr: Do we support to grant a user to be GOD?

src/interface/common.thrift
```diff
@@ -112,5 +112,50 @@ struct Pair {
...
+    GUEST  = 0x05,
+} (cpp.enum_strict)
+
+enum UserLoginType {
+   PASSWORD = 0x01,
+   LDAP     = 0x02,
+}
+
+struct RoleItem {
+    1: string        user,
```
dangleptr: If we have userID, we'd better use userId here.

src/meta/processors/usersMan/AuthenticationProcessor.cpp
```diff
@@ -74,19 +79,20 @@ void AlterUserProcessor::process(const cpp2::AlterUserReq& req) {
 
 void DropUserProcessor::process(const cpp2::DropUserReq& req) {
     folly::SharedMutex::WriteHolder wHolder(LockUtils::userLock());
-    auto ret = getUserId(req.get_account());
+    const auto& account = req.get_account();
+    auto ret = getUserId(account);
     if (!ret.ok()) {
-        if (req.get_missing_ok()) {
+        if (req.get_if_exists()) {
             handleErrorCode(cpp2::ErrorCode::SUCCEEDED);
```
dangleptr: Please recheck "handleErrorCode",  IIRC "onFinished" has been called inside?

src/graph/UserExecutor.cpp
```diff
@@ -0,0 +1,169 @@
...
+                                               ExecutionContext *ectx)
+    : Executor(ectx, "change_password") {
+    sentence_ = static_cast<ChangePasswordSentence*>(sentence);
+}
+
+Status ChangePasswordExecutor::prepare() {
+    return Status::OK();
+}
+
+void ChangePasswordExecutor::execute() {
```
dangleptr: Do we need "change password" Or just "resetPassword" ?

src/meta/MetaServiceUtils.cpp
```diff
@@ -516,48 +516,41 @@ std::string MetaServiceUtils::userKey(UserID userId) {
...
+                                             folly::StringPiece val) {
+    nebula::cpp2::UserItem oldUser;
+    apache::thrift::CompactSerializer::deserialize(val, oldUser);
```
dangleptr: Why not serialize the newUser and overwrite the old one directly

src/interface/common.thrift
bright-starry-sky: > Do we have id for user?

I've removed user id . because we can using user name to keep the uniqueness for users.

src/meta/processors/usersMan/AuthenticationProcessor.cpp
bright-starry-sky: > Do we support to grant a user to be GOD?

Only supported to grant a user to GOD in initialization phase. 
We need to initialize a root user before  the authority flag turn on.

src/interface/common.thrift
bright-starry-sky: > If we have userID, we'd better use userId here.

UserID removed.

src/meta/processors/usersMan/AuthenticationProcessor.cpp
bright-starry-sky: > Please recheck "handleErrorCode", IIRC "onFinished" has been called inside?

emm... seems It's works well now,  "onFinished" have not call "handleErrorCode" inside.

bright-starry-sky:
Refactored root user initialize logic. Now we have only one super user naming "root". 
metad will check the root user when meta service start. and then create the user if dose not exists .

src/meta/RootUserMan.h
```diff
@@ -0,0 +1,63 @@
...
+public:
+    static bool isUserExists(kvstore::KVStore* kv) {
+        auto userKey = MetaServiceUtils::userKey("root");
+        std::string val;
+        auto result = kv->get(kDefaultSpaceId, kDefaultPartId, userKey, &val);
+        if (result == kvstore::ResultCode::SUCCEEDED) {
+            LOG(INFO) << "Root user exists";
+            return true;
+        } else {
+            LOG(INFO) << "Root user is not exists";
```
dangleptr: If current instance is not leader,  we should skip the check logic.

src/meta/processors/partsMan/DropSpaceProcessor.cpp
```diff
@@ -43,7 +43,19 @@ void DropSpaceProcessor::process(const cpp2::DropSpaceReq& req) {
     deleteKeys.emplace_back(MetaServiceUtils::spaceKey(spaceId));
 
     // delete related role data.
-    // TODO(boshengchen) delete related role data under the space
+    auto rolePrefix = MetaServiceUtils::roleSpacePrefix(spaceId);
```
dangleptr: Add some LOGs when deleting the users information.

src/interface/common.thrift
dangleptr: ditto

src/interface/meta.thrift
```diff
@@ -145,26 +129,6 @@ struct HostItem {
     4: map<string, list<common.PartitionID>> (cpp.template = "std::unordered_map") all_parts,
 }
 
-struct UserItem {
```
dangleptr: We don't need it any more?

src/meta/RootUserMan.h
bright-starry-sky: > If current instance is not leader, we should skip the check logic.

Fixed. good catch. My lack of consideration.

src/interface/meta.thrift
bright-starry-sky: I don't think we need it any more, currently we only have one item of password, we can storage it direct.
If there are new elements in the future, we can add this struct at that time.

src/interface/common.thrift
bright-starry-sky: > ditto

fixed.

src/meta/processors/partsMan/DropSpaceProcessor.cpp
bright-starry-sky: > Add some LOGs when deleting the users information.

fixed.

src/daemons/MetaDaemon.cpp
```diff
@@ -235,6 +236,27 @@ int main(int argc, char *argv[]) {
         }
     }
 
+    {
+        /**
+         *  Only leader part needed.
+         */
+        auto ret = kvstore->partLeader(nebula::meta::kDefaultSpaceId,
```
dangleptr: Please check the logic on real cluster

bright-starry-sky: > Please check the logic on real cluster

Yes,  I've simple tested two meta nodes on virtual machines. 
I will do complete integration test at later.

src/meta/processors/usersMan/AuthenticationProcessor.cpp
```diff
@@ -11,51 +11,35 @@ namespace meta {
...
-        if (req.get_missing_ok()) {
-            handleErrorCode(cpp2::ErrorCode::SUCCEEDED);
+        cpp2::ErrorCode code;
+        if (req.get_if_not_exists()) {
+            code = cpp2::ErrorCode::SUCCEEDED;
         } else {
-            LOG(ERROR) << "Create User Failed :" << user.get_account() << " have existed";
-            handleErrorCode(cpp2::ErrorCode::E_EXISTED);
+            LOG(ERROR) << "Create User Failed : User " << account
+                       << " have existed!";
```
Amber1990Zhang: already existed.

PR #1844, opened 2020-03-01
LDAP authentication

panda-sheep:
Implementation of LDAP authentication method, including
1） direct bind LDAP authentication
2） search bind LDAP authentication

Todo：
1） Add test cases
2） When create user pr is merged,  perform joint debugging verification

### DIFF
--- a/src/graph/GraphService.cpp
+++ b/src/graph/GraphService.cpp
@@ -9,6 +9,7 @@
 #include "time/Duration.h"
 #include "graph/RequestContext.h"
 #include "graph/SimpleAuthenticator.h"
+ #include "graph/LdapAuthenticator.h"
 #include "storage/client/StorageClient.h"
 
 namespace nebula {
@@ -17,7 +18,8 @@ namespace graph {
 Status GraphService::init(std::shared_ptr<folly::IOThreadPoolExecutor> ioExecutor) {
     sessionManager_ = std::make_unique<SessionManager>();
     executionEngine_ = std::make_unique<ExecutionEngine>();
-    authenticator_ = std::make_unique<SimpleAuthenticator>();
+    // authenticator_ = std::make_unique<SimpleAuthenticator>();
+    authenticator_ = std::make_unique<LdapAuthenticator>();
 
     return executionEngine_->init(std::move(ioExecutor));
 }
--- a/src/graph/GraphService.cpp
+++ b/src/graph/GraphService.cpp
@@ -9,7 +9,7 @@
 #include "time/Duration.h"
 #include "graph/RequestContext.h"
 #include "graph/SimpleAuthenticator.h"
- #include "graph/LdapAuthenticator.h"
+#include "graph/LdapAuthenticator.h"
 #include "storage/client/StorageClient.h"
 
 namespace nebula {
--- a/src/graph/LdapAuthenticator.cpp
+++ b/src/graph/LdapAuthenticator.cpp
@@ -6,7 +6,6 @@
 
 #include "graph/LdapAuthenticator.h"
 #include "graph/GraphFlags.h"
-#define LDAP_DEPRECATED 1
 
 namespace nebula {
 namespace graph {
--- a/src/graph/LdapAuthenticator.cpp
+++ b/src/graph/LdapAuthenticator.cpp
@@ -6,7 +6,7 @@
 
 #include "graph/LdapAuthenticator.h"
 #include "graph/GraphFlags.h"
-#include <ldap.h>
+#define LDAP_DEPRECATED 1
 
 namespace nebula {
 namespace graph {
@@ -88,7 +88,7 @@ Status LdapAuthenticator::initLDAPConnection() {
     */
 
     // initialize the LDAP
-    ldap_ = ldap_init(FLAGS_ldap_server, FLAGS_ldap_port);
+    ldap_ = ldap_init(FLAGS_ldap_server.c_str(), FLAGS_ldap_port);
     if (!ldap_) {
         return Status::Error("Init LDAP failed.");
     }
@@ -154,7 +154,7 @@ StatusOr<bool> LdapAuthenticator::searchBindAuth() {
      * LDAP authentication disallows any characters that need to escape
      * in user name.
      */
-    for (auto& c : user) {
+    for (auto& c : userName_) {
         if (c == '*' ||
             c == '(' ||
             c == ')' ||
@@ -251,7 +251,7 @@ StatusOr<bool> LdapAuthenticator::searchBindAuth() {
         return reInit;
     }
 
-    ret = ldap_simple_bind_s(ldap_, dn, password_);
+    ret = ldap_simple_bind_s(ldap_, dn.c_str(), password_.c_str());
     if (ret != LDAP_SUCCESS) {
         ldap_unbind(ldap_);
         return Status::Error("LDAP login failed for user \"%s\" on server \"%s\".",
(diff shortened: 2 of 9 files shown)


src/graph/LdapAuthenticator.h
```diff
@@ -0,0 +1,74 @@
...
+
+    /**
+     * Execute LDAP direct bind mode authentication
+     */
+    StatusOr<bool> directBindAuth();
+
+    /**
+     * Execute LDAP authentication.
+     */
+    StatusOr<bool> auth(const std::string &user, const std::string &password) override;
```
bright-starry-sky: I think , I think only one public method is enough of auth. other should be private. right?

src/graph/LdapAuthenticator.cpp
```diff
@@ -0,0 +1,290 @@
...
+    std::string::size_type  pos = 0;
+    while ((pos = result.find(ph)) != std::string::npos) {
+        result.replace(pos, len, userName_);
+    }
+
+    return result;
+}
+
+
+Status<bool> LdapAuthenticator::searchBindAuth() {
```
dutor: If I were right, there is no such thing by `Status<bool>`.

src/graph/LdapAuthenticator.cpp
```diff
@@ -0,0 +1,290 @@
...
+        return Status::Error("LDAP login failed for user \"%s\" on server \"%s\"."
+                              dn.c_str(), FLAGS_ldap_server.c_str());
+    }
+
+    ldap_unbind(ldap_);
+    return true;
+}
+
+
+StatusOr<bool> LdapAuthenticator::directBindAuth() {
```
dutor: Please make a research on how `Status`, `StatusOr<bool>`, `StatusOr<bool>.value()` are different.

src/graph/LdapAuthenticator.cpp
panda-sheep: My mistake, it  should be StatusOr<bool>.

src/graph/LdapAuthenticator.cpp
panda-sheep: Yeah,
Status:      only has status.
StatusOr<bool>  has status or bool value， but cant have  both.
StatusOr<bool>.value():    when StatusOr<bool>.ok() is true， it has bool value and gets the bool value。

src/graph/LdapAuthenticator.cpp
```diff
@@ -0,0 +1,290 @@
...
+    }
+
+    /**
+     * Bind with a pre-defined user name and password to search.
+     * If none is specified, use the anonymous bind.
+     */
+    auto& binddn = FLAGS_ldap_binddn;
+    auto& bindPassword = FLAGS_ldap_bindPassword;
+
+    auto ret = ldap_simple_bind_s(ldap_, binddn.c_str(), bindPassword.c_str());
```
dutor: `ldap_simple_bind_s` and other interfaces here are all synchronous, right? If so, do the LDAP kits provide async ones?
We cannot do such blocking actions directly in the main working threads.

src/graph/LdapAuthenticator.h
panda-sheep: Yeah，you are right.

src/graph/LdapAuthenticator.cpp
panda-sheep: Good point.  I will research it.  
image

knightXun:
I have one question, If we deploy nebula in aws. User should expose their ldap server,

src/graph/LdapAuthenticator.cpp
panda-sheep: @dutor 
When a user logs in from the console, graphd will start a thread to process.
Here is login verification, asynchronous synchronization doesn't feel that different.
In synchronization, function A connects to the LDAP server. Before the LDAP server returns, A is blocked.
Asynchronously, function B connects to the LDAP server. Before the LDAP server returns, B can continue to execute, but it will block behind B and wait for the result from the LDAP server.
So the difference between synchronous and asynchronous is not obvious

src/graph/LdapAuthenticator.cpp
```diff
@@ -0,0 +1,362 @@
...
+    }
+    ldap_->ld_options.ldo_defport = FLAGS_ldap_port;
+
+    ret = ldap_set_option(ldap_, LDAP_OPT_HOST_NAME, FLAGS_ldap_server);
+    if (ret != LDAP_SUCCESS) {
+        ldap_ld_free(ldsp_, 1, NULL, NULL);
+        return Status::Error("Init LDAP failed.");
+    }
+    */
+    // ldap_init is deprecated, so use ldap_initialize
```
liuyu85cn: not deleted comments ?

src/graph/LdapAuthenticator.h
```diff
@@ -0,0 +1,96 @@
...
+    /**
+     * Execute LDAP simple bind mode authentication
+     */
+    StatusOr<bool> simpleBindAuth();
+
+
+    LDAP                    *ldap_{nullptr};
+
+    // Default LDAP port
+    int                      ldapPort_{389};
```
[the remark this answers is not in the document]
panda-sheep: 389 is the default port number for ldap.

src/graph/LdapAuthenticator.cpp
panda-sheep: I actually prefer the ldap_create method here.
`E.g. ldap_create (& ldap_);`
But ldap_create requires version 2.4.49 and above.

  `ldap_init (FLAGS_ldap_server.c_str (), FLAGS_ldap_port);`
ldap_init is deprecated. If you use it, you need to add` -DLDAP_DEPRECATED` in  CMakeLists.txt , so it is not recommended to use ldap_init.

I am also hesitant to remove the comments. I put this note here to remind us that the above two methods cannot be used now.

src/graph/GraphService.cpp
```diff
@@ -17,7 +18,8 @@ namespace graph {
 Status GraphService::init(std::shared_ptr<folly::IOThreadPoolExecutor> ioExecutor) {
     sessionManager_ = std::make_unique<SessionManager>();
     executionEngine_ = std::make_unique<ExecutionEngine>();
-    authenticator_ = std::make_unique<SimpleAuthenticator>();
+    // authenticator_ = std::make_unique<SimpleAuthenticator>();
+    authenticator_ = std::make_unique<LdapAuthenticator>();
```
panda-sheep: This is just for UT testing, you can ignore it temporarily, and wait for the user manager's pr to be merged and then modify it.

src/graph/LdapAuthenticator.cpp
```diff
@@ -0,0 +1,361 @@
...
+        uris.append("://");
+        uris.append(host, size);
+        uris += folly::stringPrintf(":%d", FLAGS_ldap_port);
+        host += size;
+        if (*host == ',') {
+            host++;
+        }
+        while (*host == ' ') {
+           host++;
+        }
```
critical27: Use folly::split? A little difficult to understand this block

src/graph/LdapAuthenticator.cpp
```diff
@@ -0,0 +1,361 @@
...
+    if (FLAGS_ldap_server.empty()) {
+        return Status::Error("LDAP authentication \"ldap_server\" parameter is not set.");
+    }
+
+    if (FLAGS_ldap_scheme.compare("ldap")) {
+        return Status::Error("LDAP authentication ldap_scheme only support \"ldap\".");
+    }
+
+    if (FLAGS_ldap_port == 0) {
+        FLAGS_ldap_port = ldapPort_;
```
critical27: How about just set the default port in define of gflags

src/graph/LdapAuthenticator.cpp
panda-sheep: Good catch.

src/graph/LdapAuthenticator.cpp
panda-sheep: Yeah，good

PR #1854, opened 2020-03-03
set limit for partition_number of space

knightXun:
We should set a flag to  limit partition_number of a space.


knightXun:
@dangleptr waiting for your review

dangleptr:
I want to know why we need to set the limit for the parts number.

knightXun:
> I want to know why we need to set the limit for the parts number.

If parts number is too large, storage will easily oom

Issue #1853, opened 2020-03-03
LookUpVertexIndexProcessor double free


wadeliuyi:
in function LookUpVertexIndexProcessor step 3:

```
std::for_each(req.get_parts().begin(), req.get_parts().end(), [&](auto& partId) {
        auto code = executeExecutionPlan(partId);
        if (code != kvstore::ResultCode::SUCCEEDED) {
            VLOG(1) << "Error! ret = " << static_cast<int32_t>(code)
                    << ", spaceId = " << spaceId_
                    << ", partId =  " << partId;
            if (code == kvstore::ResultCode::ERR_LEADER_CHANGED) {
                this->handleLeaderChanged(spaceId_, partId);
            } else {
                this->pushResultCode(this->to(code), partId);
            }
            this->onFinished();
            return;
        }
    });
```
this->onFinished() will delete this pointer, but the "return" just return from the lambda, so the function will excute go on to the end and call this->onFinished() again.

bright-starry-sky:
good point. let me check.

bright-starry-sky:
Thanks for your report . This bug fixed via #1855

PR #1855, opened 2020-03-03
fixed lookup bug for double free

bright-starry-sky:
fixed lookup bug .
Issus #1853

### DIFF
--- a/src/storage/index/LookUpEdgeIndexProcessor.cpp
+++ b/src/storage/index/LookUpEdgeIndexProcessor.cpp
@@ -42,8 +42,6 @@ void LookUpEdgeIndexProcessor::process(const cpp2::LookUpIndexRequest& req) {
             } else {
                 this->pushResultCode(this->to(code), partId);
             }
-            this->onFinished();
-            return;
         }
     });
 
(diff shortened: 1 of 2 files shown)


src/storage/index/LookUpEdgeIndexProcessor.cpp
```diff
@@ -42,8 +42,6 @@ void LookUpEdgeIndexProcessor::process(const cpp2::LookUpIndexRequest& req) {
             } else {
                 this->pushResultCode(this->to(code), partId);
             }
-            this->onFinished();
```
dangleptr: We don't return now?

bright-starry-sky: Emm.. I don't think we need any return here. because this is last line of lambda block. Is there any other intention?

bright-starry-sky: > We don't return now?

let me  improve the logic,  let return directly when any part error occur.

bright-starry-sky:
Improved the logic, return error directly when any part fail occur.

PR #1859, opened 2020-03-03
limit total numbers of spaces

knightXun:
We must limit the number of spaces, that's important.

### DIFF
--- a/src/interface/meta.thrift
+++ b/src/interface/meta.thrift
@@ -21,6 +21,7 @@ enum ErrorCode {
     E_LEADER_CHANGED   = -11,
 
     // Operation Failure
+    E_TOO_MANY_SPACE   = -20,
     E_NO_HOSTS         = -21,
     E_EXISTED          = -22,
     E_NOT_FOUND        = -23,
--- a/src/interface/meta.thrift
+++ b/src/interface/meta.thrift
@@ -21,7 +21,7 @@ enum ErrorCode {
     E_LEADER_CHANGED   = -11,
 
     // Operation Failure
-    E_TOO_MANY_SPACE   = -20,
+    E_TOO_MANY_SPACES  = -20,
     E_NO_HOSTS         = -21,
     E_EXISTED          = -22,
     E_NOT_FOUND        = -23,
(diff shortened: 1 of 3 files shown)


Shylock-Hg:
Could treat negative as infinity limit.

src/interface/meta.thrift
```diff
@@ -21,6 +21,7 @@ enum ErrorCode {
     E_LEADER_CHANGED   = -11,
 
     // Operation Failure
+    E_TOO_MANY_SPACE   = -20,
```
yixinglu: `E_TOO_MANY_SPACES `

knightXun: thx

src/meta/processors/partsMan/CreateSpaceProcessor.cpp
```diff
@@ -9,6 +9,8 @@
 
 DEFINE_int32(default_parts_num, 100, "The default number of parts when a space is created");
 DEFINE_int32(default_replica_factor, 1, "The default replica factor when a space is created");
+DEFINE_int32(default_space_sum, 100, "The default number of space can be created");
+DEFINE_int32(default_max_replica_factor, 1, "The default max replica factor that a space can take");
```
dutor: Please revise the naming of these two flags. _**default**_ is the value we take when user doesn't provide. You could infer from the fact how `default_parts_num` are used.

knightXun: good point, I changed these two flag names

src/meta/processors/partsMan/CreateSpaceProcessor.cpp
```diff
@@ -119,6 +137,22 @@ CreateSpaceProcessor::pickHosts(PartitionID partId,
...
+int32_t CreateSpaceProcessor::getSpaceSum() {
+    folly::SharedMutex::ReadHolder rHolder(LockUtils::spaceLock());
+    int32_t count = 0; 
+    auto prefix = MetaServiceUtils::spacePrefix();
+    std::unique_ptr<kvstore::KVIterator> iter;
+    auto ret = kvstore_->prefix(kDefaultSpaceId, kDefaultPartId, prefix, &iter);
+    if (ret != kvstore::ResultCode::SUCCEEDED) {
+        return -1;
+    }
+    std::vector<cpp2::IdName> spaces;
```
yixinglu, line 149: It seems an unuseful variables: `spaces`

knightXun: Good point, I will carefully rewrite this pr

PR #1860, opened 2020-03-04
Delete vertices with get neighbor

darionyaphet:


src/graph/DeleteVerticesExecutor.cpp
```diff
@@ -52,8 +53,44 @@ void DeleteVerticesExecutor::execute() {
         vids_.emplace_back(vid);
     }
 
-    // TODO(zlcook) Get edgeKeys of a vertex by Go
-    auto future = ectx()->getStorageClient()->getEdgeKeys(space_, vids_);
```
dangleptr: We could delete the interface in thrift now.

src/graph/DeleteVerticesExecutor.cpp
```diff
@@ -52,8 +53,44 @@ void DeleteVerticesExecutor::execute() {
...
+        return;
+    }
+
+    std::vector<EdgeType> edgeTypes;
+    std::vector<std::string> columns = {_DST, _RANK};
+    std::vector<nebula::storage::cpp2::PropDef> props;
+    for (auto &e : edgeAllStatus.value()) {
+        auto edgeStatus = ectx()->schemaManager()->toEdgeType(space_, e);
+        if (!edgeStatus.ok()) {
+            LOG(ERROR) << "Can't get all of the edge types";
```
bright-starry-sky: Is better  "Can't get edge type, edge : " << e;
It's no big deal,  LGTM.

darionyaphet: I think it's ok ~

PR #1861, opened 2020-03-04
When the compiler is GCC, use static link libasan and libubsan.

monadbobo:
When the compiler is GCC, use static link libasan and libubsan.


CMakeLists.txt
```diff
@@ -300,10 +300,14 @@ if(ENABLE_ASAN)
     if (NOT ENABLE_ASAN_OK)
         MESSAGE(FATAL_ERROR "The compiler does not support address sanitizer")
     endif()
+if("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU")
+    add_compile_options(-static-libasan)
+    add_link_options(-static-libasan)
+endif()
```
dutor: Is there any other workaround to deal with the problem you encountered?
As far as I know, statically linking against libasan is problematic for some GCC versions.

CMakeLists.txt
```diff
@@ -300,10 +300,14 @@ if(ENABLE_ASAN)
...
     add_compile_options(-fsanitize=address)
     add_compile_options(-g)
     add_compile_options(-fno-omit-frame-pointer)
-    set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -fsanitize=address")
+    add_link_options(-fsanitize=address)
```
dutor: Our requirement on CMake is 3.5.0, while `add_link_options` was introduced since 3.13, which is too restrictive for our users.

CMakeLists.txt
monadbobo: Ok, i will added the options to control this.

PR #1864, opened 2020-03-04
Using admin client in MetaServiceHandler

darionyaphet:


bright-starry-sky:
Why do this? I don't see any improvement if you do that.

darionyaphet:
Should not create another AdminClient

PR #1867, opened 2020-03-04
Fix graph client print bad when auth failed.

CPWstatic:
Actually, it prints nothing. But, it is better than printing an address.
```
$ ./bin/nebula -u=user
Password: E0304 17:57:39.003026 30347 GraphClient.cpp:49] Failed to authenticate "user":
Connection failed
```


Shylock-Hg:
Close https://github.com/vesoft-inc/nebula/pull/1326

PR #1869, opened 2020-03-04
console security

knightXun:
Remove some sensitive infomation from console.


src/console/CliManager.cpp
```diff
@@ -171,8 +171,8 @@ bool CliManager::readLine(std::string &line, bool linebreak) {
         if (!isInteractive_) {
             break;
         }
-        auto purePrompt = folly::stringPrintf("(%s@%s:%d) [%s]> ",
-                                              username_.c_str(), addr_.c_str(), port_,
+        auto purePrompt = folly::stringPrintf("(%s@) [%s]> ",
```
Shylock-Hg: You can replace IP by hostname.

knightXun: Hostname can be resolved by dns

Shylock-Hg: I mean the `gethostname` function return value which the '/etc/hostname' content. It's set by system administrator and usually not a domain could be resolved.

Shylock-Hg: I mean the `gethostname` function return value which the '/etc/hostname' content. It's set by system administrator and usually not a domain could be resolved to public IP.It could be a good label about machine in the console.

knightXun: 1. IP address could be hostname. 2.If hostname can be resolved by dns, the ip will leak

laura-ding: modify to `(%s@nebula)`

knightXun: fix this

knightXun:
@Shylock-Hg can this pr be merged?

PR #1870, opened 2020-03-04
limit console cmd length

knightXun:
For Security, Limit the max length of a single cmd.

### DIFF
--- a/src/console/CliManager.cpp
+++ b/src/console/CliManager.cpp
@@ -24,6 +24,7 @@ namespace graph {
 const int32_t kMaxAuthInfoRetries = 3;
 const int32_t kMaxUsernameLen = 16;
 const int32_t kMaxPasswordLen = 24;
+const int32_t kMaxCmdLen = 65536;
 
 CliManager::CliManager() {
     if (!fs::FileUtils::isStdinTTY()) {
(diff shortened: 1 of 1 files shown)


src/console/CliManager.cpp
```diff
@@ -142,15 +144,28 @@ void CliManager::loop() {
...
+            if (lineSize > kMaxCmdLen) {
+                cmdTooLarge = true; 
+                break;
+            }
+            line.resize(lineSize);
             cmd += line;
+            if(cmd.length() > kMaxCmdLen) {
+                cmdTooLarge = true; 
+                break;
+            }
```
dutor: Only catch the case there are line breaks?
I think you could just check right before the command gets executed/processed.

dangleptr:
Any progress about this pr?  
@knightXun @dutor @laura-ding

Judy1992:
we should take the length limitation into consideration as a whole instead of just limiting the length of cmd, e.g. r we gonna limit the length of string prop.  Maybe do in at parser 🤷‍♂️🤷‍♂️

jude-zhu:
we should take it into consideration as a whole instead of just limiting the length of cmd. For example. r we gonna limit the length of string prop. Maybe do it at parser? 🤷‍♂️🤷‍♂️

PR #1871, opened 2020-03-04
add a flag to close config cmd

knightXun:
For security, We must disable config command in some situation.


Shylock-Hg:
Disable the HTTP configuration API too?

knightXun:
good point

knightXun:
@dangleptr yeah, we have better way to do that. I close this

knightXun:
I could do this in usermanager

PR #1873, opened 2020-03-05
User permission manager

bright-starry-sky:
1,  moved `ClientSession` to common layer . and rename to `Session`.
2, Added Permission Manager .
3, Added Permission Check.

This PR depend on #1842 ，I will create more test cases after #1842 is done.


bright-starry-sky:
Ready to review. come on teachers. thanks.

src/common/permission/PermissionManager.cpp
```diff
@@ -0,0 +1,214 @@
...
+        case session::Role::INVALID_ROLE : {
+            break;
+        }
+    }
+    return havePermission;
+}
+
+// static
+bool PermissionManager::canReadSchemaData(session::Session *session) {
+    if (session->isGod()) {
```
dangleptr: why not put the check after `if (session->space() == -1) `

src/common/permission/PermissionManager.cpp
```diff
@@ -0,0 +1,214 @@
...
+}
+
+// static
+bool PermissionManager::canWriteSpace(session::Session *session) {
+    return session->isGod();
+}
+
+// static
+bool PermissionManager::canWriteSchema(session::Session *session) {
+    if (session->isGod()) {
```
dangleptr: ditto

src/common/permission/PermissionManager.cpp
```diff
@@ -0,0 +1,214 @@
...
+    }
+    return false;
+}
+
+// static
+bool PermissionManager::canWriteData(session::Session *session) {
+    if (session->isGod()) {
+        return true;
+    }
+    if (session->space() == -1) {
```
dangleptr: ditto

src/common/session/Session.h
```diff
@@ -0,0 +1,130 @@
...
+private:
+    Session() = default;
+    explicit Session(int64_t id);
+
+
+private:
+    int64_t           id_{0};
+    GraphSpaceID      space_{-1};
+    std::string       spaceName_;
+    std::string       user_;
```
dangleptr: maybe account is a better name.

src/meta/client/MetaClient.cpp
```diff
@@ -1622,6 +1622,20 @@ const std::vector<HostAddr>& MetaClient::getAddresses() {
     return addrs_;
 }
 
+StatusOr<std::vector<nebula::cpp2::RoleItem>>
+MetaClient::getRolesByUser(const std::string& user) {
```
dangleptr: We should hold a cache for User->Role map

src/common/permission/PermissionManager.cpp
bright-starry-sky: > why not put the check after `if (session->space() == -1)`

Good point.

bright-starry-sky: > why not put the check after `if (session->space() == -1)`

fixed.

src/common/session/Session.h
bright-starry-sky: > maybe account is a better name.

Good idea, changed.

src/meta/client/MetaClient.cpp
bright-starry-sky: > We should hold a cache for User->Role map

Very good advice. changed the role map to cache.

src/common/session/Session.h
```diff
@@ -0,0 +1,130 @@
...
+            }
+            case nebula::cpp2::RoleType::GUEST : {
+                return Role::GUEST;
+            }
+        }
+        return Role::INVALID_ROLE;
+    }
+
+    bool isGod() const {
+        for (auto iter = roles_.begin(); iter != roles_.end(); iter++) {
```
dangleptr: Currently, we only have one user as GOD, that is root.

src/graph/ShowExecutor.cpp
```diff
@@ -243,6 +244,14 @@ void ShowExecutor::showSpaces() {
         resp_->set_column_names(std::move(header));
 
         for (auto &space : retShowSpaces) {
+            if (FLAGS_enable_authorize) {
```
dangleptr: Maybe we could put the FLAGS_enable_authorize inside PermissionManager

src/meta/client/MetaClient.cpp
```diff
@@ -1622,6 +1649,20 @@ const std::vector<HostAddr>& MetaClient::getAddresses() {
...
+MetaClient::getRolesByUserFromCache(const std::string& user) {
+    auto iter = userRolesMap_.find(user);
+    if (iter == userRolesMap_.end()) {
+        return std::vector<nebula::cpp2::RoleItem>(0);
+    }
+    return iter->second;
+}
+
+bool MetaClient::authenticationCheck(std::string account, std::string password) {
+    auto ret = authCheck(std::move(account), std::move(password)).get();
```
dangleptr: I think we should use cache for account->password too.

src/meta/processors/usersMan/AuthenticationProcessor.cpp
```diff
@@ -198,19 +198,20 @@ void ListUsersProcessor::process(const cpp2::ListUsersReq& req) {
...
 void ListRolesProcessor::process(const cpp2::ListRolesReq& req) {
-    auto spaceRet = getSpaceId(req.get_space());
-    if (!spaceRet.ok()) {
-        LOG(ERROR) << "space dose not found";
-        handleErrorCode(MetaCommon::to(spaceRet.status()));
-        onFinished();
-        return;
+    auto spaceId = req.get_space_id();
+    /**
+     * Because the god role's space id is kDefaultSpaceId.
```
dangleptr: I am confused about this changes. 
"show roles" must be invoked under some space, right?  So why use defaultSpaceId here.

src/common/session/Session.h
```diff
@@ -0,0 +1,130 @@
...
+            }
+        }
+        return false;
+    }
+
+    void setRole(GraphSpaceID space, Role role) {
+        roles_.emplace(space, role);
+    }
+
+    uint64_t idleSeconds() const;
```
dangleptr: Where is the logic about taking care of expired session?

src/meta/processors/usersMan/AuthenticationProcessor.cpp
bright-starry-sky: > I am confused about this changes.
> "show roles" must be invoked under some space, right? So why use defaultSpaceId here.

Good point.
Actual the "show roles" must be invoked under some space. show the use the defaultSpaceId  is redundant.
let me correct .

src/common/session/Session.h
bright-starry-sky: > Where is the logic about taking care of expired session?

No session expired implement yet. 
Because nebula1.0 is a simple SessionManager. This version will not support any management of sessions. 
And the `idleSeconds` keep it here because from original `SessionClient`  class.

bright-starry-sky:
Addressed dangleptr's comments as below : 
1,Cached user password via meta client; 
2,Check god by user name root;
3,Moved FLAGS_enable_authorize to PermissionManager;
4,Skip space check for kDefaultSpaceId when list roles

src/graph/PermissionCheck.cpp
```diff
@@ -0,0 +1,139 @@
...
+    auto kind = sentence->kind();
+    switch (kind) {
+        case Sentence::Kind::kUnknown : {
+            return false;
+        }
+        case Sentence::Kind::kUse :
+        case Sentence::Kind::kDescribeSpace : {
+            /**
+             * Use space and Describe space are special operations.
+             * Permission checking needs to be done in their executor.
```
dangleptr: why? Miss the spaceId?

src/graph/PermissionCheck.cpp
```diff
@@ -0,0 +1,139 @@
...
+        case Sentence::Kind::kLookup :
+        case Sentence::Kind::kYield :
+        case Sentence::Kind::kOrderBy :
+        case Sentence::Kind::kFetchVertices :
+        case Sentence::Kind::kFetchEdges :
+        case Sentence::Kind::kFindPath :
+        case Sentence::Kind::kLimit :
+        case Sentence::Kind::KGroupBy :
+        case Sentence::Kind::kReturn : {
+            return permission::PermissionManager::canReadSchemaData(session);
```
dangleptr: canReadData?

src/graph/PermissionCheck.cpp
bright-starry-sky: > why? Miss the spaceId?

Yes, Usually we need to get the space id from the session when check permission. 
We can't get the space ID form session when `Use Space` or `Describe space `, so we need get the space id in their executor.

src/graph/PermissionCheck.cpp
bright-starry-sky: > Maybe canReadSchemaOrData is a better name.

Good idea. Changed method name from `canReadSchemaData` to `canReadSchemaOrData`

src/common/permission/PermissionManager.cpp
```diff
@@ -0,0 +1,223 @@
...
+bool PermissionManager::canReadSpace(session::Session *session, GraphSpaceID spaceId) {
+    if (!FLAGS_enable_authorize) {
+        return true;
+    }
+    if (session->isGod()) {
+        return true;
+    }
+    bool havePermission = false;
+    switch (session->roleWithSpace(spaceId)) {
+        case session::Role::GOD :
```
panda-sheep: What is the difference between  `session->isGod()`   and session::Role::GOD ？

bright-starry-sky: > What is the difference between `session->isGod()` and session::Role::GOD ？

`session->isGod` just check if the user is god. 
`session->roleWithSpace(spaceId) `check what role of in this space.
`case session::Role::GOD` still there because the compilation requirement of switch block .

src/graph/SequentialExecutor.cpp
```diff
@@ -19,10 +19,21 @@ SequentialExecutor::SequentialExecutor(SequentialSentences *sentences,
...
     for (auto i = 0U; i < sentences_->sentences_.size(); i++) {
         auto *sentence = sentences_->sentences_[i].get();
         auto executor = makeExecutor(sentence);
+        if (FLAGS_enable_authorize) {
+            auto *session = executor->ectx()->rctx()->session();
+            /**
+             * Skip special operations check at here. they are :
+             * kUse, kDescribeSpace, kRevoke and kGrant.
+             */
+            if (!PermissionCheck::permissionCheck(session, sentence)) {
```
critical27: What if the query is "use space test; sth else", can we work? But this is a quite usual.

critical27:
Clean code!

src/common/session/Session.h
```diff
@@ -0,0 +1,128 @@
...
+
+
+private:
+    int64_t           id_{0};
+    GraphSpaceID      space_{-1};
+    std::string       spaceName_;
+    std::string       account_;
+    time::Duration    idleDuration_;
+    /*
+     * map<space name, role>
```
panda-sheep: map<space Id, role> ?

src/graph/SequentialExecutor.cpp
bright-starry-sky: > What if the query is "use space test; sth else", can we work? But this is a quite usual.

Good question. you are right . it's fail when "use space ; sth else" . 
because our permission check is in the preparation phase,  have not set the space id in the session yet in the preparation phase. so the second statement should be fail.
I originally intended to check permissions in each executor, but this would result in too much code clutter.

src/common/session/Session.h
```diff
@@ -0,0 +1,128 @@
...
+            }
+            case nebula::cpp2::RoleType::USER : {
+                return Role::USER;
+            }
+            case nebula::cpp2::RoleType::GUEST : {
+                return Role::GUEST;
+            }
+        }
+        return Role::INVALID_ROLE;
+    }
```
panda-sheep: move `toRole ` to  `Session.cpp`  better ?

src/graph/SequentialExecutor.cpp
critical27: ok.

PR #1884, opened 2020-03-06
Fix we don't support balance remove in single replica

critical27:
1. To support `balance data remove`, as long as a host is still alive, it could be the source of a BalanceTask.
2. So we need to know the replica factor of a space, when it is single relica, we need to specify itself as the target of transLeader.
3. Fix bugs of leader balance when there are plenty of space and part.
4. Fix when we try to balance remove the same host twice, it would failed, e.g. `balance data remove 127.0.0.1:10000, 127.0.0.1:10000`
5. Remove leader lease flag


src/meta/processors/admin/BalanceTask.cpp
```diff
@@ -53,7 +53,10 @@ void BalanceTask::invoke() {
             LOG(INFO) << taskIdStr_ << "Ask the src to give up the leadership.";
             SAVE_STATE();
             if (srcLived_) {
-                client_->transLeader(spaceId_, partId_, src_).thenValue([this](auto&& resp) {
+                // For balance task of single replica, the target leader is src_ itself
+                HostAddr targetLeader = singleReplica_ ? src_ : kRandomPeer;
+                client_->transLeader(spaceId_, partId_, src_, targetLeader)
```
wadeliuyi: if the targetLeader is src_ itself, it is really need call transLeader funciton?

src/meta/processors/admin/BalanceTask.cpp
```diff
@@ -299,7 +303,9 @@ BalanceTask::parseVal(const folly::StringPiece& rawVal) {
     auto start = *reinterpret_cast<const int64_t*>(rawVal.begin() + offset);
     offset += sizeof(int64_t);
     auto end = *reinterpret_cast<const int64_t*>(rawVal.begin() + offset);
-    return std::make_tuple(status, ret, srcLived, start, end);
+    offset += sizeof(bool);
```
wadeliuyi: here is offset += sizeof(int64_t)

src/meta/processors/admin/BalanceTask.cpp
critical27: In rare case we would call transLeader if src is the target, only when balance data in single replica we would do this to keep code work in same way.

src/meta/processors/admin/BalanceTask.cpp
critical27: My bad, good catch.

src/meta/processors/admin/Balancer.cpp
```diff
@@ -206,32 +206,27 @@ Balancer::genTasks(GraphSpaceID spaceId, std::vector<HostAddr>& hostDel) {
...
+
     // We have two parts need to balance, the first one is parts on lost hosts and deleted hosts
     // The seconds one is parts on unbalanced host in newHostParts.
     std::vector<BalanceTask> tasks;
+    // 1. Iterate through all hosts that would not be included in newHostParts,
+    //    move all parts in them to host with minimum part in newHostParts
     for (auto& h : hostDel) {
         auto& lostParts = hostParts[h];
         for (auto& partId : lostParts) {
-            auto srcRet = hostWithPart(newHostParts, partId);
```
dangleptr: We should keep this.  Because the deleted host maybe dead. If so, we should not choose it as "src"

critical27: I misunderstood it as well.

In this code block, we need to move each part of each host in `hostDel` to another host. So the host in `hostDel` would be the src. In a `BalanceTask`, `src` is the host to close that part, and the one to be removed on meta. On the other hand, `dst` is the host to open the part, and the one to be added on meta.

dangleptr: You are right.

src/meta/processors/admin/Balancer.cpp
```diff
@@ -383,15 +391,23 @@ Balancer::sortedHostsByParts(const std::unordered_map<HostAddr,
...
             PartitionID partId) {
-    for (auto it = hostParts.begin(); it != hostParts.end(); it++) {
-        if (std::find(it->second.begin(), it->second.end(), partId) != it->second.end()) {
-            return it->first;
-        }
+    // check host hold the part and alive
+    auto checkPart = [&] (const auto& entry) {
+        auto& host = entry.first;
+        auto& peers = entry.second;
+        return std::find(peers.begin(), peers.end(), partId) != peers.end() &&
```
darionyaphet: Nice~

darionyaphet:
LGTM~

knightXun:
good job

PR #1892, opened 2020-03-08
plugin  feature [POC-55]

### LINKED ISSUE #1890: Implementing dynamic installation or uninstallation of plugins
Implementing dynamic installation or uninstallation of plugins
Includes the following features:
1) INSTALL PLUGIN Statement
`INSTALL PLUGIN plugin_name SONAME 'shared_library_name'`
2) UNINSTALL PLUGIN Statement
`UNINSTALL PLUGIN plugin_name`

3) SHOW PLUGINS Statement
`SHOW PLUGINS`

panda-sheep:
When each graphd uses a plugin, it first goes to meta to get the plugin information to see if the plugin exists.  
If the plugin does not exist, an error is reported directly. 
If the plugin exists, then see if the local pluginMamager has previously opened the plugin. 
If it has been opened, uses the previously opened information directly. If it has not been opened, open the plugin and saves the information to pluginMamager.

When each graphd is closed, all open plugins in this graphd  are closed.

close #1890

Support the following functions
```
install plugin pluginName soname "soname"

uninstall plugin pluginName;

show plugins;

pluginMamager
```
merge   LDAP framework #1905


src/parser/parser.yy
```diff
@@ -1638,6 +1640,9 @@ show_sentence
     }
     | KW_SHOW KW_COLLATION {
         $$ = new ShowSentence(ShowSentence::ShowType::kShowCollation);
+    }
+	| KW_SHOW KW_PLUGINS {
```
darionyaphet: Please alignment the space :)

src/parser/scanner.lex
```diff
@@ -157,6 +157,12 @@ SNAPSHOT                    ([Ss][Nn][Aa][Pp][Ss][Hh][Oo][Tt])
 SNAPSHOTS                   ([Ss][Nn][Aa][Pp][Ss][Hh][Oo][Tt][Ss])
 FORCE                       ([Ff][Oo][Rr][Cc][Ee])
 BIDIRECT                    ([Bb][Ii][Dd][Ii][Rr][Ee][Cc][Tt])
+INSTALL                     ([Ii][Nn][Ss][Tt][Aa][Ll][Ll])
+UNINSTALL                   ([Uu][Nn][Ii][Nn][Ss][Tt][Aa][Ll][Ll])
+PLUGIN                      ([Pp][Ll][Uu][Gg][Ii][Nn])
+SONAME                      ([Ss][Oo][Nn][Aa][Mm][Ee])
+PLUGINS					    ([Pp][Ll][Uu][Gg][Ii][Nn][Ss])
```
darionyaphet: Remove the space

src/parser/parser.yy
```diff
@@ -121,6 +121,7 @@ class GraphScanner;
 %token KW_SNAPSHOT KW_SNAPSHOTS KW_LOOKUP
 %token KW_JOBS KW_JOB KW_RECOVER KW_FLUSH KW_COMPACT KW_SUBMIT
 %token KW_BIDIRECT
+%token KW_INSTALL KW_UNINSTALL KW_PLUGIN KW_SONAME KW_PLUGINS
```
darionyaphet: `.so`'s name is a key word ?

src/parser/scanner.lex
```diff
@@ -303,10 +309,14 @@ RECOVER                     ([Rr][Ee][Cc][Oo][Vv][Ee][Rr])
...
 {BIDIRECT}                  { return TokenType::KW_BIDIRECT; }
-
 {JOBS}                      { return TokenType::KW_JOBS; }
 {JOB}                       { return TokenType::KW_JOB; }
 {RECOVER}                   { return TokenType::KW_RECOVER; }
+{INSTALL}                   { return TokenType::KW_INSTALL; }
+{UNINSTALL}                 { return TokenType::KW_UNINSTALL; }
+{PLUGIN}                    { return TokenType::KW_PLUGIN; }
+{SONAME}                    { return TokenType::KW_SONAME; }
+{PLUGINS}                    { return TokenType::KW_PLUGINS; }
```
darionyaphet: Remove the space

src/parser/test/ParserTest.cpp
```diff
@@ -1694,4 +1694,20 @@ TEST(Parser, Return) {
         ASSERT_TRUE(result.ok()) << result.status();
     }
 }
+
+TEST(Parser, InstallOrUninstallPlugin) {
+    {
+        GQLParser parser;
+        std::string query = "INSTALL PLUGIN authentication SONAME \"authentication_ldap.so\"";
```
darionyaphet: Could we control the so file's name same with the plugin name？  

if the plugin is named `authentication` and the so is `authentication.so`

src/parser/parser.yy
panda-sheep: Good catch，it may cause by mixing tabs or spaces.

src/parser/scanner.lex
panda-sheep: good catch

src/parser/scanner.lex
panda-sheep: good catch!

src/meta/test/InstallUninstallPluginTest.cpp
```diff
@@ -0,0 +1,146 @@
...
+#include "meta/processors/admin/UninstallPluginProcessor.h"
+#include "meta/processors/admin/InstallPluginProcessor.h"
+#include "meta/processors/admin/ListPluginsProcessor.h"
+#include <unistd.h>
+
+namespace nebula {
+namespace meta {
+
+TEST(InstallUninstallPluginTest, Test) {
+    fs::TempDir rootPath("/tmp/CreateUserTest.XXXXXX");
```
darionyaphet: why create user?

src/meta/test/InstallUninstallPluginTest.cpp
```diff
@@ -0,0 +1,146 @@
+/* Copyright (c) 2020 vesoft inc. All rights reserved.
```
darionyaphet: This file named PluginTest.cpp is better ?

src/graph/InstallPluginExecutor.h
```diff
@@ -0,0 +1,37 @@
...
+        return "InstallPluginExecutor";
+    }
+
+    Status MUST_USE_RESULT prepare() override;
+
+    void execute() override;
+
+private:
+    InstallPluginSentence                *sentence_{nullptr};
+    const std::string                          *pluginName_{nullptr};
```
darionyaphet: remove the space

src/graph/UninstallPluginExecutor.h
```diff
@@ -0,0 +1,36 @@
...
+        return "UninstallPluginExecutor";
+    }
+
+    Status MUST_USE_RESULT prepare() override;
+
+    void execute() override;
+
+private:
+    UninstallPluginSentence                    *sentence_{nullptr};
+    const std::string                                *pluginName_{nullptr};
```
darionyaphet: remove the space

src/parser/parser.yy
panda-sheep: so file name is not a key word.
There is a keyword called `sonam`.
This keyword is followed by the name of the so file.
e.g 
`INSTALL PLUGIN auth SONAME "ldap.so"`

src/graph/InstallPluginExecutor.cpp
```diff
@@ -0,0 +1,78 @@
...
+namespace graph {
+
+InstallPluginExecutor::InstallPluginExecutor(Sentence *sentence,
+                                             ExecutionContext *ectx)
+    : Executor(ectx, "install_plugin") {
+    sentence_ = static_cast<InstallPluginSentence*>(sentence);
+}
+
+Status InstallPluginExecutor::prepare() {
+    // Move this part of the operation to meta
```
critical27: Why we `Move this part of the operation to meta`, If this PR is later for ldap, I suppose the graphd is the one to load?

src/meta/processors/admin/InstallPluginProcessor.cpp
```diff
@@ -0,0 +1,67 @@
...
+        LOG(ERROR) << "Install plugin Failed : Get plugin id failed";
+        handleErrorCode(nebula::error(idRet));
+        onFinished();
+        return;
+    }
+
+    auto pluginId = nebula::value(idRet);
+    std::vector<kvstore::KV> data;
+
+    data.emplace_back(MetaServiceUtils::indexPluginKey(pluginName),
```
critical27: Do we need to identify that which host has load a plugin or not? If A has install a plugin, I don't think that B has install it as well. Do I understand this correctly?

src/graph/InstallPluginExecutor.cpp
panda-sheep: please ignore the comments, i will delete.

The current implementation logic is:
The install plugin just goes to the meta registration information,
uninstall plugin go to metad to delete registration information

The real place to open the so file is when in use, close the so file after use
For example,  ldap
graphd goes to metad to get the plugin information, then open the so file, and close the so file after use.
so here open so file and use so in graphd

src/meta/processors/admin/InstallPluginProcessor.cpp
panda-sheep: Good point.  👍

The current implementation is that so files need  be stored in the corresponding directories of all machines.

The way you said maybe be supported in the future.

src/meta/test/InstallUninstallPluginTest.cpp
panda-sheep: good ponit， I will fix it

src/parser/test/ParserTest.cpp
panda-sheep: The plugin  name  and the so file name can be different.
But the so file name must be the same as the function name  we call  in the so file .

src/meta/processors/admin/InstallPluginProcessor.cpp
```diff
@@ -0,0 +1,67 @@
...
+namespace nebula {
+namespace meta {
+
+// When installing, just register it, do not always open it, open it when using
+void InstallPluginProcessor::process(const cpp2::InstallPluginReq& req) {
+    const auto& pluginItem = req.get_item();
+    auto& pluginName = pluginItem.get_plugin_name();
+    auto& soName = pluginItem.get_so_name();
+    folly::SharedMutex::WriteHolder wHolder(LockUtils::pluginLock());
+
```
bright-starry-sky: A question? using so file in graph layer, and register so file in meta layer. Therefore, how to ensure that the file exists in the graph server? because some times maybe graphd and metad on different systems.

panda-sheep: good question.

At present, there are so files in the corresponding directories of all machines, include graphd and metad.

panda-sheep:
> Well done.
> 
> Please consider that there might be other kinds of plugins in the future. I mean the plugin mechanism is not only for authentication. You have to take care of this situation in the design of the interfaces and logics.

Good idea.
In the future, UDF functions in the so file format can use the plugin mechanism.
Any good ideas, welcome suggestions.

src/meta/processors/admin/InstallPluginProcessor.cpp
```diff
@@ -0,0 +1,67 @@
...
+// When installing, just register it, do not open it. open it when using
+void InstallPluginProcessor::process(const cpp2::InstallPluginReq& req) {
+    const auto& pluginItem = req.get_item();
+    auto& pluginName = pluginItem.get_plugin_name();
+    auto& soName = pluginItem.get_so_name();
+    folly::SharedMutex::WriteHolder wHolder(LockUtils::pluginLock());
+
+    // Check if file exists
+    using fs::FileUtils;
```
dangleptr: IMO we'd better put the so files in the install dir of nebula.

dutor:
> > Well done.
> > Please consider that there might be other kinds of plugins in the future. I mean the plugin mechanism is not only for authentication. You have to take care of this situation in the design of the interfaces and logics.
> 
> Good idea.
> In the future, UDF functions in the so file format can use the plugin mechanism.
> Any good ideas, welcome suggestions.

For example, your current syntax definition is not expandable to plugins for other purposes.

src/meta/processors/admin/InstallPluginProcessor.cpp
panda-sheep: In order to put multiple so files, we currently put it in the share directory of the installation directory

panda-sheep:
> For example, your current syntax definition is not expandable to plugins for other purposes.

That is what I thought:
1） Here the syntax is to learn mysql
2） Only a plugin is defined here (that is, a so file). The functions in the so file can be used as global functions with a lower priority than built-in functions. Need to be implemented here in the future

src/graph/InstallPluginExecutor.cpp
```diff
@@ -0,0 +1,61 @@
...
+#include "base/Base.h"
+#include "graph/InstallPluginExecutor.h"
+#include <dlfcn.h>
+
+namespace nebula {
+namespace graph {
+
+InstallPluginExecutor::InstallPluginExecutor(Sentence *sentence,
+                                             ExecutionContext *ectx)
+    : Executor(ectx, "install_plugin") {
```
laura-ding: `InstallPluginExecutor` doesn't need stats, here doesn't set statName.

src/graph/InstallPluginExecutor.cpp
```diff
@@ -0,0 +1,61 @@
...
+
+void InstallPluginExecutor::execute() {
+    pluginName_ = sentence_->pluginName();
+    soName_ = sentence_->soName();
+    auto future = ectx()->getMetaClient()->installPlugin(*pluginName_, *soName_);
+    auto *runner = ectx()->rctx()->runner();
+
+    auto cb = [this] (auto &&resp) {
+        if (!resp.ok()) {
+            doError(Status::Error("Install plugin %s soname \"%s\" failed: %s",
```
laura-ding: `doError` and `doFinish` will judge if it needs to do stat, so here doesn't need.

src/graph/UninstallPluginExecutor.cpp
```diff
@@ -0,0 +1,58 @@
...
+
+#include "base/Base.h"
+#include "graph/UninstallPluginExecutor.h"
+
+namespace nebula {
+namespace graph {
+
+UninstallPluginExecutor::UninstallPluginExecutor(Sentence *sentence,
+                                                 ExecutionContext *ectx)
+    : Executor(ectx, "uninstall_plugin") {
```
laura-ding: ditto

panda-sheep:
Improve the following logic

Install plugins register in meta and then automatically loaded so file in graph
Uninstall plugins log out in meta then automatically dlclose so file in graphd

When restarting
Installed plugins loaded automatically in graphd
When the graphd is closing, opened plugins are automatically closed

src/graph/GraphService.cpp
```diff
@@ -10,14 +10,40 @@
...
+    bool loadDataOk = metaClient_->waitForMetadReady(3);
+    if (!loadDataOk) {
+        // Resort to retrying in the background
+        LOG(WARNING) << "Failed to synchronously wait for meta service ready";
+    }
+
     sessionManager_ = std::make_unique<SessionManager>();
-    executionEngine_ = std::make_unique<ExecutionEngine>();
     authenticator_ = std::make_unique<SimpleAuthenticator>();
+    pluginManager_ = std::make_unique<PluginManager>(metaClient_.get());
```
dangleptr: Why not define pluginManager inside ExecutionEngine directly.

src/meta/client/MetaClient.cpp
```diff
@@ -190,6 +191,11 @@ bool MetaClient::loadData() {
             return false;
         }
 
+        if (!loadPlugins(pluginItems)) {
```
dangleptr: 1.  You don't need to load plugins for each space.
2.  After loading the new plugins,  you should open them. Otherwise we have to restart the service.

src/graph/GraphService.cpp
panda-sheep: Because the authentication module outside ExecutionEngine also needs pluginManager

src/meta/client/MetaClient.cpp
panda-sheep: good catch, I will fix it

src/graph/InstallPluginExecutor.cpp
panda-sheep: Good catch.

src/graph/InstallPluginExecutor.cpp
panda-sheep: Good catch. I got.

src/graph/UninstallPluginExecutor.cpp
panda-sheep: fixed

src/meta/client/MetaClient.cpp
panda-sheep: has fixed.

src/graph/GraphService.cpp
dangleptr: authentication module is also inside ExecutionEngine

jude-zhu:
[POC-55]

PR #1898, opened 2020-03-10
fix pull request re-run failed

laura-ding:
When the pr run ci failed, we want to re-run, it will fail on checkout. 
reference from `https://github.com/actions/checkout/issues/23`


Shylock-Hg:
If so why not update all checkout action to v2

laura-ding:
@Shylock-Hg I think the others are unaffected, so I only modify the pull request.

PR #1908, opened 2020-03-12
add production conf into install

laura-ding:


etc/CMakeLists.txt
```diff
@@ -3,8 +3,11 @@
 install(
     FILES
         nebula-graphd.conf.default
```
whitewum: is it possible nebula-graph.*?

laura-ding: no

PR #1921, opened 2020-03-15
Dns resolve meta

monadbobo:

1. Refactored the HostAddr.
2. If the meta server's address is configured as a domain name, you can sense the change in the IP behind the domain name.


src/meta/client/MetaClient.cpp
```diff
@@ -107,6 +110,35 @@ void MetaClient::stop() {
     isRunning_ = false;
 }
 
+void MetaClient::dnsQUeryThreadFunc() {
+    SCOPE_EXIT {
+        bgThread_->addDelayTask(
+            FLAGS_meta_domain_ttl_secs * 1000, &MetaClient::dnsQUeryThreadFunc, this);
```
knightXun: If metad ip changes, We will wait for 1000s to get right ipaddress, right?

src/meta/client/MetaClient.cpp
```diff
@@ -107,6 +110,35 @@ void MetaClient::stop() {
...
+            } catch (std::exception& e) {
+                LOG(ERROR) << "update meta addr failed, err: " << e.what();
+            }
+            continue;
+        }
+    }
+    folly::SharedMutexReadPriority::WriteHolder holder(addrsLock_);
+    for (auto index : addrsIndex) {
+        LOG(INFO) << "meta will update host from " << addrs_[index] << " to: " << addrs[index];
+        addrs_[index] = std::move(addrs[index]);
```
critical27: Is this correct? Can we make sure that dns return the `addrs` in the order when we create `MetaClient`? I suppose that the `addrs_` is loaded from configuration.

src/kvstore/raftex/RaftPart.cpp
```diff
@@ -324,7 +323,7 @@ void RaftPart::stop() {
     {
         std::unique_lock<std::mutex> lck(raftLock_);
         status_ = Status::STOPPED;
-        leader_ = {0, 0};
+        leader_ = {};
```
critical27: How about `network::InetAddress(0, 0)`, it hard to tell what it is if we use `{}`;

src/common/thrift/ThriftClientManager.inl
```diff
@@ -32,21 +32,16 @@ std::shared_ptr<ClientType> ThriftClientManager<ClientType>::client(
...
     // Need to create a new client
-    auto ipAddr = network::NetworkUtils::intToIPv4(host.first);
-    auto port = host.second;
-    VLOG(2) << "There is no existing client to "
-            << ipAddr << ":" << port
-            << ", trying to create one";
+    VLOG(2) << "There is no existing client to " << host << ", trying to create one";
     auto channel = apache::thrift::ReconnectingRequestChannel::newChannel(
-        *evb, [compatibility, ipAddr, port, timeout] (folly::EventBase& eb) mutable {
+        *evb, [compatibility, &host, timeout] (folly::EventBase& eb) mutable {
```
dangleptr: Pass host by value.

src/common/network/InetAddress.h
```diff
@@ -0,0 +1,107 @@
...
+    bool isInitialized() const {
+        return addrs_.isInitialized();
+    }
+
+    size_t hash() const {
+        return addrs_.hash();
+    }
+
+private:
+    folly::SocketAddress addrs_;
```
dangleptr: I want to know why not use folly::SocketAddress directly.

src/meta/client/MetaClient.cpp
```diff
@@ -107,6 +110,35 @@ void MetaClient::stop() {
     isRunning_ = false;
 }
 
+void MetaClient::dnsQUeryThreadFunc() {
```
dangleptr: I am a little confused about this function. 
`addrs_` contains all meta servers addresses and all of them are represented by hostname currently,  right?
If so,  why we resolve them again?  The connection will be rebuild in ThriftClientManager

src/meta/client/MetaClient.cpp
```diff
@@ -107,6 +110,35 @@ void MetaClient::stop() {
...
+                if (newAddr != (*it)) {
+                    addrs.emplace_back(newAddr);
+                    addrsIndex.emplace_back(std::distance(addrs.cbegin(), it));
+                }
+            } catch (std::exception& e) {
+                LOG(ERROR) << "update meta addr failed, err: " << e.what();
+            }
+            continue;
+        }
+    }
```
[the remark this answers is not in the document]
monadbobo: this is my mistake, i have fixed this.

src/meta/client/MetaClient.cpp
monadbobo: The parameter of addDelayTask is milliseconds, so here needs to be multiplied by 1000.

src/meta/client/MetaClient.cpp
monadbobo: Addrs_ will be used in updateActive and updateLeader to update the corresponding IP address.

src/common/network/InetAddress.h
monadbobo: folly :: SocketAddress does not save the original domain name, and we need to resolve the domain name again.

src/meta/client/MetaClient.cpp
monadbobo: The IP address behind the domain name may change, and thrift does not automatically resolve the corresponding domain name.

@knightXun

src/common/network/InetAddress.h
```diff
@@ -0,0 +1,107 @@
...
+
+#include <folly/SocketAddress.h>
+#include "interface/gen-cpp2/common_types.h"
+
+namespace nebula {
+namespace network {
+class InetAddress final {
+public:
+    InetAddress() : InetAddress(0, 0) {}
+    explicit InetAddress(const std::string& host, uint16_t port, bool resolved = false)
```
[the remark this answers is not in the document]
monadbobo: Why is this redundant?

src/meta/client/MetaClient.cpp
dangleptr: > thrift does not automatically resolve the corresponding domain name.

I notice the ctor about folly::SocketAddress looks like 
```
 SocketAddress(const char* host, uint16_t port, bool allowNameLookup = false) 
```
So the third params means we could resolve the dns?

src/common/network/InetAddress.h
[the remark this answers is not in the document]
Shylock-Hg: AFAK, the explicit is used for single parameter constructor.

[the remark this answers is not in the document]
monadbobo: > A constructor with a single non-default parameter (until C++11) that is declared without the function specifier explicit is called a converting constructor.

https://en.cppreference.com/w/cpp/language/explicit @Shylock-Hg

src/meta/client/MetaClient.cpp
monadbobo: Yes, this parameter is to decide whether to perform dns lookup when constructing folly :: SocketAddress.

src/meta/client/MetaClient.cpp
critical27: Still a little confused about this code block. I suppose that what we need to do is try to resolve each meta hostname, and check the corresponding ip has updated or not.

Just like `bright-starry-sky  said, the `addrs` is empty, how can we update them? I guess we need to iterate `addrs_`, not `addrs`.

src/meta/client/MetaClient.cpp
dangleptr: So why not resolve the dns when rebuilding connection.

src/meta/client/MetaClient.cpp
monadbobo: yes, i have fixed this.

src/meta/client/MetaClient.cpp
monadbobo: This DNS lookup is performed only when the constructor is called, that is, after the object is constructed, the DNS lookup is no longer implicitly performed.

dangleptr: There are many ways to solve the problem.  For example, you could reset the original 'SocketAddress' instance or construct a new one. 

We don't need the loop here.

monadbobo: > There are many ways to solve the problem. For example, you could reset the original 'SocketAddress' instance or construct a new one.
> 
> We don't need the loop here.

First of all, dns query is blocking, so we must minimize the impact of blocking, so we need to do it in a separate thread.
Secondly, we don't know when the dns lookup is needed, so here we need to loop to keep looking up.

dangleptr: > First of all, dns query is blocking, so we must minimize the impact of blocking, so we need to do it in a separate thread.

Only build a new connection needs the dns LookUp. I think it is ok for the blocking. 

> Secondly, we don't know when the dns lookup is needed, so here we need to loop to keep looking up.

Actually, we know it.  It will be used on cloud, that means only the old host dead and replace the old host with a new one.  So when the old connection broken down,  we should retry to resolve the dns. 

The loop for dns resolve is not a good way.  The main reasons are:
1. The loop here is useless in most time because the hosts IPs are not always changed. It costs too much.
2. How to deal with the storaged hosts IPs with your approach?


WDTY?

src/meta/client/MetaClient.cpp
```diff
@@ -107,6 +110,35 @@ void MetaClient::stop() {
...
+void MetaClient::dnsQUeryThreadFunc() {
+    SCOPE_EXIT {
+        bgThread_->addDelayTask(
+            FLAGS_meta_domain_ttl_secs * 1000, &MetaClient::dnsQUeryThreadFunc, this);
+    };
+    // we need update addr from dns.
+    std::vector<network::InetAddress> addrs;
+    std::vector<int> addrsIndex;
+    for (auto it = addrs.cbegin(); it != addrs.cend(); ++it) {
+        if (it->isResolved()) {
```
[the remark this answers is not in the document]
monadbobo: If it fails then keep the original IP address.

src/meta/client/MetaClient.cpp
monadbobo: > > First of all, dns query is blocking, so we must minimize the impact of blocking, so we need to do it in a separate thread.
> 
> Only build a new connection needs the dns LookUp. I think it is ok for the blocking.
> 
> > Secondly, we don't know when the dns lookup is needed, so here we need to loop to keep looking up.
> 
> Actually, we know it. It will be used on cloud, that means only the old host dead and replace the old host with a new one. So when the old connection broken down, we should retry to resolve the dns.
> 
> The loop for dns resolve is not a good way. The main reasons are:
> 
> 1. The loop here is useless in most time because the hosts IPs are not always changed. It costs too much.
1. If only a certain IP address has a temporary network failure, then we do not necessarily need to re-block to do a DNS query.
2. In most cases, the DNS query is useless, but the current implementation will not affect any other operation of the system, because the corresponding addrs_ structure will be updated only when the IP address changes.
> 2. How to deal with the storaged hosts IPs with your approach?
This issue has not been resolved for the time being, I will fix it next
> 
> WDTY?

dangleptr: If you really worry about the blocking problem.  I suggest you put the loop inside ThriftClientManager,  because it has hold all hosts used in current service. And it could handle the storaged IP changes too.

monadbobo: I think ThriftClientManager is better to handle IP addresses only.

1. currently we only need to dynamically handle the domain name of the meta server, because storage sends its own IP to the meta server every time the storage is started, which means we do not support a storage server that does not restart and change its IP address.
2 Since clientMap_ is ThreadLocal, if you want to update the IP of a domain name in ThriftClientManager, it means we need to update clientMap_ in all threads within one thread, which I don't think is necessary.

src/meta/client/MetaClient.cpp
```diff
@@ -107,6 +110,34 @@ void MetaClient::stop() {
...
+    }
+
+    if (addrsIndex.empty()) {
+        return;
+    }
+
+    folly::SharedMutexReadPriority::WriteHolder holder(addrsLock_);
+    for (auto index : addrsIndex) {
+        LOG(INFO) << "meta will update host from " << addrs_[index] << " to: " << addrs[index];
+        addrs_[index] = std::move(addrs[index]);
```
wadeliuyi, line 137: 这里可能会越界吧？假如addrs_有a/b/c三个地址，在b调用network::NetworkUtils::updateHost(*it);的时候返回的是boost::none，a/c正常获得了值，这样addrs数组有两个值，但是addrsIndex是[0,2]，这样addrs[2]就越界了吧？

src/kvstore/NebulaStore.h
```diff
@@ -42,7 +42,7 @@ class NebulaStore : public KVStore, public Handler {
 public:
     NebulaStore(KVOptions options,
                 std::shared_ptr<folly::IOThreadPoolExecutor> ioPool,
-                HostAddr serviceAddr,
+                network::InetAddress serviceAddr,
```
wadeliuyi, line 45: const network::InetAddress &serviceAddr是不是更合适一点？

src/kvstore/raftex/RaftPart.cpp
```diff
@@ -367,38 +366,62 @@ AppendLogResult RaftPart::canAppendLogs() {
...
-void RaftPart::addLearner(const HostAddr& addr) {
+std::vector<std::shared_ptr<Host>>::iterator RaftPart::updateAndFindHost(
+    std::function<bool(std::shared_ptr<Host> &h)> f) {
+    CHECK(!raftLock_.try_lock());
+
+    std::vector<std::shared_ptr<Host>>::iterator itera = hosts_.end();
+    for (auto it = hosts_.begin(); it != hosts_.end(); ++it) {
+        auto hostOptions = network::NetworkUtils::updateHost((*it)->address());
+        if (hostOptions == boost::none) {
+            continue;
```
wadeliuyi, line 377: continue之前是不是需要调用一下f，不然有可能it的地址是查询过dns了的，返回了boost::none，但是iter的地址和输入的addr相等，但是直接continue了就不会返回正确的iter了

src/kvstore/raftex/RaftPart.cpp
```diff
@@ -468,50 +491,53 @@ void RaftPart::addPeer(const HostAddr& peer) {
...
     if (it == hosts_.end()) {
-        hosts_.emplace_back(std::make_shared<Host>(peer, shared_from_this()));
-        updateQuorum();
-        LOG(INFO) << idStr_ << "Add peer " << peer;
-    } else {
-        if ((*it)->isLearner()) {
-            LOG(INFO) << idStr_ << "The host " << peer
-                      << " has been existed as learner, promote it!";
-            (*it)->setLearner(false);
+        it = updateAndFindHost([&peer](const auto& h) { return h->address() == peer; });
```
wadeliuyi, line 497: 不需要每次都调用updateAndFindHost的吧？

dangleptr:
The pr has been committed into cloud branch.

PR #1922, opened 2020-03-16
Auto check keyword

laura-ding:
- If your code adds keywords, it will check the new keyword if in the reserved keyword or the unreserved keyword when you commit the code.
- If users want to use the reserved keyword, they can use `keyword`.

###  Does this PR introduce any user-facing change?
Yes @Amber1990Zhang

### DIFF
--- a/.linters/cpp/checkKeyword.py
+++ b/.linters/cpp/checkKeyword.py
@@ -1,0 +1,139 @@
+# signout--coding:utf-8--
+
+# Copyright (c) 2019 vesoft inc. All rights reserved.
+#
+# This source code is licensed under Apache 2.0 License,
+# attached with Common Clause Condition 1.0, found in the LICENSES directory.
+
+import os
+import sys
+import re
+
+
+PASER_FILE_PATH = 'src/parser/parser.yy'
+SCANNER_FILE_PATH = 'src/parser/scanner.lex'
+
+reserved_key_words = [
+    'KW_GO',
+    'KW_AS',
+    'KW_TO',
+    'KW_OR',
+    'KW_AND',
+    'KW_XOR',
+    'KW_USE',
+    'KW_SET',
+    'KW_FROM',
+    'KW_WHERE',
+    'KW_MATCH',
+    'KW_INSERT',
+    'KW_YIELD',
+    'KW_RETURN',
+    'KW_DESCRIBE',
+    'KW_DESC',
+    'KW_VERTEX',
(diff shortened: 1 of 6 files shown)


src/parser/parser.yy
```diff
@@ -247,6 +247,7 @@ class GraphScanner;
 
 name_label
      : LABEL { $$ = $1; }
+     | STRING { $$ = $1; }
```
Shylock-Hg: Empty string for filed name workable?
And I think the empty filed will be conflict with `ttl_col = ""`.

laura-ding: Good catch, the name_lable can't be empty.

.linters/cpp/checkKeyword.py
```diff
@@ -0,0 +1,139 @@
...
+    'KW_STOP',
+    'KW_LIMIT',
+    'KW_OFFSET',
+    'KW_GROUP',
+    'KW_IS',
+    'KW_NULL',
+    'KW_FORCE',
+    'KW_RECOVER'
+]
+
```
panda-sheep: Good!

I think when defining tokens in `scanner.lex`，  you should distinguish between `keywords `and` unreserved keywords`,    rather than mixing them together.

In this way, when adding a token later,  you know that it is added to the `keyword `or the` unreserved keyword.`

src/parser/scanner.lex
```diff
@@ -169,6 +169,7 @@ RECOVER                     ([Rr][Ee][Cc][Oo][Vv][Ee][Rr])
 
 %%
 
+ /* Reserved keyword */
```
panda-sheep: Good!

src/parser/scanner.lex
```diff
@@ -364,6 +368,17 @@ RECOVER                     ([Rr][Ee][Cc][Oo][Vv][Ee][Rr])
...
+                                yylval->strval = new std::string(yytext + 1, yyleng - 2);
+                                if (yylval->strval->size() > MAX_STRING) {
+                                    auto error = "Out of range of the LABEL length, "
+                                                  "the  max length of LABEL is " +
+                                                  std::to_string(MAX_STRING) + ":";
+                                    delete yylval->strval;
+                                    throw GraphParser::syntax_error(*yylloc, error);
+                                }
+                                return TokenType::LABEL;
+                            }
```
panda-sheep: Excellent！

PR #1928, opened 2020-03-16
reduce duplicate write

darionyaphet:
When we insert some vertex or edge into nebula, we can skip some duplicate sentence in the batch mode.


dangleptr:
We should not do this in server-side.  It is dangerous. We can't ensure the real purpose about our users when receiving duplicate rows. 

So close it.  Thanks

PR #1933, opened 2020-03-17
[raft] fix follower would start election even leader exists

critical27:
* Suppose we have A and B, A has higher `lastLogTerm`, B's `lastLogId` and `commitLogId` are greater. A start election, B would vote for A, A becomes leader.
* When B vote for A, B would set its term to A's term. And when A start to replicate logs, B would return E_LOG_STALE because its `commitLogId` is greater.
* When A see the E_LOG_STALE, it would set `lastLogIdSent_` in `Host` to B's `lastLogId`.

Then no matter what A send to B, it would just skip it including heartbeat (A thinks
 that it has been sent before). So B would keep starting election all the time.

So I modify that:
1. A would only set `lastLogIdSent_` to a reasonable value, `min(resp.get_last_log_id(), self->logIdToSend_ - 1)`
2. When B vote for A, it found its `commitLogId` is greater, it just reset it self.


src/kvstore/raftex/Host.cpp
```diff
@@ -119,6 +119,7 @@ folly::Future<cpp2::AppendLogResponse> Host::appendLogs(
 
         if (requestOnGoing_ && res == cpp2::ErrorCode::SUCCEEDED) {
             if (cachingPromise_.size() <= FLAGS_max_outstanding_requests) {
+                logIdToSend_ = logId;
```
dangleptr: Good catch

src/kvstore/raftex/Host.cpp
```diff
@@ -303,7 +304,8 @@ void Host::appendLogsInternal(folly::EventBase* eb,
                         r.set_error_code(cpp2::ErrorCode::SUCCEEDED);
                         self->setResponse(r);
                     } else {
-                        self->lastLogIdSent_ = resp.get_last_log_id();
+                        self->lastLogIdSent_ = std::min(resp.get_last_log_id(),
```
dangleptr: If SUCCEEDED returned,  it means follower accept all logs,  so resp.get_last_log_id() should be less than self->logIdToSend_, right?

src/kvstore/raftex/Host.cpp
```diff
@@ -369,7 +371,8 @@ void Host::appendLogsInternal(folly::EventBase* eb,
...
```
dangleptr: The line 367 should take the same way?

src/kvstore/raftex/RaftPart.cpp
```diff
@@ -1395,6 +1395,14 @@ void RaftPart::processAskForVoteRequest(
             });
     }
 
+    if (req.get_last_log_id() < committedLogId_) {
```
dangleptr: why not use req.get_committed_log_id()

src/kvstore/raftex/Host.cpp
critical27: Yeah, so when SUCCEEDED returned, we just set `lastLogIdSent_ = resp.get_last_log_id()`, but in E_LOG_GAP, I think we'd better compare them?

src/kvstore/raftex/Host.cpp
critical27: Good catch

src/kvstore/raftex/RaftPart.cpp
critical27: I think leader's lastLogId is greater than its commitLogId, and we don't have a commitLogId in `AskForVoteRequest`, do you think it is necessary to add one?

dangleptr: >  leader's lastLogId is greater than its commitLogId

Not exactly,  when catching up data,  maybe lastLogId is less than commitLogId

src/kvstore/raftex/Host.cpp
dangleptr: Fine.  Look the same.

src/kvstore/raftex/RaftPart.cpp
```diff
@@ -1395,6 +1395,14 @@ void RaftPart::processAskForVoteRequest(
...
+        LOG(INFO) << idStr_ << "I believe that the leader has higher term " << term_
```
dangleptr: Maybe  WARNING is better.

src/kvstore/raftex/RaftPart.cpp
critical27: You're right.

src/interface/raftex.thrift
```diff
@@ -51,6 +51,8 @@ struct AskForVoteRequest {
     5: TermID              term;               // Proposed term
     6: LogID               last_log_id;        // The last received log id
     7: TermID              last_log_term;      // The term receiving the last log
+    8: LogID               committed_log_id;   // Last committed Log ID
+    9: bool                is_transfer;        // Trigger by transfer leader or not
```
dangleptr: What's the purpose about 'is_transfer'?

critical27: There is a corner case, if we start election because of transLeader, the target host's committedLogId would be smaller than old leader's, which would cause old leader `reset`. I have not figured out a better way, do you have any suggestion?

wadeliuyi:
I do not clear what case can let a raft group go into "Suppose we have A and B, A has higher lastLogTerm, B's lastLogId and commitLogId are greater. A start election, B would vote for A, A becomes leader."  If A has bigger lastLogTerm, that mean A must receive append log request from leader, and then we can derive that A must have the log which id==B(commitLogId), because if B commit the log, that mean leader must commit the log too.

wadeliuyi:
"When B vote for A, it found its commitLogId is greater, it just reset it self." in raft commit mean the log has been "Durability" storage in a majority member, maybe we can not reset the commit log directly, it break the rules of raft.

src/kvstore/raftex/RaftPart.cpp
```diff
@@ -1395,6 +1398,14 @@ void RaftPart::processAskForVoteRequest(
             });
     }
 
+    if (!req.get_is_transfer() && req.get_committed_log_id() < committedLogId_) {
```
wadeliuyi: commited mean the log has been write to the state machine, it is a "Durability" log, maybe we can not reset it directly.

critical27:
> I do not clear what case can let a raft group go into "Suppose we have A and B, A has higher lastLogTerm, B's lastLogId and commitLogId are greater. A start election, B would vote for A, A becomes leader." If A has bigger lastLogTerm, that mean A must receive append log request from leader, and then we can derive that A must have the log which id==B(commitLogId), because if B commit the log, that mean leader must commit the log too. So maybe we should find the root cause why this happen?

In theory you are right. But if it really happen, you need to make the cluster stable. This happened when A and C were in same state, B is weird with extreme high lastLogId and commitLogId, you do need to make a choice.

src/interface/raftex.thrift
dangleptr: Em...   Let me think about it.  It is a read good case.

PR #1937, opened 2020-03-17
fix fetch prop on * vertexID with ttl

### LINKED ISSUE #1932: fetch * returns expired data
nebula version Git: daa23e09, Build Time: Feb 26 2020

create two tags with ttl
insert a vertex labeled with these two tags
`fetch * ` returns the tag props which should be expired.

panda-sheep:
fix   #1932   fetch * returns expired data
close #1932


bright-starry-sky:
Ready to review?

panda-sheep:
> Ready to review?

Yeah，I forgot to tag  “ ready-for-testing”

panda-sheep:
> LGTM !
> A suggest, extracting the buildEdgeTTLInfo same with buildTagTTLInfo as well.

Good catch. 
I thought about what you said.
Different from tag, because build edge ttl info is used in one place.

For consistency of code, I will fixed it

src/storage/query/QueryVertexPropsProcessor.cpp
```diff
@@ -88,6 +88,9 @@ kvstore::ResultCode QueryVertexPropsProcessor::collectVertexProps(
             // Already found the latest version.
             continue;
         }
+        // Build TTL info
+        buildTagTTLInfo(tagId);
```
dangleptr: Could you combine it with getTagTTLInfo.  We don't need to lookup the hash map twice.

panda-sheep: Good catch，fixed it.
Please review it again.

panda-sheep:
> > LGTM !
> > A suggest, extracting the buildEdgeTTLInfo same with buildTagTTLInfo as well.
> 
> Good catch.
> I thought about what you said.
> Different from tag, because build edge ttl info is used in one place.
> 
> For consistency of code, I will fixed it
not extract the buildEdgeTTLInfo  now,   please ignore it

panda-sheep:
please block it
I simplify the code again

panda-sheep:
ok ，please review again

src/storage/query/QueryBaseProcessor.inl
```diff
@@ -171,10 +171,49 @@ cpp2::ErrorCode QueryBaseProcessor<REQ, RESP>::checkAndBuildContexts(const REQ&
 
 template<typename REQ, typename RESP>
 folly::Optional<std::pair<std::string, int64_t>>
-QueryBaseProcessor<REQ, RESP>::getTagTTLInfo(TagID tagId) {
+QueryBaseProcessor<REQ, RESP>::getTagTTLInfo(TagID tagId, bool addTagInfo) {
     folly::Optional<std::pair<std::string, int64_t>> ret;
     auto tagFound = tagTTLInfo_.find(tagId);
-    if (tagFound != tagTTLInfo_.end()) {
+
+    if (tagFound == tagTTLInfo_.end()) {
```
dangleptr: You have filled the tagTTLInfo_ in buildTTLInfoAndRespSchema, right?
So the changes here is just for fetch *,  right?

I suggest you extract the common logic about filling tagTTLInfo_.   And move the function into QueryVertexPropsProcessor

panda-sheep: Yeah，`buildTTLInfoAndRespSchema `  function fill  the tagTTLInfo_ 
But` fetch * ` do not  do   `buildTTLInfoAndRespSchema ` in `QueryVertexPropsProcessor::process`

panda-sheep: `I suggest you extract the common logic about filling tagTTLInfo_. And move the function into QueryVertexPropsProcessor`
ok

panda-sheep:
pr is ready, please continue

PR #1941, opened 2020-03-18
Modify the Dockerfile which package docker

laura-ding:


docker/Dockerfile.console
```diff
@@ -1,14 +1,15 @@
 FROM vesoft/nebula-dev:centos7 as builder
 
+RUN echo `pwd`
```
Shylock-Hg: Why add this?

laura-ding: Good catch. I will delete it.

PR #1946, opened 2020-03-20
Modify packaging script to support static link sanitizer.

monadbobo:

modify packaging script to support static link sanitizer.


monadbobo:
depend #1861

PR #1947, opened 2020-03-20
change default wal ttl from 1 day to 4 hours.

whitewum:

4 hours wal recycle is enough for most cases.

wal takes too much ssd disk spaces.

No.

Not necessary.

### DIFF
--- a/src/kvstore/raftex/RaftPart.cpp
+++ b/src/kvstore/raftex/RaftPart.cpp
@@ -28,7 +28,7 @@ DEFINE_uint64(raft_snapshot_timeout, 60 * 5, "Max seconds between two snapshot r
 
 DEFINE_uint32(max_batch_size, 256, "The max number of logs in a batch");
 
-DEFINE_int32(wal_ttl, 86400, "Default wal ttl");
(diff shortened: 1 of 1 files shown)


critical27:
I suggest change this parameter in storage.conf instead of here, it won't work if we don't update them in console.

whitewum:
> I suggest change this parameter in storage.conf instead of here, it won't work if we don't update them in console.

As far as I know, all the (old) users have changed their conf file, or they are ok with 1 day. So there is no upgrade-and-change-congfigure-file problem.

But for new users, a new parameter is enough.

Why do you prefer parameter (1 day) in the source code?

critical27:
> > I suggest change this parameter in storage.conf instead of here, it won't work if we don't update them in console.
> 
> As far as I know, all the (old) users have changed their conf file, or they are ok with 1 day. So there is no upgrade-and-change-congfigure-file problem.
> 
> But for new users, a new parameter is enough.
> 
> Why do you prefer parameter (1 day) in the source code?

After a second thought, I think it's ok to replace it with 4h.

dangleptr:
The issue has been addressed in #1959  
Close it.

PR #1948, opened 2020-03-20
upgrade rocksdb to 6.7.3

dangleptr:
Because we need some new features in the latest rocksdb,  so upgrade it to the newest version. 

Test it locally with gcc 7.5.0 on centos 7,  works quite well.


Shylock-Hg:
The CI seems not running really.But not sure lead by this PR.

bright-starry-sky:
Do you want to upgrade in rc4? If so, merge as soon as possible. The code will freeze today.

dangleptr:
> Do you want to upgrade in rc4? If so, merge as soon as possible. The code will freeze today.

Not necessary for RC4.

PR #1951, opened 2020-03-21
Detect OS type and exit if not built in Linux

JaySon-Huang:

Make it more friendly to tell users that now building nebula under MacOS X is not supported.

When I try to build nebula in my macbook, it print error like this.
```
>  cmake -DENABLE_TESTING=OFF ..
ld: unknown option: --version
CMake Error at cmake/LinkerConfig.cmake:12 (if):
  if given arguments:

    "STREQUAL" "ld"

  Unknown arguments specified
Call Stack (most recent call first):
  CMakeLists.txt:35 (include)
```
After ` cmake/LinkerConfig.cmake` fixed, it print error like this. And I found that "Only Linux is supported".
```
>  cmake -DENABLE_TESTING=OFF ..
ld: unknown option: --version
-- NEBULA_USE_LINKER: 
-- ENABLE_ASAN: OFF
-- ENABLE_TESTING: OFF
-- ENABLE_UBSAN: OFF
-- ENABLE_FUZZ_TEST: OFF
-- ENABLE_TSAN: OFF
-- ENABLE_STATIC_ASAN : OFF
-- ENABLE_STATIC_UBSAN : OFF
-- CCACHE: enabled but not found
-- CMAKE_BUILD_TYPE = Debug (Options are: Debug, Release, RelWithDebInfo, MinSizeRel)
-- CMAKE_INSTALL_PREFIX: /usr/local/nebula
-- Downloading prebuilt third party automatically...
usage: basename string [suffix]
       basename [-a] [-s suffix] string [...]
usage: basename string [suffix]
       basename [-a] [-s suffix] string [...]
usage: basename string [suffix]
       basename [-a] [-s suffix] string [...]
usage: basename string [suffix]
       basename [-a] [-s suffix] string [...]
usage: basename string [suffix]
       basename [-a] [-s suffix] string [...]
usage: basename string [suffix]
       basename [-a] [-s suffix] string [...]
```

I think it is better to detect OS type in CMake and let it exit more earlier with friendly message.

No

Manual test

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -30,6 +30,13 @@ project("Nebula Graph" C CXX)
 
 # Set the project home dir
 set(NEBULA_HOME ${CMAKE_CURRENT_SOURCE_DIR})
+
+if(${CMAKE_SYSTEM_NAME} STREQUAL "Darwin")
+	MESSAGE(FATAL_ERROR "Only Linux is supported")
+elseif (${CMAKE_SYSTEM_NAME} STREQUAL "Windows")
+	MESSAGE(FATAL_ERROR "Only Linux is supported")
+endif ()
+
 # To include customized FindXXX.cmake modules
 set(CMAKE_MODULE_PATH "${NEBULA_HOME}/cmake" ${CMAKE_MODULE_PATH})
 include(LinkerConfig)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -31,9 +31,7 @@ project("Nebula Graph" C CXX)
 # Set the project home dir
 set(NEBULA_HOME ${CMAKE_CURRENT_SOURCE_DIR})
 
-if(${CMAKE_SYSTEM_NAME} STREQUAL "Darwin")
-	MESSAGE(FATAL_ERROR "Only Linux is supported")
-elseif (${CMAKE_SYSTEM_NAME} STREQUAL "Windows")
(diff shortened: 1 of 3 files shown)


cmake/LinkerConfig.cmake
```diff
@@ -9,7 +9,7 @@ execute_process(
     OUTPUT_STRIP_TRAILING_WHITESPACE
 )
 
-if (${default_linker_type} STREQUAL "ld")
+if ("${default_linker_type}" STREQUAL "ld")
```
dutor: Good catch.

CMakeLists.txt
```diff
@@ -30,6 +30,13 @@ project("Nebula Graph" C CXX)
 
 # Set the project home dir
 set(NEBULA_HOME ${CMAKE_CURRENT_SOURCE_DIR})
+
+if(${CMAKE_SYSTEM_NAME} STREQUAL "Darwin")
```
dutor: I think it would be better to detect if the host OS is one of our supported candidates, and print it out and exit if not.
Otherwise, _FreeBSD_ should be considered in this unsupported list.

JaySon-Huang: I am not sure about other Linux system, just disable for Darwin(MacOS), FreeBSD and Windows for now.

dutor: Fair enough.

PR #1956, opened 2020-03-24
Simplify index storage

darionyaphet:
The tag and edge processing in storage layer looks very similar so I try to merge them into one


src/interface/storage.thrift
```diff
@@ -513,6 +509,6 @@ service StorageService {
     GetUUIDResp getUUID(1: GetUUIDReq req);
 
     // Interfaces for edge and vertex index scan
-    LookUpVertexIndexResp lookUpVertexIndex(1: LookUpIndexRequest req);
-    LookUpEdgeIndexResp   lookUpEdgeIndex(1: LookUpIndexRequest req);
+    LookUpIndexResp   lookUpVertexIndex(1: LookUpIndexRequest req);
+    LookUpIndexResp   lookUpEdgeIndex(1: LookUpIndexRequest req);
```
bright-starry-sky, line 513: Overall looks good to me. 
Would be more perfect if `lookUpVertexIndex` and `lookUpVertexIndex` unify into `lookupIndex`

src/storage/client/StorageClient.cpp
```diff
@@ -513,15 +513,15 @@ StorageClient::get(GraphSpaceID space,
                            });
 }
 
-folly::SemiFuture<StorageRpcResponse<storage::cpp2::LookUpVertexIndexResp>>
+folly::SemiFuture<StorageRpcResponse<storage::cpp2::LookUpIndexResp>>
 StorageClient::lookUpVertexIndex(GraphSpaceID space,
                                  IndexID indexId,
```
bright-starry-sky, line 518: Could we unify `StorageClient::lookUpVertexIndex` and `StorageClient::lookUpEdgeIndex` to `StorageClient::lookupIndex`.
They are have same code now.

src/storage/client/StorageClient.cpp
```diff
@@ -544,15 +545,15 @@ StorageClient::lookUpVertexIndex(GraphSpaceID space,
                            });
 }
 
-folly::SemiFuture<StorageRpcResponse<storage::cpp2::LookUpEdgeIndexResp>>
+folly::SemiFuture<StorageRpcResponse<storage::cpp2::LookUpIndexResp>>
 StorageClient::lookUpEdgeIndex(GraphSpaceID space,
                                IndexID indexId,
                                std::string filter,
                                std::vector<std::string> returnCols,
                                folly::EventBase *evb) {
```
bright-starry-sky, line 553: Here , add A new parameter isEdge as better? so we can unify the similar client api. WTYT?

bright-starry-sky:
By the way, Is this improvement to be merged to nebula 1.0?

src/storage/client/StorageClient.cpp
darionyaphet: Because I have support `is_edge` in `LookUpIndexRequest` so we can merge the two interface into one. I think it's a good idea :D

src/storage/index/LookUpIndexProcessor.cpp
```diff
@@ -1,20 +1,21 @@
...
 namespace storage {
 
-void LookUpEdgeIndexProcessor::process(const cpp2::LookUpIndexRequest& req) {
+void LookUpIndexProcessor::process(const cpp2::LookUpIndexRequest& req) {
     /**
      * step 1 : prepare index meta and structure of return columns.
      */
     auto ret = prepareRequest(req);
     if (ret != cpp2::ErrorCode::SUCCEEDED) {
+        LOG(ERROR) << "Prepare Request " << "";
```
bright-starry-sky, line 18: `LOG(ERROR) << "Prepare Request failed"`

src/storage/index/LookUpIndexProcessor.cpp
```diff
@@ -24,6 +25,7 @@ void LookUpEdgeIndexProcessor::process(const cpp2::LookUpIndexRequest& req) {
      */
     ret = buildExecutionPlan(req.get_filter());
     if (ret != cpp2::ErrorCode::SUCCEEDED) {
+        LOG(ERROR) << "Build Execution Plan";
```
bright-starry-sky, line 28: ditto

src/storage/index/LookUpIndexProcessor.cpp
```diff
@@ -59,8 +61,16 @@ void LookUpEdgeIndexProcessor::process(const cpp2::LookUpIndexRequest& req) {
         }
         s.set_columns(std::move(cols));
         this->resp_.set_schema(std::move(s));
+    } else {
+        LOG(ERROR) << "schema is null";
     }
```
bright-starry-sky, line 66: I think this is redundant. If schema is null means the yield clause is null , There is no need to return schema . This is not a error.

src/storage/index/LookUpIndexProcessor.h
```diff
@@ -0,0 +1,40 @@
...
+    void process(const cpp2::LookUpIndexRequest& req);
+
+private:
+    explicit LookUpIndexProcessor(kvstore::KVStore* kvstore,
+                                  meta::SchemaManager* schemaMan,
+                                  meta::IndexManager* indexMan,
+                                  stats::Stats* stats,
+                                  VertexCache* cache = nullptr)
+        : IndexExecutor<cpp2::LookUpIndexResp>
+                (kvstore, schemaMan, indexMan, stats, cache) {}
```
bright-starry-sky, line 34: alignment ?

src/graph/LookupExecutor.cpp
```diff
@@ -316,10 +316,7 @@ LookupExecutor::findValidIndex() {
 void LookupExecutor::stepEdgeOut() {
     auto *sc = ectx()->getStorageClient();
     auto filter = Expression::encode(sentence_->whereClause()->filter());
```
bright-starry-sky, line 318: Could we unify the `stepEdgeOut` and `stepVertexOut` to `stepOut` ?

src/storage/index/LookUpIndexProcessor.cpp
darionyaphet: 👍

src/graph/LookupExecutor.cpp
darionyaphet: perhaps `stepOut` is not very suitable for this context, ·lookUp· is a good name? 

What do you think?

src/storage/index/LookUpIndexProcessor.cpp
liuyu85cn: also print the ret code ?

src/graph/LookupExecutor.cpp
```diff
@@ -615,25 +572,25 @@ LookupExecutor::toThriftResponse(EdgeRpcResponse&& eRpcResp,
...
-bool LookupExecutor::processFinalEdgeResult(EdgeRpcResponse &rpcResp, const Callback& cb) const {
+bool LookupExecutor::processFinalEdgeResult(RpcResponse &rpcResp, const Callback& cb) const {
     auto& all = rpcResp.responses();
     std::vector<nebula::cpp2::SupportedType> colTypes;
     std::vector<VariantType> record;
     record.reserve(returnCols_.size() + 3);
     std::shared_ptr<ResultSchemaProvider> schema = nullptr;
     for (auto &resp : all) {
-        if (!resp.__isset.rows || resp.get_rows() == nullptr || resp.rows.empty()) {
+        if (!resp.__isset.edges || resp.get_edges() == nullptr || resp.get_edges()->empty()) {
```
liuyu85cn, line 593: may be the second check is redundant ?

src/graph/LookupExecutor.cpp
bright-starry-sky: > perhaps `stepOut` is not very suitable for this context, ·lookUp· is a good name?
> 
> What do you think?

Aha, I think the setpOut is good . because it unify with go statement.

bright-starry-sky:
Well done. LGTM. 
because the thrift was change . do we need to sync it to nebula-common ?

darionyaphet:
ok I will update nebula common later ~

src/graph/LookupExecutor.cpp
darionyaphet: make sure the elements is not empty

PR #1959, opened 2020-03-24
use bloom filter, add event listener to collect stats of compaction, improve update rocksdb options

### LINKED ISSUE #1966: Some rocksdb option wrong, storaged upgrade and can't start
ERROR:

update rocksdb_column_family_options from {"write_buffer_size":"67108864","max_write_buffer_number":"4","max_bytes_for_level_base":"268435456","bloom_locality":"1"} to {"max_background_compactions":"4","level0_file_num_compaction_trigger":"2","bloom_locality":"1","disable_auto_compactions":"false","max_background_jobs":"8","max_bytes_for_level_base":"268435456","max_write_buffer_number":"4","write_buffer_size":"67108864"}

I upgrade storaged version from Feb.20 to Mar 24.

I've update config several times.

I think it's due to rocksdb options check.

critical27:
1. Use bloom filter by default
2. Add some options to open partitioned index/filter
3. Add a EventListener to collect stats of compaction.
4. Improve udpate rocksdb configs, for those options which can't change dynamically, it would take effect after reboot.

@Amber1990Zhang, would you help to update the doc `configs-syntax`, because some options is not supported anymore, thx.

close #1966


src/kvstore/EventListner.h
```diff
@@ -0,0 +1,28 @@
...
+ */
+
+#include "base/Base.h"
+#include "rocksdb/db.h"
+#include "rocksdb/listener.h"
+
+namespace nebula {
+namespace kvstore {
+
+class EventListener : public rocksdb::EventListener {
```
darionyaphet: Good Job

src/kvstore/raftex/RaftPart.cpp
```diff
@@ -28,7 +28,7 @@ DEFINE_uint64(raft_snapshot_timeout, 60 * 5, "Max seconds between two snapshot r
 
 DEFINE_uint32(max_batch_size, 256, "The max number of logs in a batch");
 
-DEFINE_int32(wal_ttl, 86400, "Default wal ttl");
+DEFINE_int32(wal_ttl, 14400, "Default wal ttl");
```
darionyaphet: current is 4 hour ?

critical27: Current is 1 day, @whitewum suggest change it shorter.

src/kvstore/RocksEngineConfig.cpp
```diff
@@ -87,6 +91,15 @@ rocksdb::Status initRocksdbOptions(rocksdb::Options &baseOpts) {
     static std::shared_ptr<rocksdb::Cache> blockCache
         = rocksdb::NewLRUCache(FLAGS_rocksdb_block_cache * 1024 * 1024);
     bbtOpts.block_cache = blockCache;
+    bbtOpts.filter_policy.reset(rocksdb::NewBloomFilterPolicy(10, false));
+    if (FLAGS_enable_partitioned_index_filter) {
```
whitewum: how about add prefix bloom filter? for the prefix 12bytes

critical27: Good suggestion. Later I will check if it improves perf.

src/meta/processors/configMan/SetConfigProcessor.cpp
```diff
@@ -14,7 +14,7 @@ std::unordered_set<std::string> SetConfigProcessor::mutableFields_ = {
     // rocksdb_column_family_options
     "disable_auto_compactions",
     // TODO: write_buffer_size will cause rocksdb crash
-    // "write_buffer_size",
+    "write_buffer_size",
```
bright-starry-sky: "write_buffer_size" works well now ? if yes , do we need to uncomment in `MetaServerBasedPartManager::onSpaceOptionUpdated`?

bright-starry-sky: > "write_buffer_size" works well now ? if yes , do we need to uncomment in `MetaServerBasedPartManager::onSpaceOptionUpdated`?

Sorry, the comment has been removed. I'm wrong.

src/common/base/Configuration.cpp
```diff
@@ -160,7 +164,7 @@ Status Configuration::fetchAsSubConf(const char *key, Configuration &subconf) co
 }
 
 
-Status Configuration::updateStringField(const char* key, const std::string& val) {
+Status Configuration::upsertStringField(const char* key, const std::string& val) {
```
bright-starry-sky: A question, Why naming `upsert`?
What I understand that upsert is : should be override the old value  if key exists;  should be insert  a new value if the key doesn't exist.

critical27: Exactly, so we can insert some option which doesn't exists, and update some option with a new value. The function actually is a upsert, a bad name.

src/meta/processors/configMan/SetConfigProcessor.cpp
critical27: Forgot to remove the TODO : )

src/kvstore/PartManager.cpp
```diff
@@ -96,25 +96,19 @@ void MetaServerBasedPartManager::onSpaceOptionUpdated(
...
         "max_bytes_for_level_base",
         "max_bytes_for_level_multiplier",
-        "ttl",
-        "block_size",
-        "block_restart_interval"
     };
     static std::unordered_set<std::string> supportedDbOpt = {
         "max_total_wal_size",
         "delete_obsolete_files_period_micros",
         "max_background_jobs",
```
dangleptr: support sub_maxcompactions

critical27: it can't be changed dynamically, for those options, it would take effect when reboot.

PR #1962, opened 2020-03-24
Fixed a bug caused by not filtering the versions of edges

monadbobo:

When over * was used, multiple versions of edge are not filtered, so an error will be reported at this time.


src/graph/GoExecutor.cpp
```diff
@@ -233,6 +234,12 @@ Status GoExecutor::prepareOverAll() {
             return status;
         }
 
+        if (edges.find(std::abs(v)) != edges.cend()) {
```
CPWstatic: We need to make sure that we get the latest version, a set may not be helpful.

laura-ding: I agree with @CPWstatic, Keep get from by getAllEdge is lastest

monadbobo: Since our keys are ordered, the latest version of each key was the first. 

for example  in GetEdgeProcessor.cpp:
  ```
  // Get the lastest version
    if (req.get_version() < 0) {
        auto edgePrefix = MetaServiceUtils::schemaEdgePrefix(req.get_space_id(), edgeType);
        auto ret = doPrefix(edgePrefix);
        if (!ret.ok()) {
            LOG(ERROR) << "Get Edge SpaceID: " << req.get_space_id() << ", edgeName: "
                       << req.get_edge_name() << ", version " << req.get_version() << " not found";
            handleErrorCode(cpp2::ErrorCode::E_NOT_FOUND);
            onFinished();
            return;
        }
        schemaValue = ret.value()->val().str();
    }
```

CPWstatic: Well, better to add comments here to tell this.

PR #1964, opened 2020-03-24
Support cloud account authentication

panda-sheep:


src/common/encryption/Base64.cpp
```diff
@@ -0,0 +1,118 @@
...
+const char Base64::encodeTab_[] =
+    "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/";
+
+const int Base64::decodeTab_[] = {
+    -1, -1, -1, -1, -1, -1, -1, -1, -1, -1,
+    -1, -1, -1, -1, -1, -1, -1, -1, -1, -1,
+    -1, -1, -1, -1, -1, -1, -1, -1, -1, -1,
+    -1, -1, -1, -1, -1, -1, -1, -1, -1, -1,
+    -1, -1, -1, 62,    // '+'
+    -1, -1, -1, 63,    // '/'
```
bright-starry-sky: Why not use `proxygen::base64Encode() ` ？

src/graph/CloudAuthenticator.cpp
```diff
@@ -0,0 +1,66 @@
...
+    // Normal passwords and tokens use different prefixes
+
+    // First, go to meta to check if the shadow account exists
+    if (!metaClient_->checkShadowAccountFromCache(user)) {
+        return false;
+    }
+
+    const char* tmp = "http://%s:%d/mock/12/api/account/login";
+    auto url = folly::stringPrintf(tmp, FLAGS_cloud_server_ip.c_str(),
+                                   FLAGS_cloud_server_port);
```
bright-starry-sky: Is this mock code ?

etc/nebula-graphd.conf.production
```diff
@@ -55,4 +55,6 @@
 
 ########## authorization ##########
 # Enable authorization
---enable_authorize=false
\ No newline at end of file
+--enable_authorize=false
```
bright-starry-sky: Good !

etc/nebula-graphd.conf.production
```diff
@@ -55,4 +55,6 @@
...
+# Authentication type
+--auth_type=password
```
bright-starry-sky: It's better to add some comments  here. right ?

src/graph/CloudAuthenticator.cpp
```diff
@@ -0,0 +1,66 @@
...
+        std::string header = "-H \"Content-Type: application/json\"  -H \"Authorization:Basic ";
+        header =  header + base64Str + "\"";
+        result = http::HttpClient::post(url, header);
+    } else if (password[0] == 'T') {
+        // Use token method
+        std::string passwd = password.substr(1);
+        std::string header = "-H \"Content-Type: application/json\"  -H \"Authorization:Bearer ";
+        header = header + passwd + "\"";
+        result = http::HttpClient::post(url, header);
+    }
```
bright-starry-sky: else return error.

src/graph/CloudAuthenticator.cpp
panda-sheep: The path in the url ` /mock/12/api/account/login ` may need to be changed.
Currently ut uses mock, the test passed.

src/common/encryption/Base64.cpp
panda-sheep: You reminded me yesterday.
I went to proxygen and took a look. I felt that proxygen was available, but the best one felt like using it.
Reference proxygen notes are as follows:
```
// This is a wrapper around openssl base64 encoding.  It is a temporary wrapper
// around openssl, until a more optimized version lands in folly

class Base64 {
  public:
    static std::string decode(const std::string& b64message, int padding);
    static std::string urlDecode(const std::string& b64message);
    static std::string encode(folly::ByteRange buffer);
    static std::string urlEncode(folly::ByteRange buffer);
};
```

panda-sheep: Can follow the progress of folly

etc/nebula-graphd.conf.production
panda-sheep: Good catch

src/graph/CloudAuthenticator.cpp
panda-sheep: Good catch

src/graph/CloudAuthenticator.cpp
panda-sheep: I will add Flags to handle it.

src/graph/CloudAuthenticator.cpp
```diff
@@ -0,0 +1,66 @@
...
+bool CloudAuthenticator::auth(const std::string& user, const std::string& password) {
+    // The shadow account on the nebula side has been created
+    // Normal passwords and tokens use different prefixes
+
+    // First, go to meta to check if the shadow account exists
+    if (!metaClient_->checkShadowAccountFromCache(user)) {
+        return false;
+    }
+
+    const char* tmp = "http://%s:%d/mock/12/api/account/login";
```
nianiaJR: you  can make the whole address including ip、port and url path  into a single address which can be configurable

panda-sheep: Good, use a flags to do it.

src/graph/test/CMakeLists.txt
```diff
@@ -409,3 +409,23 @@ nebula_add_test(
         wangle
         gtest
 )
+
+nebula_add_test(
+    DISABLED
```
bright-starry-sky: Good . because the cloud server is private Intranet. so disable the test.

src/graph/GraphFlags.h
```diff
@@ -29,6 +29,11 @@ DECLARE_string(meta_server_addrs);
 
 DECLARE_string(default_charset);
 DECLARE_string(default_collate);
+
 DECLARE_bool(enable_authorize);
+DECLARE_string(auth_type);
+
+DECLARE_string(cloud_server_ip);
```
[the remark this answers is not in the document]
panda-sheep: Certification has been provided

Issue #1967, opened 2020-03-25
[p1 bug] storaged coredumped


whitewum:
I'll assign this to laura and give the hosts access to her.

laura-ding:
After a discussion with @darionyaphet, he will solve it.

whitewum:
Sorry, this pr crash again. I think we need a smoke test case to cover basic current crud during dev code.

```
[root@ctos1810-test scripts]# F0326 12:20:18.802213  3829 Part.cpp:295] [Port: 44501, Space: 27, Part: 39] Unknown operation: �
*** Check failure stack trace: ***
    @          0x149862c  google::LogMessage::Fail()
    @          0x149a3c3  google::LogMessage::SendToLog()
    @          0x149815c  google::LogMessage::Flush()
    @          0x149ad38  google::LogMessageFatal::~LogMessageFatal()
    @           0x67aead  nebula::kvstore::Part::commitLogs()
    @           0x6c9ee7  nebula::raftex::RaftPart::processAppendLogResponses()
    @           0x6c8d8c  _ZZN6nebula6raftex8RaftPart13replicateLogsEPN5folly9EventBaseENS0_18AppendLogsIteratorElllllENUlONS2_3TryISt6vectorISt4pairImNS0_4cpp217AppendLogResponseEESaISB_EEEEE1_clESF_
    @           0x6db89a  _ZN5folly7futures6detail17CoreCallbackStateISt6vectorISt4pairImN6nebula6raftex4cpp217AppendLogResponseEESaIS9_EEZNS6_8RaftPart13replicateLogsEPNS_9EventBaseENS6_18AppendLogsIteratorElllllEUlONS_3TryISB_EEE1_E6invokeIJSH_EEEDaDpOT_
    @           0x6db8db  _ZN5folly7futures6detail27detail_msvc_15_7_workaround6invokeILb1ENS1_17CoreCallbackStateISt6vectorISt4pairImN6nebula6raftex4cpp217AppendLogResponseEESaISB_EEZNS8_8RaftPart13replicateLogsEPNS_9EventBaseENS8_18AppendLogsIteratorElllllEUlONS_3TryISD_EEE1_EESD_SK_EEDcRT0_RNSI_IT1_EE
[...]

PR #1970, opened 2020-03-25
Fix delete verticse crash

darionyaphet:


whitewum:
close #1967

src/storage/mutate/DeleteVerticesProcessor.cpp
```diff
@@ -59,11 +59,10 @@ void DeleteVerticesProcessor::process(const cpp2::DeleteVerticesRequest& req) {
         }
     } else {
         callingNum_ = req.parts.size();
-        std::for_each(req.parts.begin(), req.parts.end(), [&](auto &partVerticse) {
-            auto partId = partVerticse.first;
-            const auto &vertices = partVerticse.second;
-            auto atomic = [&]() -> std::string {
-                return deleteVertices(spaceId, partId, vertices);
+        std::for_each(req.parts.begin(), req.parts.end(), [&](auto &pv) {
```
Shylock-Hg: I still don't think `&` in closure capture list is a good way, better list the reference capture.

darionyaphet: Good point

PR #1973, opened 2020-03-26
Fix the read performance issue after delete range with rocksdb-5.15.10

dangleptr:
The problem comes from the implementation about deleteRange in rocksdb 5.15.10.

In version 5.15.10,  when calling deleteRange, rocksdb save tombstone as a whole range,  which means for each read, we should scan the range with O(n).

After upgrade the version to the newest one,  rocksdb do split the tombstone range which could allow to do a binary search with O(log(n))

This is why i upgrade the rocksdb's version. 

For more information about deleteRange,  please refer to 
https://github.com/facebook/rocksdb/wiki/DeleteRange-Implementation

### DIFF
--- a/src/storage/mutate/DeleteEdgesProcessor.cpp
+++ b/src/storage/mutate/DeleteEdgesProcessor.cpp
@@ -23,10 +23,11 @@ void DeleteEdgesProcessor::process(const cpp2::DeleteEdgesRequest& req) {
         std::for_each(req.parts.begin(), req.parts.end(), [this](auto &partEdges) {
             this->callingNum_ += partEdges.second.size();
         });
-        std::for_each(req.parts.begin(), req.parts.end(), [spaceId, this](auto &partEdges) {
+        std::vector<std::string> keys;
+        keys.reserve(16);
+        for (auto& partEdges : req.parts) {
             auto partId = partEdges.first;
-            std::for_each(partEdges.second.begin(), partEdges.second.end(),
-                          [spaceId, partId, this](auto &edgeKey) {
+            for (auto& edgeKey : partEdges.second) {
                 auto start = NebulaKeyUtils::edgeKey(partId,
                                                      edgeKey.src,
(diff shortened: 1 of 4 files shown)


src/storage/mutate/DeleteEdgesProcessor.cpp
```diff
@@ -38,9 +41,22 @@ void DeleteEdgesProcessor::process(const cpp2::DeleteEdgesRequest& req) {
...
-                doRemoveRange(spaceId, partId, start, end);
-            });
-        });
+                std::unique_ptr<kvstore::KVIterator> iter;
+                auto ret = this->kvstore_->range(spaceId, partId, start, end, &iter);
+                if (ret != kvstore::ResultCode::SUCCEEDED) {
+                    VLOG(3) << "Error! ret = " << static_cast<int32_t>(ret)
+                            << ", spaceID " << spaceId;
+                    this->onFinished();
+                    return;
```
bright-starry-sky: do we need handle LeaderChanged?

Shylock-Hg: Seems it's not a part of raft.

bright-starry-sky: > Seems it's not a part of raft.

Yes，I mixed up.

Shylock-Hg: > > Seems it's not a part of raft.
> 
> Yes，I mixed up.

Sorry, I think I mixed up it. :) This will check leader.

critical27:
https://rocksdb.org/blog/2018/11/21/delete-range.html
This blog is easier to read.

src/storage/mutate/DeleteEdgesProcessor.cpp
dangleptr: Make sense. Let me handle it.

PR #1978, opened 2020-03-26
Remove the LOG(FATAL) when encountering the unknown log.

### LINKED ISSUE #1967: [p1 bug] storaged coredumped

dangleptr:
For the backward compatibility, we should not crash directly when encountering unknown log. Instead, we just skip it and give a LOG message. 

close #1967


whitewum:
to close #1967

PR #1985, opened 2020-03-26
fixed unstable testcase for rebuild index

bright-starry-sky:
fixed unstable testcase for rebuild index.
Rebuildindex is an asynchronous program, and checking the state immediately after execution maybe have several states.
 just check the status != "FAILED".  expected "RUNNING" or "SUCCEEDED".

**Error log info :**
```
I0326 14:43:11.336046 19444 ExecutionPlan.cpp:16] Parsing query: SHOW EDGE INDEX STATUS
/__w/nebula/nebula/src/graph/test/IndexTest.cpp:313: Failure
Value of: verifyResult(resp, expected)
  Actual: false ([multi_friend_index, RUNNING] vs. [multi_friend_index, SUCCEEDED])
Expected: true
[  FAILED  ] IndexTest.EdgeIndex (4064 ms)
```


src/graph/test/IndexTest.cpp
```diff
@@ -125,11 +125,13 @@ TEST_F(IndexTest, TagIndex) {
         std::string query = "SHOW TAG INDEX STATUS";
         auto code = client->execute(query, resp);
         ASSERT_EQ(cpp2::ErrorCode::SUCCEEDED, code);
-        std::vector<uniform_tuple_t<std::string, 2>> expected{
-            {"single_person_index", "SUCCEEDED"},
-            {"multi_person_index",  "SUCCEEDED"},
-        };
-        ASSERT_TRUE(verifyResult(resp, expected));
+        /*
```
Shylock-Hg: I think you can try 3 time checking.

bright-starry-sky: > I think you can try 3 time checking.

Because the final index data is not check at here, I don't think it is necessary to wait for the execution to complete.

src/graph/test/IndexTest.cpp
```diff
@@ -125,11 +125,13 @@ TEST_F(IndexTest, TagIndex) {
...
+         * Currently , expected the index status is "RUNNING" or "SUCCEEDED"
+         */
+        for (auto& row : *resp.get_rows()) {
```
dangleptr: I want to know the original purpose about this test case.

bright-starry-sky: > I want to know the original purpose about this test case.

Actually, I think this test case without any value at here(except for syntax checking). Because the data validity has been verified in other test cases. Therefore, the results of this asynchronous execution are best not checked here.
If it were me, I'd rather delete it.

dangleptr: :( 
Fine.  Let's leave it here right now.

PR #1992, opened 2020-03-26
Fix meta would send empty raft snapshot

critical27:
In poc project, 3 replica meta. If we remove the data dir of one. It would failed to start because it won't get the clusterId. It turned out that the leader send an empty snapshot which have nothing it. (Because it only scan the data with prefix space + part, however nothing start with them in meta)

### DIFF
--- a/etc/nebula-metad.conf.production
+++ b/etc/nebula-metad.conf.production
@@ -33,3 +33,6 @@
 ########## storage ##########
 # Root data path, here should be only single path for metad
 --data_path=data/meta
+
+############## rocksdb Options ##############
+--rocksdb_disable_wal=false
--- a/etc/nebula-metad.conf.production
+++ b/etc/nebula-metad.conf.production
@@ -36,3 +36,4 @@
 
 ############## rocksdb Options ##############
 --rocksdb_disable_wal=false
+--rocksdb_wal_sync=true
--- a/src/kvstore/SnapshotManagerImpl.cpp
+++ b/src/kvstore/SnapshotManagerImpl.cpp
@@ -17,7 +17,7 @@ void SnapshotManagerImpl::accessAllRowsInSnapshot(GraphSpaceID spaceId,
                                                   raftex::SnapshotCallback cb) {
     CHECK_NOTNULL(store_);
     std::unique_ptr<KVIterator> iter;
-    auto prefix = NebulaKeyUtils::prefix(partId);
+    auto prefix = NebulaKeyUtils::snapshotPrefix(partId);
     std::vector<std::string> data;
     int64_t totalSize = 0;
     int64_t totalCount = 0;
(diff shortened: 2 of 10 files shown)


dangleptr:
My bad,  forget the meta server....

Besides that,  we should open the rocksdb WAL on metad.  So in most cases,  it will not need to catch up the data. 

WDYT?

src/kvstore/NebulaStore.cpp
```diff
@@ -39,7 +39,9 @@ bool NebulaStore::init() {
     LOG(INFO) << "Start the raft service...";
     bgWorkers_ = std::make_shared<thread::GenericThreadPool>();
     bgWorkers_->start(FLAGS_num_workers, "nebula-bgworkers");
-    snapshot_.reset(new SnapshotManagerImpl(this));
+    if (options_.snapManBuilder_) {
```
dangleptr: Em....
Any better idea here?

critical27:
> My bad, forget the meta server....
> 
> Besides that, we should open the rocksdb WAL on metad. So in most cases, it will not need to catch up the data.
> 
> WDYT?

SGTM

src/kvstore/NebulaStore.cpp
critical27: Add a `gflag` in SnapshotManagerImpl? That would make this PR very small.

dangleptr: I think the key problem is we don't know the key format outside in kvstore.  So the prefix generator should be defined outside.

critical27: But we still to pass a lambda to NebulaStore, something like:
```
    [] (spaceId, partId) -> std::unique_ptr<KVIterator> iter {
        std::unique_ptr<KVIterator> iter;
        auto ret = store_->prefix(spaceId, partId, prefix, &iter);
        return iter;
    }
```

The KVIterator is what we need here, it seems not a big difference.

dangleptr: The difference is for kvstore,  it should not know anything about meta or data.

src/kvstore/SnapshotManagerImpl.cpp
```diff
@@ -17,7 +17,7 @@ void SnapshotManagerImpl::accessAllRowsInSnapshot(GraphSpaceID spaceId,
                                                   raftex::SnapshotCallback cb) {
     CHECK_NOTNULL(store_);
     std::unique_ptr<KVIterator> iter;
-    auto prefix = NebulaKeyUtils::prefix(partId);
+    auto prefix = NebulaKeyUtils::snapshotPrefix(partId);
```
dangleptr: If the prefix is empty,  what's the behavior for seek of rocksdb.

critical27: Just the same as SeekToFirst and iterate over all.

etc/nebula-metad.conf.production
```diff
@@ -33,3 +33,7 @@
 ########## storage ##########
 # Root data path, here should be only single path for metad
 --data_path=data/meta
+
+############## rocksdb Options ##############
+--rocksdb_disable_wal=false
```
Shylock-Hg, line 38: Why not enable for storage?I think there is secure problem to loss data without replica (or with replica but all node crash).

Issue #1998, opened 2020-03-26
Fetch on get wrong data


Qubutol:
**Describe the bug (__must be provided__)**

Fetch on get wrong data

**Your Environments (__must be provided__)**

* OS: `centos7`
* version:nebula-2020.03.20-nightly.el7-5.x86_64

**How To Reproduce(__must be provided__)**

1. Step 1
```
CREATE TAG player(name string, age int);
CREATE TAG INDEX player_index_0 on player(name,age);
INSERT VERTEX player(name, age) VALUES 121:("Useless", 60);
```
2. Step 2
```
CREATE TAG person(name string, age int);
CREATE TAG INDEX person_index_0 on player(name,age);
INSERT VERTEX person(name, age) VALUES 121:("Useless", 61);
```
3. Step 3
```
FETCH PROP ON player 121;
```
![image](https://user-images.githubusercontent.com/46239712/77673696-775c4e80-6fc5-11ea-8fc9-b3ee51abb5e2.png)

```
FETCH PROP ON person 121;
```
![image](https://user-images.githubusercontent.com/46239712/77673710-7d522f80-6fc5-11ea-81d0-7b591b515757.png)
It's right result;

4.Step 4
```
LOOKUP ON player WHERE player.name == "Useless" YIELD player.name, person.age;
```
![image](https://user-images.githubusercontent.com/46239712/77673809-a1ae0c00-6fc5-11ea-809c-f2ca3df5b217.png)

Why I search player and yield person can get such result?

PR #2000, opened 2020-03-27
Add alias check for lookup

bright-starry-sky:
Issue #1998 .


bright-starry-sky:
> please add an e2e case in the test framework

👌

Issue #1996, opened 2020-03-26
Can't delete vertex


Qubutol:
**Describe the bug (__must be provided__)**
when I delete a vertex,I got:
```
[ERROR (-8)]: SpaceId `98'  is nonexistent
```
**Your Environments (__must be provided__)**

* OS: `centos7`
* version:nebula-2020.03.20-nightly.el7-5.x86_64

**How To Reproduce(__must be provided__)**

1. Step 1
```
use nba
CREATE TAG person(name string, age int);
CREATE TAG INDEX person_index on person(name,age);
```
2. Step 2
```
INSERT VERTEX person(name, age) VALUES 101:("Tony Parker", 36);
```
3. Step 3
```
DELETE VERTEX 101
```
I got:
```
[ERROR (-8)]: SpaceId `98'  is nonexistent
```
but I describe space nba :
```
DESCRIBE SPACE nba
```
![image](https://user-images.githubusercontent.com/46239712/77661664-e7fb6f00-6fb5-11ea-9580-a19c871ea671.png)

**Expected behavior**

delete vertex 101

log:
E0326 22:56:47.638962 30576 ExecutionPlan.cpp:76] Execute failed: SpaceId `98'  is nonexistent

jude-zhu:
(*￣︶￣) we run into this problem yesterday. Simon is working on it.

Qubutol:
> (*￣︶￣) we run into this problem yesterday. Simon is working on it.


Will this bug be fixed today? I see that the corresponding pull request has been read for test. Today's nightly version will merge this part of the code?

PR #1983, opened 2020-03-26
remove  unreasonable error message

monadbobo:
### What changes were proposed in this pull request?

 if you execute "go from xx over *", and when there is no edge under a space,
an unreasonable error message is returned (SpaceId `xxx 'is nonexistent)， so we need remove this.

### How was this patch tested?

close #1996

PR #2001, opened 2020-03-27
fixed bug for delete vertex without edge

monadbobo:

When deleting a vertex, if the edge does not exist, then we should continue to delete the vertex instead of reporting an error.


monadbobo:
depend #1983 close #1996

src/graph/DeleteVerticesExecutor.cpp
```diff
@@ -111,20 +111,18 @@ void DeleteVerticesExecutor::execute() {
...
             std::unordered_map<EdgeType, std::shared_ptr<ResultSchemaProvider>> edgeSchema;
             auto *eschema = response.get_edge_schema();
-            if (eschema != nullptr) {
-                std::transform(eschema->cbegin(), eschema->cend(),
-                               std::inserter(edgeSchema, edgeSchema.begin()), [](auto &schema) {
-                                   return std::make_pair(
-                                       schema.first,
-                                       std::make_shared<ResultSchemaProvider>(schema.second));
-                               });
+            if (eschema == nullptr || eschema->empty()) {
```
dangleptr: Well done.  Could you give a UT for it.

src/graph/DeleteVerticesExecutor.cpp
```diff
@@ -111,20 +111,18 @@ void DeleteVerticesExecutor::execute() {
...
+                continue;
             }
 
-            if (edgeSchema.empty()) {
```
darionyaphet: When the edge not exist, it's not a bug 👍

darionyaphet:
Generally LGTM

Issue #1991, opened 2020-03-26
show jobs can't show all the jobs if there are too many jobs


whitewum:
**Describe the bug (__must be provided__)**

I submit a bunch of jobs, maybe 50 times. And when I show jobs, I can't see my current job.

**Your Environments (__must be provided__)**

* OS: `uname -a`
* Compliler: `g++ --version` or `clang++ --version`
* CPU: `lscpu`
* Commit id (e.g. `a3ffc7d8`)

**How To Reproduce(__must be provided__)**

1. Step 1
```
(user@127.0.0.1:3699) [rdf_5]> submit job compact
==============
| New Job Id |
==============
| 54         |
--------------
Got 1 rows (Time spent: 1.5/2.305 ms)
```

2. Step 2
```
(user@127.0.0.1:3699) [rdf_5]> show jobs;
==============================================================================
| Job Id | Command        | Status   | Start Time        | Stop Time         |
==============================================================================
| 8      | compact rdf_4  | finished | 03/24/20 16:42:34 | 03/24/20 16:43:50 |
------------------------------------------------------------------------------
| 9      | compact rdf_4  | finished | 03/24/20 16:43:50 | 03/24/20 16:44:33 |
------------------------------------------------------------------------------
| 10     | compact rdf_4  | finished | 03/24/20 16:44:33 | 03/24/20 16:45:20 |
------------------------------------------------------------------------------
| 11     | compact rdf_4  | finished | 03/24/20 16:47:48 | 03/24/20 16:48:52 |
------------------------------------------------------------------------------
| 12     | compact rdf_4  | finished | 03/24/20 16:49:35 | 03/24/20 16:50:47 |
------------------------------------------------------------------------------
| 13     | compact rdf_4  | finished | 03/24/20 16:50:47 | 03/24/20 16:52:11 |
------------------------------------------------------------------------------
| 14     | compact rdf_4  | finished | 03/24/20 16:57:29 | 03/24/20 16:59:52 |
------------------------------------------------------------------------------
| 15     | compact rdf_4  | finished | 03/24/20 17:04:25 | 03/24/20 17:07:12 |
------------------------------------------------------------------------------
| 16     | compact rdf_4  | finished | 03/24/20 17:20:43 | 03/24/20 17:25:06 |
------------------------------------------------------------------------------
| 17     | compact rdf_4  | finished | 03/24/20 17:26:48 | 03/24/20 17:31:08 |
------------------------------------------------------------------------------
[...]

Issue #1990, opened 2020-03-26
SUBMIT JOB not working


Amber1990Zhang:
I can't use `SUBMIT JOB FLUSH/COMPACT`.
![image](https://user-images.githubusercontent.com/42762957/77635643-d69e6c80-6f8d-11ea-94ed-b075020db936.png)
There are not enough logs. I use docker-comose. The 3 metas are not leaders.

Amber1990Zhang:
From @whitewum 
`SUBMIT JOB` succeeded but `SHOW JOB id` shows nothing.

PR #2007, opened 2020-03-27
fix bug 1990/1991

liuyu85cn:

1. meta client submit job only to leader
2. show jobs in desc job id order

fix issue [SUBMIT JOB not working](https://github.com/vesoft-inc/nebula/issues/1990)

part fix issue 1991 [show jobs can't show all the jobs if there are too many jobs](https://github.com/vesoft-inc/nebula/issues/1991)


Shylock-Hg:
Why not #1990 retry?

whitewum:
Do you have any case to cover these bug?

Even a UT or an E2E case is required.

Issue #1982, opened 2020-03-26
bidirect query returns double items as configured in sampling 


jude-zhu:
`Max_edge_returned_per_vertex ` is configures as 2, `go from ... over ...` and `go from ... over ... reversely` returns 2 items. However, `go from xxx over ... bidirect` gets 4 item.

Amber1990Zhang:
I encountered the same problem

PR #2013, opened 2020-03-29
Sampling apply to all edges starting from a vertex.

CPWstatic:


src/storage/query/QueryBaseProcessor.inl
```diff
@@ -491,15 +491,6 @@ kvstore::ResultCode QueryBaseProcessor<REQ, RESP>::collectEdgeProps(
...
     Getters getters;
-    std::unique_ptr<nebula::algorithm::ReservoirSampling<
-        std::pair<std::unique_ptr<RowReader>, std::string>>> sampler;
-    if (FLAGS_enable_reservoir_sampling) {
-        sampler = std::make_unique<
-            nebula::algorithm::ReservoirSampling<
-                std::pair<std::unique_ptr<RowReader>, std::string>
-            >
-        >(FLAGS_max_edge_returned_per_vertex);
-    }
```
panda-sheep: 498  line  need handle ?

src/storage/query/QueryBoundProcessor.cpp
```diff
@@ -35,17 +35,19 @@ kvstore::ResultCode QueryBoundProcessor::processEdgeImpl(const PartitionID partI
     edges.reserve(FLAGS_reserved_edges_one_vertex);
     auto ret = collectEdgeProps(
         partId, vId, edgeType, props, &fcontext,
-        [&, this](RowReader* reader, folly::StringPiece k, const std::vector<PropContext>& p) {
+        [&, this](std::unique_ptr<RowReader> reader,
+                  folly::StringPiece k,
+                  const std::vector<PropContext>& p) {
             cpp2::IdAndProp edge;
```
dangleptr: Could we unify the code here?
```
auto ret = sample(xxx)
if (ret) {
     xxxx
}
```

jude-zhu:
close  #1982

src/storage/query/QueryBaseProcessor.inl
CPWstatic: We need it, or we would not the whole samples.

src/storage/query/QueryBoundProcessor.cpp
CPWstatic: As we talked about this offline.

PR #2023, opened 2020-03-30
Fix timestamp type for default value

### LINKED ISSUE #2020: timestamp default value cannot be set as negative
image
image
image
image

laura-ding:
close #2020


src/graph/SchemaHelper.cpp
```diff
@@ -246,5 +251,42 @@ Status SchemaHelper::alterSchema(const std::vector<AlterSchemaOptItem*>& schemaO
...
+            return Status::Error("Invalid timestamp type");
+        }
+        struct tm time;
+        memset(&time, 0, sizeof(time));
+        time.tm_year = atoi(result[1].str().c_str()) - 1900;
+        time.tm_mon = atoi(result[2].str().c_str()) - 1;
+        time.tm_mday = atoi(result[3].str().c_str());
+        time.tm_hour = atoi(result[4].str().c_str());
+        time.tm_min = atoi(result[5].str().c_str());
+        time.tm_sec = atoi(result[6].str().c_str());
```
panda-sheep: I think it's best to add a `TODO `here
Time type validity check, such as `2020-02-29` or` 2020-04-31`

src/parser/MaintainSentences.h
```diff
@@ -42,62 +42,22 @@ class ColumnSpecification final {
...
-            return Status::Error("Wrong type");
-        }
-        return Value::toDouble(v);
+    bool hasDefault() {
+        return defaultExpr_ != nullptr;
     }
 
-    StatusOr<std::string> getStringValue(Getters& getter) {
+    OptVariantType getDefault(Getters& getter) {
         auto r = defaultExpr_->eval(getter);
```
panda-sheep: Using folly :: option instead of  `hasDefault` and` getDefault`  is better ？

src/graph/SchemaHelper.cpp
```diff
@@ -246,5 +251,42 @@ Status SchemaHelper::alterSchema(const std::vector<AlterSchemaOptItem*>& schemaO
...
+        timestamp = mktime(&time);
+    } else {
```
Shylock-Hg, line 280: `else if (value.which() == VAR_INT)` ?

laura-ding: line 255 has done the checkup

Shylock-Hg: So I don't think check one logical in two `if` statement is a good style.

laura-ding: Check is not the same, one is the data type, one is the data format.

src/graph/SchemaHelper.cpp
laura-ding: Good catch, pr #1450 has done it.

src/parser/MaintainSentences.h
laura-ding: Here need to return the error message.

src/graph/SchemaHelper.cpp
Shylock-Hg: I think this just check the data type.

src/graph/SchemaHelper.cpp
```diff
@@ -246,5 +251,42 @@ Status SchemaHelper::alterSchema(const std::vector<AlterSchemaOptItem*>& schemaO
...
+                                    "(20|21|22|23|[0-1]\\d|\\d):"
+                                    "([0-5]\\d|\\d):"
+                                    "([0-5]\\d|\\d)$");
+        std::smatch result;
+        if (!std::regex_match(boost::get<std::string>(value), result, reg)) {
+            return Status::Error("Invalid timestamp type");
+        }
+        struct tm time;
+        memset(&time, 0, sizeof(time));
+        time.tm_year = atoi(result[1].str().c_str()) - 1900;
```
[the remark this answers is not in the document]
dutor: > Why 1900?

Please refer to `man ctime`

PR #2030, opened 2020-03-31
fix leader change unstable UT

panda-sheep:
fix  https://github.com/vesoft-inc/nebula/pull/2012/checks?check_run_id=547541664
and other similar UT.


src/storage/test/AddEdgesTest.cpp
```diff
@@ -17,7 +17,10 @@ namespace storage {
 
 TEST(AddEdgesTest, SimpleTest) {
     fs::TempDir rootPath("/tmp/AddEdgesTest.XXXXXX");
-    std::unique_ptr<kvstore::KVStore> kv = TestUtils::initKV(rootPath.path());
+    constexpr int32_t partitions = 6;
+    std::unique_ptr<kvstore::KVStore> kv = TestUtils::initKV(rootPath.path(), partitions,
+        {0, network::NetworkUtils::getAvailablePort()});
+    TestUtils::waitUntilAllElected(kv.get(), 0, {0, 1, 2, 3, 4, 5}/*partitions*/);
```
dangleptr, line 23: I suggest you move the waitUntilAllElected into init::KV() method.

panda-sheep: Good point， 👍

src/storage/test/TestUtils.h
```diff
@@ -74,7 +74,31 @@ class TestUtils {
                                                             localhost,
                                                             workers);
         store->init();
-        sleep(1);
+
+        // When useMetaServer is ture we deal with it using waitUntilAllElected
```
Shylock-Hg, line 78: Typo `true`.

panda-sheep: Bright eyes,  👍

src/storage/test/TestUtils.h
```diff
@@ -74,7 +74,31 @@ class TestUtils {
...
+                        }
+                    }
+                }
+                if (readyNum == partitionNumber) {
+                    LOG(INFO) << "All leaders have been elected!";
+                    break;
+                }
+                usleep(100000);
+            }
+        }
```
bright-starry-sky, line 101: Here , Why not call waitUntilAllElected direct?

src/storage/test/AddVerticesTest.cpp
```diff
@@ -17,7 +17,9 @@ namespace storage {
 
 TEST(AddVerticesTest, SimpleTest) {
     fs::TempDir rootPath("/tmp/AddVerticesTest.XXXXXX");
-    std::unique_ptr<kvstore::KVStore> kv = TestUtils::initKV(rootPath.path());
+    constexpr int32_t partitions = 6;
+    std::unique_ptr<kvstore::KVStore> kv = TestUtils::initKV(rootPath.path(), partitions,
+        {0, network::NetworkUtils::getAvailablePort()});
```
bright-starry-sky, line 22: You can avoid changing the code if there are parameter defaults. right ？

src/storage/test/TestUtils.h
panda-sheep: when  useMetaServer is true,  use ` waitUntilAllElected(kvstore::KVStore* kvstore, GraphSpaceID spaceId, int partNum)`
when   useMetaServer  is flase, use ```
 waitUntilAllElected(kvstore::KVStore* kvstore, GraphSpaceID spaceId,
        std::set<PartitionID> parts)

The difference between these two functions is whether to use metaserver, so the start of partId is also different, one starts from 0 and the other starts from 1.

```

File: src/storage/test/AddEdgesTest.cpp [hunk 5]
```diff
@@ -17,7 +17,9 @@ namespace storage {
 
 TEST(AddEdgesTest, SimpleTest) {
     fs::TempDir rootPath("/tmp/AddEdgesTest.XXXXXX");
-    std::unique_ptr<kvstore::KVStore> kv = TestUtils::initKV(rootPath.path());
+    constexpr int32_t partitions = 6;
+    std::unique_ptr<kvstore::KVStore> kv = TestUtils::initKV(rootPath.path(), partitions,
+        {0, network::NetworkUtils::getAvailablePort()});
```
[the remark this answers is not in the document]
panda-sheep: Good point, fixed.

src/storage/test/AddVerticesTest.cpp
panda-sheep: 👍

src/storage/test/TestUtils.h
```diff
@@ -74,7 +74,31 @@ class TestUtils {
                                                             localhost,
                                                             workers);
         store->init();
-        sleep(1);
+
+        // When useMetaServer is true, we deal with it using waitUntilAllElected.
+        if (useMetaServer) {
+            sleep(1);
```
Shylock-Hg, line 80: `sleep(1)` here?

panda-sheep: Yeah, when useMetaServer is true, follow the original logic

Shylock-Hg: Yes, I think it's useless.

panda-sheep: I think  here `sleep`, and the  following `waitUntilAllElected `  basically does not need to `sleep`  anymore.

panda-sheep: This pr is not needed.

PR #2033, opened 2020-04-01
Fix unable to stop while storage is doing compaction

critical27:
If we trigger compaction by WebService, storaged will be unable to stop because HttpServer will wait for all on-going requests finish. So we need to CancelAllBackgroundWork before we stop WebService.

In fact, when we call rocksdb::Close(), it will call `CancelAllBackgroundWork` internally. But we Close at first and then stop raft, it would fail to commit. So we CancelAllBackgroundWork explicitly here.


whitewum:
Finally, it can replace my  `kill -9` script

PR #2046, opened 2020-04-01
Redesign spark tools

darionyaphet:


dangleptr:
Do we still need it in 1.x ?   @darionyaphet @whitewum @jude-zhu

whitewum:
> Do we still need it in 1.x ? @darionyaphet @whitewum @jude-zhu

Only if 2.0 can be released on schedule.

dangleptr:
> > Do we still need it in 1.x ? @darionyaphet @whitewum @jude-zhu
> 
> Only if 2.0 can be released on schedule.

After talked with @darionyaphet ,  the code about spark writer has been moved to nebula-java.
So the tool refactor don't depend on 2.0   @whitewum

PR #2050, opened 2020-04-02
index advance for storage layer.

bright-starry-sky:
Support new features as : 
for example `tag(col1, col2, col3); index1 on tag (col1, col2); index2 on tag (col3)`

1. Support range scan . (where col1 > 1 and col2 < 5)
2. Support prefix-range scan (where col1 == 1 and (col2 > 1 and col2 < 5))
3. Support merger-index  scan (where col1 == 1 or col1 == 2 , Scan twice and merge the result set)
4. Support multiple index (where col1 == 1 or col3==1, Using index1 and index2,Scan twice and merge the result set)


src/interface/storage.thrift
```diff
@@ -457,11 +458,12 @@ struct LookUpEdgeIndexResp {
 }
 
 struct LookUpIndexRequest {
```
darionyaphet, line 460: If we modify the thrift protocol structure, maybe cause incompatible ?

bright-starry-sky: > If we modify the thrift protocol structure, I’m afraid it maybe cause incompatible ?

No, go ahead. let me resolve this conflict later.

dangleptr:
Reimplement it inside storage-2.0

PR #2051, opened 2020-04-02
RC4 Cherrypick #2031 #2042 #2025 #2013 #2038

jude-zhu:
Cherrypick #2031 #2042 #2025 #2013 #2038


src/graph/TraverseExecutor.cpp
```diff
@@ -142,7 +142,7 @@ nebula::cpp2::SupportedType TraverseExecutor::calculateExprType(Expression* exp)
         case Expression::kInputProp: {
             auto* propExp = static_cast<const AliasPropertyExpression*>(exp);
             const auto* propName = propExp->prop();
-            if (inputs_ == nullptr) {
+            if (inputs_ == nullptr || !inputs_->hasData()) {
```
bright-starry-sky, line 145: if inputs_ == nullptr , Is it will leads to crash by  `inputs_->hasData()` ?

Shylock-Hg: I think won't because the short circuit ||

PR #2064, opened 2020-04-07
session manager 2.0 

panda-sheep:
Implement session manager version 2.0

### DIFF
--- a/etc/nebula-graphd.conf.default
+++ b/etc/nebula-graphd.conf.default
@@ -24,8 +24,8 @@
 ########## networking ##########
 # Meta Server Address
 --meta_server_addrs=127.0.0.1:45500
-# Local ip
---local_ip=127.0.0.1
+# Graph server ip, Local ip
+--graph_server_ip=127.0.0.1
 # Network device to listen on
 --listen_netdev=any
 # Port to listen on
--- a/etc/nebula-graphd.conf.production
+++ b/etc/nebula-graphd.conf.production
@@ -24,8 +24,8 @@
 ########## networking ##########
 # Meta Server Address
 --meta_server_addrs=127.0.0.1:45500
-# Local ip
---local_ip=127.0.0.1
+# Graph server ip, Local ip
+--graph_server_ip=127.0.0.1
 # Network device to listen on
 --listen_netdev=any
 # Port to listen on
--- a/src/common/network/NetworkUtils.cpp
+++ b/src/common/network/NetworkUtils.cpp
@@ -388,6 +388,22 @@ StatusOr<std::string> NetworkUtils::getLocalIP(std::string defaultIP) {
         }
     }
     return Status::Error("No IPv4 address found!");
+ }
+
+StatusOr<bool> NetworkUtils::isLocalValidIP(std::string defaultIP) {
+    if (defaultIP.empty()) {
+        return Status::Error("IPv4 address is empty!");
+    }
+    auto result = network::NetworkUtils::listDeviceAndIPv4s();
+    if (!result.ok()) {
+        return std::move(result).status();
+    }
+    for (auto& deviceIP : result.value()) {
+        if (!defaultIP.compare(deviceIP.second)) {
+            return true;
+        }
+    }
+    return false;
 }
 
 }  // namespace network
--- a/src/graph/SessionManager.cpp
+++ b/src/graph/SessionManager.cpp
@@ -7,11 +7,16 @@
 #include "base/Base.h"
 #include "graph/SessionManager.h"
 #include "graph/GraphFlags.h"
+#include <folly/executors/CPUThreadPoolExecutor.h>
 
 namespace nebula {
 namespace graph {
 
-SessionManager::SessionManager() {
+SessionManager::SessionManager(meta::MetaClient* client) {
+    metaClient_ = client;
+    executor_.reset(new folly::CPUThreadPoolExecutor(2));
+    runner_.reset(new folly::CPUThreadPoolExecutor(2));
+    graphServerAddr_ = FLAGS_graph_server_ip + ":" + folly::to<std::string>(FLAGS_port);
     scavenger_ = std::make_unique<thread::GenericWorker>();
     auto ok = scavenger_->start("session-manager");
     DCHECK(ok);
@@ -36,7 +41,7 @@ SessionManager::findSession(int64_t id) {
     if (iter == activeSessions_.end()) {
         return Status::Error("Session `%ld' has expired", id);
     }
-    return iter->second;
+    return iter->second.first;
 }
 
 
@@ -52,11 +57,78 @@ std::shared_ptr<session::Session> SessionManager::createSession() {
     }
     DCHECK_NE(sid, 0L);
     auto session = session::Session::create(sid);
-    activeSessions_[sid] = session;
+    auto now = time::WallClock::fastNowInSec();
+    activeSessions_[sid] = std::make_pair(session, now);
+    activeSessForMeta_[sid] = now;
     session->charge();
+
+    // sessionId, FLAGS_graph_server_ip:FLAGS_port, start_time
+    executor_->add([sid, now, this] () {
+        this->doAddSession(sid, graphServerAddr_, now);
+    });
+
     return session;
 }
 
+void SessionManager::doAddSession(int64_t sid, std::string graphServerAddr, int64_t startTime) {
+    std::vector<nebula::meta::cpp2::SessionItem> sessionItem;
+    nebula::meta::cpp2::SessionItem item;
+    item.set_session_id(sid);
+    item.set_addr(graphServerAddr);
+    item.set_start_time(startTime);
+    sessionItem.emplace_back(std::move(item));
+
+    auto future = metaClient_->addSession(std::move(sessionItem));
+    auto cb = [this] (auto &&resp) {
+        if (!resp.ok()) {
+            LOG(ERROR) << "Add to global session failed.";
+            return;
+        }
+        auto ret = std::move(resp).value();
+        if (!ret) {
+            LOG(ERROR) << "Add to global session failed.";
+        }
+        return;
+    };
+
+    auto error = [this] (auto &&e) {
+        LOG(ERROR) << "Exception caught: "  << e.what();
+        return;
+    };
+
+    std::move(future).via(runner_.get()).thenValue(cb).thenError(error);
+}
+
+void SessionManager::doRemoveSession(std::unordered_map<int64_t, int64_t> removeSess) {
+    std::vector<nebula::meta::cpp2::SessionItem> sessionItem;
+    for (auto& e : removeSess) {
+        nebula::meta::cpp2::SessionItem item;
+        item.set_session_id(e.first);
+        item.set_addr(graphServerAddr_);
+        item.set_start_time(e.second);
+        sessionItem.emplace_back(std::move(item));
+    }
+
+    auto future = metaClient_->removeSession(std::move(sessionItem));
+    auto cb = [this] (auto &&resp) {
+        if (!resp.ok()) {
+            LOG(ERROR) << "Romove from global session failed.";
+            return;
+        }
+        auto ret = std::move(resp).value();
+        if (!ret) {
+            LOG(ERROR) << "Romove from global session failed.";
+        }
+        return;
+    };
+
+    auto error = [this] (auto &&e) {
+        LOG(ERROR) << "Exception caught: "  << e.what();
+        return;
+    };
+
+    std::move(future).via(runner_.get()).thenValue(cb).thenError(error);
+}
 
 std::shared_ptr<session::Session> SessionManager::removeSession(int64_t id) {
     folly::RWSpinLock::WriteHolder holder(rwlock_);
@@ -64,8 +136,18 @@ std::shared_ptr<session::Session> SessionManager::removeSession(int64_t id) {
     if (iter == activeSessions_.end()) {
         return nullptr;
     }
-    auto session = std::move(iter->second);
+    auto session = std::move(iter->second.first);
+    auto startTime = iter->second.second;
+
     activeSessions_.erase(iter);
+    activeSessForMeta_.erase(id);
+
+    std::unordered_map<int64_t, int64_t> removeSess;
+    removeSess.emplace(id, startTime);
+    executor_->add([removeSess, this] () {
+        this->doRemoveSession(removeSess);
+    });
+
     return session;
 }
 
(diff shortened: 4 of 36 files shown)


Amber1990Zhang:
Snytax:
```
SHOW SESSIONS
```

src/graph/SessionManager.cpp
```diff
@@ -52,20 +57,109 @@ std::shared_ptr<session::Session> SessionManager::createSession() {
...
+    activeSessions_[sid] = std::make_pair(session, now);
+    activeSessForMeta_[sid] = now;
     session->charge();
+
+    // sessionId, FLAGS_graph_server_ip:FLAGS_port, start_time
+    executor_->add([sid, now, this] () {
+        this->doAddSession(sid, now);
+    });
+
     return session;
```
laura-ding, line 70: If the metad is abnormal, here also return the session?

src/graph/ShowExecutor.cpp
```diff
@@ -1180,6 +1183,54 @@ std::string ShowExecutor::roleToStr(nebula::cpp2::RoleType type) {
...
+            }
+            return a.get_addr() < b.get_addr();
+        });
+
+        for (auto& item : sessionItems) {
+            std::vector<cpp2::ColumnValue> row;
+            row.resize(4);
+            row[0].set_str(item.get_addr());
+            row[1].set_integer(item.get_session_id());
+            row[2].set_integer(item.get_start_time());
```
laura-ding, line 1215: it‘s better to show by data time format, wdyt

src/graph/test/CMakeLists.txt
```diff
@@ -58,6 +58,10 @@ nebula_add_test(
     SOURCES
         SessionManagerTest.cpp
     OBJECTS
+        $<TARGET_OBJECTS:graph_test_common_obj>
```
laura-ding, line 61: use ${GRAPH_TEST_CLIENT_LIBS}

src/graph/test/SessionTest.cpp
```diff
@@ -0,0 +1,42 @@
...
+};
+
+TEST_F(SessionTest, ShowSessions) {
+    auto client = gEnv->getClient();
+    ASSERT_NE(nullptr, client);
+    {
+        cpp2::ExecutionResponse resp;
+        auto *fmt = "SHOW SESSIONS";
+        auto query = fmt;
+        auto code = client->execute(query, resp);
```
laura-ding, line 36: checkout the result, such as the number of the clients and sessionid

src/meta/client/MetaClient.cpp
```diff
@@ -157,6 +158,36 @@ bool MetaClient::loadUsersAndRoles() {
...
+        folly::RWSpinLock::WriteHolder holder(localCacheLock_);
+        sessionsCache_ = std::move(SessionMap);
+    }
+    return true;
+}
+
+bool MetaClient::uploadSession() {
+    auto ret = updateSession(sessions_).get();
+    if (!ret.ok()) {
+        LOG(ERROR) << "List sesssions failed, status:" << ret.status();
```
laura-ding, line 185: Update

src/meta/client/MetaClient.cpp
```diff
@@ -1059,6 +1095,76 @@ StatusOr<int32_t> MetaClient::partsNum(GraphSpaceID spaceId) {
...
+}
+
+folly::Future<StatusOr<std::vector<cpp2::SessionItem>>>
+MetaClient::listSessions() {
+    cpp2::ListSessionsReq req;
+    folly::Promise<StatusOr<std::vector<cpp2::SessionItem>>> promise;
+    auto future = promise.getFuture();
+    getResponse(std::move(req), [] (auto client, auto request) {
+        return client->future_listSessions(request);
+    }, [] (cpp2::ListSessionsResp&& resp) -> decltype(auto) {
```
laura-ding, line 1162: indentation

src/meta/client/MetaClient.cpp
```diff
@@ -130,6 +130,7 @@ void MetaClient::heartBeatThreadFunc() {
             localLastUpdateTime_ = metadLastUpdateTime_;
         }
     }
+    uploadSession();
```
laura-ding, line 133: why not do it in `reclaimExpiredSessions`? and `pushSessionToCache` can delete

src/graph/SessionManager.cpp
panda-sheep: Yes.
Graphd manages the local session, even if metad is not running, you can connect to graphd through the console  and create a session

panda-sheep: So here use  a background thread  to do `doAddSession `.
 Of course there must be a time delay.

src/graph/ShowExecutor.cpp
panda-sheep: Good point,  👍

src/graph/test/CMakeLists.txt
panda-sheep: Good catch.

src/graph/test/SessionTest.cpp
panda-sheep: Good.

src/meta/client/MetaClient.cpp
panda-sheep: The way you speak at the moment is feasible.
Currently, metad is updated using graphd session information.
If we support` kill session`, it needs metad information to update the graphd.

src/meta/client/MetaClient.cpp
panda-sheep: Bright eyes

src/meta/client/MetaClient.cpp
panda-sheep: Good catch

etc/nebula-graphd.conf.default
```diff
@@ -24,8 +24,8 @@
 ########## networking ##########
 # Meta Server Address
 --meta_server_addrs=127.0.0.1:45500
-# Local ip
---local_ip=127.0.0.1
```
Shylock-Hg, line 28: Why rename this option?

etc/nebula-graphd.conf.production
```diff
@@ -24,8 +24,8 @@
 ########## networking ##########
 # Meta Server Address
 --meta_server_addrs=127.0.0.1:45500
-# Local ip
---local_ip=127.0.0.1
```
Shylock-Hg, line 28: dito.

etc/nebula-graphd.conf.default
panda-sheep: For graphd, `local_ip  `is not used.
Here graphd uses the local ip specified by graph_server_ip to start.

src/graph/SessionManager.h
```diff
@@ -47,10 +56,15 @@ class SessionManager final {
...
-    std::unordered_map<int64_t, SessionPtr>     activeSessions_;
-    std::unique_ptr<thread::GenericWorker>      scavenger_;
+    std::atomic<int64_t>                                            nextId_{0};
+    // TODO(dutor) writer might starve
+    folly::RWSpinLock                                               rwlock_;
+    std::unordered_map<int64_t, std::pair<SessionPtr, int64_t>>     activeSessions_;
+    std::unordered_map<int64_t, int64_t>                            activeSessForMeta_;
+    std::unique_ptr<thread::GenericWorker>                          scavenger_;
+    meta::MetaClient                                               *metaClient_{nullptr};
+    std::unique_ptr<folly::Executor>                                executor_;
```
wadeliuyi, line 66: why need two thread pool scavenger_ and executor_?

src/graph/ShowExecutor.cpp
```diff
@@ -1180,6 +1183,65 @@ std::string ShowExecutor::roleToStr(nebula::cpp2::RoleType type) {
     return role;
 }
 
+std::string ShowExecutor::timestampToStr(int64_t t) {
```
wadeliuyi, line 1186: maybe this function can move to a common file

src/graph/ShowExecutor.cpp
```diff
@@ -1180,6 +1183,65 @@ std::string ShowExecutor::roleToStr(nebula::cpp2::RoleType type) {
...
+
+        for (auto& item : sessionItems) {
+            std::vector<cpp2::ColumnValue> row;
+            row.resize(4);
+            row[0].set_str(item.get_addr());
+            row[1].set_integer(item.get_session_id());
+            row[2].set_str(timestampToStr(item.get_start_time()));
+            row[3].set_str(timestampToStr(item.get_update_time()));
+
+            rows.emplace_back();
```
wadeliuyi, line 1229: rows can resize sessionItems.size() first?

src/meta/SessionMan.cpp
```diff
@@ -0,0 +1,94 @@
...
+
+void SessionMan::removeSession(const std::string& addr, int64_t sessionId) {
+    auto iter = globalSessions_.find(addr);
+    if (iter == globalSessions_.end()) {
+        return;
+    }
+
+    auto idIter = iter->second.find(sessionId);
+    if (idIter != iter->second.end()) {
+        globalSessions_[addr].erase(idIter);
```
wadeliuyi, line 51: globalSessions_[addr] is iter?

src/meta/SessionMan.cpp
```diff
@@ -0,0 +1,94 @@
...
+        if (globalSessions_[addr].empty()) {
+            globalSessions_.erase(addr);
```
wadeliuyi, line 53: maybe globalSessions_.erase(iter) is fast

src/meta/SessionMan.cpp
```diff
@@ -0,0 +1,94 @@
...
+        auto idIter = idMap.find(sessionId);
+        // SessionId not exists
+        if (idIter == idMap.end()) {
+            globalSessions_[addr][sessionId] = std::make_pair(startTime, updateTime);
+        } else {
+        // Update update time
+            if (globalSessions_[addr][sessionId].first == startTime) {
+                globalSessions_[addr][sessionId].second = updateTime;
+            } else {
+                // Previously invalid data
```
wadeliuyi, line 35: here why do not care about invalid data? is there any chance to delete the invalid data?

src/meta/SessionMan.cpp
```diff
@@ -0,0 +1,94 @@
...
+                if (siter->second.second == UpdateTime) {
+                    globalSessions[gsIter->first][siter->first] =
+                        std::make_pair(siter->second.first, siter->second.second);
+                } else {
+                    continue;
+                }
+            }
+        }
+    }
+    globalSessions_ = std::move(globalSessions);
```
wadeliuyi, line 90: is there any performance data that time cost compare between recreate a hasmap and modify a hash map?

src/meta/client/MetaClient.cpp
```diff
@@ -1665,6 +1771,16 @@ MetaClient::getRolesByUserFromCache(const std::string& user) {
     return iter->second;
 }
 
+bool MetaClient::pushSessionToCache(std::string& addr,
+                                    std::unordered_map<int64_t, int64_t> sessions) {
+    if (!ready_) {
+        return false;
+    }
+    folly::RWSpinLock::WriteHolder holder(sessionCacheLock_);
+    sessions_[addr] = sessions;
```
wadeliuyi, line 1780: std::move(sessions)

src/meta/client/MetaClient.cpp
```diff
@@ -157,6 +158,36 @@ bool MetaClient::loadUsersAndRoles() {
...
+    }
+    {
+        folly::RWSpinLock::WriteHolder holder(localCacheLock_);
+        sessionsCache_ = std::move(SessionMap);
+    }
+    return true;
+}
+
+bool MetaClient::uploadSession() {
+    auto ret = updateSession(sessions_).get();
```
wadeliuyi, line 183: maybe there is no need pass the class member sessions_ by value, the member function updateSession can use it directly

src/meta/processors/sessionMan/GlobalSessionProcessor.cpp
```diff
@@ -0,0 +1,93 @@
...
+    const auto& sessionItems = req.get_session_items();
+    auto now = time::WallClock::fastNowInSec();
+
+    std::unordered_set<std::string> addrs;
+    for (auto& sessionItem : sessionItems) {
+        auto& addr = sessionItem.get_addr();
+        addrs.emplace(addr);
+        auto sessionId = sessionItem.get_session_id();
+        auto startTime = sessionItem.get_start_time();
+        SessionMan::addOrUpdateSession(addr, sessionId, startTime, now);
```
wadeliuyi, line 58: maybe singleton about SessionMan is more readability than static function and class

src/graph/ShowExecutor.cpp
panda-sheep: Good catch
I will do it in nebula  2.0

src/graph/ShowExecutor.cpp
panda-sheep: I do not think so.
Because emplace_back () creates an element and inserts it at the end.

src/meta/SessionMan.cpp
panda-sheep: Yes, invalid data will be deleted

src/meta/SessionMan.cpp
panda-sheep: not a iter

src/graph/test/SessionTest.cpp
panda-sheep: session id is a changing value and should not be displayed

src/meta/SessionMan.cpp
panda-sheep: Do you have a better way to delete data multiple times in the map?

src/meta/client/MetaClient.cpp
panda-sheep: Good point

src/meta/client/MetaClient.cpp
panda-sheep: Good catch

src/meta/processors/sessionMan/GlobalSessionProcessor.cpp
panda-sheep: Makes sense

src/common/network/NetworkUtils.cpp
```diff
@@ -388,6 +388,22 @@ StatusOr<std::string> NetworkUtils::getLocalIP(std::string defaultIP) {
         }
     }
     return Status::Error("No IPv4 address found!");
+ }
+
+StatusOr<bool> NetworkUtils::isLocalValidIP(std::string defaultIP) {
```
wadeliuyi, line 393: const std::string &defaultIP?

src/meta/SessionMan.cpp
wadeliuyi: I mean globalSessions_[addr] can replace by iter->second, else globalSessions_[addr] will iterator the map again.

src/meta/SessionMan.cpp
wadeliuyi: you have get the position of addr in the map, so you can erase the iter directly, else will iterate the tree again.

src/meta/SessionMan.cpp
wadeliuyi: I do not the use case about this function and how many members in the map. if the just one member in addrs, but each time need iterator all the member of the globalSessions_.
and if the member of globalSessions_ is very much, then we can save a list save member that sorted by siter->second.second

dangleptr:
Move to 2.0.

PR #1801, opened 2020-02-22
Task Manager

liuyu85cn:

PR #2067, opened 2020-04-08
Rebuild index with JobManager

darionyaphet:
dependent on [Task Manager](https://github.com/vesoft-inc/nebula/pull/1801/)


dangleptr:
Close it,   it will be reimplemented in storage 2.0

PR #2069, opened 2020-04-09
Support for using dynamic libraries

laura-ding:
The project's debug compiling size reaches about 13GB, and we can't test the master on GitHub anymore because we're running out of space. So we modify binary to use dynamic libraries.
With the dynamic library, the size reduces to 5.9GB.


CMakeLists.txt
```diff
@@ -78,6 +81,13 @@ if (ENABLE_TESTING)
     enable_testing()
 endif()
 
+if (ENABLE_TESTING OR ENABLE_ONLY_SHARED)
+    add_compile_options(-fPIC)
+    message(STATUS "${CMAKE_BINARY_DIR}/lib")
+    set(ENV{LD_LIBRARY_PATH} "${CMAKE_BINARY_DIR}/lib:$ENV{LD_LIBRARY_PATH}")
+    set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -Wl,-rpath-link=${CMAKE_BINARY_DIR}/lib:/opt/vesoft/toolset/clang/9.0.0/lib64/")
```
dangleptr, line 88: Could we not use the hard code path inside our CMakeLists.txt

CMakeLists.txt
```diff
@@ -619,3 +647,7 @@ ENDIF()
 MESSAGE(STATUS "ENABLE_PACK_ONE: ${ENABLE_PACK_ONE}")
 include(CPackage)
 package(${ENABLE_PACK_ONE})
+
+if (ENABLE_TESTING OR ENABLE_ONLY_SHARED)
+    MESSAGE(STATUS "You could run 'export LD_LIBRARY_PATH=${CMAKE_BINARY_DIR}/lib:/opt/vesoft/toolset/clang/9.0.0/lib64/:\$LD_LIBRARY_PATH' if you want to run ctest")
```
dangleptr, line 652: ditto

CMakeLists.txt
```diff
@@ -469,6 +487,16 @@ macro(nebula_add_library name type)
     )
 endmacro()
 
+# Generate object or shared lib
+macro(nebula_add_all_library name)
+    if (NOT ENABLE_ONLY_SHARED)
```
dangleptr, line 492: The logic is right?
If ENABLE_TESTING is true and ENABLE_ONLY_SHARED is false, what happens? link them both?

CMakeLists.txt
```diff
@@ -78,6 +81,13 @@ if (ENABLE_TESTING)
     enable_testing()
 endif()
 
+if (ENABLE_TESTING OR ENABLE_ONLY_SHARED)
```
dangleptr, line 84: To my understanding,  you want to use shared libs for UTs and give an option for binaries,  right? 
If so, why not set ENABLE_ONLY_SHARED to be true directly when ENABLE_TESTING switch on.  So you check logic below will be much simpler. 
Correct me if i am wrong.

src/interface/CMakeLists.txt
```diff
@@ -7,20 +7,28 @@ set(THRIFT1 ${Fbthrift_BIN})
 set(THRIFT_TEMPLATES ${Fbthrift_INCLUDE_DIR}/thrift/templates)
 include(ThriftGenerate)
 
-# Target object name : common_thrift_obj
-thrift_generate("common" "" ${CMAKE_CURRENT_SOURCE_DIR} ${CMAKE_CURRENT_BINARY_DIR} "interface")
+set(type "OBJECT")
+if (ENABLE_TESTING AND (NOT ENABLE_ONLY_SHARED))
+    set(type "ALL")
```
dangleptr, line 12: Why we need to generate libs and objects all in this situation.

laura-ding:
1.0 doesn't need it anymore

PR #2070, opened 2020-04-10
Disabled PCH by default

dutor:
PCH(Precompiled Header) has introduced too many troubles to us. Here are a few:
  * It paralyzes the out-of-source build machanism, since it forces the generated file to be at the same folder with its source header.
  * We are mandated to apply exactly the same set of compile options to all of the source files.
  * It cannot work with PIE(Position Independent Executable) and ASLR(Address Space Layout Randomization), which are now the common configurations for most recent Linux systems(for the sake of security).
  * It is problematic along with ccache.

Now, to enable PCH, you have to provide an additional CMake option `-DENABLE_PCH=ON`.


Amber1990Zhang:
Please add doc in this [repo](https://github.com/vesoft-inc/nebula-docs) if needed. Build by source doc link: https://github.com/vesoft-inc/nebula-docs/blob/master/docs/manual-EN/3.build-develop-and-administration/1.build/1.build-source-code.md

PR #2073, opened 2020-04-14
Supports older version of schema for index write

bright-starry-sky:
When insert vertex, need to check if an old index exists, if so, drop the old index and write a new index. At this point, an old index key needs to be assembled by old prop for index check, however, the old property values do not have this new column. so leads to `storaged` crash by `CHECK(prop)`.
Improved this logic，When an invalid property value appears, write index skipped .

repeat step : 

- create space
- use space
- create tag1 (col1, col2)
- create index1 on tag1 (col1)
- insert vertex tag1 values (...)
- alter tag add column col3
- create index2 on tag1 (col3)
- insert vertex tag1 values (...)

### DIFF
--- a/src/graph/test/IndexTest.cpp
+++ b/src/graph/test/IndexTest.cpp
@@ -687,6 +687,83 @@ TEST_F(IndexTest, EdgeIndexTTL) {
     }
 }
 
+TEST_F(IndexTest, AlterTag) {
+    auto client = gEnv->getClient();
+    ASSERT_NE(nullptr, client);
+    {
+        cpp2::ExecutionResponse resp;
+        std::string query = "CREATE SPACE tag_index_space(partition_num=1, replica_factor=1)";
+        auto code = client->execute(query, resp);
+        ASSERT_EQ(cpp2::ErrorCode::SUCCEEDED, code);
+
+        query = "USE tag_index_space";
+        code = client->execute(query, resp);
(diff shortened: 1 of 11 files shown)


src/graph/test/IndexTest.cpp
```diff
@@ -687,6 +687,83 @@ TEST_F(IndexTest, EdgeIndexTTL) {
...
+    {
+        cpp2::ExecutionResponse resp;
+        auto query = "INSERT VERTEX person(name, age, gender, email) VALUES "
+                     "100:  (\"Tim\",  18, \"M\", \"tim@ve.com\")";
+        auto code = client->execute(query, resp);
+        ASSERT_EQ(cpp2::ErrorCode::SUCCEEDED, code);
+    }
+    {
+        cpp2::ExecutionResponse resp;
+        auto query = "ALTER TAG person ADD (col1 int)";
```
CPWstatic, line 725: Does it work well when we change the prop name or type? It's better to test it.

bright-starry-sky: > Does it work well when we change the prop name or type? It's better to test it.

Don't allow to `alter tag change column` or `alter tag drop column` when the tag attached index. An error message will be return.
If you want to `alter tag change column` or `alter tag drop column` , need drop index firstly.

PR #2076, opened 2020-04-15
enhance helm deploy for k8s cluster

dockerzhang:
1,change pullPolicy for limited network;
2,remove redundant specified namespace;
3,refactor MetadHosts config from svc value, it could deploy on standalone k8s cluster;
4,change storage setting more configurable, it could deploy if there is no available storageClass for test;

this patch is tested on helm v3.1.2, and the k8s cluster is standalone with one node.


knightXun:
thanks for this pr, But i cannot approve this, Because we are waiting for dns feature.  See #1921.

knightXun:
You cloud try this on k8s with 3 nodes, Single node nebula is not our common case

PR #2077, opened 2020-04-15
Add operator contains.

CPWstatic:


src/graph/test/GoTest.cpp
```diff
@@ -2205,6 +2255,101 @@ TEST_P(GoTest, DuplicateColumnName) {
...
+        std::vector<std::tuple<std::string, int64_t, int64_t, std::string>> expected = {
+            {player.name(), 2003, 2005, "Hawks"},
+        };
+        ASSERT_TRUE(verifyResult(resp, expected));
+    }
+    {
+        cpp2::ExecutionResponse resp;
+        auto &player = players_["Boris Diaw"];
+        auto *fmt = "GO FROM %ld OVER serve "
+                    "WHERE (string)serve.start_year CONTAINS \"05\" "
```
laura-ding, line 2283: add `upper` or `lower` function, and Chinese？

src/common/filter/Expressions.h
```diff
@@ -313,6 +313,10 @@ class Expression {
         return std::abs(left - right) < EPSILON;
     }
 
+    static bool contains(const std::string& left, const std::string& right) {
```
wadeliuyi, line 316: is there any case that right is empty string?

src/graph/test/GoTest.cpp
CPWstatic: OK.

src/common/filter/Expressions.h
CPWstatic: It's possible.

whitewum:
@Amber1990Zhang

Amber1990Zhang:
doc done.

PR #2079, opened 2020-04-16
double support scientific notation.

xuguruogu:
Parser can deal with scientific notation like 123.456E1.


xuguruogu:
1E10 is supposed to be regard as integer or double? The format is rarely used.

Shylock-Hg:
> 1E10 is supposed to be regard as integer or double? The format is rarely used.

I think it's a double, such as the `.3` or `3.` are treat as double also in c/c++ grammar. So I think it's intuitive.

Amber1990Zhang:
doc done

PR #2090, opened 2020-04-23
Upgrade guava version

darionyaphet:


src/tools/spark-sstfile-generator/src/main/scala/com/vesoft/nebula/tools/generator/v2/SparkClientGenerator.scala
```diff
@@ -222,7 +222,8 @@ object SparkClientGenerator {
             records.foreach { row =>
               val exec = row.getString(0)
               if (rateLimiter.tryAcquire(rateTimeout, TimeUnit.MILLISECONDS)) {
-                val future = client.execute(exec)
+                val future  = client.execute(exec)
+                val service = Executors.newFixedThreadPool(1);
```
dutor, line 226: What's this for?

dangleptr:
Is there any problem with the original guava version? 
If not,  we'd better upgrade the version inside nebula-java directly. 
@darionyaphet

PR #2092, opened 2020-04-24
Skip the space check for grant when authority type is cloud

bright-starry-sky:
When cloud authority is turnned on, A god user needs to be initialized by meta client, and grant this user to GOD  with meta default space.  Then need to ignore check whether the space exists or not. 
The space exists check has been processed at the graph layer. so skip the check on meta later for cloud authority.


bright-starry-sky:
Through offline discussion, meta still need check space . 
Now only skip the God role check by meta client api. if grant a user to GOD through graph layer. an error will be return.

PR #2093, opened 2020-04-24
modify etc configs for production usage

whitewum:
### Checklist


etc/nebula-metad.conf.default
```diff
@@ -34,6 +34,6 @@
 
 ########## Misc #########
 # The default number of parts when a space is created
---default_parts_num=100
+# --default_parts_num=100
```
critical27, line 37: How about just remove them from conf.

etc/nebula-storaged.conf.default
```diff
@@ -27,25 +27,36 @@
...
 # The default block cache size used in BlockBasedTable.
 # The unit is MB.
---rocksdb_block_cache=4
+--rocksdb_block_cache=1
 # The type of storage engine, `rocksdb', `memory', etc.
 --engine_type=rocksdb
 
 ############## rocksdb Options ##############
 --rocksdb_disable_wal=true
+--wal_ttl=3600
```
critical27, line 51: How about move wal_ttl to storage?

etc/nebula-metad.conf.default
whitewum: done

etc/nebula-storaged.conf.default
whitewum: refact a bit

etc/nebula-metad.conf.production
```diff
@@ -31,9 +31,3 @@
 ########## storage ##########
 # Root data path, here should be only single path for metad
 --data_path=data/meta
-
```
critical27, line 34: Sorry for I miss the `heartbeat_interval_secs` in `nebula-storaged.conf.production`, please also add heartbeat_interval_secs=10 here, it need to be matched in two configuration. 

That's why the `show hosts` would show different things over different times.

PR #2111, opened 2020-04-30
To improve 'Show roles in space'

bright-starry-sky:
To improve 'Show roles in space' : 

- If the logged-in user is role GOD or ADMIN , show all roles with permission space. 
- If the logged-in user is role DBA, USER or GUEST , show only his own role.


Amber1990Zhang:
doc done

PR #2113, opened 2020-05-06
Added hostname support for Nebula

monadbobo:
summary:

There are two parts to use hostname in Nebula, the first part is the meta server saves the partition information of storage, the second part is the raft.

For the first case, in order to be compatible with the old version, every time the heartbeat packet of storage is passed to meta server,
it will bring the local hostname with it,  and after receiving the heartbeat packet, meta server will save a kv pair (key is hostname, value is ip address),
meta will judge whether the IP address corresponding to the current hostname changes, if it does, it will update all the information corresponding to the old IP address.

In the second case, for compatibility with the old version,
it will bring the local hostname with it when voting, and for the receiver,
if the received IP address is not in its own host list, it will update the IP address corresponding to the current hostname in the list.

Finally, for a better representation of the IP address, the folly:IPAddress is encapsulated here as a representation of the address in Nebula.

### DIFF
--- a/src/meta/ActiveHostsMan.cpp
+++ b/src/meta/ActiveHostsMan.cpp
@@ -25,7 +25,6 @@ int ActiveHostsMan::updateHostIPaddress(kvstore::KVStore* kv,
     LOG(INFO) << "updateHostIPaddress will update " << ipAddress << " to " << hostAddr;
 
     std::string leaderValue;
-    std::vector<kvstore::KV> data;
     auto leaderKey = MetaServiceUtils::leaderKey(ipAddress.toLongHBO(), ipAddress.getPort());
     ret = kv->get(kDefaultSpaceId, kDefaultPartId, leaderKey, &leaderValue);
     if (ret != kvstore::SUCCEEDED && ret != kvstore::ERR_KEY_NOT_FOUND) {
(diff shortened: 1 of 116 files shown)


src/meta/MetaServiceUtils.cpp
```diff
@@ -139,6 +140,13 @@ std::string MetaServiceUtils::hostKey(IPv4 ip, Port port) {
     return key;
 }
 
+std::string MetaServiceUtils::domainKey(const std::string& domain) {
```
liuyu85cn, line 143: If i want start 3 storaged in one host, the 1st and 2nd will be overwrite.

src/meta/ActiveHostsMan.cpp
```diff
@@ -13,21 +13,142 @@ DEFINE_int32(expired_threshold_sec, 10 * 60,
...
+    std::vector<kvstore::KV> data;
+    auto leaderKey = MetaServiceUtils::leaderKey(ipAddress.toLongHBO(), ipAddress.getPort());
+    ret = kv->get(kDefaultSpaceId, kDefaultPartId, leaderKey, &leaderValue);
+    if (ret != kvstore::SUCCEEDED && ret != kvstore::ERR_KEY_NOT_FOUND) {
+        return -1;
+    }
+
+    std::vector<std::string> removeData;
+    folly::Baton<true, std::atomic> baton;
+    removeData.emplace_back(MetaServiceUtils::hostKey(ipAddress.toLongHBO(), ipAddress.getPort()));
```
liuyu85cn, line 37: may be we need to clean the domain table also.

PR #2117, opened 2020-05-09
fix get pid

### LINKED ISSUE #2118: Startup script failed to ignore comments

Please check the [FAQ](https://github.com/vesoft-inc/nebula/blob/master/docs/manual-EN/1.overview/2.quick-start/2.FAQ.md) documentation and old issues before raising an issue in case someone has asked the same question that you are asking.

Startup script failed to ignore comments

Added a comment to the pid file path in the graph's configuration file, but it failed to achieve the desired effect. As a result, the graph service cannot start and stop the service normally

**Your Environments (__must be provided__)**

* OS: `centos 7.8`

laura-ding:
close #2118

### DIFF
--- a/scripts/utils.sh
+++ b/scripts/utils.sh
@@ -98,7 +98,7 @@ function wait_for_exit {
 # To read a config item's value from the config file
 # args: <config file> <config item name>
 function get_item_from_config {
-    sed -n -e "s/--${2}=\(.*\)$/\1/p" ${1} 2>/dev/null
+    sed -n -e "s/^--${2}=\(.*\)$/\1/p" ${1} 2>/dev/null | sed -n "1p"
 }
 
 


scripts/utils.sh
```diff
@@ -98,7 +98,7 @@ function wait_for_exit {
 # To read a config item's value from the config file
 # args: <config file> <config item name>
 function get_item_from_config {
-    sed -n -e "s/--${2}=\(.*\)$/\1/p" ${1} 2>/dev/null
+    sed -n -e "s/^--${2}=\(.*\)$/\1/p" ${1} 2>/dev/null | sed -n "1p"
```
Shylock-Hg, line 101: `sed -n -e "s/^--pid_file=\(.*\)$/\1/p" etc/nebula-storaged.conf 2>/dev/null | sed -n "1p" | cut -d " " -f1
` better.

monadbobo:
LGTM.

PR #2124, opened 2020-05-13
Correct the index write benchmark

bright-starry-sky:


bright-starry-sky:
> The benchmark numbers should be updated?

Don't need . Only corrected the  small errors. These errors are caused by another PR merges.

dangleptr:
> > The benchmark numbers should be updated?
> 
> Don't need . Only corrected the small errors. These errors are caused by another PR merges.

Fine.

Issue #1631, opened 2020-01-15
LDBC_SNB benchmark for Nebula


ouzensang:
Hi, Nebula
    Recently, I am doing some experiments based on ldbc_snb benchmark. And I wan to test Nebula on ldbc. So I want to know if you have implemented ldbc_snb benchmark for Nebula? If so, could you please provide me a solution to run ldbc_snb benchmark on Nebula? If not, now the query language in Nebula whether can support interactive complex and business intelligence queries in LDBC_SNB benchmark(https://ldbc.github.io/ldbc_snb_docs/ldbc-snb-specification.pdf)?
    Thanks for your kindly support and I'm looking forward to your early reply!

dutor:
Sorry about so late to reply.

Recently, we've been working on applying the LDBC SNB workload on Nebula Graph.
Here now I can provide a schema for SNB workload, not the final complete one though:
```
CREATE SPACE IF NOT EXISTS ldbc_snb(PARTITION_NUM = 1024)

USE ldbc_snb

CREATE TAG IF NOT EXISTS person(first_name string, last_name string, gender string, ip string, browser string)
CREATE TAG IF NOT EXISTS place(name string, type string, url string)
CREATE TAG IF NOT EXISTS organization(name string, type string, url string)
CREATE TAG IF NOT EXISTS post(time string, image string, ip string, browser string, language string, content string, length int)
CREATE TAG IF NOT EXISTS comment(time string, ip string, browser string, content string, length int)
CREATE TAG IF NOT EXISTS forum(time string, title string, type string)

CREATE EDGE IF NOT EXISTS knows()
CREATE EDGE IF NOT EXISTS is_part_of()
CREATE EDGE IF NOT EXISTS is_located_in()
CREATE EDGE IF NOT EXISTS study_at(year string)
CREATE EDGE IF NOT EXISTS work_at(year string)
CREATE EDGE IF NOT EXISTS has_post(time string)
CREATE EDGE IF NOT EXISTS has_comment(time string)
CREATE EDGE IF NOT EXISTS likes_post(time string)
CREATE EDGE IF NOT EXISTS likes_comment(time string)
CREATE EDGE IF NOT EXISTS is_reply_of_post(time string)
CREATE EDGE IF NOT EXISTS is_reply_of_comment(time string)
CREATE EDGE IF NOT EXISTS has_member(time string, type string)
```

Besides, we will provide the whole procedure of the testing, along with the import and benchmark tools later.

[...]

PR #2129, opened 2020-05-15
Reduced queueing latency

dutor:
`IOThreadPoolExecutor` and `CPUThreadPoolExecutor` differ in that, threads in the latter share a single task queue, while there is one queue for each thread in the former. Given that, in the query bound(aka. GetNeighbors) scenario,  there are chances that light tasks are queued after some heavy ones. This will increase the overall latency, even worse if we split the light tasks in order to execute them concurrently. 

By using `CPUThreadPoolExecutor`, we could minimize the queueing latency. At the meantime, the concurrent execution can be fully utilized.

As a result, latency can be reduced by more than a half in the LDBC workload.


dangleptr:
> FYI. Latency can be reduced by more than a half in the LDBC workload.

What's the concurrency setting (max_handlers_per_req) in your load test.

dutor:
> > FYI. Latency can be reduced by more than a half in the LDBC workload.
> 
> What's the concurrency setting (max_handlers_per_req) in your load test.

256

dangleptr:
Em...  Typically,  we use max_handlers_per_req = 1 in load test.  That's because the larger concurrency setting will hurt the throughput in high load mode.  
What's your max throughput with 256?

whitewum:
#1631

dutor:
> Em... Typically, we use max_handlers_per_req = 1 in load test. That's because the larger concurrency setting will hurt the throughput in high load mode.
With that setting, and `GO 3 STEPS...`, there is always one thread saturated, which leads to latency burst up.

> What's your max throughput with 256?
With the same client concurrency(28 conn), this PR boosts the QPS from 400 to 800, P99 400ms to 200ms.

dangleptr:
> With that setting, and GO 3 STEPS..., there is always one thread saturated while the others are nearly idle, which leads to latency burst up.

Indeed,  with 1 max_handlers_per_req setting,  the large request will block all requests in that thread.  What's the vertices requested at most in your situation.

dutor:
> Indeed, with 1 max_handlers_per_req setting, the large request will block all requests in that thread. What's the vertices requested at most in your situation.

Upto hundreds of thousands.

dangleptr:
> With the same client concurrency(28 conn), this PR boosts the QPS from 400 to 800, P99 400ms to 200ms.

I suggest we'd better test the 1-step request with 1 max_handlers_per_req setting again.  
Because cpuThreadPool has high contention than ioThreadPool for normal requests.

dutor:
> > With the same client concurrency(28 conn), this PR boosts the QPS from 400 to 800, P99 400ms to 200ms.
> 
> I suggest we'd better test the 1-step request with 1 max_handlers_per_req setting again.
> Because cpuThreadPool has high contention than ioThreadPool for normal requests.

Agreed. I also had this concern. But according to my tests, the 1-step scenario seems no observable degradation.

dangleptr:
> > > With the same client concurrency(28 conn), this PR boosts the QPS from 400 to 800, P99 400ms to 200ms.
> > 
> > 
> > I suggest we'd better test the 1-step request with 1 max_handlers_per_req setting again.
> > Because cpuThreadPool has high contention than ioThreadPool for normal requests.
> 
> Agreed. I also had this concern. But according to my tests, the 1-step scenario seems no observable degradation.

If so,  i am OK with the changes.  
Do you mind share the numbers about your 1-step-1-handler testing here?

@critical27 Maybe we should add the concurrent setting back in 2.0.   :(

CPWstatic:
Maybe we need more test cases to evaluate this improvement.

critical27:
Em, interesting.

PR #2139, opened 2020-05-24
Limit total partitions

knightXun:
because we leave limited memory and cpu for nebula in pubilc cloud, we must limit the number of spaces and partitions. Otherwise it's very dangerous. consider trial-version take one storaged, this may cause data loss. Never miss this

### DIFF
--- a/src/graph/test/PartitionTest.cpp
+++ b/src/graph/test/PartitionTest.cpp
@@ -1,4 +1,4 @@
-/* Copyright (c) 2018 vesoft inc. All rights reserved.
+/* Copyright (c) 2020 vesoft inc. All rights reserved.
  *
  * This source code is licensed under Apache 2.0 License,
  * attached with Common Clause Condition 1.0, found in the LICENSES directory.
--- a/src/graph/test/PartitionTest.cpp
+++ b/src/graph/test/PartitionTest.cpp
@@ -10,8 +10,6 @@
 #include "meta/test/TestUtils.h"
 #include "storage/test/TestUtils.h"
 
-DECLARE_int32(heartbeat_interval_secs);
-
 namespace nebula {
 namespace graph {
 
--- a/src/graph/test/PartitionTest.cpp
+++ b/src/graph/test/PartitionTest.cpp
@@ -1,0 +1,57 @@
+/* Copyright (c) 2018 vesoft inc. All rights reserved.
+ *
+ * This source code is licensed under Apache 2.0 License,
+ * attached with Common Clause Condition 1.0, found in the LICENSES directory.
+ */
+
+#include "base/Base.h"
+#include "graph/test/TestEnv.h"
+#include "graph/test/TestBase.h"
+#include "meta/test/TestUtils.h"
+#include "storage/test/TestUtils.h"
+
+DECLARE_int32(heartbeat_interval_secs);
+
+namespace nebula {
+namespace graph {
+
+class PartitionTest : public TestBase {
+protected:
+    void SetUp() override {
+        TestBase::SetUp();
+        // ...
+    }
+
+    void TearDown() override {
+        // ...
+        TestBase::TearDown();
+    }
+};
+
+TEST_F(PartitionTest, limit_partitions) {
+    auto client = gEnv->getClient();
+    ASSERT_NE(nullptr, client);
+    {
+        cpp2::ExecutionResponse resp;
+        std::string query = "CREATE SPACE space1 (partition_num=50)";
+        auto code = client->execute(query, resp);
+        ASSERT_EQ(cpp2::ErrorCode::SUCCEEDED, code);
+        query = "CREATE SPACE space2 (partition_num=51)";
+        code = client->execute(query, resp);
+        ASSERT_NE(cpp2::ErrorCode::SUCCEEDED, code);
+    }
+    {
+        cpp2::ExecutionResponse resp;
+        std::string query = "drop SPACE space1";
+        auto code = client->execute(query, resp);
+        ASSERT_EQ(cpp2::ErrorCode::SUCCEEDED, code);
+    }
+    {
+        cpp2::ExecutionResponse resp;
+        std::string query = "CREATE SPACE space_partitions(partition_num=101)";
+        auto code = client->execute(query, resp);
+        ASSERT_NE(cpp2::ErrorCode::SUCCEEDED, code);
+    }
+}
+}   // namespace graph
+}   // namespace nebula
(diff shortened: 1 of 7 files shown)


knightXun:
If we dont limit the total number of partitions, the cpu and memory usage will very high.
image

knightXun:
cc @dangleptr @yixinglu

Shylock-Hg:
Please add test cases and check the error message.

src/meta/processors/partsMan/CreateSpaceProcessor.cpp
```diff
@@ -119,6 +129,35 @@ CreateSpaceProcessor::pickHosts(PartitionID partId,
     return pickedHosts;
 }
 
+int32_t CreateSpaceProcessor::GetPartsNumbers() {
+    int32_t count = 0;
+    auto prefix = MetaServiceUtils::spacePrefix();
+    std::unique_ptr<kvstore::KVIterator> iter;
+    auto ret = kvstore_->prefix(kDefaultSpaceId, kDefaultPartId, prefix, &iter);
+    if (ret != kvstore::ResultCode::SUCCEEDED) {
+        return -1;
```
laura-ding, line 138: I don't know what it means to return -1 here. the caller does not process it. 
If here failed, is means the rocksdb may be abnormal, and we should return and don’t continue.

src/interface/meta.thrift
```diff
@@ -49,6 +50,7 @@ enum ErrorCode {
     E_INVALID_CHARSET           = -45,
     E_INVALID_COLLATE           = -46,
     E_CHARSET_COLLATE_NOT_MATCH = -47,
+    E_TOO_LARGE_REPLICA_FACTOR   = -48,
```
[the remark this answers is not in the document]
knightXun: good point

src/meta/processors/partsMan/CreateSpaceProcessor.cpp
knightXun: If return -1, we should throw a exception here?

knightXun: fix this

src/meta/processors/partsMan/CreateSpaceProcessor.cpp
```diff
@@ -119,6 +129,35 @@ CreateSpaceProcessor::pickHosts(PartitionID partId,
...
+    }
+
+    while (iter->valid()) {
+        auto spaceId_ = MetaServiceUtils::spaceId(iter->key());
```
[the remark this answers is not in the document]
knightXun: fixed

knightXun:
> Please add test cases and check the error message.

done

src/meta/processors/partsMan/CreateSpaceProcessor.cpp
```diff
@@ -119,6 +134,38 @@ CreateSpaceProcessor::pickHosts(PartitionID partId,
...
+        auto spaceName = MetaServiceUtils::spaceName(iter->val());
+        std::unique_ptr<kvstore::KVIterator> partIter;
+        auto kvRet = kvstore_->prefix(kDefaultSpaceId, kDefaultPartId, partPrefix, &partIter);
+        if ( kvRet != kvstore::ResultCode::SUCCEEDED ) {
+            LOG(ERROR) << "List Partitions From " << spaceName
+                       << ", Space id " << spaceId << " Failed";
+            return -1;
+        }
+
+        while (partIter->valid()) {
```
laura-ding, line 159: You can get the partition_num from space properties, don't need list parts.

knightXun: good point, Done

knightXun:
cc @laura-ding done

src/graph/test/PartitionTest.cpp
```diff
@@ -0,0 +1,57 @@
+/* Copyright (c) 2018 vesoft inc. All rights reserved.
```
laura-ding, line 1: 2020

src/graph/test/PartitionTest.cpp
```diff
@@ -0,0 +1,57 @@
...
+ * attached with Common Clause Condition 1.0, found in the LICENSES directory.
+ */
+
+#include "base/Base.h"
+#include "graph/test/TestEnv.h"
+#include "graph/test/TestBase.h"
+#include "meta/test/TestUtils.h"
+#include "storage/test/TestUtils.h"
+
+DECLARE_int32(heartbeat_interval_secs);
```
laura-ding, line 13: Don't need

src/graph/test/PartitionTest.cpp
knightXun: done

src/graph/test/PartitionTest.cpp
knightXun: done

knightXun:
cc @dutor

src/meta/processors/partsMan/CreateSpaceProcessor.cpp
```diff
@@ -9,6 +9,7 @@
 
 DEFINE_int32(default_parts_num, 100, "The default number of parts when a space is created");
 DEFINE_int32(default_replica_factor, 1, "The default replica factor when a space is created");
+DEFINE_int32(default_max_parts_num, 100, "The default max partitions number");
```
Shylock-Hg, line 12: I think you can set default to max, and give it a property value in configuration when deploy.

knightXun: 100 partitions is enough for trial version, user can create up to 100 spaces.

src/meta/processors/partsMan/CreateSpaceProcessor.cpp
```diff
@@ -119,6 +134,24 @@ CreateSpaceProcessor::pickHosts(PartitionID partId,
     return pickedHosts;
 }
 
+int32_t CreateSpaceProcessor::GetPartsNumbers() {
```
wadeliuyi, line 137: Is this function mean get all part number of all spaces?

knightXun: yeap

src/meta/processors/partsMan/CreateSpaceProcessor.cpp
dangleptr: Currently,  the largest number for parts is 2^24,   the default max number should be it.

dangleptr:
That's fine for cloud branch

PR #2146, opened 2020-05-27
Add warning msg

laura-ding:
When executing `drop space` succeed, the console will print the warning msg:
Screen Shot 2020-05-27 at 15 23 01


src/graph/DropSpaceExecutor.cpp
```diff
@@ -54,5 +54,8 @@ void DropSpaceExecutor::execute() {
     std::move(future).via(runner).thenValue(cb).thenError(error);
 }
 
+void DropSpaceExecutor::setupResponse(cpp2::ExecutionResponse &resp)  {
+    resp.set_warning_msg("The data will be deleted completely after the service restart.");
```
yixinglu, line 58: restarting

PR #2152, opened 2020-05-30
Fix a bug which returns the history version data in query_bound.

CPWstatic:
The `continue`s lead to this problem.


critical27:
I'm still confused about why we return history data, you mean old data could pass ttl check but latest could not?

CPWstatic:
> I'm still confused about why we return history data, you mean old data could pass ttl check but latest could not?

Yes, the same problem with the filter. But the main problem is that we should not touch the elder version data. The `continue`s would lead to the loop to read elder version data which is unacceptable.

critical27:
> > I'm still confused about why we return history data, you mean old data could pass ttl check but latest could not?
> 
> Yes, the same problem with the filter. But the main problem is that we should not touch the elder version data. The `continue`s would lead to the loop to read elder version data which is unacceptable.

👍

PR #2154, opened 2020-06-01
Return keys when tag edge has no prop

CPWstatic:


src/graph/FetchEdgesExecutor.cpp
```diff
@@ -372,11 +366,14 @@ void FetchEdgesExecutor::processResult(RpcResponse &&result) {
             outputSchema->appendCol(edgeSrcName_, nebula::cpp2::SupportedType::VID);
             outputSchema->appendCol(edgeDstName_, nebula::cpp2::SupportedType::VID);
             outputSchema->appendCol(edgeRankName_, nebula::cpp2::SupportedType::INT);
-            auto status = getOutputSchema(eschema.get(), &*iter, outputSchema.get());
-            if (!status.ok()) {
-                LOG(ERROR) << "Get output schema failed: " << status;
-                doError(Status::Error("Get output schema failed: %s.", status.toString().c_str()));
-                return;
+            if (!(eschema == nullptr || resultColNames_.empty())) {
```
panda-sheep, line 369: I think `if (eschema != nullptr && !resultColNames_.empty())`  is more clearer?

src/graph/test/DataTest.cpp
```diff
@@ -1042,40 +1042,54 @@ class FetchEmptyPropsTest : public ::testing::Test {
...
-    ASSERT_NE(resp.get_rows(), nullptr);
-    ASSERT_TRUE(resp.get_rows()->empty());
-}
-
 // #1478
 // Fetch property from empty tag
 TEST_F(FetchEmptyPropsTest, EmptyProps) {
     {
+        cpp2::ExecutionResponse resp;
         const std::string stmt = "FETCH PROP ON empty_tag 1";
```
laura-ding, line 1050: Add a multi-tag test(contains at least one empty tag).

Issue #2153, opened 2020-06-01
”upsert ... when“ logic bug


jude-zhu:
edge `100 -> 106` of `follow` type not exists. Same as `100 -> 107`. However, the following `UPSERT` statement inserts an edge

```
>UPSERT EDGE 100 -> 106 OF follow SET degree=21 when follow.degree < 10 YIELD follow.degree

=================
| follow.degree |
=================
| 21            |
-----------------
```

while the following doesn't.

```
> UPSERT EDGE 100 -> 107 OF follow SET degree=21 when follow.degree > 10 YIELD follow.degree
=================
| follow.degree |
=================
| 0             |
-----------------

> fetch prop on follow 100-> 107 yield follow.degree
Execution succeeded (Time spent: 5.019/6.56 ms)
```

This edgetype do not have default value
![image](https://user-images.githubusercontent.com/51590253/83385130-3d6f5480-a41b-11ea-9308-28aeb856f30a.png)

btw. if nothing was inserted, I don't think it should return `0`.

PR #2156, opened 2020-06-02
upsert statement when insert, the condition is always true  NG-99

panda-sheep:
note，update and upsert only handle one tag of vertex and one edgeType


critical27:
@Amber1990Zhang, we need to update the "update" and "upsert" documents

jude-zhu:
close #2153

Amber1990Zhang:
doc done

PR #2158, opened 2020-06-04
Add more version Info in status webservice

knightXun:


src/webservice/StatusHandler.cpp
```diff
@@ -68,6 +68,8 @@ void StatusHandler::onError(ProxygenError error) noexcept {
 folly::dynamic StatusHandler::getStatus() {
     folly::dynamic json = folly::dynamic::object();
     json["status"] = "running";
+    json["version"] = folly::stringPrintf("%s", NEBULA_STRINGIFY(NEBULA_BUILD_VERSION));
```
laura-ding, line 71: #if defined(NEBULA_BUILD_VERSION)
     json["version"] = folly::stringPrintf("%s", NEBULA_STRINGIFY(NEBULA_BUILD_VERSION));
#endif

laura-ding: retract

docker/Dockerfile.console
```diff
@@ -3,7 +3,7 @@ FROM vesoft/nebula-dev:centos7 as builder
 COPY . /home/nebula/BUILD
 
 RUN cd /home/nebula/BUILD/package \
-  && ./package.sh -v $(git rev-parse --short HEAD) -n OFF
+  && ./package.sh -v v1.0.0 -n OFF
```
laura-ding, line 6: The docker uses the master code, better use the DateTime to as the version

PR #2165, opened 2020-06-05
meta client thread local access.

xuguruogu:
use thread local object when access meta client cache. folly::SingletonThreadLocal supply efficient access wrapper by using pointer.


xuguruogu:
the issue came from compaction.

before the changes:
image

after the changes:
image

dutor:
Thanks for this improvement. I also noticed this issue recently, and did some similar optimization. Many other modules may benefit from this patch.

dangleptr:
Please recheck the Failed UTs

xuguruogu:
centos7 unit tests is killed. may due to make oom

PR #2167, opened 2020-06-08
Remove some configs which be dynamic configured

knightXun:


knightXun:
cc @dutor @laura-ding we must limit dynamic configures in cloud-version

yixinglu:
Why not to give another flags.json for cloud version ?

knightXun:
> I think we may need some discussion offline about how to maintain the cloud version configuration.

agree

Shylock-Hg:
> Why not to give another flags.json for cloud version ?

Agree, I don't think make too many difference between cloud and master branch is better.

dangleptr:
If you want to change the gflags.json in your cluster.  You'd better change it directly and don't commit them into master.

PR #2184, opened 2020-06-17
fix compaction filter doesn't work in manual compaction

critical27:
1. As title
2. Make `custom_filter_interval_secs` could be updated via console.
3. Add a flag to control whether support multi-version of same key (the timestamp field in key). `enable_multi_versions`, If true, will use wall clock when insert data. If false, always insert as newest data.
4. However, there is a bug in previous `update`........ because the timestamp in `update` is wrong... @jude-zhu release a bug fix version?

```
UPSERT VERTEX 100 SET building.name = "No1"
FETCH PROP on building 100
INSERT VERTEX building(name) VALUES 100: ("No2")
FETCH PROP on building 100
```

### DIFF
--- a/src/storage/mutate/AddVerticesProcessor.cpp
+++ b/src/storage/mutate/AddVerticesProcessor.cpp
@@ -17,8 +17,7 @@ namespace storage {
 
 void AddVerticesProcessor::process(const cpp2::AddVerticesRequest& req) {
     auto version = FLAGS_enable_multi_versions ?
-        std::numeric_limits<int64_t>::max() - time::WallClock::fastNowInMicroSec() :
-        std::numeric_limits<int64_t>::max();
+        std::numeric_limits<int64_t>::max() - time::WallClock::fastNowInMicroSec() : 0L;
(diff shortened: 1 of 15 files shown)


src/storage/mutate/AddVerticesProcessor.cpp
```diff
@@ -16,8 +16,9 @@ namespace nebula {
 namespace storage {
 
 void AddVerticesProcessor::process(const cpp2::AddVerticesRequest& req) {
-    auto version =
-        std::numeric_limits<int64_t>::max() - time::WallClock::fastNowInMicroSec();
+    auto version = FLAGS_enable_multi_versions ?
+        std::numeric_limits<int64_t>::max() - time::WallClock::fastNowInMicroSec() :
+        std::numeric_limits<int64_t>::max();
```
dangleptr, line 21: it should be zero if FLAGS_enable_multi_versions is false

critical27: Yeah, may bad, hold on, there are some other ut problems.

PR #2186, opened 2020-06-19
fix update after alter

critical27:
Fix problems about upsert after alter schema, the default value not found.

```
INSERT VERTEX building(name) VALUES 100: ("No1")
ALTER TAG building ADD (new_field string default "123")
FETCH PROP on building 100
UPSERT VERTEX 100 SET building.name = "No3", building.new_field = "321"
FETCH PROP on building 100
```

Known issue: when go/fetch data of old schema versions, the default value in new schema won't work, still use int as 0, string as "".

### DIFF
--- a/src/storage/mutate/UpdateEdgeProcessor.cpp
+++ b/src/storage/mutate/UpdateEdgeProcessor.cpp
@@ -132,8 +132,8 @@ kvstore::ResultCode UpdateEdgeProcessor::collectVertexProps(
             if (!ok(res)) {
                 auto defaultVal = schema->getDefaultValue(prop.prop_.name);
                 if (!defaultVal.ok()) {
-                    VLOG(1) << "No default value of "
-                            << tagId << ", prop " << prop.prop_.name;
+                    LOG(WARNING) << "No default value of "
+                                 << tagId << ", prop " << prop.prop_.name;
                     return kvstore::ResultCode::ERR_TAG_NOT_FOUND;
                 }
                 v = std::move(defaultVal).value();
@@ -192,15 +192,15 @@ kvstore::ResultCode UpdateEdgeProcessor::collectEdgesProps(
             if (!ok(res)) {
                 auto defaultVal = constSchema->getDefaultValue(propName);
                 if (!defaultVal.ok()) {
-                    VLOG(1) << "No default value of "
-                            << edgeKey.edge_type << ", prop " << propName;
+                    LOG(WARNING) << "No default value of "
+                                 << edgeKey.edge_type << ", prop " << propName;
                     return kvstore::ResultCode::ERR_EDGE_NOT_FOUND;
                 }
                 v = std::move(defaultVal).value();
             } else {
                 v = value(std::move(res));
             }
-            edgeFilters_.emplace(propName, v);
+            edgeFilters_.emplace(propName, std::move(v));
         }
         updater_ = std::unique_ptr<RowUpdater>(new RowUpdater(std::move(reader), constSchema));
     } else if (insertable_) {
--- a/src/storage/mutate/UpdateEdgeProcessor.cpp
+++ b/src/storage/mutate/UpdateEdgeProcessor.cpp
@@ -117,15 +117,25 @@ kvstore::ResultCode UpdateEdgeProcessor::collectVertexProps(
             // load it. To protect the data, we just return failed to graphd.
             return kvstore::ResultCode::ERR_UNKNOWN;
         }
-        const auto constSchema = reader->getSchema();
+        const auto schema = this->schemaMan_->getTagSchema(this->spaceId_, tagId);
+        if (schema == nullptr) {
+            LOG(WARNING) << "Can't find the schema for tagId " << tagId;
+            return kvstore::ResultCode::ERR_TAG_NOT_FOUND;
+        }
         for (auto& prop : props) {
+            VariantType v;
             auto res = RowReader::getPropByName(reader.get(), prop.prop_.name);
             if (!ok(res)) {
-                VLOG(1) << "Skip the bad value for tag: "
-                        << tagId << ", prop " << prop.prop_.name;
-                return kvstore::ResultCode::ERR_UNKNOWN;
+                auto defaultVal = schema->getDefaultValue(prop.prop_.name);
+                if (!defaultVal.ok()) {
+                    VLOG(1) << "No default value of "
+                            << tagId << ", prop " << prop.prop_.name;
+                    return kvstore::ResultCode::ERR_TAG_NOT_FOUND;
+                }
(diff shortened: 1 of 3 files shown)


src/storage/mutate/UpdateEdgeProcessor.cpp
```diff
@@ -117,15 +117,25 @@ kvstore::ResultCode UpdateEdgeProcessor::collectVertexProps(
...
         for (auto& prop : props) {
+            VariantType v;
             auto res = RowReader::getPropByName(reader.get(), prop.prop_.name);
             if (!ok(res)) {
-                VLOG(1) << "Skip the bad value for tag: "
-                        << tagId << ", prop " << prop.prop_.name;
-                return kvstore::ResultCode::ERR_UNKNOWN;
+                auto defaultVal = schema->getDefaultValue(prop.prop_.name);
+                if (!defaultVal.ok()) {
+                    VLOG(1) << "No default value of "
```
laura-ding, line 131: use warning log?

src/storage/mutate/UpdateEdgeProcessor.cpp
```diff
@@ -165,15 +175,27 @@ kvstore::ResultCode UpdateEdgeProcessor::collectEdgesProps(
...
         for (auto index = 0UL; index < constSchema->getNumFields(); index++) {
             auto propName = std::string(constSchema->getFieldName(index));
             auto res = RowReader::getPropByName(reader.get(), propName);
+            VariantType v;
             if (!ok(res)) {
-                VLOG(1) << "Skip the bad edge value for prop " << propName;
-                return kvstore::ResultCode::ERR_UNKNOWN;
+                auto defaultVal = constSchema->getDefaultValue(propName);
+                if (!defaultVal.ok()) {
+                    VLOG(1) << "No default value of "
```
laura-ding, line 191: ditto

src/storage/mutate/UpdateVertexProcessor.cpp
```diff
@@ -113,15 +113,25 @@ kvstore::ResultCode UpdateVertexProcessor::collectVertexProps(
...
         for (auto& prop : props) {
+            VariantType v;
             auto res = RowReader::getPropByName(reader.get(), prop.prop_.name);
             if (!ok(res)) {
-                VLOG(1) << "Skip the bad value for tag: " << tagId
-                        << ", prop " << prop.prop_.name;
-                return kvstore::ResultCode::ERR_UNKNOWN;
+                auto defaultVal = schema->getDefaultValue(prop.prop_.name);
+                if (!defaultVal.ok()) {
+                    VLOG(1) << "No default value of "
```
laura-ding, line 127: ditto

src/storage/mutate/UpdateEdgeProcessor.cpp
```diff
@@ -165,15 +175,27 @@ kvstore::ResultCode UpdateEdgeProcessor::collectEdgesProps(
...
+                            << edgeKey.edge_type << ", prop " << propName;
+                    return kvstore::ResultCode::ERR_EDGE_NOT_FOUND;
+                }
+                v = std::move(defaultVal).value();
+            } else {
+                v = value(std::move(res));
             }
-            auto&& v = value(std::move(res));
             edgeFilters_.emplace(propName, v);
```
laura-ding, line 199: edgeFilters_.emplace(propName, std::move(v));

critical27: Good catch!

src/storage/mutate/UpdateEdgeProcessor.cpp
critical27: 👌

src/storage/mutate/UpdateEdgeProcessor.cpp
```diff
@@ -47,6 +47,10 @@ void UpdateEdgeProcessor::onProcessFinished(int32_t retNum) {
             return it->second;
         };
         for (auto& exp : returnColumnsExp_) {
+            if (!exp->prepare().ok()) {
+                LOG(ERROR) << "Expression::prepare failed";
+                return;
+            }
```
panda-sheep, line 53: The check here should be redundant， because `UpdateEdgeProcessor::checkAndBuildContexts` has already processed.

src/storage/mutate/UpdateVertexProcessor.cpp
```diff
@@ -39,6 +39,10 @@ void UpdateVertexProcessor::onProcessFinished(int32_t retNum) {
         };
         for (auto& exp : returnColumnsExp_) {
             auto value = exp->eval(getters);
+            if (!exp->prepare().ok()) {
+                LOG(ERROR) << "Expression::prepare failed";
+                return;
+            }
```
panda-sheep, line 45: ditto

src/storage/mutate/UpdateEdgeProcessor.cpp
critical27: Yeah, I missed that part, the result is same.

src/storage/mutate/UpdateEdgeProcessor.cpp
```diff
@@ -117,15 +121,25 @@ kvstore::ResultCode UpdateEdgeProcessor::collectVertexProps(
...
-        const auto constSchema = reader->getSchema();
+        const auto schema = this->schemaMan_->getTagSchema(this->spaceId_, tagId);
+        if (schema == nullptr) {
+            LOG(WARNING) << "Can't find the schema for tagId " << tagId;
+            return kvstore::ResultCode::ERR_TAG_NOT_FOUND;
+        }
         for (auto& prop : props) {
+            VariantType v;
             auto res = RowReader::getPropByName(reader.get(), prop.prop_.name);
             if (!ok(res)) {
```
dangleptr, line 132: The logic is right? 
The default value order should be "default value in schema" > "default value in RowReader". right?

dangleptr: Further more,  could we extract the default value logic into one place?

critical27: You think we need to add `default value in RowReader`? For now, most logic in update/upsert only check the `default value in schema`. If failed to get it, then we return error.

critical27: That would affect lots of space, I've discussed about it during a meeting with sherman-the-tank and jude-zhu, we just keep it for now.

One of the changes is that we need to modify all places that read from a row, which would face the problem of alter schema. In that place, we use the `default value in RowReader`. Indeed, it is inconsistent.

dangleptr: Thats fine.

PR #2187, opened 2020-06-21
fix SparkClientGenerator when reading file, wrong columns

knightjya:
fix SparkClientGenerator. When reading file on hdfs, wrong index was found. 

When reading file on hdfs, wrong index was found. The change fixes the logic to find colums. 
before: find colums in config file
after: find colums in hdfs file(parquet/json/csv/orc)

No.


src/tools/spark-sstfile-generator/src/main/scala/com/vesoft/nebula/tools/generator/v2/SparkClientGenerator.scala
```diff
@@ -310,7 +310,7 @@ object SparkClientGenerator {
               val values = (for {
                 property <- valueProperties if property.trim.length != 0
               } yield extraValue(row, property)).mkString(",")
-              (row.getString(vertexIndex), values)
+              (String.valueOf(extraValue(row, vertex)), values)
```
darionyaphet, line 313: If you want to transfer extraValue(row, vertex) into String `extraValue(row, vertex).toString` is better?

knightjya: String.valueOf is used to ensure null transferrd to 'null' as String instead of raising NullPointerException.
Actually the function extraValue has had the type transferred to String. No need is String.valueOf or toString to be used.

darionyaphet: Actually `extraValue` is using row.getString by the index of fields. But it's more safe. IMO it's a good idea

PR #2193, opened 2020-06-24
fix a potential raft term error

critical27:
A, B

1. B is leader of term 2
> start compaction
2. A start election serveral times, proposedTerm from 3 to 9, all timeout
3. A recv a appanedLog from B of term 2, set currentTerm and proposedTerm to 2
4. A start election again, proposedTerm = 3, this time is elected as leader of term 3
5. However, B's term is 9, it will not aceept A's log anymore, and it will keep starting election, but not one would vote for B

The key problem is how to handle timeout in election, for now we just do nothing, then bump up the `proposedTerm` and start election again. That's why B keeps starting election.

In this PR, we don't fix the core problem. Just modify the timeout, and fix a potential term problem.

I think we'd better implement a pre-vote, see the section 9.6 in [phd paper](https://web.stanford.edu/~ouster/cgi-bin/papers/OngaroPhD.pdf) and [this one](https://www.openlife.cc/sites/default/files/4-modifications-for-Raft-consensus.pdf).

This would involve some of changes and other effect based our implement, especially potentially failed to elect leader when one's term is higher and the other one's log is more updated. Also, in appendLog we need to consider more stuff.


src/kvstore/raftex/Host.cpp
```diff
@@ -80,7 +81,7 @@ folly::Future<cpp2::AskForVoteResponse> Host::askForVote(
             return resp;
         }
     }
-    auto client = tcManager().client(addr_, eb, false, FLAGS_raft_rpc_timeout_ms);
+    auto client = tcManager().client(addr_, eb, false, FLAGS_raft_heartbeat_interval_secs * 1000);
```
dangleptr, line 84: raft_heartbeat_interval_secs unit is seconds.

PR #2196, opened 2020-06-29
Support no username and password are required

laura-ding:
@Amber1990Zhang need to modify all docs about use console. the `user` and `password` need to delete.
if `enable_authorize` is false, can directly use console without input username and password, and the default user name is `root`


Amber1990Zhang:
@laura-ding Thanks for your kind reminder. Doc is fixed.

Amber1990Zhang:
doc done

PR #2198, opened 2020-07-01
fix value/step transform to values/steps automaticlly.

xuguruogu:
fix bugs: scanner transform value/step to values/steps automaticlly.

```
create tag t(value int)
```

```
desc tag t;

result in:
values int
```


dutor:
Please check on #2199 and #2205 

For the sake of compatibility, we decide not to resolve this issue for the time being(well, addressed in the next major release of course)

Thus close this PR.

Thanks a lot anyway.

PR #2217, opened 2020-07-08
Failed to stop storaged forever if try to stop it before kvstore inited

dangleptr:
Found it during chaos testing.


dangleptr:
@critical27 We need to fix the issue in nebula-storage 2.0

PR #2218, opened 2020-07-09
Fix enabling frame pointer

dutor:
Some of the library like folly and rocksdb was missed in #2214


dangleptr:
Do we really need to enable SFP in release mode?  It hurt the performance,  especially for some hot methods in our libraries.

dutor:
> Do we really need to enable SFP in release mode? It hurt the performance, especially for some hot methods in our libraries.

For some short but not inlined functions(which should have been), yes, it has bad influences in micro benchmark. But the impact can hardly be noticcable from the overall service perspective.

dangleptr:
> > Do we really need to enable SFP in release mode? It hurt the performance, especially for some hot methods in our libraries.
> 
> For some short but not inlined functions(which should have been), yes, it has bad influences in micro benchmark. But the impact can hardly be noticcable from the overall service perspective.
> 
> FYI. Frame pointer is enabled by default in the Linux kernel of most distribution, before the ORC unwinder is utilized.

Do you mind giving a benchmark for the simplest function,  for example `int f() { return 0;}`

dutor:
> > > Do we really need to enable SFP in release mode? It hurt the performance, especially for some hot methods in our libraries.
> > 
> > 
> > For some short but not inlined functions(which should have been), yes, it has bad influences in micro benchmark. But the impact can hardly be noticcable from the overall service perspective.
> > FYI. Frame pointer is enabled by default in the Linux kernel of most distribution, before the ORC unwinder is utilized.
> 
> Do you mind giving a benchmark for the simplest function, for example `int f() { return 0;}`

For such a function, enabling ebp will double the number  of instructions. But it's meaningless.
```
f:
        pushq   %rbp
        movq    %rsp, %rbp
        movl    $0, %eax
        popq    %rbp
        ret
```
```
f:
        movl    $0, %eax
        ret
```

dangleptr:
Yes, this is what i want to say.  To be honest,  the function is not meaningless, there are lots of them inside expression.

dutor:
> Yes, this is what i want to say. To be honest, the function is not meaningless, there are lots of them inside expression.

Here is the result of `duration_bm`.
Enabled rbp:
```
steady_clock_timer                                          49.82ns   20.07M
duration_timer                                   220.93%    22.55ns   44.35M
```
Diabled:
```
steady_clock_timer                                          49.56ns   20.18M
duration_timer                                   226.56%    21.88ns   45.71M
```

dutor:
I will resort to building my own copy. Maybe add version and tag on the third party package in future.

close it.

PR #2221, opened 2020-07-10
update  vertex and edge  benchmark

panda-sheep:


src/storage/test/UpdateVertexAndEdgeBenchmark.cpp
```diff
@@ -0,0 +1,416 @@
...
+update_edge     : edge data exist and update
+
+upsert_edge     : edge data (first not exist)  and upsert
+
+V1.0
+==============================================================================
+src/storage/test/UpdateVertexAndEdgeBenchmark.cpprelative  time/iter  iters/s
+==============================================================================
+update_vertex                                               322.47us    3.10K
+upsert_vertex                                               324.23us    3.08K
```
dangleptr, line 411: For upsert benchmark,  you'd better separate the cases "insert" and "update"

panda-sheep: Good suggestion, fixed.

PR #2222, opened 2020-07-11
 rewrite fetch prop on vertex: 1. support input/variable/multi-vertex-id/multi-tags. 2. keep order of input.

xuguruogu:
rewrite fetch prop on vertex.
1. support multi vids for all case.
2. support input/variable ref
3. support multi tags. if yield is not specified, return as many as columns as posible  as long as one of the vids has value.
4. keep order of input.
5. compatible with old tests.


src/parser/parser.yy
```diff
@@ -856,14 +858,32 @@ order_by_sentence
...
         $$ = new FetchVerticesSentence($4, $5, $6);
     }
-    | KW_FETCH KW_PROP KW_ON MUL vid {
-        $$ = new FetchVerticesSentence($5);
+    | KW_FETCH KW_PROP KW_ON MUL vid_list yield_clause {
+        auto labels = new nebula::FetchLabels();
+        labels->addLabel(new std::string("*"));
+        $$ = new FetchVerticesSentence(labels, $5, $6);
+    }
+    | KW_FETCH KW_PROP KW_ON MUL vid_ref_expression yield_clause {
```
dangleptr, line 872: If you want to support multi props,  maybe we need to support "KW_PROP" and "KW_PROPS" both.

src/parser/test/ParserTest.cpp
```diff
@@ -950,6 +950,30 @@ TEST(Parser, FetchVertex) {
...
+    }
+    {
+        GQLParser parser;
+        std::string query = "yield 1 as id | FETCH PROP ON person $-.id yield person.id";
+        auto result = parser.parse(query);
+        ASSERT_TRUE(result.ok()) << result.status();
+    }
+    {
+        GQLParser parser;
+        std::string query = "FETCH PROP ON person, another 1, 2, 3";
```
dangleptr, line 973: I want to know what's the response looks like in this case.  Person and Another maybe have different props.  You join them on vid together?

src/parser/Clauses.h
```diff
@@ -259,6 +259,29 @@ class OverEdges final {
...
+        std::transform(labels_.cbegin(), labels_.cend(),
+                       std::insert_iterator<std::vector<std::string *>>(result, result.begin()),
+                       [](auto &label) { return label.get(); });
+        return result;
+    }
+
+    std::string toString() const;
+
+    std::string* release() {
+        auto label = labels_[0].release();
```
dangleptr, line 277: Is there any possible the labels_ is empty?

src/parser/parser.yy
```diff
@@ -907,12 +927,12 @@ edge_key_ref:
...
-    : KW_FETCH KW_PROP KW_ON name_label edge_keys yield_clause {
-        auto fetch = new FetchEdgesSentence($4, $5, $6);
+    : KW_FETCH KW_PROP KW_ON fetch_labels edge_keys yield_clause {
+        auto fetch = new FetchEdgesSentence($4->release(), $5, $6);
         $$ = fetch;
     }
-    | KW_FETCH KW_PROP KW_ON name_label edge_key_ref yield_clause {
-        auto fetch = new FetchEdgesSentence($4, $5, $6);
+    | KW_FETCH KW_PROP KW_ON fetch_labels edge_key_ref yield_clause {
+        auto fetch = new FetchEdgesSentence($4->release(), $5, $6);
```
dangleptr, line 935: Why call release here?

src/graph/FetchVerticesExecutor.cpp
```diff
@@ -155,277 +350,293 @@ void FetchVerticesExecutor::fetchVertices() {
...
+        outputs->setInterim(std::move(rsWriter));
+    }
+
+    if (onResult_) {
+        onResult_(std::move(outputs));
+    } else {
+        resp_ = std::make_unique<cpp2::ExecutionResponse>();
+        auto colNames = outputs->getColNames();
+        resp_->set_column_names(std::move(colNames));
+        if (outputs->hasData()) {
```
dangleptr, line 628: If it is the rightest sentence,  you could avoid the interim result setting,  it will introduce extra encode/decode. 
You could take toThriftResponse inside GoExecutor for reference.

src/graph/FetchVerticesExecutor.cpp
```diff
@@ -155,277 +350,293 @@ void FetchVerticesExecutor::fetchVertices() {
...
+            if (dataMap.find(vid) == dataMap.end() && !expCtx_->hasInputProp()) {
+                return Status::OK();
+            }
+            auto& ds = dataMap[vid];
+
+            std::vector<VariantType> record;
+            auto schema = reader->getSchema().get();
+            Getters getters;
+            getters.getVariableProp = [&] (const std::string &prop) -> OptVariantType {
+                if (prop == "VertexID") {
```
dangleptr, line 473: If you define this behavior,  please add notes about it in summary of this pr.

src/graph/FetchVerticesExecutor.cpp
```diff
@@ -155,277 +350,293 @@ void FetchVerticesExecutor::fetchVertices() {
...
+                return Status::Error("Column `%s' not found", colname_->c_str());
             }
-        } else {
-            vids_.emplace_back(valInt);
-        }
-    }
+            if (dataMap.find(vid) == dataMap.end() && !expCtx_->hasInputProp()) {
+                return Status::OK();
+            }
+            auto& ds = dataMap[vid];
```
dangleptr, line 467: You could use the iterator returned by find directly.

src/parser/Clauses.h
```diff
@@ -259,6 +259,29 @@ class OverEdges final {
...
+        std::vector<std::string *> result;
+        std::transform(labels_.cbegin(), labels_.cend(),
+                       std::insert_iterator<std::vector<std::string *>>(result, result.begin()),
+                       [](auto &label) { return label.get(); });
+        return result;
+    }
+
+    std::string toString() const;
+
+    std::string* release() {
```
dangleptr, line 276: Weird about this method. Could you add some comments about it.

src/meta/client/MetaClient.cpp
```diff
@@ -921,6 +933,20 @@ StatusOr<std::vector<std::string>> MetaClient::getAllEdgeFromCache(const GraphSp
     return it->second;
 }
 
+StatusOr<std::vector<std::string>> MetaClient::getAllTagFromCache(const GraphSpaceID& space) {
```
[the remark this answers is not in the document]
xuguruogu: yelp

src/parser/parser.yy
xuguruogu: just like value/values...

src/parser/test/ParserTest.cpp
xuguruogu: we use the tag as Wide table, simply speaking: one tag hold one value. In this way, we can support large data adding and modify. 

Person and Another tags join together, display default value if one of them is null.

src/parser/parser.yy
xuguruogu: I will remove it.

src/graph/FetchVerticesExecutor.cpp
xuguruogu: Emm... I simply copy the codes from `FetchExecutor::finishExecution`. Maybe more attention will needed to fully understand.

src/graph/FetchVerticesExecutor.cpp
xuguruogu: yes

src/parser/Clauses.h
xuguruogu: remove it.

src/graph/FetchVerticesExecutor.cpp
xuguruogu: In the previous codes, every record will be attached collumn "VertexID" at beggining.  To be compatible with this behavior. I will comment on this.

xuguruogu: modify the codes to keep identical with previous codes.

src/parser/parser.yy
```diff
@@ -907,11 +927,11 @@ edge_key_ref:
     ;
 
 fetch_edges_sentence
-    : KW_FETCH KW_PROP KW_ON name_label edge_keys yield_clause {
+    : KW_FETCH KW_PROP KW_ON fetch_labels edge_keys yield_clause {
         auto fetch = new FetchEdgesSentence($4, $5, $6);
         $$ = fetch;
     }
-    | KW_FETCH KW_PROP KW_ON name_label edge_key_ref yield_clause {
+    | KW_FETCH KW_PROP KW_ON fetch_labels edge_key_ref yield_clause {
```
CPWstatic, line 934: Emm, if don't implement it in this pr, please forbid this syntax at present.

xuguruogu: As the syntax has trouble to shift and reduce before edge_key and vids identifiers, I return the error msg in the executor.

src/parser/parser.yy
```diff
@@ -856,14 +858,32 @@ order_by_sentence
...
+    : KW_FETCH KW_PROP KW_ON fetch_labels vid_list yield_clause {
         $$ = new FetchVerticesSentence($4, $5, $6);
     }
-    | KW_FETCH KW_PROP KW_ON name_label vid_ref_expression yield_clause {
+    | KW_FETCH KW_PROP KW_ON fetch_labels vid_ref_expression yield_clause {
         $$ = new FetchVerticesSentence($4, $5, $6);
     }
-    | KW_FETCH KW_PROP KW_ON MUL vid {
-        $$ = new FetchVerticesSentence($5);
+    | KW_FETCH KW_PROP KW_ON MUL vid_list yield_clause {
```
Shylock-Hg, line 867: This seem be conflict, if you fetch prop on * but specify the tag.prop when yield.

src/graph/FetchVerticesExecutor.cpp
```diff
@@ -93,35 +48,272 @@ Status FetchVerticesExecutor::prepareVids() {
...
+        return Status::Error("tags shall never be null");
+    }
+
+    auto tagNames = tags->labels();
+    if (tagNames.empty()) {
+        LOG(ERROR) << "tags shall never be empty";
+        return Status::Error("tags shall never be empty");
+    }
+
+    if (tagNames.size() == 1 && *tagNames[0] == "*") {
```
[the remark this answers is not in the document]
xuguruogu: make sense.

src/graph/test/FetchVerticesTest.cpp
```diff
@@ -83,6 +83,20 @@ TEST_F(FetchVerticesTest, Base) {
         };
         ASSERT_TRUE(verifyResult(resp, expected));
     }
+    {
+        cpp2::ExecutionResponse resp;
+        auto &player = players_["Boris Diaw"];
+        auto *fmt = "GO FROM %ld over like YIELD like._dst as id"
```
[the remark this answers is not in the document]
xuguruogu: Heavily used. Mostly, we get a sub-graph, using filters or other restriction, and then fetch more information of the nodes.

src/parser/parser.yy
xuguruogu: we have many tags in use. it is redundant to write tag-name twice. In actually implement, if yield_clause specified, tag-names may only be used as verification.

src/graph/FetchVerticesExecutor.cpp
```diff
@@ -93,35 +48,272 @@ Status FetchVerticesExecutor::prepareVids() {
...
+    } else {
+        for (auto *col : yieldClause_->columns()) {
+            if (!col->getFunName().empty()) {
+                return Status::SyntaxError("Do not support aggregated query with fetch prop on.");
+            }
+
+            if (col->expr()->isInputExpression()) {
+                auto *inputExpr = dynamic_cast<InputPropertyExpression*>(col->expr());
+                auto *colName = inputExpr->prop();
+                if (*colName == "*") {
```
dangleptr, line 179: Excellent.  Could we combine the logic about "*" both in InputPropertyExpression and VariablePropertyExpression into one?

src/graph/FetchVerticesExecutor.cpp
```diff
@@ -93,35 +48,272 @@ Status FetchVerticesExecutor::prepareVids() {
...
+            }
+            if (tagSchema->getFieldIndex(prop) == -1) {
+                return Status::Error(
+                    "`%s' is not a prop of `%s'", tagName.c_str(), prop.c_str());
+            }
+            storage::cpp2::PropDef pd;
+            pd.owner = storage::cpp2::PropOwner::SOURCE;
+            pd.name = prop;
+            pd.id.set_tag_id(tagId);
+            props_.emplace_back(std::move(pd));
```
dangleptr, line 260: Where did your support allProps on one Tag.   Such as tag1.*

src/graph/FetchVerticesExecutor.cpp
xuguruogu: may have trouble due to 

```
Expression *expr = new InputPropertyExpression(new std::string(prop));
Expression *expr = new VariablePropertyExpression(alias, new std::string(prop));
```

src/graph/FetchVerticesExecutor.cpp
xuguruogu: tag1.* can not pass syntax check.

dangleptr: Is is the compatible with the original behavior?

xuguruogu: the behavior is same.

Amber1990Zhang:
doc done

PR #2224, opened 2020-07-13
fix unstable raft ut

critical27:
See [failed log](https://github.com/vesoft-inc/nebula/pull/2223/checks?check_run_id=863918745).

In some of the ut, RaftPart would be stop and restart. When `restart`, it would add a background task `statusPolling`. And this would make part would start election at wrong time.

In the previous failed case, a part send election twice in different thread within a very short period of time, with term 5 and 6, which makes a log of term 5 failed to replicate to followers. However, this should not happen in real instance. 

BTW, fix StoragePerf tool always write the same edge.

### DIFF
--- a/src/kvstore/raftex/RaftPart.cpp
+++ b/src/kvstore/raftex/RaftPart.cpp
@@ -316,8 +316,8 @@ void RaftPart::start(std::vector<HostAddr>&& peers, bool asLearner) {
     startTimeMs_ = time::WallClock::fastNowInMilliSec();
     // Set up a leader election task
     size_t delayMS = 100 + folly::Random::rand32(900);
-    bgWorkers_->addDelayTask(delayMS, [self = shared_from_this()] {
-        self->statusPolling();
+    bgWorkers_->addDelayTask(delayMS, [self = shared_from_this(), startTime = startTimeMs_] {
+        self->statusPolling(startTime);
     });
 }
 
@@ -1208,7 +1208,7 @@ bool RaftPart::leaderElection() {
 }
 
 
-void RaftPart::statusPolling() {
+void RaftPart::statusPolling(int64_t startTime) {
     size_t delay = FLAGS_raft_heartbeat_interval_secs * 1000 / 3;
     if (needToStartElection()) {
         if (leaderElection()) {
@@ -1229,12 +1229,18 @@ void RaftPart::statusPolling() {
     }
     {
         std::lock_guard<std::mutex> g(raftLock_);
+        // If startTime is not same as the time when `statusPolling` is add to event loop,
+        // it means the part has been restarted (it only happens in ut for now), so don't
+        // add another `statusPolling`.
+        if (startTime != startTimeMs_) {
+            return;
+        }
         if (status_ == Status::RUNNING || status_ == Status::WAITING_SNAPSHOT) {
             VLOG(3) << idStr_ << "Schedule new task";
             bgWorkers_->addDelayTask(
                 delay,
-                [self = shared_from_this()] {
-                    self->statusPolling();
+                [self = shared_from_this(), startTime] {
+                    self->statusPolling(startTime);
                 });
         }
     }
--- a/src/kvstore/raftex/RaftPart.cpp
+++ b/src/kvstore/raftex/RaftPart.cpp
@@ -1209,6 +1209,15 @@ bool RaftPart::leaderElection() {
 
 
 void RaftPart::statusPolling(int64_t startTime) {
+    {
+        std::lock_guard<std::mutex> g(raftLock_);
+        // If startTime is not same as the time when `statusPolling` is add to event loop,
+        // it means the part has been restarted (it only happens in ut for now), so don't
+        // add another `statusPolling`.
+        if (startTime != startTimeMs_) {
+            return;
+        }
+    }
     size_t delay = FLAGS_raft_heartbeat_interval_secs * 1000 / 3;
     if (needToStartElection()) {
         if (leaderElection()) {
@@ -1229,12 +1238,6 @@ void RaftPart::statusPolling(int64_t startTime) {
     }
     {
         std::lock_guard<std::mutex> g(raftLock_);
-        // If startTime is not same as the time when `statusPolling` is add to event loop,
-        // it means the part has been restarted (it only happens in ut for now), so don't
-        // add another `statusPolling`.
-        if (startTime != startTimeMs_) {
-            return;
-        }
         if (status_ == Status::RUNNING || status_ == Status::WAITING_SNAPSHOT) {
             VLOG(3) << idStr_ << "Schedule new task";
             bgWorkers_->addDelayTask(
--- a/src/kvstore/raftex/RaftPart.h
+++ b/src/kvstore/raftex/RaftPart.h
@@ -328,7 +328,7 @@ private:
 
     bool needToStartElection();
 
-    void statusPolling();
+    void statusPolling(int64_t startTime);
 
     bool needToCleanupSnapshot();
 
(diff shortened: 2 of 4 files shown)


dangleptr:
> In the previous failed case, a part send election twice in different thread within a very short period of time, with term 5 and 6, which makes a log of term 5 failed to replicate to followers. However, this should not happen in real instance.

You mean the electionRequest is sent out after the part has been stopped?

critical27:
> > In the previous failed case, a part send election twice in different thread within a very short period of time, with term 5 and 6, which makes a log of term 5 failed to replicate to followers. However, this should not happen in real instance.
> 
> You mean the electionRequest is sent out after the part has been stopped?

Actually, multiple thread is trying to start election after it restarted in ut, because we add a `statusPolling` task whenever we call `start`.

dangleptr:
> > > In the previous failed case, a part send election twice in different thread within a very short period of time, with term 5 and 6, which makes a log of term 5 failed to replicate to followers. However, this should not happen in real instance.
> > 
> > 
> > You mean the electionRequest is sent out after the part has been stopped?
> 
> Actually, multiple thread is trying to start election after it restarted in ut, because we add a `statusPolling` task whenever we call `start`.

But after the part stopped,  the statusPolling task will not send vote request anymore.  You have not called `stop` yet?

critical27:
> > > > In the previous failed case, a part send election twice in different thread within a very short period of time, with term 5 and 6, which makes a log of term 5 failed to replicate to followers. However, this should not happen in real instance.
> > > 
> > > 
> > > You mean the electionRequest is sent out after the part has been stopped?
> > 
> > 
> > Actually, multiple thread is trying to start election after it restarted in ut, because we add a `statusPolling` task whenever we call `start`.
> 
> But after the part stopped, the statusPolling task will not send vote request anymore. You have not called `stop` yet?

Yeah, we don't stop `statusPolling` when we stop the `RaftPart` for now.

critical27:
image

dangleptr:
> > > > > In the previous failed case, a part send election twice in different thread within a very short period of time, with term 5 and 6, which makes a log of term 5 failed to replicate to followers. However, this should not happen in real instance.
> > > > 
> > > > 
> > > > You mean the electionRequest is sent out after the part has been stopped?
> > > 
> > > 
> > > Actually, multiple thread is trying to start election after it restarted in ut, because we add a `statusPolling` task whenever we call `start`.
> > 
> > 
> > But after the part stopped, the statusPolling task will not send vote request anymore. You have not called `stop` yet?
> 
> Yeah, we didn't stop `statusPolling` when we stop the `RaftPart` for now. One reason is that all instance share the same `bgThread_`.

Not exactly,  i think i got the root cause.  It is one ABA problem,  it means after stopping the Part,  the status has been reset to STOPPED.  Now there is still one left task inside your eventLoop,   when the tasking invoking,  you restart it and the status has been reset to RUNNING.

critical27:
> Not exactly, i think i got the root cause. It is one ABA problem, it means after stopping the Part, the status has been reset to STOPPED. Now there is still one left task inside your eventLoop, when the task is invoked, you restart the part and the status has been reset to RUNNING.



> > > > > > In the previous failed case, a part send election twice in different thread within a very short period of time, with term 5 and 6, which makes a log of term 5 failed to replicate to followers. However, this should not happen in real instance.
> > > > > 
> > > > > 
> > > > > You mean the electionRequest is sent out after the part has been stopped?
> > > > 
> > > > 
> > > > Actually, multiple thread is trying to start election after it restarted in ut, because we add a `statusPolling` task whenever we call `start`.
> > > 
> > > 
> > > But after the part stopped, the statusPolling task will not send vote request anymore. You have not called `stop` yet?
> > 
> > 
> > Yeah, we didn't stop `statusPolling` when we stop the `RaftPart` for now. One reason is that all instance share the same `bgThread_`.
> 
> Not exactly, i think i got the root cause. It is one ABA problem, it means after stopping the Part, the status has been reset to STOPPED. Now there is still one left task inside your eventLoop, when the task is invoked, you restart the part and the status has been reset to RUNNING.

Yeah exactly. The key point is we need to prevent two thread doing `statusPolling`, otherwise the problem still exists.

dangleptr:
> > Not exactly, i think i got the root cause. It is one ABA problem, it means after stopping the Part, the status has been reset to STOPPED. Now there is still one left task inside your eventLoop, when the task is invoked, you restart the part and the status has been reset to RUNNING.
> 
> > > > > > > In the previous failed case, a part send election twice in different thread within a very short period of time, with term 5 and 6, which makes a log of term 5 failed to replicate to followers. However, this should not happen in real instance.
> > > > > > 
> > > > > > 
> > > > > > You mean the electionRequest is sent out after the part has been stopped?
> > > > > 
> > > > > 
> > > > > Actually, multiple thread is trying to start election after it restarted in ut, because we add a `statusPolling` task whenever we call `start`.
> > > > 
> > > > 
> > > > But after the part stopped, the statusPolling task will not send vote request anymore. You have not called `stop` yet?
> > > 
> > > 
> > > Yeah, we didn't stop `statusPolling` when we stop the `RaftPart` for now. One reason is that all instance share the same `bgThread_`.
> > 
> > 
> > Not exactly, i think i got the root cause. It is one ABA problem, it means after stopping the Part, the status has been reset to STOPPED. Now there is still one left task inside your eventLoop, when the task is invoked, you restart the part and the status has been reset to RUNNING.
> 
> Yeah exactly. The key point is we need to prevent two thread doing `statusPolling`, otherwise the problem still exists. Do you have better solution?

I suggest passing current part version into statusPolling,   and check it at first inside the method.
'startTimeMs_' could be reused for this purpose.

critical27:
> startTimeMs_



> > > Not exactly, i think i got the root cause. It is one ABA problem, it means after stopping the Part, the status has been reset to STOPPED. Now there is still one left task inside your eventLoop, when the task is invoked, you restart the part and the status has been reset to RUNNING.
> > 
> > 
> > > > > > > > In the previous failed case, a part send election twice in different thread within a very short period of time, with term 5 and 6, which makes a log of term 5 failed to replicate to followers. However, this should not happen in real instance.
> > > > > > > 
> > > > > > > 
> > > > > > > You mean the electionRequest is sent out after the part has been stopped?
> > > > > > 
> > > > > > 
> > > > > > Actually, multiple thread is trying to start election after it restarted in ut, because we add a `statusPolling` task whenever we call `start`.
> > > > > 
> > > > > 
> > > > > But after the part stopped, the statusPolling task will not send vote request anymore. You have not called `stop` yet?
> > > > 
> > > > 
> > > > Yeah, we didn't stop `statusPolling` when we stop the `RaftPart` for now. One reason is that all instance share the same `bgThread_`.
> > > 
> > > 
> > > Not exactly, i think i got the root cause. It is one ABA problem, it means after stopping the Part, the status has been reset to STOPPED. Now there is still one left task inside your eventLoop, when the task is invoked, you restart the part and the status has been reset to RUNNING.
> > 
> > 
> > Yeah exactly. The key point is we need to prevent two thread doing `statusPolling`, otherwise the problem still exists. Do you have better solution?
> 
> I suggest passing current part version into statusPolling, and check it at first inside the method.
> 'startTimeMs_' could be reused for this purpose.

Good suggestion, fixed.

src/kvstore/raftex/RaftPart.cpp
```diff
@@ -1229,12 +1229,18 @@ void RaftPart::statusPolling() {
     }
     {
         std::lock_guard<std::mutex> g(raftLock_);
+        // If startTime is not same as the time when `statusPolling` is add to event loop,
+        // it means the part has been restarted (it only happens in ut for now), so don't
+        // add another `statusPolling`.
+        if (startTime != startTimeMs_) {
```
dangleptr, line 1235: Maybe we need to check it at first in the method.

dangleptr:
Please double check the failed UTs

PR #2225, opened 2020-07-13
go sentence support integer/timestamp type to start traverse.

xuguruogu:
support go from integer or timestamp. Just like sentence below:

```
...| go from hash($-.id) over edge

```


src/graph/InterimResult.cpp
```diff
@@ -191,11 +191,13 @@ InterimResult::buildIndex(const std::string &vidColumn) const {
         auto name = schema->getFieldName(i);
         if (vidColumn == name) {
             VLOG(1) << "col name: " << vidColumn << ", col index: " << i;
-            if (schema->getFieldType(i).type != SupportedType::VID) {
+            if (schema->getFieldType(i).type != SupportedType::INT &&
```
dangleptr, line 194: I want to know in which situation you need to traverse from "Timestamp" data.....

xuguruogu: For example: find out active users in the last hour/minute.

dangleptr: > For example: find out active users in the last hour/minute.

Sorry, i did not understand it.  The time point is one vertex in your case?

xuguruogu: yes, connect action with a timestamp(start of the minute/hour). We can find out the active actions, without index.

xuguruogu: To my knowledge, timestamp has no difference with an integer type. I use timestamp type to be displayed more human-readable

dangleptr: Interesting.  
The pr has no problems in logic.  But we need to discuss it futher.

dangleptr: We both think the improvement is reasonable. Let's go on.  
Thanks for your contribution.

dangleptr:
Please check the failed UTs

Amber1990Zhang:
doc done

PR #2226, opened 2020-07-14
Support StatsManager::readAllValue include p99

laura-ding:
`curl -G "http://127.0.0.1:13070/get_stats"` include p99

### DIFF
--- a/src/common/stats/StatsManager.cpp
+++ b/src/common/stats/StatsManager.cpp
@@ -249,6 +249,36 @@ void StatsManager::readAllValue(folly::dynamic& vals) {
                 vals.push_back(std::move(stat));
             }
         }
+
+        // add Percentile
+        for (auto range = TimeRange::FIVE_SECONDS; range <= TimeRange::ONE_HOUR;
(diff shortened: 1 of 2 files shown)


src/common/stats/StatsManager.cpp
```diff
@@ -249,6 +249,36 @@ void StatsManager::readAllValue(folly::dynamic& vals) {
                 vals.push_back(std::move(stat));
             }
         }
+
+        // add Percentile
+        for (auto range = TimeRange::FIVE_SECONDS; range <= TimeRange::ONE_HOUR;
```
dangleptr, line 254: I want to know why not add one StatsMethod kind such as P99 directly?

laura-ding: `p99` does not belong to the `StatsMethod`, it belong to HistogramMethod.

dangleptr: make sense. Maybe we need HistogramMethod kind such as p50, p90, p95, p99

laura-ding: Good point, 2.0 can add it.

PR #2228, opened 2020-07-14
aggregate function support collect_list/collect_set. add function concat_ws

xuguruogu:
Collect integer list/set is a strong demand. As not support list/set type now, a way to support the demand is needed.

```
... | group by ... yield collect_list($-.id) as ids | yield concat_ws(",", $-.ids)
```

### DIFF
--- a/src/common/filter/FunctionManager.cpp
+++ b/src/common/filter/FunctionManager.cpp
@@ -729,6 +729,26 @@ FunctionManager::FunctionManager() {
             }
         };
     }
+    {
+        auto &attr = functions_["range"];
+        attr.minArity_ = 2;
+        attr.maxArity_ = 3;
+        attr.cacheable_ = true;
+        attr.body_ = [] (const auto &args) -> OptVariantType {
+            auto start = Expression::asInt(args[0]);
+            auto end = Expression::asInt(args[1]);
+            int64_t step = 1;
+            if (args.size() > 2) {
+                step = Expression::asInt(args[2]);
+            }
+
+            FlattenListWriter w;
+            for (int64_t i = start; i <= end; i += step) {
+                w << i;
+            }
+            return OptVariantType(w.finish());
+        };
+    }
 }  // NOLINT
 
 
--- a/src/common/filter/FunctionManager.cpp
+++ b/src/common/filter/FunctionManager.cpp
@@ -9,6 +9,7 @@
 #include "filter/Expressions.h"
 #include "time/WallClock.h"
 #include "filter/geo/GeoFilter.h"
+#include "base/FlattenList.h"
 
 namespace nebula {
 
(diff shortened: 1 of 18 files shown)


src/graph/AggregateFunction.h
```diff
@@ -421,6 +423,186 @@ class BitXor final : public AggFun {
...
+//                }
+//                break;
+            default:
+                return;
+        }
+    }
+
+    cpp2::ColumnValue getResult() override {
+        cpp2::ColumnValue result;
+        result.set_str(cord_.str());
```
CPWstatic, line 499: Why not format the values into string with comma directly?

xuguruogu: The keywords are provided by users, which are identical to hive sql.

src/common/filter/FunctionManager.cpp
```diff
@@ -553,6 +553,75 @@ FunctionManager::FunctionManager() {
...
+            auto sep   = Expression::asString(args[0]);
+            auto set   = Expression::asString(args[1]);
+
+            const char *p = set.data();
+            size_t size = set.size();
+            size_t offset = 0;
+            std::string s;
+
+            if (size != 0) {
+                uint8_t type = p[0];
```
laura-ding, line 570: You need to check the `set` value, it may be an empty string or it not the type of data you expect.

src/common/filter/FunctionManager.cpp
```diff
@@ -553,6 +553,75 @@ FunctionManager::FunctionManager() {
...
+                            vec.emplace_back(v);
+                            offset += sizeof(double);
+                        }
+                        s = folly::join(sep, vec);
+                    }
+                        break;
+                    case VAR_STR:
+                    {
+                        std::vector<std::string> vec;
+                        // TODO suport string.
```
laura-ding, line 614: support

src/common/filter/FunctionManager.cpp
```diff
@@ -553,6 +553,75 @@ FunctionManager::FunctionManager() {
...
+                        return folly::join(sep, vec);
```
laura-ding, line 615: if not support, return Status::Error("……");

src/graph/AggregateFunction.h
```diff
@@ -421,6 +423,186 @@ class BitXor final : public AggFun {
...
+        return result;
+    }
+private:
+    Cord            cord_;
+    bool            has_{false};
+    uint8_t         type_ = 0;
```
laura-ding, line 505: ColumnType  type_ = ColumnType::__EMPTY__;

src/graph/AggregateFunction.h
```diff
@@ -421,6 +423,186 @@ class BitXor final : public AggFun {
...
+        return result;
+    }
+private:
+    std::set<int64_t>               int_values_;
+    std::set<bool>                  bool_values_;
+    std::set<double>                double_values_;
+//    std::set<std::string>           str_values_;
+    Cord                            cord_;
+    bool                            has_{false};
+    uint8_t                         type_ = 0;
```
laura-ding, line 603: ditto

xuguruogu:
It seems less important, I will improve it later.

CPWstatic:
Thanks, this feature had been supported in v2.0.

PR #2232, opened 2020-07-15
Enhance the failover logic for balance procedure. 

dangleptr:
1.   For leaders and followers, check the state for each part before balance. It gives a chance to recover from the last failed state.
2.  For learners,  check the peers and their state when opening. 

During chaos testing,  we found some problems when the opening part is too slow.  It always causes the members in raft group mix up its peers,  and leader election always failed.  

To recover from it,  we have this PR.


dangleptr:
Pleas hold on for a while.

PR #2233, opened 2020-07-15
new syntax: scan vertex

xuguruogu:
```
scan vertex <tag> part <partid> [from <vid>] [latest_seconds <seconds>] [limit <num>]
```


src/graph/ScanExecutor.cpp
```diff
@@ -0,0 +1,349 @@
...
+        auto expr = sentence_->limit();
+        if (expr) {
+            int64_t limit = 0;
+            auto status = _getInt(expr, limit);
+            if (!status.ok()) {
+                return status;
+            }
+            limit_ = limit;
+        } else {
+            limit_ = std::numeric_limits<int32_t>::max();
```
dangleptr, line 102: The default limit need to be reconsidered. 
It is dangerous with no limits.  Because as your logic,  all data will been scanned out,  it will block the normal OLTP requests.  Meanwhile,  on graphd-side,  it usually has not enough memory for the response.

critical27:
Where do you use `scan` for now?

src/graph/ScanExecutor.cpp
xuguruogu: make sense.

xuguruogu:
> Where do you use `scan` for now?

mostly for debug. It is may cause performance issues for OLTP requests.

dangleptr:
After discussed inside,  we both think the feature to scan all vertices is very practical,  but it is not a good idea to support it in our query language,  because our focus is on OLTP area. 

To support the feature,  we have another debug tool "db_dump" in our repo.  But currently, it is not friendly to our users due to it is only used on one storage server.

I suggest we could enhance the tool to do distribute work inside it.  

Thanks for your contribution.  @xuguruogu

dangleptr:
Close it.  Thanks for your contribution.  @xuguruogu

PR #2234, opened 2020-07-17
Set metad and storaged's reuse_port default value is false NG-199

laura-ding:
Someone may run two storage daemons use the same conf in different dir on the same host. so there will have two storages that have the same port and ip. it will cause some problems, so set the `reuse_port` to false.

#NG-199


laura-ding:
@dangleptr OK, I will let Amber add Q&A about this.

PR #2236, opened 2020-07-17
set rocksdb_disable_wal to false by default

critical27:
As title.

### DIFF
--- a/etc/nebula-metad.conf.production
+++ b/etc/nebula-metad.conf.production
@@ -35,5 +35,4 @@
 --data_path=data/meta
 
 ############## rocksdb Options ##############
---rocksdb_disable_wal=false
 --rocksdb_wal_sync=true
--- a/etc/nebula-storaged.conf.default
+++ b/etc/nebula-storaged.conf.default
@@ -65,11 +65,7 @@
 # In order to disable compression for level 0/1, set it to "no:no"
 --rocksdb_compression_per_level=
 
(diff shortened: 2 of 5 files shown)


darionyaphet:
Keeping both the RocksDB and the Raft's WAL

whitewum:
Does this flag make write much more slower? 
Do you have or need any test?

dangleptr:
> Does this flag make write much more slower?
> Do you have or need any test?

Yes, of course, we need to test it.  According to my experience,  it has just about 10~20% improvement for high load write test.  For stable,  we close it for temporarily by default. 

Thanks.

xuguruogu:
> > Does this flag make write much more slower?
> > Do you have or need any test?
> 
> Yes, of course, we need to test it. According to my experience, it has just about 10~20% improvement for high load write test. For stable, we close it for temporarily by default.
> 
> Thanks.

I also meet so much unstable issues. Hope this can help me.

dangleptr:
> > > Does this flag make write much more slower?
> > > Do you have or need any test?
> > 
> > 
> > Yes, of course, we need to test it. According to my experience, it has just about 10~20% improvement for high load write test. For stable, we close it for temporarily by default.
> > Thanks.
> 
> I also meet so much unstable issues. Hope this can help me.

Yes, you could try it.  In some cases,  we found the memory data lost will cause snapshot  pulling when restart the cluster. 
It will cause lots of unstable problems.

PR #2239, opened 2020-07-21
rename max_parts_num and set the default value of max_parts_num to 16…

knightXun:
I rename default_max_parts_num to max_parts_num, and set this default value big enough


Shylock-Hg:
The name in cloud branch should keep same too.

knightXun:
> The name in cloud branch should keep same too.

this is to cloud branch

laura-ding:
@knightXun you need to add `--max_parts_num=100` when starting the nebula servers.

Shylock-Hg:
I think you could submit the feature to master.

PR #2240, opened 2020-07-21
fix meta can't be stopped

critical27:
3-replica meta, stop two followers at first, then leader at last. If leader receive a heartbeat from storage, it will retry forever to updateHostInfo.
The root reason is we didn't stop and release kvstore when we stop daemon.


src/daemons/MetaDaemon.cpp
```diff
@@ -297,13 +298,17 @@ void signalHandler(int sig) {
...
             {
                 auto gJobMgr = nebula::meta::JobManager::getInstance();
                 gJobMgr->shutDown();
             }
+            if (gKVStore) {
+                gKVStore->stop();
+                gKVStore.reset();
+            }
+            if (gServer) {
+                gServer->stop();
```
dangleptr, line 310: Why stop kvstore firstly.

PR #2241, opened 2020-07-22
Modify the handling of the restart service in scripts

laura-ding:


scripts/utils.sh
```diff
@@ -83,15 +83,19 @@ function is_port_listened_on {
...
-    while [[ ${seconds} > 0 ]]; do
+    sleep_count=1
+    while true; do
         sleep 0.1
         is_process_running ${pid_file} || return 0
-        seconds=$(echo "${seconds} - 0.1" | bc -l)
+        if [[ $(( $sleep_count%10 )) == 0 ]]; then
+            WARN "Cost $(( $sleep_count/10 )) seconds to stop ${daemon_name}"
+        fi
+        sleep_count=$[$sleep_count+1];
```
dutor, line 98: Don't bother to print out how long it takes to stop. Just keep warning.
Besides, warning every second is too frequent, because mostly the stopping action finishes in several seconds. We don't want to be too verbose about the stopping process. Make it 5 seconds.

CPWstatic:
Not a good practice by doing a dead loop when stopping, please reconsider it in v2.0. I'll close it first.

PR #2243, opened 2020-07-23
Support to enable rocksdbs statistics

chenxu14:
I think there is a need to enable rocksdb's statistics, so we can dump some metrics to the log, and use it to find bottleneck.
We can enable this by specify --rocksdb_statistics_enable=true in nebula-storaged.conf, and declare the stats level through --rocksdb_stats_level(default value is kExceptHistogramOrTimers)

### DIFF
--- a/src/kvstore/RocksEngineConfig.cpp
+++ b/src/kvstore/RocksEngineConfig.cpp
@@ -60,6 +60,9 @@ DEFINE_string(rocksdb_compression_per_level, "", "Specify per level compression 
                                                  "e.g. \"no:no:lz4:lz4::zstd\" === "
                                                  "\"no:no:lz4:lz4:lz4:snappy:zstd:snappy\"");
 
+DEFINE_bool(enable_rocksdb_statistics, false, "Whether or not to enable rocksdb's statistics");
+DEFINE_string(rocksdb_stats_level, "kExceptHistogramOrTimers", "rocksdb statistics level");
+
 namespace nebula {
 namespace kvstore {
 
@@ -123,6 +126,10 @@ rocksdb::Status initRocksdbOptions(rocksdb::Options &baseOpts) {
     if (!s.ok()) {
         return s;
     }
+    std::shared_ptr<rocksdb::Statistics> stats = getDBStatistics();
+    if (stats) {
+        dbOpts.statistics = std::move(stats);
+    }
     dbOpts.listeners.emplace_back(new EventListener());
 
     std::unordered_map<std::string, std::string> cfOptsMap;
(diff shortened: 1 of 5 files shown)


Shylock-Hg:
If so I think you could export the option `stats_dump_period_sec` too.

src/kvstore/RocksEngineConfig.cpp
```diff
@@ -186,5 +194,27 @@ bool loadOptionsMap(std::unordered_map<std::string, std::string> &map, const std
...
+    } else if (str == "kExceptTimeForMutex") {
+        return rocksdb::StatsLevel::kExceptTimeForMutex;
+    } else {
+        return rocksdb::StatsLevel::kAll;
+    }
+}
+
+std::shared_ptr<rocksdb::Statistics> getDBStatistics() {
+    if (FLAGS_rocksdb_statistics_enable) {
+        static std::shared_ptr<rocksdb::Statistics> dbstats = rocksdb::CreateDBStatistics();
```
Shylock-Hg, line 213: Let rocksdb manage it's lifetime.

chenxu14:
> If so I think you could export the option `stats_dump_period_sec` too.

I think we can specify this in --rocksdb_db_options

src/kvstore/RocksEngineConfig.cpp
chenxu14: We should pass the same Statistics object multiple DBs, in oder to aggregated values for all those DBs. so it's better to manage it ourselfs?

Shylock-Hg: If you share this cross db instances, the statistics may be mixed. The log of one db may contains other db's metrics data.

chenxu14: Exactly true, IMHO dump to log is not the perfect way, we want report those metrics to the monitor system directly (such as Falcon), If do so, getDBStatistics is needed.
Can we disable rocksdb's dump feature and manage it ourself, I can do this in another PR.

Shylock-Hg: Even if so I suggest you manage the one DBStatistics object to one DB instance.

chenxu14: You mean we'd better monitor each DB separately?

Shylock-Hg: Yes, I thinks so.

chenxu14: I think there is no need, since they belong to the same space,  just different partitions only.

Shylock-Hg: No, there will be multiple DB instance in the storage process.

critical27: Em, since we shared block_cache among all instance, so maybe we could collect stats of all instance.

Shylock-Hg: Ok, agree now. We could collect the all storage Statistics in one currently.

src/kvstore/RocksEngineConfig.cpp
```diff
@@ -186,5 +194,27 @@ bool loadOptionsMap(std::unordered_map<std::string, std::string> &map, const std
...
+        return dbstats;
+    }
+    return nullptr;
+}
+
```
dutor, line 218: Suggest to put the creation and initialization of `rocksdb::Statistics`  into a single function.

src/kvstore/RocksEngineConfig.cpp
```diff
@@ -60,6 +60,9 @@ DEFINE_string(rocksdb_compression_per_level, "", "Specify per level compression
                                                  "e.g. \"no:no:lz4:lz4::zstd\" === "
                                                  "\"no:no:lz4:lz4:lz4:snappy:zstd:snappy\"");
 
+DEFINE_bool(rocksdb_statistics_enable, false, "Whether to export rocksdb's statistics");
```
dutor, line 63: It would be better to name this flag to conform to others, like `--enable_partitioned_index_filter`

chenxu14:
Thanks for your advise, can you help to review again.@dutor

dangleptr:
Please check your code style with cpplint.
Thanks @chenxu14

chenxu14:
> Please check your code style with cpplint.
> Thanks @chenxu14

OK, fixed

PR #2245, opened 2020-07-24
fix double quote in index

knightjya:
What changes were proposed in this pull request?
fix SparkClientGenerator. The function extraValue add duplicate double quote to index.

Why are the changes needed?
When using extraValue to transfrom index, double quote  was added. The change fixes the logic to delete duplicate double quote.
before: duplicate double quote
after: delete duplicate double quote

Does this PR introduce any user-facing change?
No.

How was this patch tested?

### DIFF
--- a/src/tools/spark-sstfile-generator/src/main/scala/com/vesoft/nebula/tools/generator/v2/SparkClientGenerator.scala
+++ b/src/tools/spark-sstfile-generator/src/main/scala/com/vesoft/nebula/tools/generator/v2/SparkClientGenerator.scala
@@ -297,7 +297,6 @@ object SparkClientGenerator {
           fields.asScala.keys.toList
         }
 
-        val vertexIndex      = sourceProperties.indexOf(vertex)
         val nebulaProperties = properties.mkString(",")
         val data             = createDataSource(spark, pathOpt, tagConfig)
 
@@ -310,7 +309,11 @@ object SparkClientGenerator {
               val values = (for {
                 property <- valueProperties if property.trim.length != 0
               } yield extraValue(row, property)).mkString(",")
-              (String.valueOf(extraValue(row, vertex)), values)
+              (if (String.valueOf(extraValue(row, vertex)).startsWith("\"") && String.valueOf(extraValue(row, vertex)).endsWith("\"")) {
+                String.valueOf(extraValue(row, vertex)).slice(1,-1)
+              } else {
+                String.valueOf(extraValue(row, vertex))
+              }, values)
             }(Encoders.tuple(Encoders.STRING, Encoders.STRING))
             .foreachPartition { iterator: Iterator[(String, String)] =>
               val service      = MoreExecutors.listeningDecorator(Executors.newFixedThreadPool(1))


src/tools/spark-sstfile-generator/src/main/scala/com/vesoft/nebula/tools/generator/v2/SparkClientGenerator.scala
```diff
@@ -310,7 +309,11 @@ object SparkClientGenerator {
               val values = (for {
                 property <- valueProperties if property.trim.length != 0
               } yield extraValue(row, property)).mkString(",")
-              (String.valueOf(extraValue(row, vertex)), values)
+              (if (String.valueOf(extraValue(row, vertex)).startsWith("\"") && String.valueOf(extraValue(row, vertex)).endsWith("\"")) {
```
darionyaphet, line 312: if the string start with quotation mark or end with quotation mark how to deal with it ?

src/tools/spark-sstfile-generator/src/main/scala/com/vesoft/nebula/tools/generator/v2/SparkClientGenerator.scala
```diff
@@ -310,7 +309,11 @@ object SparkClientGenerator {
...
+                String.valueOf(extraValue(row, vertex)).slice(1,-1)
```
darionyaphet, line 313: seems `extraValue` will append `quotation marks` at begin and end of the string

darionyaphet: By the way, you should extraValue only once

src/tools/spark-sstfile-generator/src/main/scala/com/vesoft/nebula/tools/generator/v2/SparkClientGenerator.scala
```diff
@@ -310,7 +309,11 @@ object SparkClientGenerator {
...
+              } else {
```
darionyaphet, line 314: By the way, you should extraValue only once

src/tools/spark-sstfile-generator/src/main/scala/com/vesoft/nebula/tools/generator/v2/SparkClientGenerator.scala
knightjya: U're right. Better way to fix the problem is use another function instead of extraValue to especially deal with index rather than
 using extraValue.

src/tools/spark-sstfile-generator/src/main/scala/com/vesoft/nebula/tools/generator/v2/SparkClientGenerator.scala
knightjya: New pr refers to [https://github.com/vesoft-inc/nebula/pull/2250](#2250)

PR #2247, opened 2020-07-24
Support Getting System info via http 

knightXun:


src/storage/StorageServer.cpp
```diff
@@ -87,6 +88,11 @@ bool StorageServer::initWebService() {
     router.get("/admin").handler([this](web::PathParams&&) {
         return new storage::StorageHttpAdminHandler(schemaMan_.get(), kvstore_.get());
     });
+    router.get("/sysinfo").handler([this](web::PathParams&&) {
+        auto handler = new storage::SysInfoHandler();
+        handler->init(dataPaths_[0]);
```
dangleptr, line 93: Just the first data path?

Currently, storaged support multi data paths.  So if you want to detect the disk usage,  all dataPaths should be handled.

knightXun: good points

src/common/fs/FileUtils.cpp
```diff
@@ -134,6 +135,20 @@ std::string FileUtils::basename(const char *path) {
     return result[3].str();
 }
 
+std::tuple<double, double, double, double> FileUtils::detectFilesystemUsage(std::string data_path) {
```
dutor, line 138: `getFileSystemUsage(const std::string)`

src/common/fs/FileUtils.cpp
```diff
@@ -134,6 +135,20 @@ std::string FileUtils::basename(const char *path) {
...
+    const unsigned int GB = (1024 * 1024) * 1024;
+    struct statvfs buffer;
+    int ret = statvfs(data_path.c_str(), &buffer);
+
+    if (!ret) {
+        const double total = (double)(buffer.f_blocks * buffer.f_frsize) / GB;
+        const double available = (double)(buffer.f_bfree * buffer.f_frsize) / GB;
+        const double used = total - available;
+        const double usedPercentage = (double)(used / total) * (double)100;
```
dutor, line 147: All you need to return with this function is _total_ and _available_. The caller is responsible to calculate _used_.
Similarly, you should return by bytes, instead of by GB mandatorily.

src/common/process/ProcessUtils.cpp
```diff
@@ -167,4 +170,40 @@ StatusOr<std::string> ProcessUtils::runCommand(const char* command) {
     return out.str();
 }
 
+
+double ProcessUtils::getMemUsage() {
+    struct rusage r_usage;
+    getrusage(RUSAGE_SELF,&r_usage);
+    return r_usage.ru_maxrss;
```
dutor, line 177: Please investigate what `ru_maxrss` really means.
Moreover, memory usage is not only about phsical memory. Please be more specific on the function name.

src/common/process/ProcessUtils.cpp
```diff
@@ -167,4 +170,40 @@ StatusOr<std::string> ProcessUtils::runCommand(const char* command) {
     return out.str();
 }
 
+
+double ProcessUtils::getMemUsage() {
```
dutor, line 174: Why `double`?

src/graph/SysInfoHandler.cpp
```diff
@@ -0,0 +1,83 @@
...
+folly::dynamic SysInfoHandler::getSysInfo() {
+    folly::dynamic json = folly::dynamic::object();
+    auto memUsage = ProcessUtils::getMemUsage();
+    json["memory_usage"] = memUsage;
+
+    auto cpuUsage = ProcessUtils::getCpuUsage();
+    json["cpu_usage"] = cpuUsage;
+
+    return json;
+}
```
dutor, line 80: _**System Info**_ involves many metrics. What's your plan on the design of this interface? For example, what if we want to separate the CPU usage into  _user_, _sys_, _si_ metrics in future?

src/common/fs/FileUtils.cpp
knightXun: thanks, I will do this as you advise.

src/common/process/ProcessUtils.cpp
knightXun: good point

src/common/fs/FileUtils.cpp
knightXun: good point

knightXun:
> Have you ever tested these code?

The ut test is on the way, I am afraid this pr is too large..

src/graph/SysInfoHandler.cpp
knightXun: I Just want this interface act as `docker stats`. Via this this interface, we can briefly get cpu/memory/disk consumption. We should let users collect the more detail metrics via third party tools.In the future, we can use kubernetes manage nebula, At that time, We can do more and collect more metrics via prometheus.  Since nebula 1.0 not work very well on kubernetes, I think it's better support the minimum feature set to help users build their monitor system as easy as possible.

src/common/process/ProcessUtils.cpp
knightXun: I fixed this, I fetch the memory usage from `/proc/self/stat`

src/daemons/GraphDaemon.cpp
```diff
@@ -223,6 +227,15 @@ Status setupLogging() {
     return status;
 }
 
+Status initWebService(nebula::WebService* svc) {
+    LOG(INFO) << "Starting Graph HTTP Service";
+    auto& router = svc->router();
+    router.get("/sysinfo").handler([](PathParams&&) {
```
chenxu14, line 233: Maybe we can move this to src/webservice/WebService.app, there is no need to declare it in every daemon.

knightXun: Good point, Let me try this on

knightXun:
@dutor done

src/common/process/ProcessUtils.cpp
```diff
@@ -167,4 +170,59 @@ StatusOr<std::string> ProcessUtils::runCommand(const char* command) {
...
+    tty_nr >> tpgid >> flags >> minflt >> cminflt >> majflt >> cmajflt >>
+    utime >> stime >> cutime >> cstime >> priority >> nice
+    >> ingore >> itrealvalue >> starttime >> vsize >> rss;
+    cpu_stat.close();
+
+    long page_size = sysconf(_SC_PAGE_SIZE);
+    return rss * page_size;
+}
+
+// We only support fetch memory usage from cgroup just like runc does
```
Amber1990Zhang, line 194: support fetching

knightXun: nice hint

src/common/fs/FileUtils.cpp
```diff
@@ -134,6 +135,17 @@ std::string FileUtils::basename(const char *path) {
     return result[3].str();
 }
 
+std::tuple<double, double> FileUtils::detectFilesystemUsage(const std::string data_path) {
```
liuyu85cn, line 138: why not use (const std::string& data_path) ?

src/common/fs/FileUtils.h
```diff
@@ -114,7 +114,8 @@ class FileUtils final {
     // Refer to `man 3 rename'
     // return false when rename failed
     static bool rename(const std::string& src, const std::string& dst);
-
+    //Get Disk Usage: total size, available size
+    static std::tuple<double, double> detectFilesystemUsage(const std::string data_path);
```
liuyu85cn, line 118: ditto

knightXun: good point

src/common/fs/FileUtils.cpp
```diff
@@ -134,6 +135,17 @@ std::string FileUtils::basename(const char *path) {
     return result[3].str();
 }
 
+std::tuple<double, double> FileUtils::detectFilesystemUsage(const std::string& data_path) {
```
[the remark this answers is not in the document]
knightXun: done, thanks for nice hint.

knightXun:
nicely call @dangleptr

PR #2249, opened 2020-07-25
fix VertexHolder::getDefaultProp performance issue.

xuguruogu:
VertexHolder::getDefaultProp has huge performace problem, if get some prop not exist.

before:
image

after:
image

### DIFF
--- a/src/dataman/RowReader.h
+++ b/src/dataman/RowReader.h
@@ -72,21 +72,30 @@ public:
 
 
 public:
-    static std::unique_ptr<RowReader> getTagPropReader(
+    RowReader(const RowReader&) = delete;
+    RowReader& operator=(const RowReader&) = delete;
+    RowReader(RowReader&& x) = default;
+    RowReader& operator=(RowReader&& x) = default;
+
+    static RowReader getTagPropReader(
         meta::SchemaManager* schemaMan,
         folly::StringPiece row,
         GraphSpaceID space,
         TagID tag);
 
-    static std::unique_ptr<RowReader> getEdgePropReader(
+    static RowReader getEdgePropReader(
         meta::SchemaManager* schemaMan,
         folly::StringPiece row,
         GraphSpaceID space,
         EdgeType edge);
 
-    static std::unique_ptr<RowReader> getRowReader(
+    static RowReader getRowReader(
         folly::StringPiece row,
         std::shared_ptr<const meta::SchemaProviderIf> schema);
+
+    static RowReader getEmptyRowReader() {
+        return RowReader();
+    }
 
     static StatusOr<VariantType> getDefaultProp(const meta::SchemaProviderIf* schema,
                                                 const std::string& prop) {
@@ -242,9 +251,6 @@ public:
                 return ResultType::E_DATA_INVALID;
         }
     }
-
-    virtual ~RowReader() = default;
-
     SchemaVer schemaVer() const noexcept;
     int32_t numFields() const noexcept;
 
@@ -286,6 +292,47 @@ public:
         return data_;
     }
 
+    operator bool() const noexcept {
+        return operator!=(nullptr);
+    }
+
+    bool operator==(std::nullptr_t) const noexcept {
+        return !data_.data();
+    }
+
+    bool operator==(const RowReader& x) const noexcept {
+        return data_ == x.data_;
+    }
+
+    bool operator!=(std::nullptr_t) const noexcept {
+        return static_cast<bool>(data_.data());
+    }
+
+    bool operator!=(const RowReader& x) const noexcept {
+        return data_ != x.data_;
+    }
+
+    RowReader* operator->() const noexcept {
+        return get();
+    }
+
+    RowReader* get() const noexcept {
+        return const_cast<RowReader*>(this);
+    }
+
+    RowReader* get() noexcept {
+        return this;
+    }
+
+    RowReader& operator*() const noexcept {
+        return *get();
+    }
+
+    void reset() noexcept {
+        this->~RowReader();
+        new(this) RowReader();
+    }
+
     // TODO getPath(const std::string& name) const noexcept;
     // TODO getPath(int64_t index) const noexcept;
     // TODO getList(const std::string& name) const noexcept;
@@ -308,6 +355,7 @@ private:
     mutable std::vector<int64_t> offsets_;
 
 private:
+    RowReader() = default;
     RowReader(folly::StringPiece row,
               std::shared_ptr<const meta::SchemaProviderIf> schema);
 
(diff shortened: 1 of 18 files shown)


xuguruogu:
Anyway VertexHolder::get also has performance problem, due to call malloc for every vertex prop fetch.
 
https://github.com/vesoft-inc/nebula/blob/a2d51c83f179647348ae044a73fe965ead019733/src/graph/GoExecutor.cpp#L1352-L1356

xuguruogu:
rewrite VertexHolder maybe required.

src/graph/GoExecutor.cpp
```diff
@@ -1315,28 +1315,24 @@ bool GoExecutor::processFinalResult(Callback cb) const {
     return true;
 }
 
-OptVariantType GoExecutor::VertexHolder::getDefaultProp(TagID tid, const std::string &prop) const {
-    for (auto it = data_.cbegin(); it != data_.cend(); ++it) {
-        auto it2 = it->second.find(tid);
-        if (it2 != it->second.cend()) {
-            return RowReader::getDefaultProp(std::get<0>(it2->second).get(), prop);
-        }
+OptVariantType GoExecutor::VertexHolder::getDefaultProp(
```
dangleptr, line 1318: Actually, we don't need "getDefaultProp" and "getDefaultPropType" inside VertexHolder, 

You could use RowReader::getDefaultProp directly.

dangleptr:
Thanks for your contribution.  It is really a stupid mistake.

dangleptr:
> Anyway VertexHolder::get also has performance problem, due to call malloc for every vertex prop fetch.
> 
> https://github.com/vesoft-inc/nebula/blob/a2d51c83f179647348ae044a73fe965ead019733/src/graph/GoExecutor.cpp#L1352-L1356

I found the problem too.  Not only in graphd,  but also inside storaged.  We have fixed the issue inside 2.0
If you want to fix it in 1.0,  what you need to do is to implement `reset` method inside RowReader,  for each edge/tag, 
just reset the value and schema.  Just use one RowReader for the whole request.

dangleptr:
Totally the PR looks good to me.  Do you have any plan to fix the RowReader problem in this pr?  @xuguruogu

xuguruogu:
i prefer to fix rowreader later

xuguruogu:
> Totally the PR looks good to me. Do you have any plan to fix the RowReader problem in this pr? @xuguruogu

wait for a while. Writing codes.

xuguruogu:
> > Anyway VertexHolder::get also has performance problem, due to call malloc for every vertex prop fetch.
> > https://github.com/vesoft-inc/nebula/blob/a2d51c83f179647348ae044a73fe965ead019733/src/graph/GoExecutor.cpp#L1352-L1356
> 
> I found the problem too. Not only in graphd, but also inside storaged. We have fixed the issue inside 2.0
> If you want to fix it in 1.0, what you need to do is to implement `reset` method inside RowReader, for each edge/tag,
> just reset the value and schema. Use only one RowReader for the whole request.

expose RowReader with the interface of std::unique_ptr<RowReader> may be a better choice. Without changing large amount of code, all functions using RowReader can benefit from malloc free codes.

src/storage/mutate/AddEdgesProcessor.cpp
```diff
@@ -87,7 +87,7 @@ std::string AddEdgesProcessor::addEdges(int64_t version, PartitionID partId,
     });
     for (auto& e : newEdges) {
         std::string val;
-        std::unique_ptr<RowReader> nReader;
+        RowReader nReader = RowReader::getEmptyRowReader();
```
dangleptr, line 90: You'd better use pointer of RowReader this place. Because there are lots of check null for rowReader in the code.

src/storage/query/QueryBoundProcessor.cpp
```diff
@@ -88,7 +87,7 @@ kvstore::ResultCode QueryBoundProcessor::processEdgeSampling(const PartitionID p
     using Sample = std::tuple<
         EdgeType, /* type */
         std::string, /* key */
-        std::unique_ptr<RowReader>, /* val */
+        RowReader, /* val */
```
dangleptr, line 90: RowReader size is large. Pointer is a better choice.

xuguruogu: make sense

src/storage/mutate/AddEdgesProcessor.cpp
xuguruogu: https://github.com/vesoft-inc/nebula/blob/e5cefb2e68e189181ce56f74fcc89fe3037e8bcf/src/dataman/RowReader.h#L299-L313

resolve it with CPP magic. This can really help to improve performance, the total batch computing costs reduces from 13min to 5min, for about 2.6X improvement.

Access data set from the stack is much faster than from heap, for CPU hardware cache optimization.

src/dataman/RowReader.h
```diff
@@ -286,6 +292,47 @@ class RowReader {
...
+
+    RowReader* get() noexcept {
+        return this;
+    }
+
+    RowReader& operator*() const noexcept {
+        return *get();
+    }
+
+    void reset() noexcept {
```
dangleptr, line 331: It seems we dont need reset any more.  But never mind, leave it here.

src/storage/query/QueryBoundProcessor.cpp
```diff
@@ -112,11 +111,12 @@ kvstore::ResultCode QueryBoundProcessor::processEdgeSampling(const PartitionID p
...
                 partId, vId, edgeType, &fcontext,
-                [&](std::unique_ptr<RowReader> reader,
-                    folly::StringPiece k) {
+                [&](RowReader reader, folly::StringPiece k) {
                     sampler->sampling(
-                        std::make_tuple(edgeType, k.str(), std::move(reader),
-                                        currEdgeSchema, props));
+                        std::make_tuple(
+                            edgeType, k.str(),
+                            std::make_unique<RowReader>(std::move(reader)),
```
dangleptr, line 118: HaHa. Very tricky one~

dangleptr:
Please check the code style.

xuguruogu:
> Awesome work!! The pr looks good to me.
> 
> Besides the RowReader, there is another point we have optimized in 2.0. You could take it as your reference.
> 
> That is MetaClient::getEdgeSchemaFromCache
> For each row, inside the RowReader, it will call this method to get the related schema.
> 
> It is costly, because there are two hashMap inside it.
> So in 2.0, we copy the related edge schema before scanning each edgeType. And use a vector to store the multi versions schema for the edgeType.
> 
> So two hash lookup will be replaced by one random array access.

emm... It may not give the expected performance improvement. I am familiar with commonly used optimization method. Maybe we can talk about it later in a specified scenario.

dangleptr:
>> emm... It may not give the expected performance improvement. I am familiar with commonly used optimization method. Maybe we can talk about it later in a specified scenario.

For RowReader, it has about 1.3x improvement.

PR #2250, opened 2020-07-27
fix double quote in index

knightjya:
What changes were proposed in this pull request?
fix SparkClientGenerator. The function extraValue add duplicate double quote to index. Use extraIndexValue specially for index colums.

Why are the changes needed?
When using extraValue to transfrom index, double quote was added. The change add the function extraIndexValue  to avoid duplicate double quote.
before: duplicate double quote
after: aviod duplicate double quote

Does this PR introduce any user-facing change?
No.

How was this patch tested?

### DIFF
--- a/src/tools/spark-sstfile-generator/src/main/scala/com/vesoft/nebula/tools/generator/v2/SparkClientGenerator.scala
+++ b/src/tools/spark-sstfile-generator/src/main/scala/com/vesoft/nebula/tools/generator/v2/SparkClientGenerator.scala
@@ -297,7 +297,6 @@ object SparkClientGenerator {
           fields.asScala.keys.toList
         }
 
-        val vertexIndex      = sourceProperties.indexOf(vertex)
         val nebulaProperties = properties.mkString(",")
         val data             = createDataSource(spark, pathOpt, tagConfig)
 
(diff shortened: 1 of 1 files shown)


darionyaphet:
Hi @knightjya thanks for take care of that problem. I'm strange why I don't find it. This PR seems proposed when you using key policy and the quote is duplicated. 

`INSERT_VALUE_TEMPLATE_WITH_POLICY` and `ENDPOINT_TEMPLATE` have quote in it. 

Maybe you should remove it. thanks a lot.

knightjya:
> Hi @knightjya thanks for take care of that problem. I'm strange why I don't find it. This PR seems proposed when you using key policy and the quote is duplicated.
> 
> `INSERT_VALUE_TEMPLATE_WITH_POLICY` and `ENDPOINT_TEMPLATE` have quote in it.
> 
> Maybe you should remove it. thanks a lot.

Thx for review.

IMO, it's not a good idea to remove quote in `INSERT_VALUE_TEMPLATE_WITH_POLICY` and `ENDPOINT_TEMPLATE`. 

When using key policy, you cannot distinguish if Index is `StringType` or not. 

Simply remove the quote in `INSERT_VALUE_TEMPLATE_WITH_POLICY` and `ENDPOINT_TEMPLATE` may run into error when Index is `DoubleType`,`ArrayType` etc. The function `extraIndexValue` aviods the underlying problem.

darionyaphet:
> > Hi @knightjya thanks for take care of that problem. I'm strange why I don't find it. This PR seems proposed when you using key policy and the quote is duplicated.
> > `INSERT_VALUE_TEMPLATE_WITH_POLICY` and `ENDPOINT_TEMPLATE` have quote in it.
> > Maybe you should remove it. thanks a lot.
> 
> Thx for review.
> 
> IMO, it's not a good idea to remove quote in `INSERT_VALUE_TEMPLATE_WITH_POLICY` and `ENDPOINT_TEMPLATE`.
> 
> When using key policy, you cannot distinguish if Index is `StringType` or not.
> 
> Simply remove the quote in `INSERT_VALUE_TEMPLATE_WITH_POLICY` and `ENDPOINT_TEMPLATE` may run into error when Index is `DoubleType`,`ArrayType` etc. The function `extraIndexValue` aviods the underlying problem.

When you using `uuid` or `hash` function the argument should be String. 

If the argument is double or any other something, it's an error.

PR #2251, opened 2020-07-27
compaction threads/rate limiter

xuguruogu:
If many sub dir is used in the storaged process, a total limiter is required.

### DIFF
--- a/src/kvstore/RocksEngineConfig.cpp
+++ b/src/kvstore/RocksEngineConfig.cpp
@@ -14,6 +14,8 @@
 #include "rocksdb/slice_transform.h"
 #include "rocksdb/filter_policy.h"
 #include "base/Configuration.h"
+#include "rocksdb/concurrent_task_limiter.h"
+#include "rocksdb/rate_limiter.h"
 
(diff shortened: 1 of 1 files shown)


src/kvstore/RocksEngineConfig.cpp
```diff
@@ -63,6 +65,10 @@ DEFINE_string(rocksdb_compression_per_level, "", "Specify per level compression
 DEFINE_bool(enable_rocksdb_statistics, false, "Whether or not to enable rocksdb's statistics");
 DEFINE_string(rocksdb_stats_level, "kExceptHistogramOrTimers", "rocksdb statistics level");
 
+DEFINE_int32(num_compaction_threads, 16, "Number of total compaction threads");
+
+DEFINE_int32(rate_limit, 512, "write limit in bytes per sec. The unit is MB");
```
dangleptr, line 70: Well done.  Maybe the default value should be unlimited.

xuguruogu: make sense

PR #2252, opened 2020-07-27
bugfix-sparkclient-improve the error record for retry

datian9966:
improve the error record for retry


darionyaphet:
While the dataset is well-structured, the insert sentence should be formatted and running in nebula server successfully.

Otherwise it will report a lot of error until reach the limit.

Currently, the error buffer actually hold the token allocation failed. IMO it's unnecessary.

datian9966:
ok thanks

PR #2260, opened 2020-07-29
reimplement nebula::http::HttpClient with CURL API instead of CURL command

GestureWei:

1. Reimplement nebula::http::HttpClient with libcurl API

1. Run HttpClient in an external CURL command is inefficient. 

No

No

This patch keeps the interface compatibility. We can use the previous http_client_test use-case to test this patch.
### Checklist

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -266,6 +266,7 @@ find_package(FLEX REQUIRED)
 find_package(Readline REQUIRED)
 find_package(NCURSES REQUIRED)
 find_package(LibLZMA MODULE)
+find_package(CURL REQUIRED)
 if(ENABLE_PCH AND NOT ENABLE_ASAN AND NOT ENABLE_NATIVE)
     find_package(PCHSupport)
     add_compile_options(-Winvalid-pch)
@@ -482,6 +483,7 @@ macro(nebula_add_library name type)
     )
 endmacro()
 
+include_directories(AFTER ${CURL_INCLUDE_DIRS})
 include_directories(AFTER ${NEBULA_HOME}/src)
 include_directories(AFTER src/common)
 include_directories(AFTER src/interface)
@@ -565,6 +567,7 @@ macro(nebula_link_libraries target)
         ${libatomic_link_flags}
         -pthread
         ${COVERAGES}
+        ${CURL_LIBRARIES}
     )
 endmacro(nebula_link_libraries)
 
--- a/src/common/http/HttpClient.cpp
+++ b/src/common/http/HttpClient.cpp
@@ -7,31 +7,87 @@
 #include "http/HttpClient.h"
 #include "process/ProcessUtils.h"
 
+#include <curl/curl.h>
+
+static struct __curl_init_helper {
+    __curl_init_helper() {
+        curl_global_init(CURL_GLOBAL_ALL);
+    }
+    ~__curl_init_helper() {
+        curl_global_cleanup();
+    }
+}curl_init_helper;
+
+struct CxxCurl {
+    CxxCurl():curl(curl_easy_init()) {}
+    ~CxxCurl() {
+        if (curl) {
+            curl_easy_cleanup(curl);
+        }
+    }
+    bool isValid()const {
+        return curl != nullptr;
+    }
+    const std::string &getResponse()const {
+        return response;
+    }
+    CURLcode get(const std::string &path) {
+        curl_easy_setopt(curl, CURLOPT_URL, path.c_str());
+        curl_easy_setopt(curl, CURLOPT_FOLLOWLOCATION, 1L);
+        curl_easy_setopt(curl, CURLOPT_WRITEFUNCTION, response_callback);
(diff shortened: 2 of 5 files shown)


CMakeLists.txt
```diff
@@ -266,6 +266,7 @@ find_package(FLEX REQUIRED)
 find_package(Readline REQUIRED)
 find_package(NCURSES REQUIRED)
 find_package(LibLZMA MODULE)
+find_package(CURL REQUIRED)
```
dangleptr, line 269: Thanks for your contribution. 
If you want to introduce new third party libs, we need to repackage our third-party.rpm.  

Of course, i like the way using the libs directly.

chenxu14: > Of course, i like the way using the libs directly.
+1

GestureWei: CURL is one of common libraries provided by many Linux distributions. Why don't we make full use of these system-provided libraries?

dangleptr: Indeed, that is.  
But nebula's all libraries are recompiled by higher gcc version (gcc-7.x above). I am not sure the system-provided  one works well,  have you tested it in your env?

GestureWei: In my env(CentOS 6), system-provided CURL shows good compatibility and stability. Nebula works well with it.

GestureWei: CURL and some other system-provided libs(zlib, bzip2, libevent, etc) are written in C. Whether compiled by higher or lower gcc, they will finally be linked with system-provided glibc(2.12 on CentOS6). So i make a suggestion that nebula can use some of system-provided libs if they meet the requirements of nebula, fbthrift, folly, etc.

dangleptr: Currently,  almost all libs depended by nebula are recompiled includes "zlib, bzip2, zstd, libevent, openssl" etc. Even the "ld" are replaced by ourselves in some Env(For example, centos 6.x).    

I notice libcurl has some dependencies such as openssl.  
To be safe, i suggest we implement the build script under third-party/externals likes other libs. 

wdyt?  @GestureWei

GestureWei: I think we can use this way to use libcurl

dangleptr: That's ok.
Don't block the improvement by it.  Let's go on.

src/common/http/HttpClient.cpp
```diff
@@ -7,31 +7,87 @@
...
+        if (curl) {
+            curl_easy_cleanup(curl);
+        }
+    }
+    bool isValid()const {
+        return curl != nullptr;
+    }
+    const std::string &getResponse()const {
+        return response;
+    }
```
dutor, line 33: Please add a rvalue overload so that the caller has the choice to move the response instead of copy.

src/common/http/HttpClient.cpp
```diff
@@ -7,31 +7,87 @@
...
+
+private:
+    static size_t response_callback(char *data, size_t size, size_t nmemb, void *userptr);
+    CURL *curl;
+    std::string response;
+};
+
+size_t CxxCurl::response_callback(char *data, size_t size, size_t nmemb, void *userptr) {
+    std::string *response = (std::string*)userptr;
```
dutor, line 57: Use `reinterpret_cast`

src/common/http/HttpClient.cpp
```diff
@@ -7,31 +7,87 @@
...
-    } else {
-        return Status::Error(folly::stringPrintf("Http Get Failed: %s", path.c_str()));
+    (void)options;
+    CxxCurl curl;
+    if (!curl.isValid()) {
+        return Status::Error("curl_easy_init() failed");
+    }
+    CURLcode res = curl.get(path);
+    if (res != CURLE_OK) {
+        return Status::Error(
```
dutor, line 74: `Status::Error` accepts printf-like format string, `folly::stringPrintf` is not necessary.

src/common/http/HttpClient.cpp
```diff
@@ -7,31 +7,87 @@
...
+        curl_easy_setopt(curl, CURLOPT_POSTFIELDS, data.c_str());
+        curl_easy_setopt(curl, CURLOPT_POSTFIELDSIZE, (int64_t)data.size());
+        curl_easy_setopt(curl, CURLOPT_WRITEFUNCTION, response_callback);
+        curl_easy_setopt(curl, CURLOPT_WRITEDATA, &response);
+        return curl_easy_perform(curl);
+    }
+
+private:
+    static size_t response_callback(char *data, size_t size, size_t nmemb, void *userptr);
+    CURL *curl;
```
dutor, line 52: Better initialize it with _default member initializer_

GestureWei:
> Mistaken approval.
I have followed your comments to modify the code and refactor the interface for HttpClient::post.

GestureWei:
> Please try building the project in your local environtment

Now I have fixed CloudAuthenticator with refactored HttpClient::post. The build is paased on my env.

Sophie-Xie:

PR #2261, opened 2020-07-29
raft use seperated worker threads

xuguruogu:
issues from raft failure, logs as follow: 

```
I0729 15:02:37.842998  2919 RaftPart.cpp:1266] [Port: 21241, Space: 42, Part: 10] Recieved a VOTING request: space = 42, partition = 10, candidateAddr = 10.51.144.215:21241, term = 1746, lastLogId = 147468491, lastLogTerm = 797
I0729 15:02:37.843086  2919 RaftPart.cpp:1299] [Port: 21241, Space: 42, Part: 10] The partition currently is a Leader, lastLogId 4490575, lastLogTerm 1738, committedLogId 4490575, term 1738
I0729 15:02:37.843163  2919 RaftPart.cpp:1334] [Port: 21241, Space: 42, Part: 10] The partition's last term to receive a log is 1738, which is newer than the candidate's log 797. So the candidate will be rejected
```

raft timeout problem has fixed in #2193. But I meet this problem again and again. For now, I believe this may due to ScanVertexProcessor occupy all work threads for as long as a few minutes. This bug may be can reproduce by call scan of a tag type with empty data set, with more threads more than worker threads.

Surely, the scan process should use reader threads instead of worker threads. Let raft use separated threads alone is also a good option.

### DIFF
--- a/src/storage/StorageServer.cpp
+++ b/src/storage/StorageServer.cpp
@@ -26,6 +26,8 @@ DEFINE_int32(num_io_threads, 16, "Number of IO threads");
 DEFINE_int32(num_worker_threads, 32, "Number of workers");
 DEFINE_int32(storage_http_thread_num, 3, "Number of storage daemon's http thread");
 DEFINE_bool(local_config, false, "meta client will not retrieve latest configuration from meta");
+DEFINE_int32(raft_num_worker_threads, 0,
+             "Number of workers for raft alone. 0 means reuse worker threads");
 
 namespace nebula {
 namespace storage {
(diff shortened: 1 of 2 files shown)


xuguruogu:
As support. ScyllaDB use 4 types of rpc_clients.

https://github.com/scylladb/scylla/blob/c4cb3817cf44ce9f884e0f8791e095d9b7d1dede/message/messaging_service.cc#L468-L527

dangleptr:
My bad. 
You could reuse the readerPool_ inside StorageServiceHandler. That's what we did for QueryBoundProcessor.
Because we don't expose ScanXXXProcessor in graphd, so i just leave it there. You need to separate it from raft threads if you want to use it in your case.

src/storage/StorageServer.cpp
```diff
@@ -48,10 +50,11 @@ std::unique_ptr<kvstore::KVStore> StorageServer::getStoreInstance() {
...
     if (FLAGS_store_type == "nebula") {
-        auto nbStore = std::make_unique<kvstore::NebulaStore>(std::move(options),
-                                                              ioThreadPool_,
-                                                              localHost_,
-                                                              workers_);
+        auto nbStore = std::make_unique<kvstore::NebulaStore>(
+            std::move(options),
+            ioThreadPool_,
+            localHost_,
+            raftWorkers_ != nullptr ? raftWorkers_ : workers_);
```
dangleptr, line 57: It is not enough.
Because the works here is only used by followers,  for leader,  it will reuse the works with outside.

A better way is for read requests,  use readPool_ inside StorageServiceHandler.

xuguruogu: make sense.

PR #2262, opened 2020-07-29
Exposing rocksdb statistics through webservice

chenxu14:
After #2243, rocksdb’s statistics can be enabled by specify --enable_rocksdb_statistics=true, And the statistics info will be dumped into each DB’s log file periodically.
Since all DB instances share the same Statistics obj, there is no need to dump it every where.
So it's better to manage it ourself, and exposing the statistics through webservice.
usage example:
(1) get all stats
http://host:prot/rocksdb_stats
(2) get specify stats
http://host:prot/rocksdb_stats?stats=rocksdb.bytes.read,rocksdb.block.cache.add
(3) return json format
http://host:prot/rocksdb_stats?stats=rocksdb.bytes.read,rocksdb.block.cache.add&returnjson

### DIFF
--- a/src/kvstore/RocksEngineConfig.cpp
+++ b/src/kvstore/RocksEngineConfig.cpp
@@ -129,6 +129,7 @@ rocksdb::Status initRocksdbOptions(rocksdb::Options &baseOpts) {
     std::shared_ptr<rocksdb::Statistics> stats = getDBStatistics();
     if (stats) {
         dbOpts.statistics = std::move(stats);
+        dbOpts.stats_dump_period_sec = 0;  // exposing statistics ourself
     }
     dbOpts.listeners.emplace_back(new EventListener());
 
--- a/src/storage/CMakeLists.txt
+++ b/src/storage/CMakeLists.txt
@@ -33,6 +33,7 @@ nebula_add_library(
     http/StorageHttpIngestHandler.cpp
     http/StorageHttpDownloadHandler.cpp
     http/StorageHttpAdminHandler.cpp
+    http/StorageHttpStatsHandler.cpp
 )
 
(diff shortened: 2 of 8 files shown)


src/common/stats/StatsReporter.cpp
```diff
@@ -0,0 +1,63 @@
...
+		"In OpenFalcon case, this may be http://127.0.0.1:1988/v1/push");
+
+namespace nebula {
+namespace stats {
+
+void reportMetrics(const MetricRecored& record) {
+    if (!FLAGS_statistics_reporter_address.empty()) {
+        std::string json;
+        generateJson(record, &json);
+        StatusOr<std::string> result = http::HttpClient::postData(
```
dangleptr, line 25: Very good feature.

But i have concern about the push mode maybe cause lots of subprocesses (That's what we did inside HttpClient) on server side. 

Could we replace the push mode with pull?

chenxu14: make sense, let me try

chenxu14: Add StorageHttpStatsHandler to exposing rocksdb statistics through webservice, so we can pull metrics from it.
Do you think there is a need to reserve the push logic(since it's configurable, and disabled by default), or just discard it. @dangleptr

dangleptr: IMO we should remove the useless code.  There are too many codes already....

chenxu14: OK, remove the push logic to make the PR simple.

dangleptr:
Please check the code style with cpplint.

Amber1990Zhang:
doc done

PR #2264, opened 2020-07-31
Let us talk about replace the LRU cache with cuckoo hash cache.

xuguruogu:
Use a new kind of cache instead of LRU. :

1. LRU is not efficient for small objects.
2. TTL is needed except cache coherence is guaranteed.
3. Similar to cuckoo hash, use 64K spinlock as concurrency control.
4. Reject to insert to the cache if full is enough for most cases.

The cuckoo hash cache is copied from the cuckoo hash map, removing rehash and cuckoo_run procedures, and act as a cache.

### DIFF
--- a/src/common/base/CuckooHashCache.h
+++ b/src/common/base/CuckooHashCache.h
@@ -1,0 +1,859 @@
+#ifndef COMMON_BASE_CUCKOOHASHCACHE_H_
+#define COMMON_BASE_CUCKOOHASHCACHE_H_
+
+#include <algorithm>
+#include <array>
+#include <atomic>
+#include <bitset>
+#include <cassert>
+#include <cstdint>
+#include <cstdlib>
+#include <functional>
+#include <iostream>
+#include <iterator>
+#include <limits>
+#include <list>
+#include <memory>
+#include <mutex>
+#include <stdexcept>
+#include <string>
+#include <thread>
+#include <type_traits>
+#include <utility>
+#include <vector>
+
+#include "time/WallClock.h"
+
+namespace nebula {
+
+//! The default maximum number of keys per bucket
+constexpr size_t DEFAULT_SLOT_PER_BUCKET = 4;
+//! The default number of elements in an empty hash table
+constexpr size_t DEFAULT_SIZE = (1U << 20) * DEFAULT_SLOT_PER_BUCKET;
+constexpr size_t DEFAULT_TTL_SECONDS = 10 * 60;
+
+/**
+ * bucket_container manages storage of key-value pairs for the table.
+ * It stores the items inline in uninitialized memory, and keeps track of which
+ * slots have live data and which do not. It also stores a partial hash for
+ * each live key. It is sized by powers of two.
+ *
+ * @tparam Key type of keys in the table
+ * @tparam T type of values in the table
+ * @tparam Partial type of partial keys
+ * @tparam SLOT_PER_BUCKET number of slots for each bucket in the table
+ */
+template <class Key, class T, class Partial, std::size_t SLOT_PER_BUCKET>
+class bucket_container {
+public:
+    using key_type = Key;
+    using mapped_type = T;
+    using value_type = std::pair<const Key, T>;
+    using partial_t = Partial;
+
+public:
+    /*
+     * The bucket type holds SLOT_PER_BUCKET key-value pairs, along with their
+     * partial keys and occupancy info. It uses aligned_storage arrays to store
+     * the keys and values to allow constructing and destroying key-value pairs
+     * in place. The lifetime of bucket data should be managed by the container.
+     * It is the user's responsibility to confirm whether the data they are
+     * accessing is live or not.
+     */
+    class bucket {
+    public:
+        bucket() noexcept : occupied_() {}
+
+        const value_type &kvpair(size_t ind) const {
+            return *static_cast<const value_type *>(
+                static_cast<const void *>(&values_[ind]));
+        }
+        value_type &kvpair(size_t ind) {
+            return *static_cast<value_type *>(static_cast<void *>(&values_[ind]));
+        }
+
+        const key_type &key(size_t ind) const {
+            return storage_kvpair(ind).first;
+        }
+
+        const mapped_type &mapped(size_t ind) const {
+            return storage_kvpair(ind).second;
+        }
+        mapped_type &mapped(size_t ind) { return storage_kvpair(ind).second; }
+
+        partial_t partial(size_t ind) const { return partials_[ind]; }
+        partial_t &partial(size_t ind) { return partials_[ind]; }
+
+        uint32_t expire(size_t ind) const { return expires_[ind]; }
+        uint32_t &expire(size_t ind) { return expires_[ind]; }
+
+        bool occupied(size_t ind) const { return occupied_[ind]; }
+        bool &occupied(size_t ind) { return occupied_[ind]; }
+    private:
+        friend class bucket_container;
+
+        using storage_value_type = std::pair<Key, T>;
+
+        const storage_value_type &storage_kvpair(size_t ind) const {
+            return *static_cast<const storage_value_type *>(
+                static_cast<const void *>(&values_[ind]));
+        }
+        storage_value_type &storage_kvpair(size_t ind) {
+            return *static_cast<storage_value_type *>(
+                static_cast<void *>(&values_[ind]));
+        }
+
+        std::array<typename std::aligned_storage<sizeof(storage_value_type),
+            alignof(storage_value_type)>::type, SLOT_PER_BUCKET> values_;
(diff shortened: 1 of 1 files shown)


xuguruogu:
With the help of cuckoo cache and cache non-exist tags, I achieved a larger performance improvement. I can not even count out the improvement factors.

Lock collision of LRU cache and cache miss(including non exist tag cache miss) rate are the main performance bottlenecks for now.

xuguruogu:
With the cuckoo hash cache difference only, I achieved to reduce the costs mentioned in #2249 from 5min to less than 2min. Regarding the spark task management and preprocessing costs, I believe it can contribute more than 3X improvement.

dangleptr:
Very good feature.  
We'd better use an option to control which kind of cache used.

dangleptr:
> The cuckoo hash cache is copied from the cuckoo hash map, removing rehash and cuckoo_run procedures, and act as a cache.

We should add some comments about it.

src/common/base/CuckooHashCache.h
```diff
@@ -0,0 +1,875 @@
...
+    }
+
+    // snapshot_and_lock_two loads locks the buckets associated with the given
+    // hash value, making sure the hashpower doesn't change before the locks are
+    // taken. Thus it ensures that the buckets and locks corresponding to the
+    // hash value will stay correct as long as the locks are held. It returns
+    // the bucket indices associated with the hash value and the current
+    // hashpower.
+    TwoBuckets snapshot_and_lock_two(const hash_value &hv) const {
+        while (true) {
```
dangleptr, line 667: why while(true) here?

src/common/base/CuckooHashCache.h
```diff
@@ -0,0 +1,875 @@
...
+            return table_position{b.i2, static_cast<size_t>(res2),
+                                  failure_key_duplicated};
+        }
+        if (res1 != -1) {
+            return table_position{b.i1, static_cast<size_t>(res1), ok};
+        }
+        if (res2 != -1) {
+            return table_position{b.i2, static_cast<size_t>(res2), ok};
+        }
+        return table_position{0, 0, failure_table_full};
```
dangleptr, line 781: If two slots are occupied, why not move the element on the original slot instead of return full directly. 

In some rare cases, there may be lots of empty slots, but return full just due to one key's  two slots are occupied?

src/common/base/CuckooHashCache.h
```diff
@@ -0,0 +1,875 @@
...
+                static_cast<const void *>(&values_[ind]));
+        }
+        storage_value_type &storage_kvpair(size_t ind) {
+            return *static_cast<storage_value_type *>(
+                static_cast<void *>(&values_[ind]));
+        }
+
+        std::array<typename std::aligned_storage<sizeof(storage_value_type),
+            alignof(storage_value_type)>::type, SLOT_PER_BUCKET> values_;
+        std::array<uint32_t, SLOT_PER_BUCKET> expires_;
```
dangleptr, line 110: What's the value about expires_ here?  The expired timestamp in seconds?

src/common/base/CuckooHashCache.h
```diff
@@ -0,0 +1,875 @@
...
+    // in `slot` and return true. If no duplicate key is found and no empty slot
+    // is found, we store -1 in `slot` and return true.
+    template <typename K>
+    bool try_find_insert_bucket(const size_t bucket_ind, int &slot,
+                                const partial_t partial, const K &key) {
+        bucket &b = buckets_[bucket_ind];
+        slot = -1;
+        for (int i = 0; i < static_cast<int>(slot_per_bucket()); ++i) {
+            if (try_reclaim(bucket_ind, i)) {
+                slot = i;
```
dangleptr, line 808: why not return true here?

src/common/base/CuckooHashCache.h
```diff
@@ -0,0 +1,875 @@
...
+                if (!is_simple() && partial != b.partial(i)) {
+                    continue;
+                }
+                if (key_eq()(b.key(i), key)) {
+                    slot = i;
+                    ++locks_[lock_ind(bucket_ind)].hit_counter();
+                    return false;
+                }
+            } else {
+                slot = i;
```
dangleptr, line 819: ditto

src/common/base/CuckooHashCache.h
```diff
@@ -0,0 +1,875 @@
...
+     * the keys and values to allow constructing and destroying key-value pairs
+     * in place. The lifetime of bucket data should be managed by the container.
+     * It is the user's responsibility to confirm whether the data they are
+     * accessing is live or not.
+     */
+    class bucket {
+    public:
+        bucket() noexcept : occupied_() {}
+
+        const value_type &kvpair(size_t ind) const {
```
dangleptr, line 66: what's the difference with`const storage_value_type &storage_kvpair(size_t ind) const `

src/common/base/CuckooHashCache.h
```diff
@@ -0,0 +1,875 @@
...
+        }
+
+        const key_type &key(size_t ind) const {
+            return storage_kvpair(ind).first;
+        }
+        key_type &&movable_key(size_t ind) {
+            return std::move(storage_kvpair(ind).first);
+        }
+
+        const mapped_type &mapped(size_t ind) const {
```
dangleptr, line 81: Why name it "mapped" instead of "value" ?  Make me confused.

dangleptr:
> With the cuckoo hash cache difference only, I achieved to reduce the costs mentioned in #2249 from 5min to less than 2min. Regarding the spark task management and preprocessing costs, I believe it can contribute more than 3X improvement.

Ha, i think the performance comes from TTL feature, not just cuckoo hash cache.

src/common/base/CuckooHashCache.h
xuguruogu: storage_value_type is a std::pair with key const. It should not be visited outside.

xuguruogu:
It acts like google sparse_hash_map, using continuous memory, go forward a limited space for insert and lookup. With no memory malloc if key/value is simple. With good spatial locality features.

dangleptr:
> It acts like google sparse_hash_map, using continuous memory, go forward a limited space for insert and lookup. With no memory malloc if key/value is simple. With good spatial locality features.

Yes,  we should satisfy different purposes with different kind cache.   Let's talk about it offline.

src/common/base/CuckooHashCache.h
dangleptr: I got your purpose.

src/common/base/CuckooHashCache.h
```diff
@@ -0,0 +1,875 @@
...
+    template <typename K>
+    bool try_find_insert_bucket(const size_t bucket_ind, int &slot,
+                                const partial_t partial, const K &key) {
+        bucket &b = buckets_[bucket_ind];
+        slot = -1;
+        for (int i = 0; i < static_cast<int>(slot_per_bucket()); ++i) {
+            if (try_reclaim(bucket_ind, i)) {
+                slot = i;
+            } else if (b.occupied(i)) {
+                if (!is_simple() && partial != b.partial(i)) {
```
dangleptr, line 810: We could eliminate the `if` here. Because we could ensure `is_simple` at compile time.

src/common/base/CuckooHashCache.h
```diff
@@ -0,0 +1,875 @@
+#pragma once
```
dangleptr, line 1: `#pragma once` is not standard,  please use include guard as other headers in our project.

src/common/base/CuckooHashCache.h
```diff
@@ -0,0 +1,859 @@
...
+ * @tparam Key type of keys in the table
+ * @tparam T type of values in the table
+ * @tparam Partial type of partial keys
+ * @tparam SLOT_PER_BUCKET number of slots for each bucket in the table
+ */
+template <class Key, class T, class Partial, std::size_t SLOT_PER_BUCKET>
+class bucket_container {
+public:
+    using key_type = Key;
+    using mapped_type = T;
```
bright-starry-sky, line 50: Could you explain the usage of T?

src/common/base/CuckooHashCache.h
```diff
@@ -0,0 +1,859 @@
...
+    bool erase_fn(const K &key, F fn) {
+        const hash_value hv = hashed_key(key);
+        const auto b = snapshot_and_lock_two(hv);
+        const table_position pos = cuckoo_find(key, hv.partial, b.i1, b.i2);
+        if (pos.status == ok) {
+            if (fn(buckets_[pos.index].mapped(pos.slot))) {
+                del_from_bucket(pos.index, pos.slot);
+            }
+            return true;
+        } else {
```
monadbobo, line 319: No need for an else here.

Sophie-Xie:

PR #2267, opened 2020-08-03
Fix the wrong usage for LRUCache

dangleptr:
https://discuss.nebula-graph.com.cn/t/topic/948

As current usage, we can't ensure the data consistency between cache and data.

For example,   reader data ->  evict cache (write) -> write cache(in reader)  will cause the dirty data forever before next write comes.  

We should fix the problem with putIfAbsent when reading but insert overwrite when writing. 

Of course, update/delete still has the problem, will fix them later.


src/storage/mutate/AddVerticesProcessor.cpp
```diff
@@ -30,78 +30,66 @@ void AddVerticesProcessor::process(const cpp2::AddVerticesRequest& req) {
...
+            const auto& tags = v.get_tags();
+            std::for_each(tags.begin(), tags.end(), [&](auto& tag) {
+                VLOG(3) << "PartitionID: " << partId << ", VertexID: " << v.get_id()
+                        << ", TagID: " << tag.get_tag_id() << ", TagVersion: " << version;
+                auto key = NebulaKeyUtils::vertexKey(partId, v.get_id(),
+                                                     tag.get_tag_id(), version);
+                if (FLAGS_enable_vertex_cache && vertexCache_ != nullptr) {
+                    cacheData.emplace_back(v.get_id(), tag.get_tag_id(), tag.get_props());
+                }
+                data.emplace_back(std::move(key), std::move(tag.get_props()));
```
bright-starry-sky, line 52: One question,Does it cause duplicate indexes to be inserted? for example : 
insert vertex ({vid1_tag1_version1, prop1}, {vid1_tag1_version2, prop2}...)

dangleptr: When you insert {vid1_tag1_version, prop2}, the index generated by ({vid1_tag1_version, prop1} will be deleted as our logic.

dangleptr: Good one.  I miss the logic.  Maybe we could combine the inserts before calling kvstore::asyncPut

src/common/base/ConcurrentLRUCache.h
```diff
@@ -206,19 +206,21 @@ class LRU {
...
+        }
+        VLOG(3) << "Insert key " << key << ", val " << value;
+        typename map_type::iterator it = map_.find(key);
+        if (it == map_.end()) {
             list_.push_front(key);
             map_.emplace(std::forward<key_type>(key),
                          std::make_tuple(std::forward<value_type>(value), list_.begin()));
+        } else {
+            // Overwrite the value
+            std::get<0>(it->second) = std::move(value);
```
liuyu85cn, line 223: also need to move this value to front of list ?

dangleptr: If so, the operation becomes evict the old key then insert the new key.  
I am not sure which way is better.

src/common/base/ConcurrentLRUCache.h
```diff
@@ -206,19 +206,21 @@ class LRU {
...
-            // insert item into the cache, but first check if it is full
-            if (size() >= capacity_) {
-                VLOG(3) << "Size:" << size() << ", capacity " << capacity_;
-                // cache is full, evict the least recently used item
-                evict();
-            }
-            VLOG(3) << "Insert key " << key << ", val " << value;
-            // insert the new item
+        // insert item into the cache, but first check if it is full
+        if (size() >= capacity_) {
```
liuyu85cn, line 210: what if I have a 10 capacity LRU, 
which I have key 1-10 in it. (list 1 to 10)
Then if I insert a key 5,
it will pop 10, which is not really necessary

src/common/base/ConcurrentLRUCache.h
panda-sheep: 👍

dangleptr: Make sense

src/common/base/ConcurrentLRUCache.h
dangleptr: make sense

PR #2268, opened 2020-08-03
Go on with go optimization

xuguruogu:
1. preserve vector/map
2. remove storage lock
3. avoid VertexHolder string copy.


src/graph/GoExecutor.cpp
```diff
@@ -1317,37 +1328,27 @@ bool GoExecutor::processFinalResult(Callback cb) const {
...
-    return RowReader::getDefaultProp(schema.get(), prop);
-}
-
-SupportedType GoExecutor::VertexHolder::getDefaultPropType(
-    TagID tagId, const std::string &prop) const {
-    auto space = ectx_->rctx()->session()->space();
-    auto schema = ectx_->schemaManager()->getTagSchema(space, tagId);
-    if (schema == nullptr) {
-        return nebula::cpp2::SupportedType::UNKNOWN;
+    auto iter = tagSchemaMap_.find(tagId);
```
dangleptr, line 1331: Please rebase the pr, it seems there are some overlap with last pr.

src/graph/GoExecutor.cpp
```diff
@@ -1356,39 +1357,56 @@ OptVariantType GoExecutor::VertexHolder::get(VertexID id, TagID tid,
...
+        for (auto &vdata : *vertices) {
+            auto vid = vdata.get_vertex_id();
+            for (auto &td : vdata.tag_data) {
+                DCHECK(td.__isset.data);
+                auto tagId = td.tag_id;
+                auto it = tagSchemaMap_.find(tagId);
+                DCHECK(it != tagSchemaMap_.end());
+                auto &vschema = it->second;
+                auto vreader = RowReader::getRowReader(td.data, vschema);
+                data_.emplace(std::make_pair(vid, tagId), std::move(vreader));
```
dangleptr, line 1407: Well done.   Indeed, we only need to decode once. 
You'd better add some comments here,  because RowReader does not hold any data,  the usage for VertexHolder currently should ensure the lifetime of the "responses".  
This is different with the original behavior for VertexHolder.

src/graph/GoExecutor.cpp
```diff
@@ -1356,39 +1357,56 @@ OptVariantType GoExecutor::VertexHolder::get(VertexID id, TagID tid,
...
-    for (auto &vdata : *vertices) {
-        std::unordered_map<TagID, VData> m;
-        for (auto &td : vdata.tag_data) {
-            DCHECK(td.__isset.data);
-            auto it = vertexSchema->find(td.tag_id);
-            DCHECK(it != vertexSchema->end());
-            m[td.tag_id] = {std::make_shared<ResultSchemaProvider>(it->second), td.data};
-        }
-        data_[vdata.vertex_id] = std::move(m);
+        std::transform(
```
dangleptr, line 1383: Em...  Mostly, each response has the same vertexSchema returned.  
Do we really need to do the transform for every response?

src/graph/GoExecutor.h
```diff
@@ -178,13 +178,12 @@ class GoExecutor final : public TraverseExecutor {
...
         OptVariantType get(VertexID id, TagID tid, const std::string &prop) const;
-        void add(const storage::cpp2::QueryResponse &resp);
-        nebula::cpp2::SupportedType getDefaultPropType(TagID tid, const std::string &prop) const;
-        nebula::cpp2::SupportedType getType(VertexID id, TagID tid, const std::string &prop);
+        void add(const std::vector<storage::cpp2::QueryResponse> &responses);
 
     private:
-        using VData = std::tuple<std::shared_ptr<ResultSchemaProvider>, std::string>;
-        std::unordered_map<VertexID, std::unordered_map<TagID, VData>> data_;
+        std::unordered_map<std::pair<VertexID, TagID>, RowReader> data_;
```
dangleptr, line 184: Well done.  VertexId + TagID should be the key.

src/storage/query/QueryBoundProcessor.cpp
```diff
@@ -197,8 +205,7 @@ kvstore::ResultCode QueryBoundProcessor::processVertex(PartitionID partId, Verte
     }
 
     if (onlyVertexProps_) {
-        std::lock_guard<std::mutex> lg(this->lock_);
```
dangleptr, line 200: Well done.  I want to how much the perf being improved without the lock?

xuguruogu: much smaller than the cache optimization.

xuguruogu: storaged:
image

graphd:
image

What else I can to to help improve the performance.

src/graph/GoExecutor.cpp
xuguruogu: moving data also can make confusion if use it later.

src/storage/query/QueryBoundProcessor.cpp
xuguruogu: Need to write a custom prefix bloom filter.

dangleptr: For GoExecutor,  current logic is not reasonable when yield dest properties.  Actually we don't need VertexHolder. 

When the last step getNeighbors invoked,  the edge props have been fetched,  so we could combine them directly after fetch dest points props.  VertexHolder is needless.

dangleptr: > Need to write a custom prefix bloom filter.

The problem about prefix_extractor is we store some system information (such as parts hold by one rocksdb instance). The format is not suitable for prefix_extractor.

dangleptr: More further,  the InterimResult has big performance issue.  It will introduce extra encode/decode for the response. 
If we fix the problem,  "Go .... yield dstId | fetch props $- ...."  will be a better choice for you.

Currently,  you could use two queries "Go ..." and "Fetch props" on client separately which will acquire better performance.

xuguruogu: > For GoExecutor, current logic is not reasonable when yield dest properties. Actually we don't need VertexHolder.
> 
> When the last step getNeighbors invoked, the edge props have been fetched, so we could combine them directly after fetch dest points props. VertexHolder is needless.

VertexHolder is needed currently.

src/graph/GoExecutor.cpp
dangleptr: what's the meaning?  we don't use the response anymore.

src/storage/query/QueryBoundProcessor.cpp
dangleptr: Anyway,  the pr totally looks good to me.  Let's open another one for it.

src/graph/GoExecutor.cpp
```diff
@@ -639,7 +639,20 @@ void GoExecutor::onVertexProps(RpcResponse &&rpcResp) {
 
 std::vector<VertexID> GoExecutor::getDstIdsFromResps(std::vector<RpcResponse>::iterator begin,
                                                      std::vector<RpcResponse>::iterator end) const {
+    size_t num = 0;
+    for (auto it = begin; it != end; ++it) {
+        for (const auto &resp : it->responses()) {
+            auto *vertices = resp.get_vertices();
+            if (vertices == nullptr) {
+                continue;
+            }
```
laura-ding, line 648: if (!resp.__isset.vertices) {
    continue;
}

src/graph/GoExecutor.cpp
```diff
@@ -639,7 +639,20 @@ void GoExecutor::onVertexProps(RpcResponse &&rpcResp) {
...
+            num += vertices->size();
+        }
+    }
+
     std::unordered_set<VertexID> set;
+    set.reserve(num);
+
     for (auto it = begin; it != end; ++it) {
         for (const auto &resp : it->responses()) {
             auto *vertices = resp.get_vertices();
```
laura-ding, line 658: ditto

src/graph/GoExecutor.cpp
```diff
@@ -1356,39 +1357,53 @@ OptVariantType GoExecutor::VertexHolder::get(VertexID id, TagID tid,
...
-
-    auto iter2 = iter->second.find(tid);
-    if (iter2 == iter->second.end()) {
-        return getDefaultPropType(tid, prop);
+void GoExecutor::VertexHolder::add(
+    const std::vector<storage::cpp2::QueryResponse> &responses) {
+    size_t num = 0;
+    for (auto& resp : responses) {
+        auto *vertices = resp.get_vertices();
+        auto *vertexSchema = resp.get_vertex_schema();
```
laura-ding, line 1365: ditto

PR #2270, opened 2020-08-04
Add GetNeighborsBenchmark

critical27:
Add GetNeighborsBenchmark, no concurrency, used to compare with 2.0 under almost same data.

The copy of lastRank and lastDstId has trivial impact in 1.0, so leave it alone. @dangleptr.

### DIFF
--- a/src/storage/test/GetNeighborsBenchmark.cpp
+++ b/src/storage/test/GetNeighborsBenchmark.cpp
@@ -1,0 +1,317 @@
+/* Copyright (c) 2020 vesoft inc. All rights reserved.
+ *
+ * This source code is licensed under Apache 2.0 License,
+ * attached with Common Clause Condition 1.0, found in the LICENSES directory.
+ */
+
+#include "base/Base.h"
+#include <folly/Benchmark.h>
+#include "fs/TempDir.h"
+#include "utils/NebulaKeyUtils.h"
+#include <gtest/gtest.h>
+#include <rocksdb/db.h>
+#include "storage/query/QueryBoundProcessor.h"
+#include "storage/test/TestUtils.h"
+#include "dataman/RowReader.h"
+#include "dataman/RowWriter.h"
+#include "meta/SchemaManager.h"
+#include "common/filter/Expressions.h"
+
+DEFINE_int64(max_rank, 1000, "max rank of each edge");
+DEFINE_double(filter_ratio, 0.5, "ratio of data would pass filter");
+
+std::unique_ptr<nebula::kvstore::KVStore> gKV;
+std::unique_ptr<nebula::storage::AdHocSchemaManager> gSchemaMan;
+
+namespace nebula {
+namespace storage {
+
+// only add two tag and two edge
+void mockData(kvstore::KVStore* kv) {
+    GraphSpaceID spaceId = 0;
+    PartitionID partId = 0;
+    TagID tagId = 1;
+    EdgeType edgeType = 101;
+    std::vector<kvstore::KV> data;
+    auto ver = std::numeric_limits<int32_t>::max() - 0;
+    ver = folly::Endian::big(ver);
+
+    // tags
+    for (VertexID src = 1; src <= 10; src++) {
+        VertexID dst = src + 100;
+        {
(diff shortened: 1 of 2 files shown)


src/storage/test/GetNeighborsBenchmark.cpp
```diff
@@ -0,0 +1,317 @@
...
+
+void go(int32_t iters,
+        const std::vector<nebula::VertexID> vIds,
+        const std::vector<std::string>& playerProps,
+        const std::vector<std::string>& serveProps) {
+    nebula::storage::cpp2::GetNeighborsRequest req;
+    BENCHMARK_SUSPEND {
+        req = nebula::storage::buildRequest(vIds, playerProps, serveProps);
+    }
+    auto executor = std::make_unique<folly::CPUThreadPoolExecutor>(1);
```
CPWstatic, line 224: I think we should build the thread pool outside the test case.

critical27: my bad, fixed.

src/storage/test/GetNeighborsBenchmark.cpp
```diff
@@ -0,0 +1,320 @@
...
+TenVertexOneProperty                                         6.00ms   166.76
+TenVertexOnePropertyWithFilter                    89.82%     6.68ms   149.78
+============================================================================
+
+--max_rank=1000 --filter_ratio=0.5
+============================================================================
+GetNeighborsBenchmark.cpprelative                         time/iter  iters/s
+============================================================================
+OneVertexOneProperty                                       629.54us    1.59K
+OneVertexOnePropertyWithFilter                    72.33%   870.39us    1.15K
```
dangleptr, line 306: It seems the more data filtered out, the better performance acquired.   
So the bottleneck is encoder(RowWriter) ?

critical27: Hard to say whether bottleneck is encoder or not. With bigger `filter_ratio`, it will need to eval the exp which is more likely to be useless. But with a small `filter_ratio`, we can filter out more data, so there will be less data to encode. We can only say with a bit `filter_ratio`, it has worse performance.

The result in 2.0 is same.

dangleptr: What if we comment the collector?  What's the performance looks like?

critical27: Looks like this, do not collect any value.
image

src/storage/test/GetNeighborsBenchmark.cpp
```diff
@@ -0,0 +1,320 @@
...
+DEFINE_double(filter_ratio, 0.5, "ratio of data would pass filter");
+
+std::unique_ptr<nebula::kvstore::KVStore> gKV;
+std::unique_ptr<nebula::storage::AdHocSchemaManager> gSchemaMan;
+std::unique_ptr<folly::CPUThreadPoolExecutor> gExecutor;
+
+namespace nebula {
+namespace storage {
+
+// only add two tag and two edge
```
panda-sheep, line 30: This comment is not right,
Here have 10 tags and 10,000 edges

src/storage/test/GetNeighborsBenchmark.cpp
panda-sheep: Good job.
By the way, I think that even with the same data, there is no comparison between filters and no filters, right?

critical27: Yeah, you're right.

dangleptr: > Looks like this, do not collect any value.
> image

Em.... So we can ensure the cost comes from encoder.

src/storage/test/GetNeighborsBenchmark.cpp
critical27: My bad.

src/storage/test/GetNeighborsBenchmark.cpp
dangleptr: Let's run the benchmark for a long time,  and check the perf top about it.  @critical27

PR #2272, opened 2020-08-04
Support TTL cache inside ConcurrentLRUCache

dangleptr:
In many cases,  users could load all vertices.  
LRU is not suitable,  so we support Map cache for all data with TTL feature. 
The performance has been improved by 2.7x compare to LRU cache.

### DIFF
--- a/src/common/base/CMakeLists.txt
+++ b/src/common/base/CMakeLists.txt
@@ -10,6 +10,7 @@ add_library(
     SanitizerOptions.cpp
     SignalHandler.cpp
     SlowOpTracker.cpp
+    ConcurrentLRUCache.cpp
 )
 
 IF(${PCHSupport_FOUND})
--- a/src/common/base/ConcurrentLRUCache.cpp
+++ b/src/common/base/ConcurrentLRUCache.cpp
@@ -1,0 +1,15 @@
+/* Copyright (c) 2020 vesoft inc. All rights reserved.
+ *
+ * This source code is licensed under Apache 2.0 License,
+ * attached with Common Clause Condition 1.0, found in the LICENSES directory.
+ */
+
+#include "base/ConcurrentLRUCache.h"
+
+DEFINE_int64(cache_ttl_ms, -1, "TTL for cache entry, only valid for MapTTLCache");
+
+namespace nebula {
+
+}  // namespace nebula
+
+
--- a/src/common/base/ConcurrentLRUCache.h
+++ b/src/common/base/ConcurrentLRUCache.h
@@ -348,7 +348,8 @@ public:
                                          time::WallClock::fastNowInMilliSec()));
         } else {
             // Overwrite the value
-            std::get<0>(it->second) = std::move(value);
+            it->second = std::make_tuple(std::move(value),
+                                         time::WallClock::fastNowInMilliSec());
         }
     }
 
--- a/src/common/base/ConcurrentLRUCache.h
+++ b/src/common/base/ConcurrentLRUCache.h
@@ -13,13 +13,18 @@
 #include <utility>
 #include <boost/optional.hpp>
 #include <gtest/gtest_prod.h>
+#include <folly/container/F14Map.h>
+#include <folly/SpinLock.h>
+#include "time/WallClock.h"
+
+DECLARE_int64(cache_ttl_ms);
 
 namespace nebula {
 
 template<class Key, class Value>
 class LRU;
 
-template<typename K, typename V>
+template<typename K, typename V, typename C = LRU<K, V>>
 class ConcurrentLRUCache final {
     FRIEND_TEST(ConcurrentLRUCacheTest, SimpleTest);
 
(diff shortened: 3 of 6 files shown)


dangleptr:
@xuguruogu Do you want to try it.

xuguruogu:
what if some keys never access, how to evict them

dangleptr:
> what if some keys never access, how to evict them

Em... The cache is used for situation all the vertices could be hold in cache. So is it really a problem?

xuguruogu:
> > what if some keys never access, how to evict them
> 
> Em... The cache is used for situation all the vertices could be hold in cache. So is it really a problem?

Does every space share the same cache?

dangleptr:
> > > what if some keys never access, how to evict them
> > 
> > 
> > Em... The cache is used for situation all the vertices could be hold in cache. So is it really a problem?
> Does every space share the same cache?

That's a good question. Currently, all spaces share the same vertexCache.  Maybe "one space one cache"  is  a better choice.  wdyt?

xuguruogu:
> > > > what if some keys never access, how to evict them
> > > 
> > > 
> > > Em... The cache is used for situation all the vertices could be hold in cache. So is it really a problem?
> > > Does every space share the same cache?
> 
> That's a good question. Currently, all spaces share the same vertexCache. Maybe "one space one cache" is a better choice. wdyt?

I prefer to share the cache for all spaces. Otherwise, it will have trouble with memory control.

dangleptr:
> > > > > what if some keys never access, how to evict them
> > > > 
> > > > 
> > > > Em... The cache is used for situation all the vertices could be hold in cache. So is it really a problem?
> > > > Does every space share the same cache?
> > 
> > 
> > That's a good question. Currently, all spaces share the same vertexCache. Maybe "one space one cache" is a better choice. wdyt?
> 
> I prefer to share the cache for all spaces. Otherwise, it will have trouble with memory control.

OK.  Could you try it in your env?  F14Map is also an  implementation of cuckoo Hash with vector instructions used to accelerate the scan inside one bucket.  I think the performance should be same with yours in #2264

xuguruogu:
> > > > > > what if some keys never access, how to evict them
> > > > > 
> > > > > 
> > > > > Em... The cache is used for situation all the vertices could be hold in cache. So is it really a problem?
> > > > > Does every space share the same cache?
> > > 
> > > 
> > > That's a good question. Currently, all spaces share the same vertexCache. Maybe "one space one cache" is a better choice. wdyt?
> > 
> > 
> > I prefer to share the cache for all spaces. Otherwise, it will have trouble with memory control.
> 
> OK. Could you try it in your env? F14Map is also an implementation of cuckoo Hash with vector instructions used to accelerate the scan inside one bucket. I think the performance should be same with yours in #2264

As we may change the space for test frequently. It seems the cache may has no effect if space changed.

dangleptr:
> > > > > > > what if some keys never access, how to evict them
> > > > > > 
> > > > > > 
> > > > > > Em... The cache is used for situation all the vertices could be hold in cache. So is it really a problem?
> > > > > > Does every space share the same cache?
> > > > 
> > > > 
> > > > That's a good question. Currently, all spaces share the same vertexCache. Maybe "one space one cache" is a better choice. wdyt?
> > > 
> > > 
> > > I prefer to share the cache for all spaces. Otherwise, it will have trouble with memory control.
> > 
> > 
> > OK. Could you try it in your env? F14Map is also an implementation of cuckoo Hash with vector instructions used to accelerate the scan inside one bucket. I think the performance should be same with yours in #2264
> 
> As we may change the space for test frequently. It seems the cache may has no effect if space changed.

Em... If so, how do you test #2264

xuguruogu:
> > > > > > > > what if some keys never access, how to evict them
> > > > > > > 
> > > > > > > 
> > > > > > > Em... The cache is used for situation all the vertices could be hold in cache. So is it really a problem?
> > > > > > > Does every space share the same cache?
> > > > > 
> > > > > 
> > > > > That's a good question. Currently, all spaces share the same vertexCache. Maybe "one space one cache" is a better choice. wdyt?
> > > > 
> > > > 
> > > > I prefer to share the cache for all spaces. Otherwise, it will have trouble with memory control.
> > > 
> > > 
> > > OK. Could you try it in your env? F14Map is also an implementation of cuckoo Hash with vector instructions used to accelerate the scan inside one bucket. I think the performance should be same with yours in #2264
> > 
> > 
> > As we may change the space for test frequently. It seems the cache may has no effect if space changed.
> 
> Em... If so, how do you test #2264

The cache in #2264 will try to reclaim the slots that contain expired data. If space changed, new data will replace the data never access.

dangleptr:
> > > > > > > > > what if some keys never access, how to evict them
> > > > > > > > 
> > > > > > > > 
> > > > > > > > Em... The cache is used for situation all the vertices could be hold in cache. So is it really a problem?
> > > > > > > > Does every space share the same cache?
> > > > > > 
> > > > > > 
> > > > > > That's a good question. Currently, all spaces share the same vertexCache. Maybe "one space one cache" is a better choice. wdyt?
> > > > > 
> > > > > 
> > > > > I prefer to share the cache for all spaces. Otherwise, it will have trouble with memory control.
> > > > 
> > > > 
> > > > OK. Could you try it in your env? F14Map is also an implementation of cuckoo Hash with vector instructions used to accelerate the scan inside one bucket. I think the performance should be same with yours in #2264
> > > 
> > > 
> > > As we may change the space for test frequently. It seems the cache may has no effect if space changed.
> > 
> > 
> > Em... If so, how do you test #2264
> 
> The cache in #2264 will try to reclaim the slots that contain expired data. If space changed, new data will replace the data never access.

Em..... I see.

dangleptr:
Close it.  #2264  is a better choice with TTL feature. Let's go back there.

PR #2273, opened 2020-08-05
Remove lock in FunctionManager

xuguruogu:
atomic operation acts like dead lock, and can not go out of the while loop if multi-thread access it frequently.

image

### DIFF
--- a/src/common/filter/FunctionManager.h
+++ b/src/common/filter/FunctionManager.h
@@ -60,7 +60,6 @@ private:
         Function                body_;
     };
 
-    mutable folly::RWSpinLock                           lock_;
     std::unordered_map<std::string, FunctionAttributes> functions_;
 };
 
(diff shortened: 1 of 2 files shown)


dutor:
But given your post of `perf top`, I am curious about why `FunctionManager::instance` is so hot. Except for the first several invokes, it does not involve any atomic or memory barrier thing.

xuguruogu:
> But given your post of `perf top`, I am curious about why `FunctionManager::instance` is so hot. Except for the first several invokes, it does not involve any atomic or memory barrier thing.

call function as filter with a little part of data returned.

src/common/filter/FunctionManager.h
```diff
@@ -60,7 +60,6 @@ class FunctionManager final {
         Function                body_;
     };
 
-    mutable folly::RWSpinLock                           lock_;
     std::unordered_map<std::string, FunctionAttributes> functions_;
```
Shylock-Hg, line 63: I think you better mark it const.

xuguruogu: This may not help due to ref counter use atomic operation.

PR #2274, opened 2020-08-06
Add prefix bloom filter support

chenxu14:
Currently, The DataModel of Vertex and Edge are as follows

(1)Vertex: **type(1byte)_PartID(3bytes)_VertexID(8bytes)**_TagID(4bytes)_Timestamp(8bytes)
(2)Edge: **type(1byte)_PartID(3bytes)_VertexID(8bytes)**_EdgeType(4bytes)_Rank(8bytes)_otherVertexId(8bytes)_Timestamp(8bytes)

We can use “**type(1byte)_PartID(3bytes)_VertexID(8bytes)**” as the value for prefix bloom filter, QueryEdgePropsProcessor and QueryVertexPropsProcessor can benefit from this a lot

### DIFF
--- a/etc/nebula-storaged.conf.default
+++ b/etc/nebula-storaged.conf.default
@@ -76,6 +76,14 @@
 #   * kAll, Collect all stats
 --rocksdb_stats_level=kExceptHistogramOrTimers
 
+# Whether or not to enable rocksdb's prefix bloom filter, disabled by default.
+--enable_prefix_filtering=false
+# Whether or not to enable the whole key filtering.
+--enable_whole_key_filtering=true
+# The prefix length for each key to use as the filter value.
+# can be 12 bytes(PartitionId + VertexID), or 16 bytes(PartitionId + VertexID + TagID/EdgeType).
+--prefix_length=12
+
 ############## rocksdb Options ##############
 # rocksdb DBOptions in json, each name and value of option is a string, given as "option_name":"option_value" separated by comma
 --rocksdb_db_options={"max_subcompactions":"1","max_background_jobs":"1"}
--- a/etc/nebula-storaged.conf.default
+++ b/etc/nebula-storaged.conf.default
@@ -77,12 +77,12 @@
 --rocksdb_stats_level=kExceptHistogramOrTimers
 
 # Whether or not to enable rocksdb's prefix bloom filter, disabled by default.
---enable_prefix_filtering=false
+--enable_rocksdb_prefix_filtering=false
 # Whether or not to enable the whole key filtering.
---enable_whole_key_filtering=true
+--enable_rocksdb_whole_key_filtering=true
 # The prefix length for each key to use as the filter value.
 # can be 12 bytes(PartitionId + VertexID), or 16 bytes(PartitionId + VertexID + TagID/EdgeType).
---prefix_length=12
+--rocksdb_filtering_prefix_length=12
 
 ############## rocksdb Options ##############
 # rocksdb DBOptions in json, each name and value of option is a string, given as "option_name":"option_value" separated by comma
--- a/src/kvstore/RocksEngine.cpp
+++ b/src/kvstore/RocksEngine.cpp
@@ -192,6 +192,7 @@ ResultCode RocksEngine::range(const std::string& start,
                               const std::string& end,
                               std::unique_ptr<KVIterator>* storageIter) {
     rocksdb::ReadOptions options;
+    options.total_order_seek = true;
     rocksdb::Iterator* iter = db_->NewIterator(options);
     if (iter) {
         iter->Seek(rocksdb::Slice(start));
@@ -204,6 +205,7 @@ ResultCode RocksEngine::range(const std::string& start,
 ResultCode RocksEngine::prefix(const std::string& prefix,
                                std::unique_ptr<KVIterator>* storageIter) {
     rocksdb::ReadOptions options;
+    options.prefix_same_as_start = true;
     rocksdb::Iterator* iter = db_->NewIterator(options);
     if (iter) {
         iter->Seek(rocksdb::Slice(prefix));
@@ -217,6 +219,7 @@ ResultCode RocksEngine::rangeWithPrefix(const std::string& start,
                                         const std::string& prefix,
                                         std::unique_ptr<KVIterator>* storageIter) {
     rocksdb::ReadOptions options;
+    options.prefix_same_as_start = true;
     rocksdb::Iterator* iter = db_->NewIterator(options);
     if (iter) {
         iter->Seek(rocksdb::Slice(start));
--- a/src/kvstore/RocksEngineConfig.cpp
+++ b/src/kvstore/RocksEngineConfig.cpp
@@ -72,11 +72,11 @@ DEFINE_int32(num_compaction_threads, 0,
 DEFINE_int32(rate_limit, 0,
             "write limit in bytes per sec. The unit is MB. 0 means unlimited.");
 
-DEFINE_bool(enable_prefix_filtering, false,
+DEFINE_bool(enable_rocksdb_prefix_filtering, false,
             "Whether or not to enable rocksdb's prefix bloom filter.");
-DEFINE_bool(enable_whole_key_filtering, true,
+DEFINE_bool(enable_rocksdb_whole_key_filtering, true,
             "Whether or not to enable the whole key filtering.");
-DEFINE_int32(prefix_length, 12,
+DEFINE_int32(rocksdb_filtering_prefix_length, 12,
             "The prefix length, default value is 12 bytes(PartitionID+VertexID).");
 
 namespace nebula {
@@ -224,10 +224,11 @@ rocksdb::Status initRocksdbOptions(rocksdb::Options &baseOpts) {
         bbtOpts.pin_l0_filter_and_index_blocks_in_cache =
             baseOpts.compaction_style == rocksdb::CompactionStyle::kCompactionStyleLevel;
     }
-    if (FLAGS_enable_prefix_filtering) {
-        baseOpts.prefix_extractor.reset(new GraphPrefixTransform(FLAGS_prefix_length));
+    if (FLAGS_enable_rocksdb_prefix_filtering) {
+        baseOpts.prefix_extractor.reset(
+                new GraphPrefixTransform(FLAGS_rocksdb_filtering_prefix_length));
     }
-    bbtOpts.whole_key_filtering = FLAGS_enable_whole_key_filtering;
+    bbtOpts.whole_key_filtering = FLAGS_enable_rocksdb_whole_key_filtering;
     baseOpts.table_factory.reset(NewBlockBasedTableFactory(bbtOpts));
     baseOpts.create_if_missing = true;
     return s;
--- a/src/kvstore/RocksEngineConfig.cpp
+++ b/src/kvstore/RocksEngineConfig.cpp
@@ -84,22 +84,22 @@ namespace kvstore {
 
 class GraphPrefixTransform : public rocksdb::SliceTransform {
 private:
-    size_t prefix_len_;
+    size_t prefixLen_;
     std::string name_;
 
 public:
-    explicit GraphPrefixTransform(size_t prefix_len)
-        : prefix_len_(prefix_len),
-        name_("nebula.GraphPrefix." + std::to_string(prefix_len_)) {}
+    explicit GraphPrefixTransform(size_t prefixLen)
+        : prefixLen_(prefixLen),
+        name_("nebula.GraphPrefix." + std::to_string(prefixLen_)) {}
 
     const char* Name() const override { return name_.c_str(); }
 
     rocksdb::Slice Transform(const rocksdb::Slice& src) const override {
-        return rocksdb::Slice(src.data(), prefix_len_);
+        return rocksdb::Slice(src.data(), prefixLen_);
     }
 
     bool InDomain(const rocksdb::Slice& src) const override {
-        return src.size() >= prefix_len_ && NebulaKeyUtils::isDataKey(
+        return src.size() >= prefixLen_ && NebulaKeyUtils::isDataKey(
                 folly::StringPiece(src.data(), 1));
     }
 };
--- a/src/kvstore/RocksEngineConfig.cpp
+++ b/src/kvstore/RocksEngineConfig.cpp
@@ -99,7 +99,7 @@ public:
     }
 
     bool InDomain(const rocksdb::Slice& src) const override {
-    	return src.size() >= prefix_len_ && NebulaKeyUtils::isDataKey(
+        return src.size() >= prefix_len_ && NebulaKeyUtils::isDataKey(
                 folly::StringPiece(src.data(), 1));
     }
 };
--- a/src/kvstore/RocksEngineConfig.cpp
+++ b/src/kvstore/RocksEngineConfig.cpp
@@ -16,6 +16,7 @@
 #include "rocksdb/concurrent_task_limiter.h"
 #include "rocksdb/rate_limiter.h"
 #include "base/Configuration.h"
+#include "utils/NebulaKeyUtils.h"
 
 // [WAL]
 DEFINE_bool(rocksdb_disable_wal,
@@ -71,8 +72,37 @@ DEFINE_int32(num_compaction_threads, 0,
 DEFINE_int32(rate_limit, 0,
             "write limit in bytes per sec. The unit is MB. 0 means unlimited.");
 
+DEFINE_bool(enable_prefix_filtering, false,
+            "Whether or not to enable rocksdb's prefix bloom filter.");
+DEFINE_bool(enable_whole_key_filtering, true,
+            "Whether or not to enable the whole key filtering.");
+DEFINE_int32(prefix_length, 12,
+            "The prefix length, default value is 12 bytes(PartitionID+VertexID).");
+
 namespace nebula {
 namespace kvstore {
+
+class GraphPrefixTransform : public rocksdb::SliceTransform {
+private:
+    size_t prefix_len_;
+    std::string name_;
+
+public:
+    explicit GraphPrefixTransform(size_t prefix_len)
+        : prefix_len_(prefix_len),
+        name_("nebula.GraphPrefix." + std::to_string(prefix_len_)) {}
+
+    const char* Name() const override { return name_.c_str(); }
+
+    rocksdb::Slice Transform(const rocksdb::Slice& src) const override {
+        return rocksdb::Slice(src.data(), prefix_len_);
+    }
+
+    bool InDomain(const rocksdb::Slice& src) const override {
+    	return src.size() >= prefix_len_ && NebulaKeyUtils::isDataKey(
+                folly::StringPiece(src.data(), 1));
+    }
+};
 
 static rocksdb::Status initRocksdbCompression(rocksdb::Options &baseOpts) {
     static std::unordered_map<std::string, rocksdb::CompressionType> m = {
@@ -194,6 +224,10 @@ rocksdb::Status initRocksdbOptions(rocksdb::Options &baseOpts) {
         bbtOpts.pin_l0_filter_and_index_blocks_in_cache =
             baseOpts.compaction_style == rocksdb::CompactionStyle::kCompactionStyleLevel;
     }
+    if (FLAGS_enable_prefix_filtering) {
+        baseOpts.prefix_extractor.reset(new GraphPrefixTransform(FLAGS_prefix_length));
+    }
+    bbtOpts.whole_key_filtering = FLAGS_enable_whole_key_filtering;
     baseOpts.table_factory.reset(NewBlockBasedTableFactory(bbtOpts));
     baseOpts.create_if_missing = true;
     return s;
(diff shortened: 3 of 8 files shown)


src/kvstore/RocksEngineConfig.cpp
```diff
@@ -196,6 +199,10 @@ rocksdb::Status initRocksdbOptions(rocksdb::Options &baseOpts) {
     }
     baseOpts.table_factory.reset(NewBlockBasedTableFactory(bbtOpts));
     baseOpts.create_if_missing = true;
+    if (FLAGS_enable_prefix_bloom_filter) {
+        baseOpts.prefix_extractor.reset(rocksdb::NewFixedPrefixTransform(
```
dangleptr, line 203: I have tried this way, the problem we need to fix is how to deal with the system keys inside rocksdb.  The format is not with "partId + vertexId" as its prefix.

chenxu14:
Maybe there is a need to override SliceTransform#InDomain to judge if the target key is NebulaKeyType::kData?

dangleptr:
> Maybe there is a need to override SliceTransform#InDomain to judge if the target key is NebulaKeyType::kData?

Yes, we need to take care of it.  Another way is to pad the system key to at least 12 bytes.

src/kvstore/RocksEngine.cpp
```diff
@@ -153,6 +154,7 @@ ResultCode RocksEngine::get(const std::string& key, std::string* value) {
 std::vector<Status> RocksEngine::multiGet(const std::vector<std::string>& keys,
                                           std::vector<std::string>* values) {
     rocksdb::ReadOptions options;
+    options.total_order_seek = true;
```
xuguruogu, line 157: total_order_seek with prefix bloom filter will result in huge performance issue.

src/kvstore/RocksEngineConfig.cpp
```diff
@@ -196,6 +199,10 @@ rocksdb::Status initRocksdbOptions(rocksdb::Options &baseOpts) {
...
+                sizeof(PartitionID) + sizeof(VertexID)));
```
xuguruogu, line 204: I prefer to use rocksdb.FixedPrefix.16, for most of the case prefix is used by partid+vid+typeid.

src/kvstore/RocksEngine.cpp
chenxu14: Code in BlockBasedTable::NewIterator
```
    return new BlockBasedTableIterator<DataBlockIter>(
        this, read_options, rep_->internal_comparator,
        NewIndexIterator(
            read_options,
            need_upper_bound_check &&
                rep_->index_type == BlockBasedTableOptions::kHashSearch,
            /*input_iter=*/nullptr, /*get_context=*/nullptr, &lookup_context),
        !skip_filters && !read_options.total_order_seek &&
            prefix_extractor != nullptr,
        need_upper_bound_check, prefix_extractor, BlockType::kData, caller,
        compaction_readahead_size);
```
According to code，if total_order_seek = true, means we don't consider prefix bloom filter, so there won't be any performance issue?

critical27: IIRC, total_order_seek need to be true when we use prefix bloom filter? Of course, we can set it according to `FLAGS_enable_prefix_bloom_filter`

src/kvstore/RocksEngineConfig.cpp
chenxu14:
```
kvstore::ResultCode QueryVertexPropsProcessor::collectVertexProps(
                            PartitionID partId,
                            VertexID vId,
                            std::vector<cpp2::TagData> &tds) {
    auto prefix = NebulaKeyUtils::vertexPrefix(partId, vId);
```
If the prefix is 16 bytes, QueryVertexPropsProcessor can't use the filter? since it use partId+vId as the seek key

dangleptr: Exactly, we support scan all tags for one vertex.

src/kvstore/RocksEngine.cpp
dangleptr: If we implement custom Transformer,  we could set it to be false.

chenxu14: The whole key filters for Get() and MultiGet is enough, so it's better to set it to be true to disable the prefix filter?

dangleptr:
Another point we need to consider is the balance procedure. 
It will scan the data with prefix "partId".

src/kvstore/RocksEngine.cpp
dangleptr: Make sense.

chenxu14:
> Another point we need to consider is the balance procedure.
> It will scan the data with prefix "partId".

Since rocksdb can only specify one type of prefix bloom filter, and partId + vertexId are more useful, can be ignore the balance case?

src/kvstore/RocksEngineConfig.cpp
xuguruogu: QueryVertexPropsProcessor::collectVertexProps is only used in `fetch prop on <ids> <no yield>`. For most of case, it is used for debug and test.

dangleptr:
> > Another point we need to consider is the balance procedure.
> > It will scan the data with prefix "partId".
> 
> Since rocksdb can only specify one type of prefix bloom filter, and partId + vertexId are more useful, can be ignore the balance case?

Yes,  for balance,  we should skip the filters.

src/kvstore/RocksEngineConfig.cpp
dangleptr: We can't have this assumption

chenxu14: If the query condition is (partid+vid+typeid), it can also use (partid+vid) as the prefix filter. do you think 12bytes is good enough? @xuguruogu

src/kvstore/RocksEngine.cpp
xuguruogu: So can block-based table assure all data in order when prefix bloom filter enabled?

chenxu14: Rocksdb can guaranteed a total order of keys for the same prefix.

src/kvstore/RocksEngineConfig.cpp
xuguruogu: > If the query condition is (partid+vid+typeid), it can also use (partid+vid) as the prefix filter. do you think 12bytes is good enough? @xuguruogu

This miss rate may be high if multi edge/tag type associated with the vid. So why you introduce the prefix bloom filter.

chenxu14: IMO, (partid+vid) is good enough, or do you think there is a need to make this configuable

src/kvstore/RocksEngine.cpp
xuguruogu: According the docs of rocksdb, total_order_seek will ensure the behavior the same as no prefix bloom filter, by ordering the index. Benchmark may be required for all prefix operation with total_order_seek on.

chenxu14:
Add an custome SliceTransform to judge key's type. @dangleptr

src/kvstore/RocksEngineConfig.cpp
```diff
@@ -71,9 +72,40 @@ DEFINE_int32(num_compaction_threads, 0,
...
+        name_("nebula.GraphPrefix") {}
+
+    const char* Name() const override { return name_.c_str(); }
+
+    rocksdb::Slice Transform(const rocksdb::Slice& src) const override {
+        return rocksdb::Slice(src.data(), prefix_len_);
+    }
+
+    bool InDomain(const rocksdb::Slice& src) const override {
+        if (src.size() >= prefix_len_) {
```
dangleptr, line 100: You could just check the first byte inside key.

dangleptr:
Could you give a benchmark compare to w/o prefix_extractor?  I think it is necessary

chenxu14:
> Could you give a benchmark compare to w/o prefix_extractor? I think it is necessary

OK, let me try, do you think GetNeighborsBenchmark is OK?

dangleptr:
> > Could you give a benchmark compare to w/o prefix_extractor? I think it is necessary
> 
> OK, let me try, do you think GetNeighborsBenchmark is OK?

I think it is not suitable. Because the bloom filter is mostly used when iterator->seek.  
So our focus is to test seek() not next().  WDYT?

chenxu14:
OK, Let me improve this.

chenxu14:
Add PrefixBloomFilterBenchmark

src/kvstore/RocksEngineConfig.cpp
```diff
@@ -196,6 +224,9 @@ rocksdb::Status initRocksdbOptions(rocksdb::Options &baseOpts) {
     }
     baseOpts.table_factory.reset(NewBlockBasedTableFactory(bbtOpts));
     baseOpts.create_if_missing = true;
+    if (FLAGS_enable_prefix_bloom_filter) {
+        baseOpts.prefix_extractor.reset(new GraphPrefixTransform(FLAGS_prefix_length));
```
darionyaphet, line 228: Do you need to check the `FLAGS_prefix_length` is the specified size ? Maybe setting it as 12 or 16 is OK ？

xuguruogu: I think it is ok for users to config prefix length. As my data set is so big, I close whole key filters by default, all requests rely on prefix bloom filter heavily.

xuguruogu:
> Add PrefixBloomFilterBenchmark

What is the difference

chenxu14:
> What is the difference

sorry, I have force-pushed my branch,since there are so many changes, the core diff is:
(1) Add PrefixBloomFilterBenchmark.cpp to do prefix bloom filter benchmark test.
(2) Refactor GraphPrefixTransform#InDomain, use one bytes to judge the key type.

src/kvstore/RocksEngineConfig.cpp
chenxu14: > As my data set is so big, I close whole key filters by default.

Maybe we should make this as the default behavior? According to **BlockBasedTable::FullFilterKeyMayMatch**, prefix bloom filer seems to be enough for the nebula.

xuguruogu: Agree, prefix bloom filter is more useful than full key filter, with much less memory consumption.

dangleptr:
One extra question,  how to ensure the order for the index data.

chenxu14:
> One extra question, how to ensure the order for the index data.

For index data，prefix bloom filter will no be use? since we judge this in GraphPrefixTransform#InDomain.
So the behavior is like before?

chenxu14:
Oh, there are some conflicts now , will do a rebase later.

dangleptr:
> > One extra question, how to ensure the order for the index data.
> 
> For index data，prefix bloom filter will no be use? since we judge this in GraphPrefixTransform#InDomain.
> So the behavior is like before?

we should set total_order_seek to be true when index prefix?

chenxu14:
> > > One extra question, how to ensure the order for the index data.
> > 
> > 
> > For index data，prefix bloom filter will no be use? since we judge this in GraphPrefixTransform#InDomain.
> > So the behavior is like before?
> 
> we should set total_order_seek to be true when index prefix scan?

Sorry，I don't quite understand，please correct me if I am wrong.
The rowkey of index datas will prefix with NebulaKeyType#kIndex, So GraphPrefixTransform#InDomain will always return false.

When quering index data, rocksdb’s BlockBasedTable::PrefixMayMatch will triger.
```
bool BlockBasedTable::PrefixMayMatch(
  …
  if (!prefix_extractor->InDomain(user_key)) {
    return true;
  }
  …
}
```
According to the code, PrefixMayMatch will return true, so the prefix bloom filter are not considered at all.
So the results will not be affected no matter if total_order_seek is true?

dangleptr:
> PrefixMayMatch

Sorry,  maybe i misunderstand the behavior of prefix seek.  

As your explain,  if the key is not in domain,  the seek will fall back to the original binary search for the correct block id, right?

chenxu14:
> > PrefixMayMatch
> 
> Sorry, maybe i misunderstand the behavior of prefix seek.
> 
> As your explain, if the key is not in domain, the seek will fall back to the original binary search for the correct block id, right?

yes, that is

src/kvstore/RocksEngine.cpp
```diff
@@ -204,6 +205,7 @@ ResultCode RocksEngine::range(const std::string& start,
 ResultCode RocksEngine::prefix(const std::string& prefix,
                                std::unique_ptr<KVIterator>* storageIter) {
     rocksdb::ReadOptions options;
+    options.prefix_same_as_start = true;
```
darionyaphet, line 208: According to the explain of `prefix_same_as_start` 

```
This option is effective only for prefix seeks, i.e. prefix_extractor is non-null for the column family and total_order_seek is false.
```

src/kvstore/RocksEngine.cpp
```diff
@@ -192,6 +192,7 @@ ResultCode RocksEngine::range(const std::string& start,
                               const std::string& end,
                               std::unique_ptr<KVIterator>* storageIter) {
     rocksdb::ReadOptions options;
+    options.total_order_seek = true;
```
darionyaphet, line 195: why setting `total_order_seek ` at here?

src/kvstore/RocksEngine.cpp
chenxu14: Hi, you mean we should first judge whether prefix is null？rocksdb already judge this.
Code in db/db_iter.cc like
```
DBIter::DBIter(Env* _env, const ReadOptions& read_options,
      ...
      prefix_same_as_start_(mutable_cf_options.prefix_extractor
                                ? read_options.prefix_same_as_start
                                : false),
```

src/storage/test/PrefixBloomFilterBenchmark.cpp
```diff
@@ -0,0 +1,119 @@
...
+
+void testPrefixSeek(kvstore::KVStore *kv, int32_t iters) {
+    for (decltype(iters) i = 0; i < iters; i++) {
+        for (PartitionID partId = 0; partId < FLAGS_part_number; partId++) {
+            for (int32_t vertexId = partId * FLAGS_vertex_per_part;
+                    vertexId < (partId + 1) * FLAGS_vertex_per_part; vertexId++) {
+                auto prefix = NebulaKeyUtils::vertexPrefix(partId, vertexId);
+                std::unique_ptr<kvstore::KVIterator> iter;
+                kv->prefix(0, partId, prefix, &iter);
+                iter->next();
```
critical27, line 52: Would you post the benchmark without next? I am curious about how much we benefit from it.

src/kvstore/RocksEngine.cpp
```diff
@@ -217,6 +219,7 @@ ResultCode RocksEngine::rangeWithPrefix(const std::string& start,
                                         const std::string& prefix,
                                         std::unique_ptr<KVIterator>* storageIter) {
     rocksdb::ReadOptions options;
+    options.prefix_same_as_start = true;
```
bright-starry-sky, line 222: A question，if set options.prefix_same_as_start = true in this method, Is the check as below redundant of `RocksPrefixIter`?

```
bool valid() const override {
        return !!iter_ && iter_->Valid() && (iter_->key().starts_with(prefix_));
    }
```

src/storage/test/PrefixBloomFilterBenchmark.cpp
chenxu14: let me try

src/kvstore/RocksEngine.cpp
chenxu14: I think so, let me dig it more

src/kvstore/RocksEngine.cpp
darionyaphet: When you setting `prefix_same_as_start` as `true`, maybe you should make sure `prefix_extractor` is not null and `total_order_seek` is false ?

src/storage/test/PrefixBloomFilterBenchmark.cpp
chenxu14: image

src/kvstore/RocksEngine.cpp
chenxu14: Hi, **total_order_seek** is false by default. And as the code above, if prefix_extractor is null, **prefix_same_as_start_** will be false, no matter what **prefix_same_as_start** is.

src/kvstore/RocksEngine.cpp
chenxu14: Seems prefix_same_as_start can save one Next operation, so we’d better keeping it here?

bright-starry-sky: > Seems prefix_same_as_start can save one Next operation, so we’d better keeping it here?

I agree，good that.

xuguruogu: Can ScanProcessor work well with prefix_same_as_start on.

chenxu14: Hi, ScanProcessor will use RocksEngine::range instead of RocksEngine::rangeWithPrefix？

xuguruogu:
why not use prefix bloom filter as default？ it is so rare to access the key with version to actually use whole key filter.

chenxu14:
> why not use prefix bloom filter as default？ it is so rare to access the key with version to actually use whole key filter.

How do you think about this? @dangleptr @darionyaphet @critical27 @bright-starry-sky  Shall we enable this by default? Means enable_prefix_filtering=true and enable_whole_key_filtering=false.
What I concern is the existing bloom filters will not work until compaction happens, so let the user decide whether to turn it on or not.

dangleptr:
Agreed with @chenxu14 ,   we'd better not change any behavior in the coming version. 

Maybe we could do it in the next release one.

src/kvstore/RocksEngineConfig.cpp
```diff
@@ -71,9 +72,38 @@ DEFINE_int32(num_compaction_threads, 0,
...
+            "Whether or not to enable the whole key filtering.");
+DEFINE_int32(prefix_length, 12,
+            "The prefix length, default value is 12 bytes(PartitionID+VertexID).");
+
 namespace nebula {
 namespace kvstore {
 
+class GraphPrefixTransform : public rocksdb::SliceTransform {
+private:
+    size_t prefix_len_;
```
panda-sheep, line 87: Please pay attention to the way the variables are named, keep it consistent with our code, thx.

src/kvstore/RocksEngineConfig.h
```diff
@@ -43,6 +43,8 @@ DECLARE_string(rocksdb_compression);
 DECLARE_bool(enable_rocksdb_statistics);
 DECLARE_string(rocksdb_stats_level);
 
+DECLARE_bool(enable_prefix_filtering);
+
```
panda-sheep, line 47: why not declare `enable_whole_key_filtering`  and `prefix_length`?

src/storage/test/QueryVertexPropsTest.cpp
```diff
@@ -48,6 +49,8 @@ void mockData(kvstore::KVStore* kv,
                 baton.post();
             });
         baton.wait();
+        kvstore::ResultCode code = kv->flush(0);  // flush per partition
```
panda-sheep, line 52: What is the reason for flush here?

chenxu14: We should generate sstable filte to test prefix bloom filter take effect. @panda-sheep

src/kvstore/RocksEngineConfig.cpp
chenxu14: Changed to prefixLen

liuyu85cn:
It's OK for 1.0 implement, but in NB 2.0, different space may have different id length, if this length can be set by space, I think it will be better.

chenxu14:
> It's OK for 1.0 implement, but in NB 2.0, different space may have different id length, if this length can be set by space, I think it will be better.

Sorry, I don't quit understand, where is the 2.0 branch?

liuyu85cn:
> > It's OK for 1.0 implement, but in NB 2.0, different space may have different id length, if this length can be set by space, I think it will be better.
> 
> Sorry, I don't quit understand, where is the 2.0 branch?

just release to public yesterday~

https://github.com/vesoft-inc/nebula-storage/

still in progress.

etc/nebula-storaged.conf.default
```diff
@@ -76,6 +76,14 @@
 #   * kAll, Collect all stats
 --rocksdb_stats_level=kExceptHistogramOrTimers
 
+# Whether or not to enable rocksdb's prefix bloom filter, disabled by default.
+--enable_prefix_filtering=false
```
Shylock-Hg, line 80: For the option name format, pls start with `rocksdb_` like others.

chenxu14: Since HBase also support prefix-bloom-filter, I think we'd better make it as a general option, WDYT?

darionyaphet: If so maybe you should remove the `rocksdb` from comment

Shylock-Hg: Make sense. But we don't plan to support hbase again.

chenxu14: Changed to enable_rocksdb_prefix_filtering

dangleptr (approved):
LGTM

darionyaphet (approved):

Issue #2269, opened 2020-08-04
"graph_update_vertex_qps" is equal to "graph_update_vertex_error_qps"


GestureWei:
In `UpdateVertexExecutor.cpp`, line 71 to 73
`if (status.ok()) {
        stats::Stats::addStatsValue(stats_.get(), false, duration().elapsedInUSec());
    }
`
error_qps should be added when status is not ok. This leads to "graph_update_vertex_qps" is equal to "graph_update_vertex_error_qps".

PR #2275, opened 2020-08-06
fix bugs for "graph_update_vertex_error_qps"

GestureWei:
Statistics of `graph_update_vertex_error_qps` is incorrect. This stats should be added when status is NOT ok. #2269


chenxu14:
nice find.

PR #2276, opened 2020-08-06
raft lease is always valid when replica is one

panda-sheep:
Synchronize storage 2.0 pr 69


xuguruogu:
So, may be cannot accessed when replica set is 1

dangleptr:
> So, may be cannot accessed when replica set is 1

Sorry? I can't follow you.  If replica is 1,  the leader will always be leader.

xuguruogu:
> > So, may be cannot accessed when replica set is 1
> 
> Sorry? I can't follow you. If replica is 1, the leader will always be leader.

I encountered a bug before, request may be rejected under 1 replica set num, with the error of change leader.

PR #2277, opened 2020-08-07
To improve index scan logic since string range scan disabled.

bright-starry-sky:
Disable range scan for string type index. 
Disable contains relational expression in lookup.

### DIFF
--- a/src/common/utils/NebulaKeyUtils.h
+++ b/src/common/utils/NebulaKeyUtils.h
@@ -36,6 +36,13 @@ enum class NebulaSystemKeyType : uint32_t {
     kSystemPart        = 0x00000002,
 };
 
+enum class NebulaBoundValueType : uint8_t {
+    kMax                = 0x0001,
+    kMin                = 0x0002,
+    kSubtraction        = 0x0003,
+    kAddition           = 0x0004,
+};
+
 /**
  * This class supply some utils for transition between Vertex/Edge and key in kvstore.
  * */
@@ -246,6 +253,108 @@ public:
     static folly::StringPiece keyWithNoVersion(const folly::StringPiece& rawKey) {
         // TODO(heng) We should change the method if varint data version supportted.
         return rawKey.subpiece(0, rawKey.size() - sizeof(int64_t));
+    }
+
+    // Only int and double are supported
+    static std::string boundVariant(nebula::cpp2::SupportedType type,
+                                    NebulaBoundValueType op,
+                                    const VariantType& v = 0L) {
+        switch (op) {
+            case NebulaBoundValueType::kMax : {
+                std::vector<unsigned char> bytes = {0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF};
+                return std::string(bytes.begin(), bytes.end());
+            }
+            case NebulaBoundValueType::kMin : {
+                std::string str;
+                str.reserve(sizeof(int64_t));
+                str.append(sizeof(int64_t), '\0');
+                return str;
+            }
+            case NebulaBoundValueType::kSubtraction : {
+                std::string str;
+                if (type == nebula::cpp2::SupportedType::INT) {
+                    str = encodeInt64(boost::get<int64_t>(v));
+                } else {
+                    str = encodeDouble(boost::get<double>(v));
+                }
+                std::vector<unsigned char> bytes(str.begin(), str.end());
+                for (size_t i = bytes.size(); i > 0; i--) {
+                    if (bytes[i-1] > 0) {
+                        bytes[i-1] -= 1;
+                        break;
+                    }
+                }
+                return std::string(bytes.begin(), bytes.end());
+            }
+            case NebulaBoundValueType::kAddition : {
+                std::string str;
+                if (type == nebula::cpp2::SupportedType::INT) {
+                    str = encodeInt64(boost::get<int64_t>(v));
+                } else {
+                    str = encodeDouble(boost::get<double>(v));
+                }
+                std::vector<unsigned char> bytes(str.begin(), str.end());
+                for (size_t i = bytes.size(); i > 0; i--) {
+                    if (bytes[i-1] < 255) {
+                        bytes[i-1] += 1;
+                        break;
+                    }
+                }
+                return std::string(bytes.begin(), bytes.end());
+            }
+        }
+        return "";
+    }
+
+    static bool checkAndCastVariant(nebula::cpp2::SupportedType sType,
+                                    VariantType& v) {
+        nebula::cpp2::SupportedType type = nebula::cpp2::SupportedType::UNKNOWN;
+        switch (v.which()) {
+            case VAR_INT64: {
+                type = nebula::cpp2::SupportedType::INT;
+                break;
+            }
+            case VAR_DOUBLE: {
+                type = nebula::cpp2::SupportedType::DOUBLE;
+                break;
+            }
+            case VAR_BOOL: {
+                type = nebula::cpp2::SupportedType::BOOL;
+                break;
+            }
+            case VAR_STR: {
+                type = nebula::cpp2::SupportedType::STRING;
+                break;
+            }
+            default:
+                return false;
+        }
+        if (sType != type) {
+            switch (sType) {
+                case nebula::cpp2::SupportedType::INT:
+                case nebula::cpp2::SupportedType::TIMESTAMP: {
+                    v = Expression::toInt(v);
+                    break;
+                }
+                case nebula::cpp2::SupportedType::BOOL: {
+                    v = Expression::toBool(v);
+                    break;
+                }
+                case nebula::cpp2::SupportedType::FLOAT:
+                case nebula::cpp2::SupportedType::DOUBLE: {
+                    v = Expression::toDouble(v);
+                    break;
+                }
+                case nebula::cpp2::SupportedType::STRING: {
+                    v = Expression::toString(v);
+                    break;
+                }
+                default: {
+                    return false;
+                }
+            }
+        }
+        return true;
     }
 
     static std::string encodeVariant(const VariantType& v)  {
--- a/src/common/utils/NebulaKeyUtils.h
+++ b/src/common/utils/NebulaKeyUtils.h
@@ -261,14 +261,10 @@ public:
                                     const VariantType& v = 0L) {
         switch (op) {
             case NebulaBoundValueType::kMax : {
-                std::vector<unsigned char> bytes = {0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF};
-                return std::string(bytes.begin(), bytes.end());
+                return std::string(8, '\377');
             }
             case NebulaBoundValueType::kMin : {
-                std::string str;
-                str.reserve(sizeof(int64_t));
-                str.append(sizeof(int64_t), '\0');
-                return str;
+                return std::string(8, '\0');
             }
             case NebulaBoundValueType::kSubtraction : {
                 std::string str;
@@ -278,10 +274,12 @@ public:
                     str = encodeDouble(boost::get<double>(v));
                 }
                 std::vector<unsigned char> bytes(str.begin(), str.end());
-                for (size_t i = bytes.size(); i > 0; i--) {
-                    if (bytes[i-1] > 0) {
-                        bytes[i-1] -= 1;
-                        break;
+
+                for (size_t i = bytes.size();; i--) {
+                    if (i > 0) {
+                        if (bytes[i-1]-- != 0) break;
+                    } else {
+                        return std::string(bytes.size(), '\0');
                     }
                 }
                 return std::string(bytes.begin(), bytes.end());
@@ -294,10 +292,11 @@ public:
                     str = encodeDouble(boost::get<double>(v));
                 }
                 std::vector<unsigned char> bytes(str.begin(), str.end());
-                for (size_t i = bytes.size(); i > 0; i--) {
-                    if (bytes[i-1] < 255) {
-                        bytes[i-1] += 1;
-                        break;
+                for (size_t i = bytes.size();; i--) {
+                    if (i > 0) {
+                        if (bytes[i-1]++ != 255) break;
+                    } else {
+                        return std::string(bytes.size(), '\377');
                     }
                 }
                 return std::string(bytes.begin(), bytes.end());
--- a/src/graph/LookupExecutor.cpp
+++ b/src/graph/LookupExecutor.cpp
@@ -5,6 +5,7 @@
  */
 
 #include "graph/LookupExecutor.h"
+#include <interface/gen-cpp2/common_types.h>
 
 namespace nebula {
 namespace graph {
@@ -170,23 +171,37 @@ Status LookupExecutor::optimize() {
     return status;
 }
 
-Status LookupExecutor::traversalExpr(const Expression *expr) {
+Status LookupExecutor::traversalExpr(const Expression *expr, const meta::SchemaProviderIf* schema) {
     switch (expr->kind()) {
         case nebula::Expression::kLogical : {
+            Status ret = Status::OK();
             auto* lExpr = dynamic_cast<const LogicalExpression*>(expr);
-            if (lExpr->op() == LogicalExpression::Operator::XOR) {
-                return Status::SyntaxError("Syntax error : %s", lExpr->toString().c_str());
+            if (lExpr->op() == LogicalExpression::Operator::XOR ||
+                lExpr->op() == LogicalExpression::Operator::OR) {
+                return Status::SyntaxError("OR and XOR are not supported "
+                                           "in lookup where clause ：%s",
+                                           lExpr->toString().c_str());
             }
             auto* left = lExpr->left();
-            traversalExpr(left);
+            ret = traversalExpr(left, schema);
+            if (!ret.ok()) {
+                return ret;
+            }
             auto* right = lExpr->right();
-            traversalExpr(right);
+            ret = traversalExpr(right, schema);
+            if (!ret.ok()) {
+                return ret;
+            }
             break;
         }
         case nebula::Expression::kRelational : {
             std::string prop;
             VariantType v;
             auto* rExpr = dynamic_cast<const RelationalExpression*>(expr);
+            auto ret = relationalExprCheck(rExpr->op());
+            if (!ret.ok()) {
+                return ret;
+            }
             auto* left = rExpr->left();
             auto* right = rExpr->right();
             /**
@@ -212,15 +227,18 @@ Status LookupExecutor::traversalExpr(const Expression *expr) {
                 return Status::SyntaxError("Unsupported expression ：%s",
                                            rExpr->toString().c_str());
             }
+
+            if (rExpr->op() != RelationalExpression::Operator::EQ) {
+                auto type = schema->getFieldType(prop).type;
+                if (!dataTypeCheckForRange(type)) {
+                    return Status::SyntaxError("Data type of field %s not support range scan",
+                                               prop.c_str());
+                }
+            }
             break;
         }
         case nebula::Expression::kFunctionCall : {
-            auto* fExpr = dynamic_cast<const FunctionCallExpression*>(expr);
-            auto* name = fExpr->name();
-            if (*name == "udf_is_in") {
-                return Status::SyntaxError("Unsupported function ： %s", name->c_str());
-            }
-            break;
+            return Status::SyntaxError("Function expressions are not supported yet");
         }
         default : {
             return Status::SyntaxError("Syntax error ： %s", expr->toString().c_str());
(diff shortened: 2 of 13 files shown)


src/storage/index/IndexPolicyMaker.cpp
```diff
@@ -151,6 +153,81 @@ bool IndexPolicyMaker::exprEval(Getters &getters) {
...
+        case RelationalExpression::Operator::GE: {
+            return RelationalExpression::Operator::LE;
+        }
+        default : {
+            return op;
+        }
+    }
+}
+
+bool IndexPolicyMaker::writeScanItem(const std::string& prop, const OperatorItem& item) {
```
dangleptr, line 178: Please add some comments about this method.  It looks some bit complex.

src/graph/LookupExecutor.cpp
```diff
@@ -676,5 +702,35 @@ bool LookupExecutor::processFinalVertexResult(RpcResponse &rpcResp,
...
+    // Need to consider the logic in method 'traversalExpr' after new relational operation added.
+    switch (op) {
+        case RelationalExpression::Operator::EQ :
+        case RelationalExpression::Operator::GE :
+        case RelationalExpression::Operator::GT :
+        case RelationalExpression::Operator::LE :
+        case RelationalExpression::Operator::LT : {
+            return Status::OK();
+        }
+        case RelationalExpression::Operator::NE :
```
dangleptr, line 716: We don't support NE any more?

bright-starry-sky: > We don't support NE any more?

Will need full scan index when a single NE condition occur. We need very complex optimization logic if we want to support NE. I recommend cut this NE in 1.0 .  and you? What do you think？

src/storage/index/IndexPolicyMaker.h
```diff
@@ -22,6 +22,11 @@ namespace storage {
  */
 using OperatorItem = std::tuple<std::string, VariantType, RelationalExpression::Operator>;
 
+/**
+ * Range scan item tuple<less_than , begin_value, greater_than, end_value>
+ **/
+using ScanItem = std::tuple<bool, OptVariantType, bool, OptVariantType>;
```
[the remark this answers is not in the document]
bright-starry-sky: Sure.

src/storage/index/IndexPolicyMaker.cpp
bright-starry-sky: Added. please take a look in source code. thanks.

src/graph/LookupExecutor.cpp
dangleptr: I don't think so.  IMO for NE, we need to scan the whole index.   It is a very import case for index.
And i suggest we disable range scan for string inside graphd.

src/storage/index/IndexPolicyMaker.h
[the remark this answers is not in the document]
dangleptr: Weird!  I suggest we use a more meaningful structure for it.

src/graph/LookupExecutor.cpp
bright-starry-sky: > I don't think so. IMO for NE, we need to scan the whole index. It is a very import case for index.
> And i suggest we disable range scan for string inside graphd.

got, let me improve it.

bright-starry-sky:
1，Optimized index scan logic;
2，Some of the structures have been refactored to make them easier to understand

bright-starry-sky:
Ready to review.

src/common/utils/NebulaKeyUtils.h
```diff
@@ -251,6 +258,108 @@ class NebulaKeyUtils final {
         return rawKey.subpiece(0, rawKey.size() - sizeof(int64_t));
     }
 
+    // Only int and double are supported
+    static std::string boundVariant(nebula::cpp2::SupportedType type,
+                                    NebulaBoundValueType op,
+                                    const VariantType& v = 0L) {
+        switch (op) {
+            case NebulaBoundValueType::kMax : {
+                std::vector<unsigned char> bytes = {0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF};
```
dangleptr, line 267: Is it right for int64_t and double?

src/common/utils/NebulaKeyUtils.h
```diff
@@ -251,6 +258,108 @@ class NebulaKeyUtils final {
...
+                return std::string(bytes.begin(), bytes.end());
+            }
+            case NebulaBoundValueType::kMin : {
+                std::string str;
+                str.reserve(sizeof(int64_t));
+                str.append(sizeof(int64_t), '\0');
+                return str;
+            }
+            case NebulaBoundValueType::kSubtraction : {
```
dangleptr, line 276: For integer,  what's the meaning about subtraction?

src/graph/LookupExecutor.cpp
```diff
@@ -240,76 +241,155 @@ Status LookupExecutor::checkFilter() {
...
 }
 
-Status
-LookupExecutor::findValidIndex() {
-    std::vector<std::shared_ptr<nebula::cpp2::IndexItem>> indexes;
-    std::set<std::string> filterCols;
-    for (auto& filter : filters_) {
-        filterCols.insert(filter.first);
-    }
+Status LookupExecutor::findValidIndex() {
```
dangleptr, line 244: It seems some bit complex,  let's talk about it offline.

src/graph/test/LookupTest.cpp
```diff
@@ -344,17 +346,6 @@ TEST_F(LookupTest, VertexConditionScan) {
         std::vector<std::tuple<VertexID>> expected = {};
         ASSERT_TRUE(verifyResult(resp, expected));
     }
-    {
```
dangleptr, line 347: Why remove it?

src/graph/test/LookupTest.cpp
```diff
@@ -402,18 +393,6 @@ TEST_F(LookupTest, EdgeConditionScan) {
         };
         ASSERT_TRUE(verifyResult(resp, expected));
     }
-    {
```
dangleptr, line 405: ditto

src/storage/index/IndexExecutor.inl
```diff
@@ -115,16 +117,96 @@ cpp2::ErrorCode IndexExecutor<RESP>::checkReturnColumns(const std::vector<std::s
...
+                                       const ScanBound& item) {
+    std::string begin, end;
+    auto type = field.get_type().type;
+    // if begin == end, means the scan is equivalent scan.
+    if (item.beginBound_.rel_ != RelationType::kGTRel &&
+        item.endBound_.rel_ != RelationType::kGTRel &&
+        item.beginBound_.rel_ != RelationType::kNull &&
+        item.endBound_.rel_ != RelationType::kNull &&
+        item.beginBound_.val_ == item.endBound_.val_) {
+        begin = end = NebulaKeyUtils::encodeVariant(item.beginBound_.val_);
```
dangleptr, line 169: em....    
if begin.val == end.val
It seems  only the begin.rel == end.rel is a valid operation, right?

src/storage/index/IndexExecutor.inl
```diff
@@ -115,16 +117,96 @@ cpp2::ErrorCode IndexExecutor<RESP>::checkReturnColumns(const std::vector<std::s
...
+        return std::make_pair(begin, end);
+    }
+    // normalize begin and end value using ScanItem. for example :
```
dangleptr, line 172: So,  the scope is. [being, end)  ?

src/common/utils/NebulaKeyUtils.h
bright-starry-sky: > Is it right for int64_t and double?

int64 and double convert to binary as below, I think using byte of filled with '0xFF'  is ok. WDYT?
std::numeric_limits<int64_t>::max() -->0xfffffffffffff
std::numeric_limits<double>::max() --> -nan(0xfffffffffefff)

src/storage/index/IndexExecutor.inl
bright-starry-sky: Yes, that's right.

bright-starry-sky:
Simplified method findValidIndex.

bright-starry-sky:
To improved NE condition optimizer logic for storage layer.
If NE expression occur , after fields will no longer be optimized.

src/common/utils/NebulaKeyUtils.h
```diff
@@ -251,6 +258,108 @@ class NebulaKeyUtils final {
...
     }
 
+    // Only int and double are supported
+    static std::string boundVariant(nebula::cpp2::SupportedType type,
+                                    NebulaBoundValueType op,
+                                    const VariantType& v = 0L) {
+        switch (op) {
+            case NebulaBoundValueType::kMax : {
+                std::vector<unsigned char> bytes = {0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF};
+                return std::string(bytes.begin(), bytes.end());
```
wadeliuyi, line 268: return the static string directly?

src/common/utils/NebulaKeyUtils.h
```diff
@@ -251,6 +258,108 @@ class NebulaKeyUtils final {
...
+            }
+            case NebulaBoundValueType::kMin : {
+                std::string str;
+                str.reserve(sizeof(int64_t));
+                str.append(sizeof(int64_t), '\0');
+                return str;
```
wadeliuyi, line 274: static std::string str("\0\0\0\0\0\0\0\0\0\0\0", sizeof(int64_t)); return str;

src/common/utils/NebulaKeyUtils.h
```diff
@@ -251,6 +258,108 @@ class NebulaKeyUtils final {
...
+            }
+            case NebulaBoundValueType::kSubtraction : {
+                std::string str;
+                if (type == nebula::cpp2::SupportedType::INT) {
+                    str = encodeInt64(boost::get<int64_t>(v));
+                } else {
+                    str = encodeDouble(boost::get<double>(v));
+                }
+                std::vector<unsigned char> bytes(str.begin(), str.end());
```
wadeliuyi, line 283: Is it correct when v is a negative number?

src/common/utils/NebulaKeyUtils.h
```diff
@@ -251,6 +258,108 @@ class NebulaKeyUtils final {
...
+            case NebulaBoundValueType::kAddition : {
+                std::string str;
+                if (type == nebula::cpp2::SupportedType::INT) {
+                    str = encodeInt64(boost::get<int64_t>(v));
+                } else {
+                    str = encodeDouble(boost::get<double>(v));
+                }
+                std::vector<unsigned char> bytes(str.begin(), str.end());
+                for (size_t i = bytes.size(); i > 0; i--) {
+                    if (bytes[i-1] < 255) {
```
wadeliuyi, line 301: ditto

src/storage/index/IndexPolicyMaker.cpp
```diff
@@ -151,6 +161,108 @@ bool IndexPolicyMaker::exprEval(Getters &getters) {
...
+        }
+        default : {
+            auto v = scanItems_.find(prop);
+            if (v == scanItems_.end()) {
+                scanItems_[prop] = ScanBound(Bound(toRel(op), std::get<1>(item)),
+                                             Bound(toRel(op), std::get<1>(item)));
+            } else {
+                // If this field appears in scanItems_ ,
+                // means that the filter conditions are wrong, for example :
+                // c1 == 1 and c2 == 2
```
wadeliuyi, line 258: maybe is // c1 == 1 and c1 == 2?

src/storage/index/IndexPolicyMaker.cpp
```diff
@@ -41,41 +41,58 @@ cpp2::ErrorCode IndexPolicyMaker::decodeExpression(const std::string &filter) {
...
         }
     }
-    if (optimizedPolicy_ && hintNum == operatorList_.size() && !hasStr) {
-        requiredFilter_ = false;
+    // re-check operatorList_.
+    // if operatorList_ is not empty, that means there are still fields to filter
+    if (!requiredFilter_ && operatorList_.size() > 0) {
+        requiredFilter_ = true;
     }
+    return true;
```
wadeliuyi, line 91: need here check wrong filter condition like c1 > 2 and c1 < 1?

src/storage/index/IndexPolicyMaker.cpp
bright-starry-sky: Yes, good point.

src/common/utils/NebulaKeyUtils.h
bright-starry-sky: > static std::string str("\0\0\0\0\0\0\0\0\0\0\0", sizeof(int64_t)); return str;

What's the difference?

src/common/utils/NebulaKeyUtils.h
bright-starry-sky: Yes. the method encodeDouble can be handle  negative number.

src/storage/index/IndexPolicyMaker.cpp
bright-starry-sky: > need here check wrong filter condition like c1 > 2 and c1 < 1?

I don't think we need to check it here, but if we do, it's better to put it in other method .

bright-starry-sky: > need here check wrong filter condition like c1 > 2 and c1 < 1?

I think it's ok even if no error message is returned. It just doesn't return the result set. Current processing  same with some relational databases. WDYT?

src/common/utils/NebulaKeyUtils.h
wadeliuyi: less a constructor function about string call and memory allocate and copy

bright-starry-sky:
Improved string index scan logic, Use the length of the byte as prefix scan or range scan when the field is string type.And avoid expression filtering.

bright-starry-sky:
Add some test cases for index scan of string type.

src/common/utils/NebulaKeyUtils.h
```diff
@@ -251,6 +258,108 @@ class NebulaKeyUtils final {
...
+            }
+            case NebulaBoundValueType::kSubtraction : {
+                std::string str;
+                if (type == nebula::cpp2::SupportedType::INT) {
+                    str = encodeInt64(boost::get<int64_t>(v));
+                } else {
+                    str = encodeDouble(boost::get<double>(v));
+                }
+                std::vector<unsigned char> bytes(str.begin(), str.end());
+                for (size_t i = bytes.size(); i > 0; i--) {
```
dangleptr, line 284: Is this logic correct? 
For example,  the bytes are "0x01  0x00"
As your code, the lower bound is "0x00 0x00" ??

src/graph/LookupExecutor.cpp
```diff
@@ -240,77 +241,212 @@ Status LookupExecutor::checkFilter() {
...
-    std::map<int32_t, IndexID> indexHint;
+    bool hasStringCol = false;
+    for (const auto& filter : filters_) {
+        auto type = schema->getFieldType(filter.first);
+        if (type.get_type() == nebula::cpp2::SupportedType::STRING) {
+            hasStringCol = true;
+            break;
+        }
+    }
+    return hasStringCol ? findValidIndexWithStr() : findValidIndexNoStr();
```
dangleptr, line 372: Add more comments about each method. I can't figure out what your logic in your code.

src/storage/index/IndexPolicyMaker.cpp
```diff
@@ -41,41 +41,54 @@ cpp2::ErrorCode IndexPolicyMaker::decodeExpression(const std::string &filter) {
     return code;
 }
 
-void IndexPolicyMaker::buildPolicy() {
-    prefix_.reserve(256);
-    decltype(operatorList_.size()) hintNum = 0;
-    bool hasStr = false;
+bool IndexPolicyMaker::buildPolicy() {
```
dangleptr, line 44: What's the difference with the original logic?
Add comments about this method.

src/storage/BaseProcessor.h
```diff
@@ -65,9 +65,6 @@ class BaseProcessor {
 
     void doRemove(GraphSpaceID spaceId, PartitionID partId, std::vector<std::string> keys);
 
-    kvstore::ResultCode doRange(GraphSpaceID spaceId, PartitionID partId, std::string start,
```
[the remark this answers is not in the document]
bright-starry-sky: > Why remove it?

Because exists  as below, this is redundant. 

```
kvstore::ResultCode doRange(GraphSpaceID spaceId, PartitionID partId, const std::string& start,
                                const std::string& end, std::unique_ptr<kvstore::KVIterator>* iter);
```

src/common/utils/NebulaKeyUtils.h
bright-starry-sky: > Is this logic correct?
> For example, the bytes are "0x01 0x00"
> As your code, the lower bound is "0x00 0x00" ??

Good point， "0x01 0x00" - 1 should be "0x00 0xFF".

src/storage/index/IndexPolicyMaker.cpp
bright-starry-sky: > What's the difference with the original logic?
> Add comments about this method.

original logic : if has string type, need to do expression filter.
current logic : don't need to  expression filter when has string type, because the query contains all column of index, and the scan string appended the string length.

src/common/utils/NebulaKeyUtils.h
bright-starry-sky: because here also handle double type. so to do it as std::vector<unsigned char> bytes

src/common/utils/NebulaKeyUtils.h
bright-starry-sky: > If so, why not use the bytes encoded by (int - 1) directly?

Yes, we can do that if just have int type. this method also has double. so using bytes to do it at here.

bright-starry-sky:
- Addressed dangleptr's comments.
- Improved logic for bound value.

dangleptr (approved):
Generally, the pr looks good to me.

src/storage/index/IndexPolicyMaker.cpp
```diff
@@ -151,6 +157,106 @@ bool IndexPolicyMaker::exprEval(Getters &getters) {
...
+                if (v->second.endBound_.rel_ == RelationType::kNull) {
+                    v->second.endBound_ = Bound(toRel(op), std::get<1>(item));
+                } else if (v->second.endBound_.val_ > std::get<1>(item)) {
+                    // This might be the case where c1 < 1 and c1 < 5 , so the 1 should be save.
+                    v->second.endBound_.val_ = std::get<1>(item);
+                }
+            }
+            break;
+        }
+        case RelationalExpression::Operator::NE : {
```
critical27, line 241: NE should be illegal based on `buildPolicy`?

src/storage/index/IndexPolicyMaker.cpp
```diff
@@ -151,6 +157,106 @@ bool IndexPolicyMaker::exprEval(Getters &getters) {
...
+            break;
+        }
+        default : {
```
critical27, line 244: Better to use EQ here?

src/graph/LookupExecutor.cpp
```diff
@@ -240,77 +241,225 @@ Status LookupExecutor::checkFilter() {
...
-    return Status::OK();
+    // Sort the priorityIdxs for equivalent condition.
+    std::vector<std::shared_ptr<nebula::cpp2::IndexItem>> priorityIdxs;
+    auto comp = [] (std::pair<int32_t, std::shared_ptr<nebula::cpp2::IndexItem>>& lhs,
+                    std::pair<int32_t, std::shared_ptr<nebula::cpp2::IndexItem>>& rhs) {
+        return lhs.first > rhs.first;
+    };
+    std::sort(eqIndexHint.begin(), eqIndexHint.end(), comp);
+    // Get the index with the highest hit rate from eqIndexHint.
+    int32_t maxHint = eqIndexHint[0].first;
```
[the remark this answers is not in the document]
bright-starry-sky: > Why don't return the first one? Is this reserved on purpose?

To return the highest-rated elements at here, maybe there are multiple, maybe there is one. for example

```
{3, index1}
{3, index2}
{2, index3}
```

so index1 and index2 should be return. for the next step selection.

src/storage/index/IndexPolicyMaker.cpp
bright-starry-sky: Sure. thanks.

src/storage/index/IndexPolicyMaker.cpp
bright-starry-sky: Yes, the NE should be expression.eval(), it not based on scan pair.

src/storage/index/IndexExecutor.inl
```diff
@@ -115,16 +117,108 @@ cpp2::ErrorCode IndexExecutor<RESP>::checkReturnColumns(const std::vector<std::s
...
+}
+
+template <typename RESP>
+std::pair<std::string, std::string>
+IndexExecutor<RESP>::normalizeScanPair(const nebula::cpp2::ColumnDef& field,
+                                       const ScanBound& item) {
+    std::string begin, end;
+    auto type = field.get_type().type;
+    // if begin == end, means the scan is equivalent scan.
+    if (item.beginBound_.rel_ != RelationType::kGTRel &&
```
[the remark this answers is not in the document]
bright-starry-sky: Good point. the end should be LE. And your question led me to other problem. fixed. thanks a lot.

bright-starry-sky:
Addressed critical27's comments

src/graph/LookupExecutor.cpp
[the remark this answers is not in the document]
critical27: Ah, I got it. 👍

critical27 (approved):
Good job!

dangleptr (approved):

PR #2281, opened 2020-08-08
Fixed bug of snapshot from chaos test.

bright-starry-sky:
1,  Fixed checkpoint bug. 
2, Refused to create snapshot when index rebuilding.

### DIFF
--- a/src/meta/processors/admin/AdminClient.cpp
+++ b/src/meta/processors/admin/AdminClient.cpp
@@ -586,12 +586,13 @@ folly::Future<Status> AdminClient::getLeaderDist(HostLeaderMap* result) {
     return future;
 }
 
-folly::Future<Status> AdminClient::createSnapshot(GraphSpaceID spaceId, const std::string& name) {
+folly::Future<Status> AdminClient::createSnapshot(GraphSpaceID spaceId,
+                                                  const std::string& name,
+                                                  const HostAddr& host) {
     if (injector_) {
         return injector_->createSnapshot();
     }
 
-    auto allHosts = ActiveHostsMan::getActiveHosts(kv_);
     storage::cpp2::CreateCPRequest req;
     req.set_space_id(spaceId);
     req.set_name(name);
@@ -603,7 +604,7 @@ folly::Future<Status> AdminClient::createSnapshot(GraphSpaceID spaceId, const st
      * Don't need retry.
      * Because existing checkpoint directories leads to fail again.
      **/
-    getResponse(std::move(allHosts), 0, std::move(req), [] (auto client, auto request) {
+    getResponse({host}, 0, std::move(req), [] (auto client, auto request) {
         return client->future_createCheckpoint(request);
     }, 0, std::move(pro), 0);
     return f;
@@ -611,7 +612,7 @@ folly::Future<Status> AdminClient::createSnapshot(GraphSpaceID spaceId, const st
 
 folly::Future<Status> AdminClient::dropSnapshot(GraphSpaceID spaceId,
                                                 const std::string& name,
-                                                const std::vector<HostAddr>& hosts) {
+                                                const HostAddr& host) {
     if (injector_) {
         return injector_->dropSnapshot();
     }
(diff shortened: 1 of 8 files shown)


dangleptr:
Could you explain the bug in detail ?

src/meta/processors/admin/SnapShot.cpp
```diff
@@ -21,38 +21,42 @@ cpp2::ErrorCode Snapshot::createSnapshot(const std::string& name) {
                    << static_cast<int32_t>(ret);
         return cpp2::ErrorCode::E_STORE_FAILURE;
     }
-    for (auto& space : spaces) {
-        auto status = client_->createSnapshot(space, name).get();
-        if (!status.ok()) {
-            return cpp2::ErrorCode::E_RPC_FAILURE;
+    auto hosts = ActiveHostsMan::getActiveHosts(kv_);
```
dangleptr, line 24: What's the difference with the original logic.

src/meta/processors/admin/SnapShot.cpp
```diff
@@ -77,78 +81,37 @@ bool Snapshot::getAllSpaces(std::vector<GraphSpaceID>& spaces, kvstore::ResultCo
 }
 
 cpp2::ErrorCode Snapshot::blockingWrites(storage::cpp2::EngineSignType sign) {
+    folly::SharedMutex::ReadHolder rHolder(LockUtils::spaceLock());
+    // check the index rebuild. not allowed to block write when index rebuilding.
+    auto prefix = MetaServiceUtils::rebuildIndexStatusPrefix();
```
dangleptr, line 86: Why not put the check inside processor.

src/meta/processors/admin/SnapShot.cpp
```diff
@@ -77,78 +81,37 @@ bool Snapshot::getAllSpaces(std::vector<GraphSpaceID>& spaces, kvstore::ResultCo
...
-
-        if (!status.ok()) {
-            return cpp2::ErrorCode::E_BLOCK_WRITE_FAILURE;
-        }
-    }
-    return cpp2::ErrorCode::SUCCEEDED;
-}
-
-std::unordered_map<HostAddr, std::vector<PartitionID>>
-Snapshot::getLeaderParts(HostLeaderMap *hostLeaderMap, GraphSpaceID spaceId) {
```
dangleptr, line 98: We don't need the method any more?

src/meta/processors/admin/AdminClient.cpp
```diff
@@ -603,15 +604,15 @@ folly::Future<Status> AdminClient::createSnapshot(GraphSpaceID spaceId, const st
      * Don't need retry.
      * Because existing checkpoint directories leads to fail again.
      **/
-    getResponse(std::move(allHosts), 0, std::move(req), [] (auto client, auto request) {
+    getResponse({host}, 0, std::move(req), [] (auto client, auto request) {
```
darionyaphet, line 607: Seems you will create snapshot at host one by one?

bright-starry-sky:
> Could you explain the bug in detail ?
A very humiliating mistake, the error leads to lost checkpoint in some storage node. 
I misunderstood the meaning of 'getResponse'

src/meta/processors/admin/AdminClient.cpp
bright-starry-sky: Yes, you are right.

src/meta/processors/admin/SnapShot.cpp
bright-starry-sky: Yes,This method is no longer used.

src/meta/processors/admin/SnapShot.cpp
bright-starry-sky: Good point.

src/meta/processors/admin/SnapShot.cpp
```diff
@@ -77,78 +81,37 @@ bool Snapshot::getAllSpaces(std::vector<GraphSpaceID>& spaces, kvstore::ResultCo
...
 cpp2::ErrorCode Snapshot::blockingWrites(storage::cpp2::EngineSignType sign) {
+    folly::SharedMutex::ReadHolder rHolder(LockUtils::spaceLock());
+    // check the index rebuild. not allowed to block write when index rebuilding.
+    auto prefix = MetaServiceUtils::rebuildIndexStatusPrefix();
+    std::unique_ptr<kvstore::KVIterator> iter;
+    auto ret = kv_->prefix(kDefaultSpaceId, kDefaultPartId, prefix, &iter);
+    if (ret != kvstore::ResultCode::SUCCEEDED) {
+        return cpp2::ErrorCode::E_BLOCK_WRITE_FAILURE;
+    }
+    while (iter->valid()) {
```
dangleptr, line 92: why use while here?

bright-starry-sky: > why use while here?

I need check whether any index is being rebuild for all spaces and all indexes.

bright-starry-sky:
Address's comments .move the index check logic to processor.

Issue #2256, opened 2020-07-28
Upsert should succeed, not an error


panda-sheep:
create tag t (a int , b int default 1, c int);
upsert vertex 5 set t1.a = 1, t1.b = $^.t1.a + 1, t1.c = 2;

edge has the same problem

2.0 will be supported

PR #2282, opened 2020-08-09
Enhance update/upsert  in nebula 1.0

panda-sheep:
fix issue #2256
three questions in nebula 1.0

```
case 1
create tag t1 (a int , b int default 1);
upsert vertex 5 set t1.a = 1, t1.b = $^.t1.a + 1;
The upsert statement should be executed successfully, but currently it has failed
The reason is that` b = $^.t1.a + 1 `requires `a` to have a default value, which is wrong

case 2
create edge e1 (c  int , d int default 1);
upsert edge  111 -> 222@0 of e1  set  c= 3,  d = e1.c+ 10;
The upsert statement should be executed successfully, but currently it has failed
The reason is that` d = e1.c+ 10 `  requires` c`  to have a default value, which is wrong

case 3
update/upsert edge 111->333 @0 of e1 set  c = 1,  d = $^.t1.a; 
$^ is different in the forward edge and the reverse edge
$^ in the positive edge is 5
$^ in the reverse edge is 10002
Cause the properties of the forward and reverse edges to be inconsistent.
Therefore, tag properties are not allowed in the` set clause` and `when clause` in update edge

case 1 and case 2  case 3   are ok in   2.0 
```

### DIFF
--- a/src/graph/test/UpdateTest.cpp
+++ b/src/graph/test/UpdateTest.cpp
@@ -239,12 +239,13 @@ TEST_P(UpdateUpsertTest, UpdateUpsertEdge) {
     }
     {   // Filter out
         // https://github.com/vesoft-inc/nebula/issues/1888
+        // Update set clause and filter clause are not allowed to manipulate tag attributes
         cpp2::ExecutionResponse resp;
         auto query = GetParam() + " EDGE 200 -> 101@0 OF select "
                     + "SET grade = select.grade + 1, year = 2000 "
                     + "WHEN select.grade > 1024 && $^.student.age > 15";
         auto code = client_->execute(query, resp);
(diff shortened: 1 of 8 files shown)


critical27:
Good job, so after this PR merged, the update is the same with 2.0?

panda-sheep:
> Good job, so after this PR merged, the update is the same with 2.0?
Haha, this pr does not synchronize 2.0 to 1.0
Just solve 3 small problems in 1.0.

panda-sheep:
> Good job, so after this PR merged, the update is the same with 2.0?

Haha, this pr does not synchronize 2.0 to 1.0
Just solve 3 small problems in 1.0.

PR #2286, opened 2020-08-10
support ref $-.`$-.label`

xuguruogu:
otherwise rename label again and again may be a big trouble if has many columns.


Sophie-Xie:

PR #2289, opened 2020-08-11
fix multi daemon can't start on same node

critical27:
Introduced in #2278, if we want to start multiple daemon with different flag file, the `FLAGS_pid_file` is the same because we have not init `gflags` yet. (all of them is the default value of `FLAGS_pid_file`, not the one in flag file)

So we need to init `gflags`, `glogs` by manual instead of `folly::init`.

### DIFF
--- a/src/daemons/MetaDaemon.cpp
+++ b/src/daemons/MetaDaemon.cpp
@@ -171,8 +171,7 @@ int main(int argc, char *argv[]) {
     }
 
     google::SetVersionString(nebula::versionString());
-    folly::SingletonVault::singleton()->registrationComplete();
-    google::InitGoogleLogging(argv[0]);
+    folly::init(&argc, &argv, true);
     if (FLAGS_data_path.empty()) {
         LOG(ERROR) << "Meta Data Path should not empty";
         return EXIT_FAILURE;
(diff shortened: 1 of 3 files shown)


src/daemons/MetaDaemon.cpp
```diff
@@ -170,7 +171,8 @@ int main(int argc, char *argv[]) {
     }
 
     google::SetVersionString(nebula::versionString());
-    folly::init(&argc, &argv, true);
```
dangleptr, line 173: why remove the init method.

critical27: Because folly::init is almost same with init `gflags + glog + singleton`. You mean we init gflags twice?

dangleptr:
> Introduced in #2278, if we want to start multiple daemon with different flag file, the `FLAGS_pid_file` is the same because we have not init `gflags` yet.
> 
> So we need to init `gflags`, `glogs` by manual instead of `folly::init`.

I am confused,  if you want to start multi instances on one host,  each instance should have its own pid file in config file, right?

critical27:
Init gflags twice, one before `check pid`, the other in `folly::init`

PR #2290, opened 2020-08-11
Add warning msg for partially performed

laura-ding:


src/graph/DropSpaceExecutor.cpp
```diff
@@ -40,6 +40,8 @@ void DropSpaceExecutor::execute() {
         if (*spaceName_ == ectx()->rctx()->session()->spaceName()) {
             ectx()->rctx()->session()->setSpace("", -1);
         }
+        ectx()->addWarningMsg("Data will be deleted completely after restarting the services");
```
dangleptr, line 43: BTW.  What's the message meaning?...

src/graph/GoExecutor.cpp
```diff
@@ -714,7 +715,9 @@ void GoExecutor::finishExecution() {
             yc.emplace_back(std::move(ptr));
         }
     }
-
+    if (!warningMsg_.empty()) {
+        ectx()->addWarningMsg(std::move(warningMsg_));
```
dangleptr, line 719: I suggest just copy the content to avoid the warningMsg_  reused after the move operation..

src/graph/DropSpaceExecutor.cpp
laura-ding: when the user executes `drop space`, it will give the warning message to the user that the data dir will delete after service restart

src/graph/GoExecutor.cpp
laura-ding: Here is the final handle, there is no place to use it again.  but i agree with you.

PR #2291, opened 2020-08-11
Make some rocksdb's option dynamically changeable

chenxu14:
According to rocksdb’s **MutableCFOptions** and **MutableDBOptions**, there are some options that can be changed dynamically through SetOptions, but nebula currently don’t supported them, maybe there is a need to enable those?


critical27:
We didn't add some option because we rarely use them in production befor. We could enable them if they are necessary.

chenxu14:
> We didn't add some option because we rarely use them in production before. Of course, we could enable them if they are necessary.

Agree with you, some options are rarely use. But I think the following options are useful, since they may stall the write.
(1) soft_pending_compaction_bytes_limit
(2) hard_pending_compaction_bytes_limit are useful
(3) max_compaction_bytes

dangleptr:
IMO we should support all mutable options in RocksDB,  of course,  some of them maybe forbidden by ConfigManager.

critical27:
> > We didn't add some option because we rarely use them in production before. Of course, we could enable them if they are necessary.
> 
> Agree with you, some options are rarely use. But I think the following options are useful, since they may stall the write.
> (1) soft_pending_compaction_bytes_limit
> (2) hard_pending_compaction_bytes_limit
> (3) max_compaction_bytes

Make sense, so we just add all of them.

Issue #2288, opened 2020-08-10
I can not understand this test.


xuguruogu:
https://github.com/vesoft-inc/nebula/blob/339d0c575e2dd035814f14b636b13281442f158c/src/graph/test/GoTest.cpp#L2872-L2898

According to my understanding, the result should have 4 records instead of 8 records.

Shylock-Hg:
First the left sentence in pipe get result with 2 records, then right sentence of pipe query from two start vid and get 4 records,
and the result join previous input get 4*2 records.

xuguruogu:
As mentioned in #2227

e1: 1:5
e2: 1:1

`go ... over e1 ... | go ... over e2`, result in 25 records. This seems unexpected.

Shylock-Hg:
> As mentioned in #2227
> 
> e1: 1:5
> e2: 1:1
> 
> `go ... over e1 ... | go ... over e2`, result in 25 records. This seems unexpected.

It follow this process too.

Issue #2227, opened 2020-07-14
go from a over e1 ... | go from a over e2. return more records than expected.


xuguruogu:
e1: 1:5
e2: 1:1
we got 1:25 after the pipeline, more than the expected 5 records.

Shylock-Hg:
Do you join previous result ?

PR #2292, opened 2020-08-11
fix go ... | go ... join bug.

xuguruogu:
fix issue #2288 #2227


dangleptr:
very efficient

src/graph/GoExecutor.cpp
```diff
@@ -1116,25 +1244,14 @@ bool GoExecutor::processFinalResult(Callback cb) const {
             for (const auto &vdata : resp.vertices) {
                 DCHECK(vdata.__isset.edge_data);
                 VLOG(1) << "Total vdata.edge_data size " << vdata.edge_data.size();
-                auto tagData = vdata.get_tag_data();
-                auto srcId = vdata.get_vertex_id();
+                tagData = &vdata.get_tag_data();
+                srcId = vdata.get_vertex_id();
                 const auto roots = getRoots(srcId, recordIn);
```
dangleptr, line 1249: We only need to get the inputRows when fromType_ is not kInstantExpr

critical27:
close #2288, #2227

PR #2297, opened 2020-08-13
Remove proposed term in raft election

critical27:
1. Remove proposed term in raft election, we always use `currentTerm + 1` to start election, once elected, use this term as the new term as well. The other election logic is the same as described in paper.
2. If any peer rejected explicitly, just give up. If quorum satisfied, step up as leader.

Hold on please, we need to test it in chaos for several days.


src/kvstore/raftex/RaftPart.cpp
```diff
@@ -219,8 +219,7 @@ RaftPart::RaftPart(ClusterID clusterId,
         , ioThreadPool_{pool}
         , bgWorkers_{workers}
         , executor_(executor)
-        , snapshot_(snapshotMan)
-        , weight_(1) {
```
bright-starry-sky, line 223: Can `weight_` be deleted?

critical27: Yeah, `weight_` is used to control elect interval before, if log is not more up-to-date than others, it will double interval (There is a log, we can rarely see it). After this PR, if I am explicitly rejected, just step down as follower, will try start election later.

src/kvstore/raftex/RaftPart.cpp
```diff
@@ -1053,13 +1052,18 @@ typename RaftPart::Role RaftPart::processElectionResponses(
...
+        }
         if (r.second.get_error_code() == cpp2::ErrorCode::SUCCEEDED) {
             ++numSucceeded;
-        } else if (r.second.get_error_code() == cpp2::ErrorCode::E_LOG_STALE) {
-            LOG(INFO) << idStr_ << "My last log id is less than " << hosts[r.first]->address()
-                      << ", double my election interval.";
-            uint64_t curWeight = weight_.load();
-            weight_.store(curWeight * 2);
+        } else if (r.second.get_error_code() == cpp2::ErrorCode::E_LOG_STALE ||
+                   r.second.get_error_code() == cpp2::ErrorCode::E_TERM_OUT_OF_DATE) {
```
darionyaphet, line 1061: If the log is older than other node, it will downgrade to follower. 👍

PR #2298, opened 2020-08-13
Fix not init argv

laura-ding:


bright-starry-sky:
Do we need to change `folly::init(&argc, &argv, true)` to `folly::init(&argc, &argv, false)` ?

laura-ding:
> Do we need to change `folly::init(&argc, &argv, true)` to `folly::init(&argc, &argv, false)` ?

Don't need, because the code does not use `argv` anymore.

PR #2305, opened 2020-08-17
Minor changes in raft, fix unstable ut, update GetNeighbors benchmark

critical27:
1. Fix unstable ut in NebulaStoreTest because we check the leader by `leader_` not the `role` in `RaftPart`
2. Update GetNeighbors benchmark,  add bench of encode and filter, similar to what we did in 2.0
3. Add a random delay after election failed, make sure only one election request is on-going.
4. Fix follower would crash in unusual case.
5. Keep one extra wal, because I check a log in test env and find out that even if the log gap is very small, when we reboot, it would trigger snapshot because there is only one wal.

### DIFF
--- a/src/kvstore/raftex/RaftPart.cpp
+++ b/src/kvstore/raftex/RaftPart.cpp
@@ -1516,10 +1516,10 @@ void RaftPart::processAppendLogRequest(
                   << ", my term is " << term_
                   << ", to make the cluster stable i will follow the high term"
                   << " candidate and clenaup my data";
-        reset();
         resp.set_committed_log_id(committedLogId_);
         resp.set_last_log_id(lastLogId_);
         resp.set_last_log_term(lastLogTerm_);
+        return;
     }
 
     // req.get_last_log_id_sent() >= committedLogId_
--- a/src/kvstore/raftex/RaftPart.cpp
+++ b/src/kvstore/raftex/RaftPart.cpp
@@ -760,6 +760,13 @@ void RaftPart::replicateLogs(folly::EventBase* eb,
             break;
         }
 
+        if (term_ != currTerm) {
+            VLOG(2) << idStr_ << "Term has been updated, previous "
+                    << currTerm << ", current " << term_;
+            currTerm = term_;
+            break;
+        }
+
         hosts = hosts_;
     } while (false);
 
@@ -1089,6 +1096,14 @@ typename RaftPart::Role RaftPart::processElectionResponses(
 bool RaftPart::leaderElection() {
     VLOG(2) << idStr_ << "Start leader election...";
     using namespace folly;  // NOLINT since the fancy overload of | operator
+
+    bool expected = false;
+    if (!inElection_.compare_exchange_strong(expected, true)) {
+        return true;
+    }
+    SCOPE_EXIT {
+        inElection_ = false;
+    };
 
     cpp2::AskForVoteRequest voteReq;
     decltype(hosts_) hosts;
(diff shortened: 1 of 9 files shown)


src/kvstore/raftex/RaftPart.cpp
```diff
@@ -1527,10 +1527,10 @@ void RaftPart::processAppendLogRequest(
                   << ", my term is " << term_
                   << ", to make the cluster stable i will follow the high term"
                   << " candidate and clenaup my data";
-        reset();
         resp.set_committed_log_id(committedLogId_);
         resp.set_last_log_id(lastLogId_);
         resp.set_last_log_term(lastLogTerm_);
+        return;
```
critical27, line 1533: image
I only have this log, crashed in v1.0.1. I think it is the only way that lastLogId in wal is 0. PS: I am not pretty sure about whether we could `reset` in this case/

src/kvstore/raftex/RaftPart.cpp
```diff
@@ -1211,7 +1211,7 @@ void RaftPart::statusPolling(int64_t startTime) {
         } else {
             // No leader has been elected, sleep for a while to continue
             VLOG(2) << idStr_ << "Wait for a while and continue the leader election";
-            delay = FLAGS_raft_heartbeat_interval_secs * 1000;
+            delay = FLAGS_raft_heartbeat_interval_secs * 1000 + folly::Random::rand32(500);
```
bright-starry-sky, line 1214: Does this avoid the problem of all nodes voting at the same time?

bright-starry-sky:
LGTM.

src/kvstore/raftex/RaftPart.cpp
critical27: Yeah, I delete it in #2297 by mistake.

bright-starry-sky (approved):

darionyaphet (approved):

src/kvstore/raftex/RaftPart.cpp
```diff
@@ -758,6 +758,13 @@ void RaftPart::replicateLogs(folly::EventBase* eb,
             break;
         }
 
+        if (term_ != currTerm) {
+            VLOG(2) << idStr_ << "Term has been updated, previous "
+                    << currTerm << ", current " << term_;
+            currTerm = term_;
+            break;
```
dangleptr, line 765: "break" is right? You skip the line 768.

src/kvstore/raftex/RaftPart.cpp
```diff
@@ -1536,6 +1551,7 @@ void RaftPart::processAppendLogRequest(
         resp.set_committed_log_id(committedLogId_);
         resp.set_last_log_id(lastLogId_);
         resp.set_last_log_term(lastLogTerm_);
+        return;
```
dangleptr, line 1554: Return here is right?  I suggest if we dont' want reset operation, maybe just return is another choice.

src/kvstore/raftex/RaftPart.cpp
critical27: My bad, good catch!

src/kvstore/raftex/RaftPart.cpp
critical27: Yeah, we do need to return in this case.

dangleptr (approved):
LGTM

liuyu85cn (dismissed):

monadbobo (approved):
LGTM.

critical27:
Hold on, still check the failed ut.

dangleptr (approved):

liuyu85cn (approved):
LGTM

PR #2307, opened 2020-08-18
Improve action workflow configuration

yixinglu:
- Limit cpu cores for building and testing.
- Use docker action to build image.
- Disable build action for push event.


.github/workflows/docker.yaml
```diff
@@ -3,12 +3,6 @@ name: docker
 on:
   schedule:
     - cron: '0 18 * * *'
-  push:
-    branches:
-      - 'v[0-9]+.*'
```
Shylock-Hg, line 8: Why remove this ?

dutor: Every PR is already required to be tested based on the upstream. At the meantime, it seems that no one cares about the status of the push event, at present.

Of course, we could add it back if necessary in future.

Shylock-Hg: Emmm. Need add it again when need release.

dutor: > Emmm. Need add it again when need release.

This is an awesome idea. We should conceive of a workflow about releasing, including CI, FO and benchmarks, etc.

dutor: I'm gonna mark it down and merge this PR.

yixinglu: this has been covered by release workflow.

PR #2311, opened 2020-08-21
Split Storaged's RPC into multiple threadpools according to their priority

chenxu14:
_StorageServer_ use _PriorityThreadManager_ as the thread manager, but currently all RPC request are treat as _NORMAL_ priority. In write heavy scenario, read requests may be blocked because there are no idle thread to process it. 

So we'd better split RPC requests into different priorities. To implements this, we should remove the **process_in_event_base** declaration in ThriftGenerate.cmake and introduce **priority** annotation for each method.

### DIFF
--- a/cmake/ThriftGenerate.cmake
+++ b/cmake/ThriftGenerate.cmake
@@ -81,7 +81,7 @@ add_custom_command(
   COMMAND ${THRIFT1}
     --strict "--allow-neg-enum-vals"
     --templates ${THRIFT_TEMPLATES}
-    --gen "mstch_cpp2:include_prefix=${include_prefix},process_in_event_base,stack_arguments"
+    --gen "mstch_cpp2:include_prefix=${include_prefix},stack_arguments"
     --gen "py"
     --gen "java:hashcode"
     --gen "go:thrift_import=github.com/facebook/fbthrift/thrift/lib/go/thrift,package_prefix=github.com/vesoft-inc/nebula-go/"
--- a/src/storage/StorageServer.cpp
+++ b/src/storage/StorageServer.cpp
@@ -24,9 +24,10 @@
 DEFINE_int32(port, 44500, "Storage daemon listening port");
 DEFINE_bool(reuse_port, true, "Whether to turn on the SO_REUSEPORT option");
 DEFINE_int32(num_io_threads, 16, "Number of IO threads");
-DEFINE_int32(num_worker_threads, 32, "Number of workers");
 DEFINE_int32(storage_http_thread_num, 3, "Number of storage daemon's http thread");
 DEFINE_bool(local_config, false, "meta client will not retrieve latest configuration from meta");
+DEFINE_string(num_threads_per_priority, "2:2:16:16:2",
+        "Number of worker threads for each request priority.");
 
 namespace nebula {
 namespace storage {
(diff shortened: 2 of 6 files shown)


src/storage/StorageServer.cpp
```diff
@@ -96,10 +97,31 @@ bool StorageServer::initWebService() {
...
 
+static std::shared_ptr<apache::thrift::concurrency::ThreadManager> newThreadManager() {
+    std::vector<size_t> threadNums;
+    folly::split(":", FLAGS_num_threads_per_priority, threadNums, false);
+    if (threadNums.size() != apache::thrift::concurrency::N_PRIORITIES) {
+        return nullptr;
+    }
+    return apache::thrift::concurrency::PriorityThreadManager::newPriorityThreadManager(
+            {{
+                threadNums[apache::thrift::concurrency::HIGH_IMPORTANT],
```
dangleptr, line 108: HIGH_IMPORTANT is in usage?

chenxu14: Hi, not in use yet, I'm not quit sure if the _priority_ declaration in storage.thrift is reasonable, do you think 0:2:32:32:2 is OK?

chenxu14:
Any suggestions? ping @xuguruogu

xuguruogu:
I want to know if this has any experimental results. Because In fact, all requests will be submitted to reader thread pools immediately.

xuguruogu:
By using PriorityThreadManager, can we split different socket connections for different request types?

chenxu14:
> I want to know if this has any experimental results. Because In fact, all requests will be submitted to reader thread pools immediately.

What do you mean "**reader thread pools**"?

chenxu14:
> By using PriorityThreadManager, can we split different socket connections for different request types?

I may know what you mean “reader thread pools", thread pools for handling Socket IO events？
PriorityThreadManager is a worker pool. So we can't different Connections by request priority.

cmake/ThriftGenerate.cmake
```diff
@@ -81,7 +81,7 @@ add_custom_command(
   COMMAND ${THRIFT1}
     --strict "--allow-neg-enum-vals"
     --templates ${THRIFT_TEMPLATES}
-    --gen "mstch_cpp2:include_prefix=${include_prefix},process_in_event_base,stack_arguments"
+    --gen "mstch_cpp2:include_prefix=${include_prefix},stack_arguments"
```
dangleptr, line 84: After digging the usage about "process_in_event_base",  i am not sure  it is not correct for current nebula thread model.

Typically,  we don't want the service function call being invoked inside IOThreadPool,  right?  And By default,  thrift will use a ThreadManager(CpuThreadPool) to take care the real job.  So all logic inside nebula are based on this thread model.  For example,  in write path,  the real processing logic happen inside ThreadManager and for some read path not normally (For example, scan path etc )are reuse this pool too.

wdyt? @chenxu14

chenxu14: Hi, thanks for your input, after tracing the code stack, here's what I found.
After remove “**process_in_event_base**”, the call stack will works like this (take getEdgeProps as an example)
```
StorageServiceAsyncProcessor::process(StorageService.cpp)
    —>apache::thrift::detail::ap::process(GeneratedCodeHelper.h)
        —> StorageServiceAsyncProcessor::getBinaryProtocolProcessMap(StorageService.cpp)
            —> process_pmap(GeneratedCodeHelper.h)
                   # get function pointer through Cpp2RequestContext#getMethodName and pmap
                   # the method signature will be _processInThread_getEdgeProps
                   —> StorageServiceAsyncProcessor::_processInThread_getEdgeProps(StorageService.tcc)
                       —> GeneratedAsyncProcessor::processInThread(AsyncProcessor.h)
                           —> ThreadManager#add(PriorityEventTask, 0, 0, true, true)
                                  # add Task to ThreadManager
```
So the service function will be invoked through **ThreadManager**, not **IOThreadPool**.
The core difference is that the task type add to ThreadManager are different. before remove “process_in_event_base”, task type is a FunctionRunner, has no priority info, so ThreadManager will treat all request as NORMAL, After remove “process_in_event_base”, task type will be PriorityEventTask which has priority info.

dangleptr: Indeed, as the name "process_in_event_base" meaning,  if pass the flag into thrift compiler,  `ThreadManager` should be disabled, right?  
But as your description above,  no matter w/o the flag, the real job always been done inside ThreadManager.

That's weired, maybe a bug for thrift?

chenxu14: There seems to be no official introduction how to use it, not easy to say if it's a bug.

Sophie-Xie:

PR #2312, opened 2020-08-24
fix scan don't return result after add enable_multi_versions

critical27:
We just return all data if enable_multi_versions is false.

### DIFF
--- a/src/storage/query/ScanEdgeProcessor.cpp
+++ b/src/storage/query/ScanEdgeProcessor.cpp
@@ -12,6 +12,7 @@
 #include "dataman/RowReader.h"
 #include "dataman/RowWriter.h"
 #include "meta/NebulaSchemaProvider.h"
+#include "storage/StorageFlags.h"
 
 DEFINE_int32(max_scan_block_size, 4 * 1024 * 1024, "Max size of a respsonse block");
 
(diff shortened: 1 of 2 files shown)


panda-sheep (approved):
Excellent!

panda-sheep:
Is there documentation for the scan interface?
If so, it is recommended to follow the documentation.

If FLAGS_enable_multi_versions is false, the `start time` and `end time ` have no effect. 
Otherwise it will cause user confusion.

src/storage/query/ScanEdgeProcessor.cpp
```diff
@@ -67,7 +68,7 @@ void ScanEdgeProcessor::process(const cpp2::ScanEdgeRequest& req) {
         // only return data within time range [start, end)
         EdgeVersion version = folly::Endian::big(NebulaKeyUtils::getVersion(key));
         int64_t ts = std::numeric_limits<int64_t>::max() - version;
-        if (ts < startTime || ts >= endTime) {
+        if (FLAGS_enable_multi_versions && (ts < startTime || ts >= endTime)) {
```
bright-starry-sky, line 71: A question, if FLAGS_enable_multi_versions == false, here should be return all versions data? right?

critical27: Yeah, we will return all versions. When `FLAGS_enable_multi_versions` is false, there should be only one version.

bright-starry-sky (approved):
LGTM

darionyaphet (approved):
LGTM

dangleptr:
Maybe we could skip the check logic inside compaction too.  Of course, we should open another pr for it.

Issue #2304, opened 2020-08-17
GoExecutor should not return succ by default when completeness not eq 100


Crazyinging:
![image](https://user-images.githubusercontent.com/5928054/90366716-3c54d780-e09a-11ea-9502-28bd39873331.png)
As shown, GoExecutor return succ when completeness not eq 100. So it seems that the data received by the caller has the same effect as the missing data!
Reference: https://discuss.nebula-graph.com.cn/t/topic/1005

Crazyinging:
I think GoExecutor should return an exception when completeness not eq 100. And before that, graphd can retry some times on the failed storageds.

To be smarter, after reading the code, I think maybe it can be changed like this:
1. GraphService add interface: 
- future_execute(int64_t sessionId, const std::string& stmt, const RequestConf req_conf);
This interface will deal req_conf and call future_execute(int64_t sessionId, const std::string& stmt);

2. Add: 
class RequestConf {
  private:
    int32_t graphd_retry_times_;    // Value range 0-3. Default is 0.
                                                       // Indicates the number of retries from graphd to storaged.
    bool require_full_data_;            // When graphd can not get the full amount of data, whether to return exception or return the partial data that has been obtained. 
                                                      // True(Default value): return exception when can not get the full amount of data. 
                                                      // False: return the partial data that has been obtained.
};

3. When GoExecutor::stepOut or GoExecutor::fetchVertexProps call storaged failed, retry according to RequestConf.

Crazyinging:
Correspondingly, I use java client and java client should add interface:
```java
public ResultSet executeQuery(String statement, RequestConf requestConf)
            throws ConnectionException, NGQLException, TException;
```

dangleptr:
Do we really need the configuration is on request level?

Crazyinging:
> Do we really need the configuration is on request level?

Indeed, request level may over -designed.
Client level maybe be better, register RequestConf when new client.

In addition, server level may not appropriate, such as adding configuration in the file nebula-graphd.conf, because the same service will have many users, and everyone's demands for stability may be different.

CPWstatic:
This won't be fixed in nebula1.0. We had a configuration for it now in nebula2.0.

PR #2313, opened 2020-08-24
Fix the problem of data loss in query statements

Crazyinging:
Fix the problem of data loss in query statements, involving GO, Fetch TAG, FETCH EDGE, LOOKUP, FIND PATH, and add error code -12 so that the client can retry on demand.

When graphd calls storaged abnormally, the client may get incomplete data, resulting in data loss.

The error code -12 has been added, so the SDK should changed together, otherwise the SDK may not recognize -12 and report an error. 
Like Java:
image

No

Method 1:
We can deploy two storaged and save the data separately, and use iptables to cut off the network of graphd and one of the storaged, so that we can simulate the situation where the completeness is greater than 0 and less than 100.
Method 2:
Use gdb to set the completeness to be greater than 0 and less than 100.

### Checklist

- [ Y ] I've run the tests to see all new and existing tests pass
- [ Y ] If this Pull Request resolves an issue, I linked to the issue in the text above
https://github.com/vesoft-inc/nebula/issues/2304
- [ N ] I've informed the technical writer about the documentation change if necessary

#### Test Cases
Case | Result | Whether meet expectations | Comment
-- | -- | -- | --
Go |image| Y | -
Go |image| Y | -
FetchEdges |image| Y | -
FetchVertices |image| Y | -
Find Path |image| Y | Set getFromFrontiers(..)'s completeness to 10
Find Path |image| Y | Set getToFrontiers(..)'s completeness to 10
Find Path |image| Y | Both set getFromFrontiers(..) and getToFrontiers(..)'s completeness to 10
Lookup |image| Y | -


src/common/base/Status.h
```diff
@@ -148,6 +149,7 @@ class Status final {
         kSyntaxError      = 201,
         kStatementEmpty   = 202,
         kMalformedRequest = 203,
+        kPartiallyFailed  = 204,
```
laura-ding, line 152: I don't think it's good to add error code, because there are other users who want to be successful in this situation, so you can add a `gflag` and let the users set whether to return success or failure in this situation.

whitewum: who wants the partial success as default? AFAIK, none.

laura-ding: > who wants the partial success as default? AFAIK, none.

We need to make sure that the user all think this, and why we return succeeded at the beginning.

Crazyinging: Yeah, as a storage, it is better to strictly prevent data loss. Most businesses cannot tolerate data loss, especially as a default behavior.
Indeed, we can do better here, because a few businesses may think that high availability is more important than data consistency.
So here we can consider two steps for improvement.

The first step, like this PR, return an error code by default when partial failure.

The second step is to consider whether to accept the proposal in this issue: https://github.com/vesoft-inc/nebula/issues/2304, such as allowing users to pass in class RequestConf when creating GraphClient to customize how to deal with partial failure.

dangleptr: Agreed with return error code if succeeded partially.

src/graph/FindPathExecutor.cpp
```diff
@@ -202,6 +202,12 @@ void FindPathExecutor::getNeighborsAndFindPath() {
         UNUSED(result);
         if (!fStatus_.ok() || !tStatus_.ok()) {
             std::string msg = fStatus_.toString() + " " + tStatus_.toString();
+            if (fStatus_ == Status::PartiallyFailed()
+                && tStatus_ == Status::PartiallyFailed()) {
```
bright-starry-sky, line 206: `|| `is better?

Crazyinging: > `|| `is better?

Thank you, `||` is better.

PR #2315, opened 2020-08-25
Yet another fix in raft election

critical27:
2 phase commit in election.

| | lastLogId | lastLogTerm | currentTerm |
|-|-----------|-------------|-------------|
|A|   100     |     15      |     15      |
|B|    90      |     15      |     15      |

We restart service, A start election with `proposed term` = `currentTerm` + 1 = 16, B will vote for A, B will set `currentTerm` to 16. However, the response is timeout. 

After that no one will win election. B's term is higher, so B won't vote for A. On the other hand, since `lastLogTerm` is same, but A's `lastLogId` is bigger, A won't vote for B.

### DIFF
--- a/src/kvstore/raftex/RaftPart.cpp
+++ b/src/kvstore/raftex/RaftPart.cpp
@@ -401,7 +401,7 @@ void RaftPart::preProcessTransLeader(const HostAddr& target) {
                         self->role_ = Role::CANDIDATE;
                         self->leader_ = HostAddr(0, 0);
                     }
-                    self->leaderElection();
+                    self->leaderElection(false);
                 });
             }
             break;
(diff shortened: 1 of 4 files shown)


src/kvstore/raftex/RaftPart.cpp
```diff
@@ -1011,48 +1012,57 @@ bool RaftPart::prepareElectionRequest(
     req.set_part(partId_);
     req.set_candidate_ip(addr_.first);
     req.set_candidate_port(addr_.second);
-    req.set_term(term_ + 1);
+    if (isPreVote) {
+        req.set_term(term_ + 1);
+    } else {
+        req.set_term(++term_);
```
dangleptr, line 1018: Bump up term_ before commit step?  What if timeout in this step?  The term_ will be increased crazily, it is same with the original purposeTerm.

critical27: Make sense, so use the term in prepare phase?

critical27:
Will fix this in another way, just close it.

PR #2316, opened 2020-08-25
[RAFT] fix raft snapshot by mistake

critical27:
1. When elected as leader, we need to reset all info in `Host`, which may trigger snapshot by mistake. Otherwise, if it is elected as leader more than once, the info in host is outdated.
2. Add more EventListener

@sherman-the-tank, we will open another PR to fix election problem.

### DIFF
--- a/src/kvstore/raftex/Host.h
+++ b/src/kvstore/raftex/Host.h
@@ -53,6 +53,18 @@ public:
         stopped_ = true;
     }
 
+    void reset() {
+        std::unique_lock<std::mutex> g(lock_);
+        noMoreRequestCV_.wait(g, [this] { return !requestOnGoing_; });
+        logIdToSend_ = 0;
+        logTermToSend_ = 0;
+        lastLogIdSent_ = 0;
+        lastLogTermSent_ = 0;
+        committedLogId_ = 0;
+        sendingSnapshot_ = false;
+        followerCommittedLogId_ = 0;
+    }
+
     void waitForStop();
 
     bool isLearner() const {
--- a/src/kvstore/raftex/RaftPart.cpp
+++ b/src/kvstore/raftex/RaftPart.cpp
@@ -562,7 +562,7 @@ folly::Future<AppendLogResult> RaftPart::appendLogAsync(ClusterID source,
     auto retFuture = folly::Future<AppendLogResult>::makeEmpty();
 
     if (bufferOverFlow_) {
-        PLOG_EVERY_N(WARNING, 30) << idStr_
+        PLOG_EVERY_N(WARNING, 100) << idStr_
                      << "The appendLog buffer is full."
                         " Please slow down the log appending rate."
                      << "replicatingLogs_ :" << replicatingLogs_;
@@ -1163,6 +1163,9 @@ bool RaftPart::leaderElection() {
                 std::lock_guard<std::mutex> g(raftLock_);
                 if (status_ == Status::RUNNING) {
                     leader_ = addr_;
+                    for (auto& host : hosts_) {
+                        host->reset();
+                    }
                     bgWorkers_->addTask([self = shared_from_this(),
                                        term = voteReq.get_term()] {
                         self->onElected(term);
--- a/src/kvstore/raftex/RaftPart.cpp
+++ b/src/kvstore/raftex/RaftPart.cpp
@@ -362,7 +362,7 @@ AppendLogResult RaftPart::canAppendLogs() {
         return AppendLogResult::E_STOPPED;
     }
     if (role_ != Role::LEADER) {
-        PLOG_EVERY_N(ERROR, 100) << idStr_ << "The partition is not a leader";
+        LOG_EVERY_N(ERROR, 100) << idStr_ << "The partition is not a leader";
         return AppendLogResult::E_NOT_A_LEADER;
     }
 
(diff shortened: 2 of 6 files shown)


src/kvstore/raftex/RaftPart.cpp
```diff
@@ -1317,8 +1320,10 @@ void RaftPart::processAskForVoteRequest(
         return;
     }
 
-    // Check candidate's proposed term is greater than current term
-    if (req.get_term() <= term_) {
+    // Check candidate's proposed term is greater than current term, there is one exception,
+    // if candidate's proposed term is same as current term and it is the leader we voted before,
+    // we could vote for it again because it maybe a retry
+    if (req.get_term() < term_ || (req.get_term() == term_ && leader_ != candidate)) {
```
dangleptr, line 1326: Maybe a better way is
```
if (req.term == term_ && leader_ == candidate) {
           resp.set_error_code(cpp2::ErrorCode::SUCCEEDED);	    
           resp.set_current_term(term_);
           return;
}
```

src/kvstore/raftex/Host.h
```diff
@@ -53,6 +53,22 @@ class Host final : public std::enable_shared_from_this<Host> {
         stopped_ = true;
     }
 
+    void reset() {
```
dangleptr, line 56: Well done.

src/kvstore/raftex/RaftPart.cpp
critical27: Make sense, more clear

src/kvstore/raftex/Host.h
```diff
@@ -53,6 +53,22 @@ class Host final : public std::enable_shared_from_this<Host> {
...
+        std::lock_guard<std::mutex> g(lock_);
+        if (!promise_.isFulfilled()) {
```
dangleptr, line 58: I am thinking about this logic. 
If we setValue for current promise here, what if the repsonse comes back,  reset the promise again? 
Is it legal?

src/kvstore/raftex/RaftPart.cpp
```diff
@@ -1080,6 +1080,9 @@ typename RaftPart::Role RaftPart::processElectionResponses(
         term_ = proposedTerm;
         role_ = Role::LEADER;
         leader_ = addr_;
+        for (auto& host : hosts_) {
+            host->reset();
+        }
```
bright-starry-sky, line 1085: Em...  I have  one question, because reset hosts_ at here, What will happen if appendLogsInternal failed？
I saw to assign values to hosts_ in method `Host::appendLogs`, means this part will not be able to vote leader again when `appendLogsInternal` cannot execute successfully . right?

dangleptr:
Please check the failed UT.  snapshot_command_test
BTW.  Maybe we should add a chaos plan to cover the case inside nebula-chaos

bright-starry-sky:
> Please check the failed UT. snapshot_command_test
> BTW. Maybe we should add a chaos plan to cover the case inside nebula-chaos

The test case is available in chaos repo. let me resolve the conflict just now. 
But I suggest interweaving this case into other complex tests.
https://github.com/vesoft-inc/nebula-chaos/pull/6

src/kvstore/raftex/RaftPart.cpp
```diff
@@ -1317,6 +1320,14 @@ void RaftPart::processAskForVoteRequest(
...
     }
 
+    // If candidate's proposed term is same as current term and it is the leader we voted before,
+    // we could vote for it again because it maybe a retry
+    if (req.get_term() == term_ && leader_ == candidate) {
+        resp.set_error_code(cpp2::ErrorCode::SUCCEEDED);
+        resp.set_current_term(term_);
+        return;
+    }
+
```
liuyu85cn, line 1330: talked offline already, there will be a problem if : 

precondition
A: Term 14, logid 100
B: Term 14, logid 90
C: term 14, logid 95.

Then, Network broken for a while, A, B, C begin leader election

B get request of C before A. 

B: term = 15, leader_ = C, and won't vote for A forever.

src/kvstore/raftex/RaftPart.cpp
critical27: Good question, we will change values of hosts when `Host::appendLogs`, the corner case is that if one is elected as leader of consecutive term, there maybe one request which may have not received response yet. So we wait until all response is received, see just a few lines above.

BTW, we don't use the values in `Host` in election, the info in `RaftPart` is what we used when we vote.

bright-starry-sky: > Good question, we will change values of hosts when `Host::appendLogs`, the corner case is that if one is elected as leader of consecutive term, there maybe one request which may have not received response yet. So we wait until all response is received, see just a few lines above.
> 
> BTW, we don't use the values in `Host` in election, the info in `RaftPart` is what we used when we vote.

got, thanks.

src/kvstore/raftex/RaftPart.cpp
```diff
@@ -1075,9 +1068,20 @@ typename RaftPart::Role RaftPart::processElectionResponses(
...
+            host->reset();
+        }
+        bgWorkers_->addTask([self = shared_from_this(), proposedTerm] {
+            self->onElected(proposedTerm);
+        });
+        lastMsgAcceptedTime_ = 0;
+        return true;
+    } else {
+        LOG(INFO) << idStr_ << "Didn't get enough votes of term " << proposedTerm;
+        role_ = Role::FOLLOWER;
```
wadeliuyi, line 1082: by raft protocol, if the votes has been divide by other candidate, maybe no need change the role, when it election timeout, add term and request vote again.

critical27: Indeed, the problem is that we use proposed term in election, which will make no leader is elected. We'll open another PR to fix it.

dangleptr (approved):

panda-sheep (approved):
Excellent，👍!

PR #2320, opened 2020-08-26
Exposing some metrics about ThreadManager

chenxu14:
Sometimes we don't know if the setting of num_worker_threads are suitable, or whether the threadpool’s usage is saturated.
We’d better exposing some metrics about the ThreadManager in order to get a feedback.

PR mainly do the following thing
(1) declare nebula:: thread::getThreadManager() in ThreadManager.cpp in order to build ThreadManager instance that can be used by metad, graphd and storaged.
(2) Exposing some metrics about ThreadManager, in order to get feedback on how many threads we should run with


Sophie-Xie:

PR #2321, opened 2020-08-26
Revert "Remove proposed term in raft election"

critical27:
Reverts vesoft-inc/nebula#2297. @sherman-the-tank, will modify according to discussion.


dangleptr:
Revert it, we have another easy way to fix the problem.

dangleptr (approved):

darionyaphet (approved):

bright-starry-sky (approved):

PR #2325, opened 2020-08-27
[RAFT] Add keepAlive heartBeat to keep the leader stable

dangleptr:
Currently, we use an empty log as heartBeat to keep alive for leader. 

In some cases,  appendLog request maybe blocked (for example,  writing stalll when doing compaction).  So we use a short path without writing wal and commiting to state machine to keep alive for leader.

In most sutiation,  we only need to send the light-weight keep-alive heartbeat because writing request is always on going.
But in some cases, we need the original way,  for example in metad there are few writes coming from outside. 

The pr is still in testing

### DIFF
--- a/src/kvstore/raftex/Host.cpp
+++ b/src/kvstore/raftex/Host.cpp
@@ -231,14 +231,18 @@ void Host::appendLogsInternal(folly::EventBase* eb,
                         VLOG(1) << self->idStr_
                                 << "We send nothing in the last request"
                                 << ", so we don't send the same logs again";
-                        self->followerCommittedLogId_ = resp.get_committed_log_id();
+                        if (resp.get_committed_log_id() > 0) {
+                            self->followerCommittedLogId_ = resp.get_committed_log_id();
+                        }
                         cpp2::AppendLogResponse r;
                         r.set_error_code(res);
                         self->setResponse(r);
                     } else {
                         self->lastLogIdSent_ = resp.get_last_log_id();
                         self->lastLogTermSent_ = resp.get_last_log_term();
-                        self->followerCommittedLogId_ = resp.get_committed_log_id();
+                        if (resp.get_committed_log_id() > 0) {
+                            self->followerCommittedLogId_ = resp.get_committed_log_id();
+                        }
                         if (self->lastLogIdSent_ < self->logIdToSend_) {
                             // More to send
                             VLOG(2) << self->idStr_
@@ -298,7 +302,9 @@ void Host::appendLogsInternal(folly::EventBase* eb,
                                 << ", so we don't send the same logs again";
                         self->lastLogIdSent_ = resp.get_last_log_id();
                         self->lastLogTermSent_ = resp.get_last_log_term();
-                        self->followerCommittedLogId_ = resp.get_committed_log_id();
+                        if (resp.get_committed_log_id() > 0) {
+                            self->followerCommittedLogId_ = resp.get_committed_log_id();
+                        }
                         cpp2::AppendLogResponse r;
                         r.set_error_code(cpp2::ErrorCode::SUCCEEDED);
                         self->setResponse(r);
@@ -306,7 +312,9 @@ void Host::appendLogsInternal(folly::EventBase* eb,
                         self->lastLogIdSent_ = std::min(resp.get_last_log_id(),
                                                         self->logIdToSend_ - 1);
                         self->lastLogTermSent_ = resp.get_last_log_term();
-                        self->followerCommittedLogId_ = resp.get_committed_log_id();
+                        if (resp.get_committed_log_id() > 0) {
+                            self->followerCommittedLogId_ = resp.get_committed_log_id();
+                        }
                         newReq = self->prepareAppendLogRequest();
                     }
                 }
@@ -335,7 +343,9 @@ void Host::appendLogsInternal(folly::EventBase* eb,
                     } else {
                         self->lastLogIdSent_ = self->committedLogId_;
                         self->lastLogTermSent_ = self->logTermToSend_;
-                        self->followerCommittedLogId_ = resp.get_committed_log_id();
+                        if (resp.get_committed_log_id() > 0) {
+                            self->followerCommittedLogId_ = resp.get_committed_log_id();
+                        }
                         newReq = self->prepareAppendLogRequest();
                     }
                 }
@@ -365,7 +375,9 @@ void Host::appendLogsInternal(folly::EventBase* eb,
                                 << "It means the request has been received by follower";
                         self->lastLogIdSent_ = self->logIdToSend_ - 1;
                         self->lastLogTermSent_ = resp.get_last_log_term();
-                        self->followerCommittedLogId_ = resp.get_committed_log_id();
+                        if (resp.get_committed_log_id() > 0) {
+                            self->followerCommittedLogId_ = resp.get_committed_log_id();
+                        }
                         cpp2::AppendLogResponse r;
                         r.set_error_code(cpp2::ErrorCode::SUCCEEDED);
                         self->setResponse(r);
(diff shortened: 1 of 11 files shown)


src/kvstore/raftex/RaftPart.h
```diff
@@ -486,37 +491,44 @@ class RaftPart : public std::enable_shared_from_this<RaftPart> {
...
+    HostAddr leader_;
+    Status status_;
+    Role role_;
     // The id and term of the last-sent log
     LogID lastLogId_{0};
     TermID lastLogTerm_{0};
-    // The id for the last globally committed log (from the leader)
-    LogID committedLogId_{0};
-
+    folly::Synchronized<std::vector<std::shared_ptr<Host>>> hosts_;
```
panda-sheep, line 517: 👍
Maybe we need to change `hosts_.size()` to `hosts->size()` in MemberChangeTest.cpp

src/kvstore/raftex/RaftPart.cpp
```diff
@@ -375,16 +376,18 @@ void RaftPart::addLearner(const HostAddr& addr) {
...
-    } else {
-        LOG(INFO) << idStr_ << "The host " << addr << " has been existed as "
-                  << ((*it)->isLearner() ? " learner " : " group member");
-    }
+    hosts_.withWLock([this, addr](auto& hosts) {
+        auto it = std::find_if(hosts.begin(), hosts.end(), [&addr] (const auto& h) {
+                    return h->address() == addr;
+                });
+        if (it == hosts.end()) {
+            hosts.emplace_back(std::make_shared<Host>(addr, shared_from_this(), true));
```
darionyaphet, line 384: Only append to the copy host ?

src/kvstore/raftex/Host.cpp
```diff
@@ -454,9 +469,30 @@ Host::prepareAppendLogRequest() {
...
                       << ", lastLogId in wal = " << part_->wal()->lastLogId();
             sendingSnapshot_ = true;
-            part_->snapshot_->sendSnapshot(part_, addr_)
-                .thenValue([self = shared_from_this()] (Status&& status) {
+            auto idAndTerm = part_->lastCommittedLogId();
+            part_->snapshot_->sendSnapshot(part_, addr_, idAndTerm)
+                .thenValue([self = shared_from_this(), idAndTerm] (Status&& status) {
                 if (status.ok()) {
+                    {
+                        std::lock_guard<std::mutex> g(self->lock_);
```
critical27, line 477: 👍

src/kvstore/raftex/SnapshotManager.cpp
```diff
@@ -90,6 +91,11 @@ folly::Future<Status> SnapshotManager::sendSnapshot(std::shared_ptr<RaftPart> pa
             p.setValue(Status::Error("Send snapshot failed!"));
             return false;
         });
+    };
+
+    folly::via(executor_.get(), [this, task = std::move(task)] () mutable {
+        auto* evb = executor_->getEventBase();
+        evb->runAfterDelay(std::move(task), FLAGS_snapshot_task_delay_ms);
```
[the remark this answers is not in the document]
dangleptr: It is to ensure the snapshot request happens-after the normal snapshot-appendLog request.

src/kvstore/raftex/Host.cpp
dangleptr: Check for what?

src/kvstore/raftex/RaftPart.h
dangleptr: Good point.

panda-sheep (approved):
👍

src/kvstore/raftex/RaftPart.cpp
```diff
@@ -1761,18 +1885,21 @@ void RaftPart::processSendSnapshotRequest(const cpp2::SendSnapshotRequest& req,
...
 std::vector<std::shared_ptr<Host>> RaftPart::followers() const {
     CHECK(!raftLock_.try_lock());
-    decltype(hosts_) hosts;
-    for (auto& h : hosts_) {
+    auto hosts = hosts_.copy();
+    std::vector<std::shared_ptr<Host>> followers;
+    for (auto& h : hosts) {
         if (!h->isLearner()) {
-            hosts.emplace_back(h);
+            followers.emplace_back(h);
```
wadeliuyi, line 1902: return followers after this for?

src/kvstore/raftex/RaftPart.cpp
```diff
@@ -1448,13 +1565,20 @@ void RaftPart::processAppendLogRequest(
     }
 
     // Reset the timeout timer
-    lastMsgRecvDur_.reset();
+    lastMsgRecvMs_ = time::WallClock::fastNowInMilliSec();
+    if (req.get_keep_alive()) {
+        resp.set_error_code(cpp2::ErrorCode::SUCCEEDED);
```
wadeliuyi, line 1570: no need return status code because leader do not care the response?

src/kvstore/raftex/RaftPart.cpp
```diff
@@ -1415,10 +1517,25 @@ void RaftPart::processAppendLogRequest(
...
+    // under raftLock again, it is a rare case.
+    if (req.get_keep_alive()) {
+        if (role_ == Role::FOLLOWER &&
+                req.get_current_term() == term_ &&
+                req.get_leader_ip() == leader_.first &&
+                req.get_leader_port() == leader_.second) {
+            // Because leader don't care the response, just return
+            lastMsgRecvMs_ = time::WallClock::fastNowInMilliSec();
+            return;
+         }
```
wadeliuyi, line 1536: else return directly?

src/kvstore/raftex/RaftPart.cpp
```diff
@@ -1448,13 +1565,20 @@ void RaftPart::processAppendLogRequest(
...
+        resp.set_error_code(cpp2::ErrorCode::SUCCEEDED);
+        return;
+    }
 
     if (req.get_sending_snapshot() && status_ != Status::WAITING_SNAPSHOT) {
         LOG(INFO) << idStr_ << "Begin to wait for the snapshot"
                   << " " << req.get_committed_log_id();
         reset();
         status_ = Status::WAITING_SNAPSHOT;
+        resp.set_committed_log_id(committedLogId_);
```
wadeliuyi, line 1579: this line is under lock "raftLock_", why call "set_commited_log_id" again after line 1542?

src/kvstore/raftex/RaftPart.cpp
```diff
@@ -954,36 +1001,73 @@ void RaftPart::processAppendLogResponses(
...
+                                lastLogId,
+                                committedId,
+                                prevLogTerm,
+                                prevLogId);
+        };
+
+        folly::via(eb, [eb,
+                        retryNum = retryNum_,
+                        task = std::move(task)] () mutable {
+            eb->runAfterDelay(std::move(task), retryNum);
```
wadeliuyi, line 1036: is the runAfterDelay is function:
/**
   * Runs the given Cob at some time after the specified number of
   * milliseconds.  (No guarantees exactly when.)
   *
   * Throws a std::system_error if an error occurs.
   */
  void runAfterDelay(
      Func cob,
      uint32_t milliseconds,
      InternalEnum internal = InternalEnum::NORMAL);

critical27 (approved):

src/kvstore/raftex/RaftPart.cpp
```diff
@@ -1091,7 +1175,7 @@ bool RaftPart::leaderElection() {
     using namespace folly;  // NOLINT since the fancy overload of | operator
 
     cpp2::AskForVoteRequest voteReq;
-    decltype(hosts_) hosts;
+    std::vector<std::shared_ptr<Host>> hosts;
     if (!prepareElectionRequest(voteReq, hosts)) {
         // Suppose we have three replicas A(leader), B, C, after A crashed,
         // B, C will begin the election. B win, and send hb, C has gap with B
```
ystaticy, line 1181: the sentence 'Meanwhile C begin the election' .It means there is a situation where C restarts the election after B win.
In fact, there is one condition for C to resume the election after B win,
that is, the heartbeat between B and C has timedout due to pressure and other reasons?

critical27:
Will implement in another way, close it for now.

PR #2327, opened 2020-08-31
[RAFT] Refactor leader election logic

dangleptr:
Forget the original logic,  just focus on the new one. 

The heart logic is we will use 2PC to do election,  otherwise we can't have a totally correct one.

### DIFF
--- a/src/kvstore/raftex/RaftPart.cpp
+++ b/src/kvstore/raftex/RaftPart.cpp
@@ -1620,7 +1620,7 @@ cpp2::ErrorCode RaftPart::verifyLeader(
     } else if (req.get_current_term() > term_) {
         // Leader stickness, no matter the term in Request is larger or not.
         // TODO(heng) Maybe we should reconsider the logic
-        if (leader_ != candidate
+        if (leader_ !=  std::make_pair(0, 0) && leader_ != candidate
                 && lastMsgRecvDur_.elapsedInMSec()
                         < FLAGS_raft_heartbeat_interval_secs * 1000) {
             LOG_EVERY_N(INFO, 100) << idStr_ << "I believe the leader " << leader_ << " exists. "
--- a/src/kvstore/raftex/RaftPart.cpp
+++ b/src/kvstore/raftex/RaftPart.cpp
@@ -1008,6 +1008,9 @@ bool RaftPart::prepareElectionRequest(
         return false;
     }
 
+    // Before start a new election, reset the votedAddr
+    votedAddr_ = std::make_pair(0, 0);
+
     req.set_space(spaceId_);
     req.set_part(partId_);
     req.set_candidate_ip(addr_.first);
@@ -1352,10 +1355,10 @@ void RaftPart::processAskForVoteRequest(
         }
     }
 
-    // If we have voted for somebody, we will reject other candidates under the votedTerm.
-    if (votedAddr_ != std::make_pair(0, 0) && votedTerm_ >= req.get_term()) {
+    // If we have voted for somebody, we will reject other candidates under the proposedTerm.
+    if (votedAddr_ != std::make_pair(0, 0) && proposedTerm_ >= req.get_term()) {
         LOG(INFO) << idStr_
-                  << "We have voted " << votedAddr_ << " on term " << votedTerm_
+                  << "We have voted " << votedAddr_ << " on term " << proposedTerm_
                   << ", so we should reject the candidate " << candidate
                   << " request on term " << req.get_term();
         resp.set_error_code(cpp2::ErrorCode::E_TERM_OUT_OF_DATE);
@@ -1377,7 +1380,7 @@ void RaftPart::processAskForVoteRequest(
 
     role_ = Role::FOLLOWER;
     votedAddr_ = candidate;
-    votedTerm_ = req.get_term();
+    proposedTerm_ = req.get_term();
     leader_ = std::make_pair(0, 0);
 
     // Reset the last message time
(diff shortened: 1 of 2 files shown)


critical27:
The roll back logic in follower need to be reconsidered. After rethink for a while, I don't think we can simply rollback to neither leader's commitId or follower's commitId.

critical27:
||lastLogId|commitId|lastLogTerm|
||||
|A|90|70|5|
|B|80|70|5|
|C|75|60|4|

liuyu85cn (approved):
LGTM

liuyu85cn:
> In this case, we need to rollback to exactly 65, which is the latest point of entry with same index and term.
||lastLogId|commitId|lastLogTerm|
|-|-|-|-|
|A|90|70|5|
|B|80|70|5|
|C|75|60|4|

liuyu85cn:
||60|62|64|65|
|-|-|-|-|-|
|S1|4|5|5|5|
|S2|4|5|5|6|
|S3|4|5|||
|S4|4|5|||
|S5|4||||
|S6|4||||
|S7|4||||

darionyaphet (approved):

dangleptr:
> The roll back logic in follower need to be reconsidered. After rethink for a while, I don't think we can simply rollback to neither leader's commitId or follower's commitId.

Let's take care of the rollback logic in another pr.

critical27 (approved):

src/kvstore/raftex/RaftPart.h
```diff
@@ -492,15 +492,27 @@ class RaftPart : public std::enable_shared_from_this<RaftPart> {
...
     // bumped up by 1 every time when sending out the AskForVote
     // Request
+
+    // If voted for somebody, the proposeTerm will be reset to the candidate
+    // propose term. So we could use it to prevent revote if someone else ask for
+    // vote for current proposedTerm.
+
+    // TODO(heng) We should persist it on the disk in the future
+    // Otherwise, after restart the whole cluster, maybe the stale
+    // leader still has the unsend log with larger term, and after other
```
darionyaphet, line 513: to avoid larger term with longer log which hold in locally?

darionyaphet (approved):

src/kvstore/raftex/RaftPart.cpp
```diff
@@ -1010,6 +1010,9 @@ bool RaftPart::prepareElectionRequest(
         return false;
     }
 
+    // Before start a new election, reset the votedAddr
+    votedAddr_ = std::make_pair(0, 0);
```
wadeliuyi, line 1014: why we do not vote myself? In the raft paper, we should vote for myself first.

src/kvstore/raftex/RaftPart.cpp
```diff
@@ -1362,6 +1357,16 @@ void RaftPart::processAskForVoteRequest(
         }
     }
 
+    // If we have voted for somebody, we will reject other candidates under the proposedTerm.
+    if (votedAddr_ != std::make_pair(0, 0) && proposedTerm_ >= req.get_term()) {
```
wadeliuyi, line 1361: because we do not vote myself, so this check will not true if we are a candidate. and then the other candidate will convert we from candidate to follower

src/kvstore/raftex/RaftPart.cpp
```diff
@@ -1632,54 +1614,48 @@ cpp2::ErrorCode RaftPart::verifyLeader(
...
+                    && leader_ == std::make_pair(0, 0)) {
+                LOG_EVERY_N(INFO, 100) << idStr_
+                                       << "I dont know who is leader for current term " << term_
+                                       << ", so accept the candidate " << candidate;
+                break;
+            }
+            // Same leader
+            if (role_ != Role::LEADER
+                    && candidate == leader_)  {
+                return cpp2::ErrorCode::SUCCEEDED;
```
wadeliuyi, line 1648: maybe in most case will return from this branch, is it better do this check first?

src/kvstore/raftex/RaftPart.cpp
```diff
@@ -1632,54 +1614,48 @@ cpp2::ErrorCode RaftPart::verifyLeader(
...
+                        < FLAGS_raft_heartbeat_interval_secs * 1000) {
+            LOG_EVERY_N(INFO, 100) << idStr_ << "I believe the leader " << leader_ << " exists. "
+                                   << "Refuse to append logs of " << candidate;
+            return cpp2::ErrorCode::E_WRONG_LEADER;
         }
+    } else {
+        // req.get_current_term() == term_
+        do {
+            if (role_ != Role::LEADER
+                    && leader_ == std::make_pair(0, 0)) {
```
wadeliuyi, line 1639: I want ask a question: I think a long time but I do not know what communication sequence about all member of raft will in this condition?

src/kvstore/raftex/RaftPart.cpp
dangleptr: Maybe the variable name is not clear enough. It is just uesed to recored whether current replica has voted for somebody else in the requested term.

src/kvstore/raftex/RaftPart.cpp
dangleptr: That's fine.  The branch elimination is not so important in current path.

src/kvstore/raftex/RaftPart.cpp
dangleptr: A(Leader)  B  C

C is fenced for a while then rejoin the raft group.

critical27:
The ut failed because leader changed after term 1, perhaps we need to also judge term in `waitUntilLeaderElected`.

Issue #2326, opened 2020-08-28
Unexpected Permission denied: USE mySpace; YIELD 1;


xuguruogu:
```
USE mySpace; YIELD 1;
```

xuguruogu:
Why this test should fail.

https://github.com/vesoft-inc/nebula/blob/87208af25f4ffe46fa3057119d1eb66c71f672f5/src/graph/test/PermissionTest.cpp#L58-L60

PR #2328, opened 2020-08-31
fix permission check accross spaces. free unused executor to save memory.

xuguruogu:
#2326

### DIFF
--- a/src/graph/AssignmentExecutor.cpp
+++ b/src/graph/AssignmentExecutor.cpp
@@ -49,11 +49,6 @@ Status AssignmentExecutor::prepare() {
 
 
 void AssignmentExecutor::execute() {
-    auto status = checkIfGraphSpaceChosen();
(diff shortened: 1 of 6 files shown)


CPWstatic:
```
USE ...; GO ...;
```
So It will work well now in such a situation when permission check was enabled, right?
If it is so, then that is what we want actually. And would you please add some test for it? I'm so sorry that we did not have much tests about these situation.

xuguruogu:
> ```
> USE ...; GO ...;
> ```
> 
> So It will work well now in such a situation when permission check was enabled, right?
> If it is so, then that is what we want actually. And would you please add some test for it? I'm so sorry that we did not have much tests about these situation.

Tests with no spaces chosen are added.

CPWstatic:
Thanks, this will not considered to merge into v1.0. Please pull another request for v2.0 if the problem remains.

PR #2329, opened 2020-08-31
enhance var: single element var can be used as primary expression. support global var stored in session.

xuguruogu:
https://discuss.nebula-graph.com.cn/t/topic/1134

```
$var = yield 1;
go ... where edge.prop == hash($var);
```

```
global $var = yield 1;

// the whole session can access it by:
yield $var;
```

### DIFF
--- a/src/common/filter/Expressions.cpp
+++ b/src/common/filter/Expressions.cpp
@@ -79,6 +79,8 @@ std::unique_ptr<Expression> Expression::makeExpr(uint8_t kind) {
             return std::make_unique<AliasPropertyExpression>();
         case kVariableProp:
             return std::make_unique<VariablePropertyExpression>();
+        case kVariableVariant:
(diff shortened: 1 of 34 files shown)


xuguruogu:
It seems depends on #2228

Amber1990Zhang:
guess doc is needed if merged. I'll take care this.

dangleptr:
Currently session implementation is not good enough.  After FO,  the session informaiton will be lost.  
So store the variable together with session is eally satisfed your requirements?

xuguruogu:
> Currently session implementation is not good enough. After FO, the session informaiton will be lost.
> So store the variable together with session is eally satisfed your requirements?

As multi spaces may created in same cluster. Session wise global variable is enough.

CPWstatic:
This will be considered in a later v2.0.

PR #2330, opened 2020-08-31
Fix the bug that storages can't exit properly in some cases.

monadbobo:
Use the variable meta_client_retry_interval_secs to control the retry of the meta client.

This fixes the problem that if you start the storage when the meta server is not running, the storage will not exit properly (the default is infinite retry).

### Checklist

### DIFF
--- a/src/storage/StorageServer.cpp
+++ b/src/storage/StorageServer.cpp
@@ -112,7 +112,7 @@ bool StorageServer::start() {
     metaClient_ = std::make_unique<meta::MetaClient>(ioThreadPool_,
                                                      metaAddrs_,
                                                      options);
-    if (!metaClient_->waitForMetadReady()) {
+    if (!metaClient_->waitForMetadReady(FLAGS_meta_client_retry_times)) {
         LOG(ERROR) << "waitForMetadReady error!";
         return false;
     }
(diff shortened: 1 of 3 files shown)


src/storage/StorageServer.cpp
```diff
@@ -112,7 +112,7 @@ bool StorageServer::start() {
     metaClient_ = std::make_unique<meta::MetaClient>(ioThreadPool_,
                                                      metaAddrs_,
                                                      options);
-    if (!metaClient_->waitForMetadReady()) {
+    if (!metaClient_->waitForMetadReady(FLAGS_meta_client_retry_times)) {
```
dangleptr, line 115: If pull message failed after some reties, we should ensure the storaged exit, otherwise it has risk to loss data.

monadbobo: If the meta retry fails, then waitForMetadReady will return false, where the storage will exit.

critical27: If we modify in this way, we need to start meta first. Why not stop the `MetaClient`?

monadbobo: The default wait time here (10 seconds) I think should also be enough for meta to start?

dangleptr: I am curious about why meta-client can't exit the loop after we stop it?

PR #2332, opened 2020-09-01
bug fix: shutdown meta server before job manager init, may core

liuyu85cn:
shutdown job manager if its Initialization finished

### DIFF
--- a/src/daemons/MetaDaemon.cpp
+++ b/src/daemons/MetaDaemon.cpp
@@ -306,7 +306,9 @@ void signalHandler(int sig) {
             }
             {
                 auto gJobMgr = nebula::meta::JobManager::getInstance();
-                gJobMgr->shutDown();
+                if (gJobMgr) {
+                    gJobMgr->shutDown();
+                }
             }
             if (gKVStore) {
                 gKVStore->stop();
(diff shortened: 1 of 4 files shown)


src/daemons/MetaDaemon.cpp
```diff
@@ -306,7 +306,9 @@ void signalHandler(int sig) {
             }
             {
                 auto gJobMgr = nebula::meta::JobManager::getInstance();
-                gJobMgr->shutDown();
+                if (gJobMgr) {
```
monadbobo, line 309: Here we also need to determine if gJobMgr has called the init function.

liuyu85cn: good catch

monadbobo (approved):
LGTM.

src/meta/test/JobManagerTest.cpp
```diff
@@ -13,6 +13,8 @@
 #include "kvstore/Common.h"
 #include "webservice/WebService.h"
 
+#define private public
```
dangleptr, line 16: What's this.....

liuyu85cn: black magic...
this will make test code modify private member directly. 
without declare any "FRIEND_TEST"

dangleptr (dismissed):

monadbobo (dismissed):
LGTM.

bright-starry-sky (dismissed):
LGTM

dangleptr (approved):

critical27 (approved):
Good catch

monadbobo (approved):
LGTM.

PR #2335, opened 2020-09-02
bug fix: delete vertices processor may core, if this vertex also have edges and index

liuyu85cn:
get one core screen from BBS. 

https://discuss.nebula-graph.com.cn/t/topic/1123/8

check all reference to getTagId, 
find deleteVerticesProcessor call this without check if it is a vertex first.

### DIFF
--- a/src/storage/mutate/DeleteVerticesProcessor.cpp
+++ b/src/storage/mutate/DeleteVerticesProcessor.cpp
@@ -96,14 +96,13 @@ DeleteVerticesProcessor::deleteVertices(GraphSpaceID spaceId,
         while (iter->valid()) {
             auto key = iter->key();
             if (!NebulaKeyUtils::isVertex(key)) {
-                break;  // there isn't any version of curr vertex, go to next
+                iter->next();
+                continue;
             }
             auto tagId = NebulaKeyUtils::getTagId(key);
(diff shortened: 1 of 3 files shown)


dangleptr (approved):
Well done. LGTM. Good catch

critical27 (approved):

src/storage/mutate/DeleteVerticesProcessor.cpp
```diff
@@ -95,6 +95,9 @@ DeleteVerticesProcessor::deleteVertices(GraphSpaceID spaceId,
         TagID latestVVId = -1;
         while (iter->valid()) {
             auto key = iter->key();
+            if (!NebulaKeyUtils::isVertex(key)) {
+                break;  // there isn't any version of curr vertex, go to next
+            }
```
panda-sheep, line 100:
```
   if (!NebulaKeyUtils::isVertex(key)) {
          iter->next();
          continue;
   }
```

panda-sheep (approved):
Excellent!

bright-starry-sky:
A question about header : "delete vertices processor may core, if this vertex also have edges and index", where is index ? I think there are only edges. right?

bright-starry-sky (approved):
LGTM

Issue #2333, opened 2020-09-02
can not build using glibc 2.32 and above / can not run using kernel 5.5 and above


Sasasu:
**Describe the bug**

```shell
[ 42%] Building CXX object src/graph/CMakeFiles/graph_obj.dir/TraverseExecutor.cpp.o
/usr/bin/ld.bfd: /home/sa/W/nebula_project/nebula/build/third-party/install/lib/libevent.a(evutil_rand.o):(.text+0x2a3): undefined reference to `sysctl'
collect2: error: ld returned 1 exit status
```

**Your Environments**

* OS: `kernel 5.8.5` `glibc 2.32`
* Compliler: `g++ 10.2.0`

**How To Reproduce**

glibc 2.30 rename the `sysctl` function to `_sysctl` and mark it as deprecated.  https://lwn.net/Articles/795127/

    * The Linux-specific <sys/sysctl.h> header and the sysctl function have been
       deprecated and will be removed from a future version of glibc.
       Application should directly access /proc instead.  For obtaining random
       bits, the getentropy function can be used. 

glibc 2.32 removed `_sysctl` fincuton. so build using glibc 2.32 and above will get `undefined reference to sysctl` during linking.

and linux kernel removed `sysctl` syscall after kernel 5.5 if `CONFIG_SYSCTL_SYSCALL` is not set.

**Possible solution**

1. do not use `sysctl`

libevent already fixed this compile error in https://github.com/libevent/libevent/commit/71fc3eb.

re-config libevent, make sure `sys/sysctl.h` is not exist.

https://github.com/libevent/libevent/blob/02428d9a2d89ee0a595166909dd6d75b5beb777b/CMakeLists.txt#L621-L622

`CHECK_CONST_EXISTS` is a macro, the definition is
https://github.com/libevent/libevent/blob/master/cmake/CheckConstExists.cmake

2. implement a mini version of `sysctl` since `sysctl` only used to generate random number.

Sasasu:
solved by rebuild third-party
```sh
export CC=(which gcc-9)
export CXX=(which g++-9)
bash third-party/build-third-party.sh
```

dutor:
Thanks a lot for reporting this issue. We will update all dependencies recently.

PR #2338, opened 2020-09-05
some fix for third-part build

Sasasu:
## What I want to do
build nebula graph on ArchLinux, I need to rebuild all third-party and create a new nebula build toolchain.

glibc 2.27 is too old #2333 

this PR fix third-party building on glibc 2.32 / gcc 10 / kernel 5.8.5 / cmake 3.18

almost in third-party cmake file, not affect existing code.

no.

there is two major change:
  - add `-fcommon` when using GCC 10
  - add `Boost_NO_SYSTEM_PATHS` to make sure using vendor boost.

the first problem is very easy to reproduce: add `-fno-common` to CFLAGS, rebuild third-party using GCC 9. krb5 and libunwind get an error when linking.

the second problem: install a new Ubuntu 19.04. run `bash third-party/build-third-party.sh` got this warning
    
    New Boost version may have incorrect or missing dependencies and imported targets.

and after check the link script, when building fbthrift and folly, LD will link system default boost, not the vendor boost. if install a new boost on Ubuntu 19.04 will get link error.

the last problem is just a warning about NCURSES library naming.

### Checklist
- [x] If this Pull Request resolves an issue, I linked to the issue in the text above #2333


dutor (commented):
Thanks for addressing these issues.

Could you please help us to test this patch on some old distro, e.g. Ubuntu 18.04?

PR #2340, opened 2020-09-07
Run command at background to avoid fork.

KinhanS:

This PR is for avoiding the process forking during importing data by SST file.

The forking during importing degrades the query performance at the same time.

No

No

### Checklist


xuguruogu:
All subcommand should run in the background process, otherwise, fork will cause huge latency caused by large page table copy.
https://discuss.nebula-graph.com.cn/t/topic/1205

dutor:
IMO. It will be more elegant to implement a faster and better `popen` via `clone(CLONE_VM)` & `pipe` & `exec`

dangleptr:
Agreed with @dutor 
Can't we use copy on write mechanism when forking sub process?

Share memory between processes is too heavy.

xuguruogu:
https://redis.io/topics/latency#latency-generated-by-fork

Fork seems to me will block the whole process. The latency linearly grows according to process memory size. I am not sure if clone(CLONE_VM) can communicate with the parent process easily, as no object the child process can inherit from the parent process.

dutor:
> Agreed with @dutor
> Can't we use copy on write mechanism when forking sub process?
> 
> Share memory between processes is too heavy.

I think the latency spikes were just introduced by the very COW mechanism of fork.

Here is the rationale: `popen` mainly performs a `fork`, creates a pipe and then an `exec`. If the hosting service is busy on writing the memory, a lot of page faults will be triggered and page frame will be copied.

dutor:
> https://redis.io/topics/latency#latency-generated-by-fork
> 
> Fork seems to me will block the whole process. The latency linearly grows according to process memory size. I am not sure if clone(CLONE_VM) can communicate with the parent process easily, as no object the child process can inherit from the parent process.

Here the benefits of using `clone` is to launch a child process without creating a separate page table and mark the PTE as write-only. Instead, the child process shares the address space with the parent, just like a thread, allowing to perform a `exec` at meantime(disallowed in the thread case).

A better engineered `popen` could be reused for other purposes. And I think it's more elegant than keeping a long running background __process__

dutor:
Anyway, this patch is a positive one to me. We could adopt it.

dangleptr:
> If the hosting service is busy on writing the memory, a lot of page faults will be triggered and page frame will be copied. Because there is a time window between the fork and exec. 

Agreed.

CPWstatic (approved):

xuguruogu:
A more common solution may be using _vfork_: 

https://github.com/ClickHouse/ClickHouse/blob/master/src/Common/ShellCommand.cpp#L108-L143

dutor (approved):

jamieliu1023:
@KinhanS Thanks for your contribution! This is Jamie Liu with Nebula Graph. To memorize this special moment, we'd like to send you the Nebula Contributor certificate which is exclusively for you via email. Also, we want to ship you a mug to mark the contribution. Could you kindly let me know your email address and where we should ship the mug? You can reach me via jamie.liu at vesoft.com  Thanks again!

PR #2341, opened 2020-09-07
fixed error for lookup after alter tag add column

bright-starry-sky:
This bug from discuss: 
https://discuss.nebula-graph.com.cn/t/topic/1209

Repeat steps： 
CREATE SPACE alter_space(partition_num=1, replica_factor=1)
USE alter_space
CREATE TAG alter_tag(id int)
CREATE TAG INDEX alter_index ON alter_tag(id)
INSERT VERTEX alter_tag(id) VALUES 100:(1), 200:(2)
ALTER TAG alter_tag ADD (type int)
LOOKUP ON alter_tag WHERE alter_tag.id == 1 YIELD alter_tag.type

### DIFF
--- a/src/storage/BaseProcessor.inl
+++ b/src/storage/BaseProcessor.inl
@@ -131,8 +131,16 @@ BaseProcessor<RESP>::collectIndexValues(RowReader* reader,
             auto val = NebulaKeyUtils::encodeVariant(value(std::move(res)));
             values.emplace_back(col.get_type().get_type(), std::move(val));
         } else {
-            LOG(ERROR) << "Skip bad column prop : " << col.get_name();
-            return Status::Error("Skip bad column prop : %s", col.get_name().c_str());
+            VLOG(1) << "Bad value for prop: " << col.get_name();
+            // TODO: Should return NULL
+            auto defaultVal = RowReader::getDefaultProp(col.get_type().get_type());
+            if (!defaultVal.ok()) {
+                // Should never reach here.
+                return Status::Error("Skip bad type of prop : %s", col.get_name().c_str());
+            } else {
+                auto val = NebulaKeyUtils::encodeVariant(std::move(defaultVal).value());
+                values.emplace_back(col.get_type().get_type(), std::move(val));
+            }
         }
     }
     return values;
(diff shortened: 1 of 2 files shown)


dangleptr:
Could you explain it in detail why crash?

critical27:
> Could you explain it in detail why crash?

It seems storage did't return the default value of the newly added field in look up.

critical27 (approved):
LGTM

dangleptr:
> > Could you explain it in detail why crash?
> 
> It seems storage did't return the default value of the newly added field in look up.

So just return the default type value no matter whether the field exist or not?

critical27:
> > > Could you explain it in detail why crash?
> > 
> > 
> > It seems storage did't return the default value of the newly added field in look up.
> 
> So just return the default type value no matter whether the field exist or not?

I would prefer return the default value defined in schema, if user didn't specify default value, then return the default value of value type. @bright-starry-sky WDYT?

dangleptr:
> > > > Could you explain it in detail why crash?
> > > 
> > > 
> > > It seems storage did't return the default value of the newly added field in look up.
> > 
> > 
> > So just return the default type value no matter whether the field exist or not?
> 
> It seems most of the logic in GO and FETCH return the default value of field type for now, see `QueryBaseProcessor<REQ, RESP>::collectProps`. I guess this PR follow this function.
> 
> In general, I would prefer return the default value defined in schema, if user didn't any specify default value, then return the default value of field type.

OK  That's fine.

dangleptr (approved):

bright-starry-sky:
> Could you explain it in detail why crash?

This crash error is caused by the yield clause of the Lookup statement, if the prop value get failed of yield field , this value is set to an empty string in storage layer, so a crash error caused when decode prop value in graph layer.
the reason of prop value get failed : when yield including no index filed , we need to get the data val, but the data val is old schema version, the new fields have not yet been included .

src/storage/BaseProcessor.inl
```diff
@@ -146,11 +154,21 @@ void BaseProcessor<RESP>::collectProps(RowReader* reader,
...
+            VariantType v;
             if (!ok(res)) {
-                VLOG(1) << "Skip the bad value for prop " << name;
-                continue;
+                VLOG(1) << "Bad value for prop: " << name;
+                // TODO: Should return NULL
+                auto defaultVal = RowReader::getDefaultProp(prop.type_.type);
+                if (!defaultVal.ok()) {
+                    // Should never reach here.
+                    LOG(FATAL) << "Get default value failed for " << name;
```
darionyaphet, line 164: If the name not found in schema and the default value is not defined ?

bright-starry-sky:
> > > > Could you explain it in detail why crash?
> > > 
> > > 
> > > It seems storage did't return the default value of the newly added field in look up.
> > 
> > 
> > So just return the default type value no matter whether the field exist or not?
> 
> It seems most of the logic in GO and FETCH return the default value of field type for now, see `QueryBaseProcessor<REQ, RESP>::collectProps`. I guess this PR follow this function.
> 
> In general, I would prefer return the default value defined in schema, if user didn't any specify default value, then return the default value of field type.

Yes, you are right . this PR follow `QueryBaseProcessor<REQ, RESP>::collectProps`， I agree with you for return the default value defined in schema. But there is a problem, reader parsed with Val corresponds to the older version of schema, which does not yet have new field. That would be a big logical change, so a small change based on the existing logic.

src/storage/BaseProcessor.inl
bright-starry-sky: > If the name not found in schema and the default value is not defined ?

Emm.. have not using schema at here, it simply returns A default value based on the type, not the default value of the schema itself.

PR #2356, opened 2020-10-03
add flags force all spaces have only one partition

knightXun:
rt, this is for cloud

### DIFF
--- a/src/meta/processors/partsMan/CreateSpaceProcessor.cpp
+++ b/src/meta/processors/partsMan/CreateSpaceProcessor.cpp
@@ -103,7 +103,7 @@ void CreateSpaceProcessor::process(const cpp2::CreateSpaceReq& req) {
         return;
     }
 
-    if ( force_single_part ) {
+    if (FLAGS_force_single_part) {
         properties.set_partition_num(1);
     }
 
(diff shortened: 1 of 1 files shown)


src/meta/processors/partsMan/CreateSpaceProcessor.cpp
```diff
@@ -102,6 +103,10 @@ void CreateSpaceProcessor::process(const cpp2::CreateSpaceReq& req) {
         return;
     }
 
+    if ( force_single_part ) {
```
yixinglu, line 106: `FLAGS_force_single_part`

knightXun: good point

CPWstatic (commented):
Would it be better put this global config in graph?

src/meta/processors/partsMan/CreateSpaceProcessor.cpp
```diff
@@ -102,6 +103,10 @@ void CreateSpaceProcessor::process(const cpp2::CreateSpaceReq& req) {
         return;
     }
 
+    if (FLAGS_force_single_part) {
```
Shylock-Hg, line 106: I think you can set the FLAGS_max_parts_num to 1 instead of add a new flags.

knightXun: No, FLAGS_max_parts_num is for total space partitions, not for single space partition

Shylock-Hg: Ok.

PR #2359, opened 2020-10-09
Fix fetch * return wrong result when the vertex with empty props tag

### LINKED ISSUE #2351: no data is returned for `fetch prop on *`

No data is returned for `fetch prop on * 1`. But `fetch prop on <tagname> <vid>` has the expected result. 

**Your Environments (__must be provided__)**

Docker-compose (with the nightly image) + Studio v1.1.1 beta

I have two data files:  
* user.csv: contains one column only. int, and from 0 to 7000.
* course.csv: contains two columns. one is of the int type (range: from 0 to 96), and the other is of the string type.

1. run `create tag user()` to create a tag without a property.
2. import data from user.csv.
3. run `fetch prop on * 1`, and the expected result is returned.
4. run `create tag course(id int, name string)` to create another tag, named course. Hash("string") is used to generate VIDs.
5. import data from course.csv. 
6. run `fetch prop on * 1`. no data is returned. But when I run `fetch prop on user 1`, the expected result is returned. 

**Expected behavior**

After both user and course data is imported, when I run `fetch prop on * 1`, I should get the expected user data. 

**Additional context**

NO

laura-ding:
close #2351

### DIFF
--- a/src/graph/FetchVerticesExecutor.cpp
+++ b/src/graph/FetchVerticesExecutor.cpp
@@ -118,9 +118,6 @@ Status FetchVerticesExecutor::prepareTags() {
             if (!tagIdStatus.ok()) {
                 return tagIdStatus.status();
             }
-            auto tagId = tagIdStatus.value();
-            tagNames_.push_back(tagName);
-            tagIds_.push_back(tagId);
             auto result = tagNameSet_.emplace(tagName);
             if (!result.second) {
                 return Status::Error(folly::sformat("tag({}) was dup", tagName));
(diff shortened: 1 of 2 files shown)


src/graph/FetchVerticesExecutor.cpp
```diff
@@ -118,9 +118,6 @@ Status FetchVerticesExecutor::prepareTags() {
             if (!tagIdStatus.ok()) {
                 return tagIdStatus.status();
             }
```
CPWstatic, line 120: Seems we don't need it anymore.

laura-ding: when `fetch prop on * vid yield tag.prop` need to check the yield

CPWstatic: I mean the tag id is no more used here.

laura-ding: Good catch!

CPWstatic (approved):

Shylock-Hg (approved):

PR #2362, opened 2020-10-10
Catch exception in set op.

CPWstatic:


src/graph/SetExecutor.cpp
```diff
@@ -114,18 +108,41 @@ void SetExecutor::execute() {
     }
 
     auto *runner = ectx()->rctx()->runner();
-    runner->add([this] () mutable { left_->execute(); });
-    runner->add([this] () mutable { right_->execute(); });
 
-    auto cb = [this] (auto &&result) {
-        UNUSED(result);
-        if (!leftS_.ok() || !rightS_.ok()) {
+    auto lTask = [this] () mutable { left_->execute(); };
```
dutor, line 112: Is `mutable` necessary?

src/graph/SetExecutor.cpp
```diff
@@ -114,18 +108,41 @@ void SetExecutor::execute() {
...
+    auto lError = [this] (auto &&e) {
+        std::stringstream ss;
+        ss << "Exception when handle left subquery: " << e.what();
+        LOG(ERROR) << ss.str();
+        leftS_ = Status::Error(ss.str());
+    };
+    auto lF = std::move(folly::makeFuture()).via(runner).then(lTask).thenError(lError);
```
dutor, line 119: If `makeFuture` returns a future by value, the `std::move` is not needed. Am I right?

dutor (dismissed):

laura-ding (approved):

src/graph/SetExecutor.cpp
CPWstatic: Not anymore.

laura-ding (approved):

Shylock-Hg (approved):

PR #2376, opened 2020-10-22
Fixed filter error for index scan

bright-starry-sky:


src/storage/index/IndexPolicyMaker.cpp
```diff
@@ -78,6 +78,12 @@ bool IndexPolicyMaker::buildPolicy() {
         if (exist == scanItems_.end()) {
             break;
         }
+        // Stop build policy when last operator is range scan,
+        // And other fields will use expression filtering.
+        auto it = scanItems_.rbegin()->second;
```
critical27, line 83: Is `rbegin` wrong? We can't make sure that the last index column is the `rbegin` of `scanItems`.

critical27:
index (c1, c2 ,c3), where c1 == 1 and c2 > 1 and c2 < 3 and c3 == 4, [begin, end) is [1_2, 1_3)

src/storage/index/IndexPolicyMaker.cpp
bright-starry-sky: > Is `rbegin` correct? We can't make sure that the last index column is the `rbegin` of `scanItems`.

I think this is correct. we just have to make sure it's the last one in scanItems_, not index fields. WDYT?

bright-starry-sky:
> index (c1, c2 ,c3), where c1 == 1 and c2 > 1 and c2 < 3 and c3 == 4, [begin, end) is [1_2, 1_3)

yeah.

src/storage/index/IndexPolicyMaker.cpp
critical27: But the scanItems_ is a map<index_column_name, ScanBound>? The `rbegin` is the last element of a map, not the column name in index order?

bright-starry-sky: > But the scanItems_ is a map<index_column_name, ScanBound>? The `rbegin` is the last element of a map, not the column name in index order?

you are right. my bad.

bright-starry-sky:
Addressed critical27's comment. Thanks.

critical27 (approved):
LGTM

panda-sheep (approved):
Well done！

PR #2377, opened 2020-10-24
Try to use whole key bloom filter when insert data with index

critical27:
As title, since the default value of enable_multi_versions is false, we could use the bloom filter to avoid extra disk read if the newly insert data does not exist.


bright-starry-sky (approved):
Great work. LGTM

liuyu85cn (approved):

PR #2379, opened 2020-10-26
Ignore existed index

critical27:
1. Add a keyword IGNORE_EXISTED_INDEX in `insert`, if it is specified, storage will not delete old index. As `update` will always try to read old data, so only `insert` is taken care of.
2. Slightly modify insert process in storage, we only need to initialize the old reader and new reader once.

### DIFF
--- a/src/graph/InsertEdgeExecutor.cpp
+++ b/src/graph/InsertEdgeExecutor.cpp
@@ -37,6 +37,7 @@ Status InsertEdgeExecutor::check() {
 
     spaceId_ = ectx()->rctx()->session()->space();
     overwritable_ = sentence_->overwritable();
+    ignoreExistedIndex_ = sentence_->ignoreExistedIndex();
     auto edgeStatus = ectx()->schemaManager()->toEdgeType(spaceId_, *sentence_->edge());
     if (!edgeStatus.ok()) {
         status = edgeStatus.status();
@@ -247,7 +248,8 @@ void InsertEdgeExecutor::execute() {
 
     auto future = ectx()->getStorageClient()->addEdges(spaceId_,
                                                        std::move(result).value(),
-                                                       overwritable_);
+                                                       overwritable_,
+                                                       ignoreExistedIndex_);
     auto *runner = ectx()->rctx()->runner();
 
     auto cb = [this] (auto &&resp) {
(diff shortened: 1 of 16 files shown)


bright-starry-sky (commented):
I think , if ignoreExistedIndex_ == true, We don't have to go through `asyncAtomicOp`. It will affect performance. WDYT?
I suggest doing it directly in method process(....)  when ignoreExistedIndex_ == true.

critical27:
> I think , if ignoreExistedIndex_ == true, We don't have to go through `asyncAtomicOp`. It will affect performance. WDYT?
> I suggest doing it directly in method process(....) when ignoreExistedIndex_ == true.

Make sense

bright-starry-sky:
Another question, We need to get the data value via indexKey for lookup. At this point, an incorrect logical condition will occur. for example : 
tag (c1, c2)
index on tag (c1)
INSERT VERTEX tag(c1, c2) VALUES 1:("c1_1", "c2_1") 
INSERT IGNORE_EXISTED_INDEX VERTEX tag(c1, c2) VALUES 1:("c1_2", "c2_2") 
lookup on tag where c1 == "CCC" yield c1, c2
the invalid result should be return {"c1_1", "c2_1"}.

critical27:
> Another question, We need to get the data value via indexKey for lookup. At this point, an incorrect logical condition will occur. for example :
> tag (c1, c2)
> index on tag (c1)
> INSERT VERTEX tag(c1, c2) VALUES 1:("c1_1", "c2_1")
> INSERT IGNORE_EXISTED_INDEX VERTEX tag(c1, c2) VALUES 1:("c1_2", "c2_2")
> lookup on tag where c1 == "c1_1" yield c1, c2
> the invalid result should be return {"c1_2", "c2_2"}.

Exactly, the lookup result will be dirty If a key is inserted several times with different value.

liuyu85cn (approved):

panda-sheep:
I have a question. Since this method causes the lookup result to be wrong,
Why not delete the index and insert the data directly?

critical27:
> I have a question. Since this method causes the lookup result to be wrong,
> Why not delete the index and insert the data directly?

Poor performance. Delete also need to read the value, build index and delete it. This PR is used to bypass delete old index, only for a few scenario in poc.

critical27:
Will move to enterprise version.

Issue #2385, opened 2020-10-30
Go n-hop query hanging and graphd takes up too much memory util OOM


runnerls:
**Describe the bug (__must be provided__)**
 
Use GO statement, n-hop query hanging and graphd takes up too much memory util Killed by OOM Killer.

**Your Environments (__must be provided__)**

* OS: `CentOS Linux release 7.1.1503 (Core)`
* Compliler: `g++ 7.5.0`
* CPU: `32U`
* Commit id (`nebula v1.1.0`)

**How To Reproduce(__must be provided__)**

1. Create graph: Use NBA data provided by Nebula. https://oss-cdn.nebula-graph.io/doc/nba.ngql
2. Execute sql: GO 1 to 40 STEPS FROM 100 OVER follow YIELD follow._dst as dst.

**Additional context**

![image](https://user-images.githubusercontent.com/6389830/97690856-86b1f600-1ad8-11eb-9c06-e26563e7f893.png)

PR #2386, opened 2020-10-30
Fix go n-hop query hanging, and graphd takes up too much memory util OOM 

runnerls:

https://github.com/vesoft-inc/nebula/issues/2385

All test cases passed
### Checklist

### DIFF
--- a/src/graph/GoExecutor.cpp
+++ b/src/graph/GoExecutor.cpp
@@ -681,7 +681,7 @@ std::vector<VertexID> GoExecutor::getDstIdsFromRespWithBackTrack(const RpcRespon
     // 7  ,  6
     // Will mistake lead to 7->6 if insert edge(dst, src) one by one
     // So read all roots of current step first , then insert them
-    std::multimap<VertexID, VertexID> backTrace;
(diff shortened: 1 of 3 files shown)


CPWstatic (dismissed):
LGTM.

src/graph/GoExecutor.cpp
```diff
@@ -681,7 +681,7 @@ std::vector<VertexID> GoExecutor::getDstIdsFromRespWithBackTrack(const RpcRespon
     // 7  ,  6
     // Will mistake lead to 7->6 if insert edge(dst, src) one by one
     // So read all roots of current step first , then insert them
-    std::multimap<VertexID, VertexID> backTrace;
+    std::multiset<std::pair<VertexID, VertexID>> curBackTrace;
```
Shylock-Hg, line 684: Why not use unordered_set?

runnerls: Using unorderd_set here looks more effecient, I have modified it, please review.

CPWstatic (approved):
Thanks for the improvement.

laura-ding (approved):
Thanks very much!

PR #2115, opened 2020-05-08
Use "-- " as comment prefix

laura-ding:

PR #2388, opened 2020-11-02
cherry-pick commits from master

knightXun:
pick-up commits before May 12


knightXun:
nicely ping @dangleptr @dutor

src/parser/scanner.lex
```diff
@@ -543,7 +543,7 @@ RECOVER                     ([Rr][Ee][Cc][Oo][Vv][Ee][Rr])
                             }
 "#".*                       // Skip the annotation
 "//".*                      // Skip the annotation
-"--".*                      // Skip the annotation
+"-- ".*                     // Skip the annotation
```
dutor, line 546: If there is no special reason, please drop this commit(#2115). Because it will not be supported any more in 2.0

knightXun: Can I Mark it and Push a pr to fix this after the all commits merged into cloud? Otherwise Testing is very time-comsuming

PR #2403, opened 2020-11-18
MetaClient always send to leader of meta server

critical27:
For safety, always send to leader.


panda-sheep (approved):
Good job，LGTM.

darionyaphet:
If we always sent to the leader still need argument to check ?

liuyu85cn (approved):
LGTM

PR #2404, opened 2020-11-23
use range in delete edges

critical27:
Use `delete range` in delete edges, there are some encode problem that we can not use it in delete vertices, contact me if interested.

Other changes:
1. lower flush log level in `EventListener`
2. fix leader not returned in scan


bright-starry-sky (approved):
LGTM

liuyu85cn (approved):
LGTM

Issue #2405, opened 2020-11-24
Raft:a server send 2 vote to different server in one term


ystaticy:
In 1.1.0.
proposedTerm_ can be set to a small value in verifyLeader as follows;
But I think it should Increase forever, because it's used to determine whether to vote for others in newest term.

------------------------------------------

if we have 3 nodes: 'A','B','C' 
At a normal state, 'C' is leader term=4,'A','B' is follower. 'C' keep sending appenlog RPC(term 4);

------------------------------------------

at the next monment B become term 5 's leader;
A log OP_TRANS_LEADER in log to commit;
B print log  'Give up my leadership'. B.role_ change to FOLLOWER;

------------------------------------------

at the next monment 'B','C' heartbeat timeout  ,'B' send askVote RPC (term=6) ,and  'C' send askVote RPC(term=6);

------------------------------------------

then 'A' received 'B' askVote RPC，A print ‘will vote for the candidate' (A return SUCCEEDED response to B),
and 'A' set **A.proposedTerm_=term6** ;But 'A'.term_ is still 4(hasn't received the heartbeat from term 5).
(this moment, appendlog RPC(term=4) from C may be still in process )

at the next monment  'A' just execute verifyleader appendlog RPC from C (term 4)
'A' execute :
`term_ = proposedTerm_ = req.get_current_term()`
now **A.proposedTerm_=4**

------------------------------------------

now 'A' received 'C' askVote RPC(term 6) , due to  A.proposedTerm_=4，A vote for C

now there are two leader in term 6，in A,B,C .
------------------------------------------
'A' log as follows
![image](https://user-images.githubusercontent.com/3659421/100064139-172bee00-2e6d-11eb-89da-4eaa812140e7.png)

ystaticy:
the pull request is  https://github.com/vesoft-inc/nebula/pull/2406

critical27:
Thanks for the feedback. It is ok to vote twice to different peer in same term. One key difference in code for now is that, the vote is more like a `prevote`, not vote in the paper. Even though B and C could become leader, A will accept one of B/C in `verifyLeader` when they send appendLogs or heartbeat, depends on which one received first. I agree with you that it could be misleading.

ystaticy:
After 'A vote for B' and 'A vote for C'
finally A accept C
but now B.role_=LEADER  and C.role_=LEADER;

[...]

PR #2406, opened 2020-11-24
a server send 2 vote to different server in one term

ystaticy:
the issue describe is https://github.com/vesoft-inc/nebula/issues/2405


Sophie-Xie:

PR #2415, opened 2020-12-09
LOOKUP sentence supports DISTINCT and supports yield EdgeType._src/._dst

LuconYang:

1. LOOKUP语句现在支持了DISTINCT功能
LOOKUP ON {<vertex_tag> | <edge_type>} WHERE <expression> YIELD DISTINCT <return_list>
2. 针对边类型的LOOKUP语句，支持yield子句中指定输出边的起始顶点(._src)和终止顶点(._dst)
LOOKUP ON <edge_type> WHERE <expression> YIELD <edge_type>._src, <edge_type>._dst
3. 在LookupTest.cpp中添加了对应的测试用例。
    边类型的lookup语句，distinct测试：TEST_F(LookupTest, EdgeDistinctTest)
    点类型的lookup语句，distinct测试：TEST_F(LookupTest, VertexDistinctTest)
    边类型的lookup语句输出起止顶点id：TEST_F(LookupTest, EdgeYieldVertexId)
     
1. 业务方的使用场景需要对LOOKUP语句的属性列进行去重
2. 业务方的使用场景中，对于边类型的LOOKUP语句，需要指定输出边的起始顶点或者终止顶点，并参与去重逻辑

语法存在变更，同时兼容旧版本。改动如下：
1. LOOKUP语句中，DISTINCT之前版本是合法关键字，但未生效，本版本使其能正常去重
2. 边类型的LOOKUP语句，支持yield子句中输出边的起始顶点(._src)和终止顶点(._dst)

1. LOOKUP语句中的DISTINCT关键字，之前没有效果，现在能起到去重作用
示例：
lookup on player where player.name=='Danny' yield distinct player.age;
2. 对边执行LOOKUP语句，现在支持输出边的起止顶点（可同时结合DISTINCT关键字，对点进行去重）
示例：
lookup on follow where follow.degree==70 yield [distinct] follow._src, follow._dst;

在LookupTest.cpp中添加了3组测试用例，均通过
### Checklist


src/graph/LookupExecutor.cpp
```diff
@@ -126,15 +130,23 @@ Status LookupExecutor::prepareYield() {
...
+            const AliasPropertyExpression* aExpr;
+            if (prop->kind() == Expression::kAliasProp) {
+                aExpr = dynamic_cast<const AliasPropertyExpression*>(prop);
+            } else if (isEdge_ && prop->kind() == Expression::kEdgeSrcId) {
+                aExpr = dynamic_cast<const EdgeSrcIdExpression*>(prop);
+            } else if (isEdge_ && prop->kind() == Expression::kEdgeDstId) {
+                aExpr = dynamic_cast<const EdgeDstIdExpression*>(prop);
+            } else {
+                return Status::SyntaxError("Expressions other than "
+                                           "AliasProp/kEdgeSrcId/kEdgeDstId are not supported");
```
bright-starry-sky, line 142: Is there a problem with the `dynamic_cast` here?

src/graph/LookupExecutor.cpp
```diff
@@ -126,15 +130,23 @@ Status LookupExecutor::prepareYield() {
...
             }
-            auto* aExpr = dynamic_cast<const AliasPropertyExpression*>(prop);
             auto st = checkAliasProperty(aExpr);
             if (!st.ok()) {
                 return st;
             }
-            if (schema->getFieldIndex(aExpr->prop()->c_str()) < 0) {
+            if (prop->kind() == Expression::kAliasProp &&
+                schema->getFieldIndex(aExpr->prop()->c_str()) < 0) {
```
bright-starry-sky, line 149: So there's some confusion here by the logic above

src/graph/LookupExecutor.cpp
```diff
@@ -491,7 +511,14 @@ std::vector<std::shared_ptr<nebula::cpp2::IndexItem>> LookupExecutor::findIndexF
...
     auto filter = Expression::encode(sentence_->whereClause()->filter());
-    auto future  = sc->lookUpIndex(spaceId_, index_, filter, returnCols_, isEdge_);
+    // Get the columns that really need to be requested
+    std::vector<std::string> requestCols_;
+    for (auto& column : returnCols_) {
+        if (column != _SRC && column != _DST) {
+            requestCols_.emplace_back(column);
+        }
+    }
+    auto future  = sc->lookUpIndex(spaceId_, index_, filter, requestCols_, isEdge_);
```
bright-starry-sky, line 521: How to identify `_ID` of vertex?

src/graph/LookupExecutor.cpp
LuconYang: Thanks for your review! EdgeSrcIdExpression and EdgeDstIdExpression both extend AliasPropertyExpression, and I tested these logical branches, no problem found here.

src/graph/LookupExecutor.cpp
LuconYang: The logic here is to judge whether the current attribute column is valid only when the value of `prop->kind() == Expression::kAliasProp` is true. In other words, when `isEdge_ && prop->kind() == Expression::kAliasProp` or `isEdge_ && prop->kind() == Expression::kEdgeDstId` is true  , there is no need to go through the logic here

src/graph/LookupExecutor.cpp
LuconYang: This PR only supports specifying  and outputing the start and end vertice ID of the edge in the Yield clause, and It's also consistent with the implementation of Go Sentence. 
Actually, the judgment of `prop->kind()` in `prepareYield()` method has ensured that it is illegal to specify the vertex id, so there is no need to judge this situation here.

src/graph/LookupExecutor.cpp
bright-starry-sky: > Thanks for your review! EdgeSrcIdExpression and EdgeDstIdExpression both extend AliasPropertyExpression, and I tested these logical branches, no problem found here.

Got , if so , I think below code is better

```
if (prop->kind() == Expression::kAliasProp) {
                 aExpr = static_cast<const AliasPropertyExpression*>(prop);
             } else if (isEdge_ && prop->kind() == Expression::kEdgeSrcId) {
                 aExpr = static_cast <const AliasPropertyExpression*>(prop);
             } else if (isEdge_ && prop->kind() == Expression::kEdgeDstId) {
                 aExpr = static_cast <const AliasPropertyExpression*>(prop);
             } else {
                      return ERROR;
             }
```

OR 

```
if (prop->kind() != Expression::kAliasProp && isEdge_) {
   if (prop->kind() != Expression::kEdgeSrcId || prop->kind() != Expression::kEdgeDstId) {
                      return ERROR;
  }
}
auto* aExpr = dynamic_cast<const AliasPropertyExpression*>(prop);
```

src/graph/LookupExecutor.cpp
bright-starry-sky: > This PR only supports specifying and outputing the start and end vertice ID of the edge in the Yield clause, and It's also consistent with the implementation of Go Sentence.
> Actually, the judgment of `prop->kind()` in `prepareYield()` method has ensured that it is illegal to specify the vertex id, so there is no need to judge this situation here.

Got, excellent.

bright-starry-sky:
Thanks for your code contribution. I have a little suggestion , Let's solve the problem once and for all : 
I thought it over, the best way is push down the YIELD-DISTINCT clause to storage layer. avoid redundant data in the result set. WDYT?

src/graph/LookupExecutor.cpp
LuconYang: All right. I have completed the modification as suggested, please check it.

LuconYang:
> Thanks for your code contribution. I have a little suggestion , Let's solve the problem once and for all :
> I thought it over, the best way is push down the YIELD-DISTINCT clause to storage layer. avoid redundant data in the result set. WDYT?

Nice advice! Given that the current implementation has met our needs, I will consider this optimization in future PR. Have completed the modification as suggested above, please check it.

src/graph/LookupExecutor.cpp
```diff
@@ -774,27 +803,61 @@ bool LookupExecutor::processFinalEdgeResult(RpcResponse &rpcResp, const Callback
...
+                    }
+                    v = value(std::move(res));
                 }
-                auto&& v = value(std::move(res));
+                yieldRecord.emplace_back(v);
                 record.emplace_back(std::move(v));
             }
+            // Check if duplicate, only for user-specified columns within yield clause
+            if (distinct_) {
+                auto ret = uniqResult->emplace(boost::hash_range(yieldRecord.begin(),
```
CPWstatic, line 855: Emm, the `DISTINCT` applies to a row, so it is better using `record` here.

LuconYang: > Emm, the `DISTINCT` applies to a row, so it is better using `record` here.

The sentence proposed by this PR is `LOOKUP ON {<vertex_tag> | <edge_type>} WHERE YIELD DISTINCT <return_list>`. In my opinion, **DISTINCT** is more reasonable to play a role in the columns in the user-specified `<return_list>`. In fact, taking the LOOKUP statement for edges as an example, the columns `SrcVID`, `DstVID`, and `Ranking` are added by default in each row of the output. In this way, all the columns of each record will be different, and **DISTINCT** will not play the expected role.

CPWstatic (approved):

bright-starry-sky (approved):
LGTM.
Thanks for your contribution. nice job. Welcome the next contribution.
I have a vision for the future, Interested?
There are two stages for `DISTINCT`：
The first stage , de-dup the result in the storage layer when the `DISTINCT` clause contains the elements vid | src | rank | dst , this stage ensures that the vid or (src, rank, dst) for all storage nodes is unique.
The second stage, de-dup the result in graph layer when the `DISTINCT` clause only contain prop. this stage ensures that the prop data for all storage nodes is unique.

LuconYang:
> LGTM.
> Thanks for your contribution. nice job. Welcome the next contribution.
> I have a vision for the future, Interested?
> There are two stages for `DISTINCT`：
> The first stage , de-dup the result in the storage layer when the `DISTINCT` clause contains the elements vid | src | rank | dst , this stage ensures that the vid or (src, rank, dst) for all storage nodes is unique.
> The second stage, de-dup the result in graph layer when the `DISTINCT` clause only contain prop. this stage ensures that the prop data for all storage nodes is unique.

Thanks for your approval and great suggestions! I will take your suggestions into consideration in the next contribution~

jamieliu1023:
@LuconYang Thanks for your contribution! This is Jamie Liu with Nebula Graph. To memorize this special moment, we'd like to  send you the Nebula Contributor certificate which is exclusively for you. Also, we want to ship you a mug to mark the contribution. Could you kindly let me know your email and shipping addresses via jamie.liu at vesoft.com? Thank again!

PR #2418, opened 2020-12-15
Not push down filter when multi edge type

CPWstatic:


laura-ding (dismissed):

critical27 (approved):
Good job. The filter push down need to be discussed in 2.0 later.

critical27 (approved):

laura-ding (approved):

PR #2422, opened 2020-12-22
support index range scan of timestamp

bright-starry-sky:
Please hold this PR.
A temporary solution for index range scan of timestamp data type. This requirement comes from community user.
This solution just converted the data type from timestamp to int64, without considering the time zone and time format.


bright-starry-sky:
Close this one, It has been completed by other PR.

PR #2423, opened 2020-12-23
fix meta follower can't read clusterId

critical27:
Introduced in https://github.com/vesoft-inc/nebula/pull/2403.


liuyu85cn (approved):

panda-sheep (approved):
Good

PR #2428, opened 2020-12-31
fix limit/offset

czpmango:
Fix limit/offset, Similar to [https://github.com/vesoft-inc/nebula-graph/pull/542](https://github.com/vesoft-inc/nebula-graph/pull/542)


laura-ding (approved):
Well done

Aiee (approved):

dutor (approved):

PR #2432, opened 2021-01-04
Fix only get dst with filter push down and max_edge_returned_per_vertex config

laura-ding:
The problem comes from https://discuss.nebula-graph.com.cn/t/topic/2245/11

### DIFF
--- a/src/storage/query/QueryBaseProcessor.inl
+++ b/src/storage/query/QueryBaseProcessor.inl
@@ -517,7 +517,7 @@ kvstore::ResultCode QueryBaseProcessor<REQ, RESP>::collectEdgeProps(
         lastRank = rank;
         lastDstId = dstId;
         RowReader reader = RowReader::getEmptyRowReader();
-        if ((!onlyStructure || retTTL.has_value()) && !val.empty()) {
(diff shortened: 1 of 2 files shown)


src/storage/query/QueryBaseProcessor.inl
```diff
@@ -517,7 +517,7 @@ kvstore::ResultCode QueryBaseProcessor<REQ, RESP>::collectEdgeProps(
         lastRank = rank;
         lastDstId = dstId;
         RowReader reader = RowReader::getEmptyRowReader();
-        if ((!onlyStructure || retTTL.has_value()) && !val.empty()) {
+        if ((!onlyStructure || exp_ != nullptr|| retTTL.has_value()) && !val.empty()) {
```
bright-starry-sky, line 520: What about the code as below?
```
if (!onlyStructure && !val.empty()) {
}
```

laura-ding: if only get dst, the `onlyStructure` is true, the filter is no use.

bright-starry-sky: > if only get dst, the `onlyStructure` is true, the filter is no use.

Got.

bright-starry-sky (approved):
Well done.

yixinglu (approved):

PR #2433, opened 2021-01-07
fixed log blowout error for compact index data

bright-starry-sky:
When compact index data,  The index key will check whether the index exists. If the index does not exist, many logs will be thrown.


critical27 (approved):

panda-sheep (approved):
LGTM

PR #2450, opened 2021-02-26
improve int convert float/doulbe/bool, bool convert int, float/double conv…

panda-sheep:
improve int convert float/doulbe/bool,  bool convert int,  float/double convert int.


panda-sheep:
In `https://discuss.nebula-graph.com.cn/t/topic/2935/14`，
although it will report an error at present, the type conversion is more reasonable.

liuyu85cn (approved):
LGTM

darionyaphet (approved):

PR #2483, opened 2021-06-30
Raft problem

sherlockkenan:

####. fix inconsistency problem in raftpart appendlog method.

*Hi, when I read  the  implementation of raft in neblua,  I found out that may exist a inconsistency problem. In the raftpart,  if  a follower receive  logs  from  leader conflicting with itself,  it will rollback  the  log to lastCommitID. such behavior may cause inconsistency commit logs. 

here is a picture to show how it may happend : 
image

I think that the follower could not deletes the log arbitrarily. because some logs may have been already committed.

I  refine the  implementation to follow the guildline in the paper of raft .
image

no 

no

### Checklist

### DIFF
--- a/src/kvstore/raftex/RaftPart.cpp
+++ b/src/kvstore/raftex/RaftPart.cpp
@@ -1584,7 +1584,7 @@ void RaftPart::processAppendLogRequest(
                                             req.get_log_str_list().end());
 
     // may be need to rollback wal_
-    if (!( req.get_last_log_id_sent() == wal_->lastLogId() && req.get_last_log_term_sent() == wal_->lastLogTerm())) {
+    if (!(req.get_last_log_id_sent() == lastLogId_ && req.get_last_log_term_sent() == lastLogTerm_)) {
         // check the diff index in log
         size_t diffIndex = 0;
         {
--- a/src/kvstore/raftex/RaftPart.cpp
+++ b/src/kvstore/raftex/RaftPart.cpp
@@ -1548,56 +1548,92 @@ void RaftPart::processAppendLogRequest(
     }
 
     // req.get_last_log_id_sent() >= committedLogId_
-    if (lastLogTerm_ > 0 && req.get_last_log_term_sent() != lastLogTerm_) {
-        LOG(INFO) << idStr_ << "The local last log term is " << lastLogTerm_
-                << ", which is different from the leader's prevLogTerm "
-                << req.get_last_log_term_sent()
-                << ", the prevLogId is " << req.get_last_log_id_sent()
-                << ". So need to rollback to last committedLogId_ " << committedLogId_;
-        if (wal_->rollbackToLog(committedLogId_)) {
+    if (req.get_last_log_id_sent() == lastLogId_ && req.get_last_log_term_sent() == lastLogTerm_) {
+        // nothing to do
+        // just append log later
+    } else if (req.get_last_log_id_sent() > lastLogId_) {
+        // There is a gap
+        LOG(INFO) << idStr_ << "Local is missing logs from id " << lastLogId_
+                  << ". Need to catch up";
+        resp.set_error_code(cpp2::ErrorCode::E_LOG_GAP);
+        return;
+    } else  {
+        // check the term
+        int reqLastLogTerm  = wal_->getLogTerm(req.get_last_log_id_sent());
+        if (req.get_last_log_term_sent() != reqLastLogTerm) {
+            LOG(INFO) << idStr_ << "The local log term is " << reqLastLogTerm
+                      << ", which is different from the leader's prevLogTerm "
+                      << req.get_last_log_term_sent()
+                      << ", the prevLogId is " << req.get_last_log_id_sent()
+                      << ". So ask leader to send logs from committedLogId " << committedLogId_;
+            TermID committedLogTerm = wal_->getLogTerm(committedLogId_);
+            if (committedLogTerm > 0 ) {
+                resp.set_last_log_id(committedLogId_);
+                resp.set_last_log_term(committedLogTerm);
+            }
+            resp.set_error_code(cpp2::ErrorCode::E_LOG_GAP);
+            return;
+        }
+    }
+
+    // request get_last_log_term_sent  == wal[get_last_log_id_sent].log_term
+    size_t numLogs = req.get_log_str_list().size();
+    LogID firstId = req.get_last_log_id_sent() + 1;
+
+    std::vector<cpp2::LogEntry> logEntries (req.get_log_str_list().begin(),
+                                            req.get_log_str_list().end());
+
+    // may be need to rollback wal_
+    if (!( req.get_last_log_id_sent() == wal_->lastLogId() && req.get_last_log_term_sent() == wal_->lastLogTerm())) {
+        // check the diff index in log
+        size_t diffIndex = 0;
+        {
+            std::unique_ptr<LogIterator> it = wal_->iterator(firstId,firstId + numLogs);
+            for (size_t i = 0; i < numLogs && it->valid(); i++, ++(*it), diffIndex ++ ) {
+                int logTerm  =  it->logTerm();
+                if (req.get_log_term() !=  logTerm) {
+                    break;
+                }
+            }
+            it.reset();
+        }
+
+        //  stale log
+        if (diffIndex == numLogs) {
+            resp.set_last_log_id(firstId + numLogs - 1);
+            resp.set_last_log_term(req.get_log_term());
+            resp.set_error_code(cpp2::ErrorCode::E_LOG_STALE);
+            //nothing to append
+            return;
+        }
+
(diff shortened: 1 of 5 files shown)


Shylock-Hg:
I have a problem in above flow in this picture, that stage 2 is impossible, the leader will commit after log committed already in major.

Shylock-Hg:
I have a problem about the flow in above picture, in that, the leader must commit after committed by major, but the stage 1,2 doesn't.

critical27:
👍 , this is a known issue (long story). I'll review later.

sherlockkenan:
> I have a problem about the flow in above picture, in that, the leader must commit after committed by major, but the stage 1,2 doesn't.

 the leader could commit  after it sends logs to major successfully. after that leader update its commit message to followers in the next request. before leader transfer its commit message, commitId between leader and followers could be different.

src/kvstore/wal/FileBasedWal.cpp
```diff
@@ -792,5 +792,14 @@ size_t FileBasedWal::accessAllBuffers(std::function<bool(BufferPtr buffer)> fn)
     return count;
 }
 
+TermID FileBasedWal::getLogTerm(LogID id) {
+    TermID term = -1;
+    auto walIter = std::make_unique<FileBasedWalIterator>(shared_from_this(), id, id);
```
critical27, line 797: You could try to read from memory cache first, it could be quite slow.

src/kvstore/raftex/RaftPart.cpp
```diff
@@ -1548,56 +1548,92 @@ void RaftPart::processAppendLogRequest(
...
+        if (req.get_last_log_term_sent() != reqLastLogTerm) {
+            LOG(INFO) << idStr_ << "The local log term is " << reqLastLogTerm
+                      << ", which is different from the leader's prevLogTerm "
+                      << req.get_last_log_term_sent()
+                      << ", the prevLogId is " << req.get_last_log_id_sent()
+                      << ". So ask leader to send logs from committedLogId " << committedLogId_;
+            TermID committedLogTerm = wal_->getLogTerm(committedLogId_);
+            if (committedLogTerm > 0 ) {
+                resp.set_last_log_id(committedLogId_);
+                resp.set_last_log_term(committedLogTerm);
```
critical27, line 1572: Could consider do not `set_last_log_term`, IIRC, it is not used, so we don't nedd to `wal_->getLogTerm(committedLogId_);`

critical27 (commented):
Thx for contribution, some place need to have a discuss, see the comments.

src/kvstore/raftex/RaftPart.cpp
```diff
@@ -1548,56 +1548,92 @@ void RaftPart::processAppendLogRequest(
...
+
+    // request get_last_log_term_sent  == wal[get_last_log_id_sent].log_term
+    size_t numLogs = req.get_log_str_list().size();
+    LogID firstId = req.get_last_log_id_sent() + 1;
+
+    std::vector<cpp2::LogEntry> logEntries (req.get_log_str_list().begin(),
+                                            req.get_log_str_list().end());
+
+    // may be need to rollback wal_
+    if (!( req.get_last_log_id_sent() == wal_->lastLogId() && req.get_last_log_term_sent() == wal_->lastLogTerm())) {
```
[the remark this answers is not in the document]
sherlockkenan: yes, using

>  req.get_last_log_id_sent() == lastLogId_ && req.get_last_log_term_sent() == lastLogTerm_

 to check is better, I will modify it later.

I agree that when it is true,  it not need to check term in each log.  that is why I add this condition check. 

>     if (!( req.get_last_log_id_sent() == wal_->lastLogId() && req.get_last_log_term_sent() == wal_->lastLogTerm()))
In most case,  above check  is false, therefore we don not need check term of log one by one,  just append the new log。

src/kvstore/wal/FileBasedWal.cpp
sherlockkenan: I will modify it later

src/kvstore/raftex/RaftPart.cpp
sherlockkenan: I have considered  it before,  but if we don't set it，then the resp is: (commmitedLogId_,last_log_term). 

 when leader receive the response, it will send a new request  with  (commmitedLogId_,last_log_term) not (commmitedLogId_, commmitedLogIdTerm_), see [  self->lastLogTermSent_ = resp.get_last_log_term();](https://github.com/vesoft-inc/nebula/blob/046b522c2be6bcdd4b27828419c5ae82d21fc844/src/kvstore/raftex/Host.cpp#L310)

append log request with  (commmitedLogId_,last_log_term)  could not pass the check in the follower ,  which will causing a dead loop

critical27: Yeah, you're right.

src/kvstore/wal/FileBasedWal.cpp
sherlockkenan: done

src/kvstore/raftex/RaftPart.cpp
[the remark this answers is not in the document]
critical27: > In most case, above check is false, therefore we don not need check term of log one by one, just append the new log。

Agree.

I think `lastLogId_` and `lastLogTerm_` will be better. A extreme example is, this host converts from follower -> leader -> follower again, in the leader phase, it could write some log in wal (although the case is rare).

src/kvstore/wal/FileBasedWal.cpp
critical27: 👌  LGTM

sherlockkenan:
I only ran CTest in the centos environment , it is ok,  but not in ubuntu clang-9. I don't known why the error happend.

src/kvstore/raftex/RaftPart.cpp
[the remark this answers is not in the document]
sherlockkenan: i get it

src/kvstore/raftex/RaftPart.cpp
```diff
@@ -1548,56 +1548,92 @@ void RaftPart::processAppendLogRequest(
...
+            resp.set_error_code(cpp2::ErrorCode::E_LOG_GAP);
+            return;
+        }
+    }
+
+    // request get_last_log_term_sent  == wal[get_last_log_id_sent].log_term
+    size_t numLogs = req.get_log_str_list().size();
+    LogID firstId = req.get_last_log_id_sent() + 1;
+
+    std::vector<cpp2::LogEntry> logEntries (req.get_log_str_list().begin(),
```
critical27, line 1583: Try use `std::make_move_iterator` here.

src/kvstore/raftex/RaftPart.cpp
```diff
@@ -1548,56 +1548,92 @@ void RaftPart::processAppendLogRequest(
...
+            LOG(ERROR) << idStr_ << "Rollback fail! lastLogId is" << lastLogId_
+                       << ", logLogTerm is " << lastLogTerm_
+                       << ", committedLogId is " << committedLogId_
+                       << ", rollback id  is " << firstId + diffIndex -1;
+            resp.set_error_code(cpp2::ErrorCode::E_WAL_FAIL);
+            return;
+        }
+
+        // update msg
+        logEntries = std::vector<cpp2::LogEntry> (req.get_log_str_list().begin() + diffIndex,
```
critical27, line 1628: This could be quite expensive as well, emm, can't figure out a better way for now.

critical27:
> I only ran CTest in the centos environment , it is ok, but not in ubuntu clang-9. I don't known why the error happend.

OK, leave it alone.

src/kvstore/raftex/RaftPart.cpp
sherlockkenan: yeah,  I have refined it

src/kvstore/raftex/RaftPart.cpp
sherlockkenan: I have refined it to avoid copy

sherlockkenan:
> > I only ran CTest in the centos environment , it is ok, but not in ubuntu clang-9. I don't known why the error happend.
> 
> OK, leave it alone.

is there any other things

critical27 (approved):
Sorry for late response, LGTM now. Thx!

liuyu85cn (approved):
LGTM

jamieliu1023:
@sherlockkenan Thanks for your contribution to the Nebula Graph community! This is Jamie with Nebula and I'd like to email you the Nebula Contributor certificate and ship you a mug to mark this special moment. Could you please kindly reach me via jamie.liu(at)vesoft.com? Again, thanks for being a part of the Nebula community!

sherlockkenan:
Hi，Jamie，I am please to accept this message


xuehan ke

PR #2493, opened 2021-08-11
Fix nebula service script.

CPWstatic:


yixinglu (commented):
This have been fixed in #2491

PR #2503, opened 2021-08-11
Remove the lock of ObjectPool, we don't need MT-safe for it's only ac…

Shylock-Hg:
…cessed by single thread.


yixinglu (commented):
Not agree with this change. Why to do that? for performance? but spinlock will not hurt the performance in single thread scenario.

PR #2505, opened 2021-08-11
Rename source directories in meta

yixinglu:
1. if one word is enough to express the meaning, we should try not to use multiple words as the directory name
2. collect meta http handlers and tests in separate folder.


bright-starry-sky (approved):

src/meta/http/MetaHttpReplaceHostHandler.h
```diff
@@ -51,5 +51,5 @@ class MetaHttpReplaceHostHandler : public proxygen::RequestHandler {
 }  // namespace meta
 }  // namespace nebula
 
-#endif  // META_METAHTTPReplaceHostHANDLER_H
+#endif  // META_HTTP_METAHTTPReplaceHostHANDLER_H
```
CPWstatic, line 54: Emmm, this comment is not right.

yixinglu: Good catch. Use `#pragma once` to replace the macro. I'll update the code style to recommend to use this `#pragma`

CPWstatic (approved):

critical27 (approved):
Generally LGTM, if we use pragma once, could use it every where.

yixinglu:
> Generally LGTM, if we use pragma once, just replace all of them.

I suggest adopting `#pragma once` in the newly added code, and we don't modify the existing codes at present.

PR #2511, opened 2021-08-12
Request the rfc to manage the memory of AST by Arena technology.

Shylock-Hg:


docs/rfcs/0001-manage-memory-of-AST-by-arena.md
```diff
@@ -0,0 +1,71 @@
...
+};
+```
+
+The memory of `AttributeExpr` itself is managed by Arena, but the `std::string` will allocate/deallocate memory by c++ default Allocator although `std::string` itself is managed by Arena.
+
+This trade-off will discussed in next section.
+
+# Rationale and alternatives
+
+The c++ stl container also provide the template argument for custom allocator. But it make container type different. So the same container with different allocator is different type, and need extra conversion.
```
Aiee, line 59: The c++ stl container also provides the template argument for the custom allocator. But it makes container type different. So the same container with a different allocator is a different type, and need extra conversion.

wey-gu:
I really loved/considered this approach a better way for nebula graph design/dev context to be shared for all contributors.

Shylock-Hg:
> I really loved/considered this approach a better way for nebula graph design/dev context to be shared for all contributors.

I think so.

PR #2512, opened 2021-08-12
Implement the arena allocator.

Shylock-Hg:
A part of #2511


cangfengzhs:
Why not use the arena in folly ?

Shylock-Hg:
> Why not use the arena in folly ?

I introduce some optimization and gain better performance than folly arena in the benchmark.

src/common/base/Arena.h
```diff
@@ -0,0 +1,85 @@
...
+    allocatedSize_ = 0;
+#endif
+    availableSize_ = 0;
+    currentPtr_ = nullptr;
+  }
+
+  // The CPU access memory with the alignment,
+  // So construct object from alignment address will reduce the CPU access count then
+  // speed up read/write
+  void *allocateAligned(const std::size_t alloc);
```
critical27, line 38: Consider only provide an interface of `allocate`. It should not be caller's duty to provide an aligned size. When user call `allocate(not aligned size)`, you can round up the size and call `allocateAligned`.

Or provide both `allocateAligned` and `allocate` LGTM as well.

src/common/base/Arena.cpp
```diff
@@ -0,0 +1,46 @@
...
+  // replace the modulo operation by bit and
+  static_assert(kAlignment && !(kAlignment & (kAlignment - 1)), "Align must be power of 2.");
+  const std::size_t pad =
+      kAlignment - (reinterpret_cast<uintptr_t>(currentPtr_) & (kAlignment - 1));
+  const std::size_t consumption = alloc + pad;
+  if (consumption > kMaxChunkSize) {
+    DLOG(FATAL) << "Arena can't allocate so large memory.";
+    return nullptr;
+  }
+  if (consumption <= availableSize_) {
```
critical27, line 24: We'd better add a LIKELY here

src/common/base/Arena.cpp
```diff
@@ -0,0 +1,46 @@
...
+    allocatedSize_ += consumption;
+#endif
+    availableSize_ -= consumption;
+    return ptr;
+  } else {
+    newChunk(std::max(alloc, kMinChunkSize));
+    // The new operator will allocate the aligned memory
+    DCHECK_EQ(reinterpret_cast<uintptr_t>(currentPtr_) & (kAlignment - 1), 0);
+    void* ptr = currentPtr_;
+    currentPtr_ += alloc;
```
critical27, line 37: `currentPtr_` need to add pad as well, otherwise, upcoming allocated won't be aligned?

critical27 (commented):
A very good start! 

BTW, what do we have in the `ObjPool` for now besides AST?

src/common/base/Arena.cpp
Shylock-Hg: The `newChunk` method will allocate the aligned memory so we don't need to compute alignment here.

Shylock-Hg:
> A very good start!
> 
> BTW, what do we have in the `ObjPool` for now besides AST?

Only AST and Plan related objects.

src/common/base/Arena.h
Shylock-Hg: The user don't need to provide aligned size, the method only accept one parameter which is allocated size.

critical27: 👌 , my bad

src/common/base/Arena.cpp
critical27: 👌 , I misunderstood previously.

src/common/base/Arena.cpp
Shylock-Hg: Added.

PR #2522, opened 2021-08-16
[raft] clean part rocksdb data when raftpart::reset

panda-sheep:
Before a part copy receives the snapshot data, execute `RaftPart::reset.`
`RaftPart::reset` clears all wal files of the current part and deletes the SystemCommit data of the current part. 
The data that has been placed in the disk is not cleared,, which may cause data inconsistency between copies.

for example:
A and B are copies of the same part, A is the leader, and B is the follower. Insert two records into the part.

```
                   A(leader)                     B(follower)
insert(key1)       key1                          key1
insert(key2)       key2                          key2
```

At this time, B has a network isolation.
A deletes key1, and then A performs many other operations. As a result, the wal that deletes key1 has been deleted by Wal TTL.
```
                       A(leader)                  B(follower)
delete(key1)           key1(X)                    key1
                       key2                       key2
                       ......
                       ......
```

At this time, B rejoins the network. 
B needs to pull a snapshot data  from A. 
Because key1 and key2 in B have been placed in the disk, they have not been deleted. 
Finally, after B’s data catches up with A’s data, B’s data has more key1 records than A’s data， resulting in inconsistent data between copies.

### DIFF
--- a/src/common/utils/IndexKeyUtils.cpp
+++ b/src/common/utils/IndexKeyUtils.cpp
@@ -86,9 +86,7 @@ std::string IndexKeyUtils::indexFirstKey(PartitionID partId) {
   int32_t item = (partId << kPartitionOffset) | static_cast<uint32_t>(NebulaKeyType::kIndex);
 
   std::string key;
-  IndexID indexId = 0;
-  key.append(reinterpret_cast<const char*>(&item), sizeof(int32_t))
-      .append(reinterpret_cast<const char*>(&indexId), sizeof(IndexID));
+  key.append(reinterpret_cast<const char*>(&item), sizeof(int32_t)).append(sizeof(IndexID), '\0');
   return key;
 }
 
@@ -97,9 +95,7 @@ std::string IndexKeyUtils::indexLastKey(PartitionID partId) {
   int32_t item = (partId << kPartitionOffset) | static_cast<uint32_t>(NebulaKeyType::kIndex);
 
   std::string key;
-  IndexID indexId = INT32_MAX;
-  key.append(reinterpret_cast<const char*>(&item), sizeof(int32_t))
-      .append(reinterpret_cast<const char*>(&indexId), sizeof(IndexID));
+  key.append(reinterpret_cast<const char*>(&item), sizeof(int32_t)).append(sizeof(IndexID), '\377');
   return key;
 }
 
--- a/src/common/utils/IndexKeyUtils.cpp
+++ b/src/common/utils/IndexKeyUtils.cpp
@@ -82,6 +82,28 @@ std::string IndexKeyUtils::edgeIndexKey(size_t vIdLen,
 }
 
 // static
+std::string IndexKeyUtils::indexFirstKey(PartitionID partId) {
+  int32_t item = (partId << kPartitionOffset) | static_cast<uint32_t>(NebulaKeyType::kIndex);
+
+  std::string key;
+  IndexID indexId = 0;
+  key.append(reinterpret_cast<const char*>(&item), sizeof(int32_t))
+      .append(reinterpret_cast<const char*>(&indexId), sizeof(IndexID));
+  return key;
+}
+
+// static
+std::string IndexKeyUtils::indexLastKey(PartitionID partId) {
+  int32_t item = (partId << kPartitionOffset) | static_cast<uint32_t>(NebulaKeyType::kIndex);
+
+  std::string key;
+  IndexID indexId = INT32_MAX;
+  key.append(reinterpret_cast<const char*>(&item), sizeof(int32_t))
+      .append(reinterpret_cast<const char*>(&indexId), sizeof(IndexID));
+  return key;
+}
+
+// static
 std::string IndexKeyUtils::indexPrefix(PartitionID partId, IndexID indexId) {
   PartitionID item = (partId << kPartitionOffset) | static_cast<uint32_t>(NebulaKeyType::kIndex);
   std::string key;
--- a/src/common/utils/IndexKeyUtils.cpp
+++ b/src/common/utils/IndexKeyUtils.cpp
@@ -82,24 +82,6 @@ std::string IndexKeyUtils::edgeIndexKey(size_t vIdLen,
 }
 
 // static
-std::string IndexKeyUtils::indexFirstKey(PartitionID partId) {
-  int32_t item = (partId << kPartitionOffset) | static_cast<uint32_t>(NebulaKeyType::kIndex);
-
-  std::string key;
-  key.append(reinterpret_cast<const char*>(&item), sizeof(int32_t)).append(sizeof(IndexID), '\0');
-  return key;
-}
-
-// static
-std::string IndexKeyUtils::indexLastKey(PartitionID partId) {
-  int32_t item = (partId << kPartitionOffset) | static_cast<uint32_t>(NebulaKeyType::kIndex);
-
-  std::string key;
-  key.append(reinterpret_cast<const char*>(&item), sizeof(int32_t)).append(sizeof(IndexID), '\377');
-  return key;
-}
-
-// static
 std::string IndexKeyUtils::indexPrefix(PartitionID partId, IndexID indexId) {
   PartitionID item = (partId << kPartitionOffset) | static_cast<uint32_t>(NebulaKeyType::kIndex);
   std::string key;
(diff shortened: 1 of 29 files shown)


src/kvstore/Part.cpp
```diff
@@ -420,11 +421,55 @@ bool Part::preProcessLog(LogID logId, TermID termId, ClusterID clusterId, const
...
+  if (ret != nebula::cpp2::ErrorCode::SUCCEEDED) {
+    LOG(ERROR) << idStr_ << "Remove the part vertex data failed, error "
+               << static_cast<int32_t>(ret);
+    return;
+  }
+  while (vertexIter && vertexIter->valid()) {
+    toDelete.emplace_back(vertexIter->key().str());
+    vertexIter->next();
+  }
+
```
cangfengzhs, line 442: use delete range may be a better way?
https://github.com/facebook/rocksdb/wiki/Delete-A-Range-Of-Keys

panda-sheep: Yeah, `DeleteRange `should be used here. I discussed it with @critical27 offline.

panda-sheep: has used `DeleteRange `

src/kvstore/test/PartTest.cpp
```diff
@@ -9,11 +9,58 @@
...
+#include "common/utils/NebulaKeyUtils.h"
 #include "kvstore/Part.h"
+#include "kvstore/RocksEngine.h"
 
 namespace nebula {
 namespace kvstore {
 
+const int32_t kDefaultVIdLen = 8;
+
+void checkVetexData(RocksEngine* engine, PartitionID partId, int expectNum, bool checkVal = false) {
```
cangfengzhs, line 22: `CheckVetexData` Spelling mistake?

src/common/utils/IndexKeyUtils.cpp
```diff
@@ -81,6 +81,28 @@ std::string IndexKeyUtils::edgeIndexKey(size_t vIdLen,
...
+      .append(reinterpret_cast<const char*>(&indexId), sizeof(IndexID));
+  return key;
+}
+
+// static
+std::string IndexKeyUtils::indexLastKey(PartitionID partId) {
+  int32_t item = (partId << kPartitionOffset) | static_cast<uint32_t>(NebulaKeyType::kIndex);
+
+  std::string key;
+  IndexID indexId = INT32_MAX;
```
cangfengzhs, line 100: Can we guarantee that index_id is always greater than 0? If not, should use 0xffffffff instead of 0x7ffffff

src/kvstore/test/PartTest.cpp
panda-sheep: Bright eyes.

src/kvstore/Part.cpp
```diff
@@ -420,11 +423,38 @@ bool Part::preProcessLog(LogID logId, TermID termId, ClusterID clusterId, const
...
+  auto startiKey = IndexKeyUtils::indexFirstKey(partId_);
+  auto endiKey = IndexKeyUtils::indexLastKey(partId_);
+  ret = engine_->removeRange(std::move(startiKey), std::move(endiKey));
+  if (ret != nebula::cpp2::ErrorCode::SUCCEEDED) {
+    LOG(ERROR) << idStr_ << "Remove the part index data failed, error "
+               << static_cast<int32_t>(ret);
+    return;
+  }
+
```
critical27, line 453: Operation key could be delete as well

src/common/utils/IndexKeyUtils.cpp
panda-sheep: The indexId is logically (0, INT32_MAX), although there is no protection in our code.

Also for the above range, 0x7ffffff is the maximum
Whether it is big-endian storage or little-endian storage, there is no bigger number than this.

In order to avoid ambiguity, the minimum and maximum value is directly stored in characters (‘\0’,'\377')

critical27 (commented):
Consider provide a unify interface
```
firstKey(partId, type, suffix, count of padding \0)
lastKey(partId, type, suffix, count of padding 0xff)
```
They generate the key of `(partId | type, suffix, n * \0)` and `(partId | type, suffix, n * 0xff)`, the `suffix` is the part you intend to delete after `partId | type`

So all these indexFirstKey/vertexFirstKey just call them.

for example, vertexFirstKey would be firstKey(partId, kVertex, "", vIdLen)

Or
```
firstKey(prefix, count of padding \0)
lastKey(prefix, count of padding 0xff)
```

This would be better I think, prefix is the share part you want to delete.

src/common/utils/IndexKeyUtils.cpp
cangfengzhs: 👌🏻I was wrong

panda-sheep:
> Consider provide a unify interface
> 
> ```
> firstKey(partId, type, suffix, count of padding \0)
> lastKey(partId, type, suffix, count of padding 0xff)
> ```
> 
> They generate the key of `(partId | type, suffix, n * \0)` and `(partId | type, suffix, n * 0xff)`, the `suffix` is the part you intend to delete after `partId | type`
> 
> So all these indexFirstKey/vertexFirstKey just call them.
> 
> for example, vertexFirstKey would be firstKey(partId, kVertex, "", vIdLen)
> 
> Or
> 
> ```
> firstKey(prefix, count of padding \0)
> lastKey(prefix, count of padding 0xff)
> ```
> 
> This would be better I think, prefix is the share part you want to delete. When we want to delete all tag of a vertex, just call
> 
> ```
> prefix = vertexPrefix(partId, vId)
> deleteRange(firstKey(prefix, count), lastKey(prefix, count))
> ```

👍
```
firstKey(prefix, count of padding \0)
lastKey(prefix, count of padding 0xff)
```
This way is better.

src/kvstore/Part.cpp
panda-sheep: fixed

cangfengzhs (approved):
LGTM

src/kvstore/Part.cpp
```diff
@@ -420,11 +424,47 @@ bool Part::preProcessLog(LogID logId, TermID termId, ClusterID clusterId, const
...
-  if (res != nebula::cpp2::ErrorCode::SUCCEEDED) {
-    LOG(WARNING) << idStr_ << "Remove the committedLogId failed, error "
-                 << static_cast<int32_t>(res);
+  LOG(INFO) << idStr_ << "Clean rocksdb part data";
+  // Remove the vertex, edge, index, systemCommitKey, operation data under the part
+  const auto& vertexPre = NebulaKeyUtils::vertexPrefix(partId_);
+  auto ret = engine_->removeRange(NebulaKeyUtils::firstKey(vertexPre, vIdLen_),
+                                  NebulaKeyUtils::lastKey(vertexPre, vIdLen_));
+  if (ret != nebula::cpp2::ErrorCode::SUCCEEDED) {
+    LOG(ERROR) << idStr_ << "Remove the part vertex data failed, error "
```
yixinglu, line 433: insert one space between idStr_ and Remove

src/kvstore/Part.cpp
```diff
@@ -420,11 +424,47 @@ bool Part::preProcessLog(LogID logId, TermID termId, ClusterID clusterId, const
...
+  if (ret != nebula::cpp2::ErrorCode::SUCCEEDED) {
+    LOG(ERROR) << idStr_ << "Remove the part edge data failed, error" << static_cast<int32_t>(ret);
+    return;
+  }
+
+  const auto& indexPre = IndexKeyUtils::indexPrefix(partId_);
+  ret = engine_->removeRange(NebulaKeyUtils::firstKey(indexPre, sizeof(IndexID)),
+                             NebulaKeyUtils::lastKey(indexPre, sizeof(IndexID)));
+  if (ret != nebula::cpp2::ErrorCode::SUCCEEDED) {
+    LOG(ERROR) << idStr_ << "Remove the part index data failed, error "
```
yixinglu, line 450: ditto

critical27 (approved):
Good job!

src/kvstore/Part.cpp
critical27: IIRC, idStr_ already has one space suffix.

panda-sheep: There is already a space at the end of `idStr_ `string, so there is no need to insert space.

bright-starry-sky (approved):
Nice job

PR #2528, opened 2021-08-17
Cleanup package workflow

yixinglu:
Need not another project level CMakeLists.txt for packaging at present.


darionyaphet:
`Storage Performance Tool` is need to set PERMISSIONS?

yixinglu:
> `Storage Performance Tool` is need to set PERMISSIONS?

This tool does not seem to be installed.

Aiee (approved):

nevermore3 (approved):

PR #2529, opened 2021-08-17
fix get target host

darionyaphet:


src/meta/processors/job/MetaJobExecutor.cpp
```diff
@@ -79,15 +79,21 @@ ErrOrHosts MetaJobExecutor::getTargetHost(GraphSpaceID spaceId) {
...
   // use vector instead of set because this can convient for next step
+  std::unordered_map<HostAddr, std::vector<PartitionID>> hostAndPart;
   std::vector<std::pair<HostAddr, std::vector<PartitionID>>> hosts;
   while (iter->valid()) {
+    auto part = MetaServiceUtils::parsePartKeyPartId(iter->key());
     auto targets = MetaServiceUtils::parsePartVal(iter->val());
     for (auto& target : targets) {
+      hostAndPart[target].emplace_back(part);
       std::vector<PartitionID> parts;
       hosts.emplace_back(std::make_pair(std::move(target), std::move(parts)));
```
yixinglu, line 90: If you use map to collect all parts of each host, it's not necessary to push them into the hosts container here and dedup them by std::unique in following steps.

If you need to dedup the partIDs, you could replace the std::vector with std::unordered_set in unordered_map.

darionyaphet: Actually I need a vector<pair> to collect the host and partition mapping

yixinglu: I see you have copied the contents of map to the vector in the code below, so need not to push them to hosts here.

src/meta/processors/job/MetaJobExecutor.cpp
```diff
@@ -79,15 +79,21 @@ ErrOrHosts MetaJobExecutor::getTargetHost(GraphSpaceID spaceId) {
   }
 
   // use vector instead of set because this can convient for next step
+  std::unordered_map<HostAddr, std::vector<PartitionID>> hostAndPart;
```
bright-starry-sky, line 82: in line 81, don't correspond.

src/meta/processors/job/MetaJobExecutor.cpp
```diff
@@ -79,15 +79,21 @@ ErrOrHosts MetaJobExecutor::getTargetHost(GraphSpaceID spaceId) {
...
+    auto part = MetaServiceUtils::parsePartKeyPartId(iter->key());
     auto targets = MetaServiceUtils::parsePartVal(iter->val());
     for (auto& target : targets) {
+      hostAndPart[target].emplace_back(part);
       std::vector<PartitionID> parts;
       hosts.emplace_back(std::make_pair(std::move(target), std::move(parts)));
     }
     iter->next();
   }
+  for (auto it = hostAndPart.begin(); it != hostAndPart.end(); it++) {
```
bright-starry-sky, line 94: line 89 and 90 , should we delete these two lines?

darionyaphet: it seems useless .

src/meta/processors/job/MetaJobExecutor.cpp
darionyaphet: the comments and codes inconsistent ?

bright-starry-sky:
Is there any progress on this PR? Need to merge to version 2.6?

darionyaphet:
> Is there any progress on this PR? Need to merge to version 2.6?

No more progress and this PR have ready to review.

critical27 (commented):
Just use a `std::unordered_map<HostAddr, std::vector<PartitionID>>` to construct the `std::vector<std::pair<HostAddr, std::vector<PartitionID>>>` to return is enough?

Both the implementation before/after the PR is hard to understand.....

darionyaphet:
> Just use a `std::unordered_map<HostAddr, std::vector<PartitionID>>` to construct the `std::vector<std::pair<HostAddr, std::vector<PartitionID>>>` to return is enough?
> 
> Both the implementation before/after the PR is hard to understand.....

Maybe it's for sequential allocation, so a list is returned

critical27:
> Maybe it's for sequential allocation, so a list is returned
> 
> Using unordered_map as the return value is ok, but getTargetHost, getLeaderHost and getListenerHost should have the same return value type.

I mean it is ok to return a vector, but why `hosts` is emplace back more than once in line 85 and 90. Just use the map to build the vector you need to return.

critical27:
We definitely should not return more than `number of hosts` in the result. Read this `MetaJobExecutor::execute`  and `MetaJobExecutor::getTargetHost` carefully, the logic in `getTargetHost` is implemented in the wrong way. The code works fine only because of coincidence.

darionyaphet:
> We definitely should not return more than `number of hosts` in the result. Read this `MetaJobExecutor::execute` and `MetaJobExecutor::getTargetHost` carefully, the logic in `getTargetHost` is implemented in the wrong way. The code works fine only because of coincidence.

I don't think it works right now.  `parts` only filled with an empty vector.

src/meta/processors/job/MetaJobExecutor.cpp
```diff
@@ -79,15 +79,19 @@ ErrOrHosts MetaJobExecutor::getTargetHost(GraphSpaceID spaceId) {
...
+      hostAndPart[target].emplace_back(part);
     }
     iter->next();
   }
+  for (auto it = hostAndPart.begin(); it != hostAndPart.end(); it++) {
+    hosts.emplace_back(std::pair(it->first, it->second));
+  }
   std::sort(hosts.begin(), hosts.end());
   auto last = std::unique(hosts.begin(), hosts.end());
   hosts.erase(last, hosts.end());
```
bright-starry-sky, line 97: I think these three lines are redundant, right? other are all ok for me.

bright-starry-sky (approved):

liuyu85cn (approved):

critical27 (approved):

PR #2530, opened 2021-08-18
Add the proposal about push down plan to storage to execute.

Shylock-Hg:


docs/rfcs/0001-edge-computation-in-storage-side.md
```diff
@@ -0,0 +1,78 @@
...
+     // Origin fields ...
+     EdgeComputer edge_computer,
+}
+```
+
+# Design explanation
+
+In storage side:
+
+The `Edge Computation` will append to the origin plan in storage, e.g. Append push-down plan into `GetNeighbors` origin plan. So storage scan data firstly, then complete the push-down computation and response data finally.
```
critical27, line 55: If the idea coms from coprocessor, as far as I know, both hbase and tidb is the query layer to generate the whole plan, not only part. The append is hard to control, for now there is only one output node for `go`, what if there are more than one (e.g. lookup may have more than one when we lookup where e1.c1 == 1 or e2.c1 == 1)?

IMO, either just use flag, or the whole plan should be generate in query engine.

Shylock-Hg: As I known the tikv coprocessor only support a subset of executor, so maybe it can't push down the whole plan to storage.

Shylock-Hg: Yes, I means a part of the whole plan generated by query layer.

critical27: I understand your idea. This is quite interesting, imagine we generate the whole plan for storage in query layer not only part of them, many optimization will be easier (e.g. filter). Query layer only need to merge/reorder the executor in logic plan.

If we only want to push down a part of plan to storage, there is not a big difference with just do optimize in query layer as long as we support agg/sort/dedup... in each storage RPC interface. We could talk offline if necessary.

czpmango: IMHO, most TP scenarios do not need to push down the calculation(sql-like) on the graph side to the storage side, if the index is implemented well (IO will not become the most critical bottleneck). So the point is to enhance the ability of the index rather than just push down.

Shylock-Hg: Yes, but I found new issue in nearly development. As report in #2583 , the order of push-down executor is not determined in current implementation. Now we could resolve #2583 by do filter before limit, but if graph layer require semantic that limit first (or other case), we can't fit it now.

Shylock-Hg: The problem is that the capacity of current implementation is not complete.

PR #2532, opened 2021-08-18
add a script to do metad folder scp

liuyu85cn:
In our forum, some user complain lack of mete transfer tools. 

this tool is a simple wrapper to scp to copy local folder of metad to another (remote)place
    it afford 4 options:   
        1. -f (from) local metad dir
        2. -t (to) remote destination
        3. -u (update) any configs need to be changed
            different configs should be seperated by ':'
            each config has to be the form of "local_ip=172.0.0.1"
       
for example
```
 ./meta-transfer-tools.sh -f /path-to-src/metad1 -t bob@172.0.0.1:/path-to-dst -u local_ip=172.0.0.1:port=10086
```
 this is not a well discussed solution for this, and I'm not very good at shell, any suggestion is welcome.

### DIFF
--- a/scripts/meta-transfer-tools.sh
+++ b/scripts/meta-transfer-tools.sh
@@ -22,10 +22,11 @@ Usage="this tool is a simple wrapper to scp to copy local folder of metad to ano
 echo -e $Usage
 }
 
-while getopts "hf:t:u:" opt; do
+while getopts ":hf:t:u:" opt; do
     case $opt in
         h)
             usage
+            exit 0
             ;;
         f)
             echo "trying to copy metad from: " $OPTARG
@@ -41,7 +42,8 @@ while getopts "hf:t:u:" opt; do
             configs=$OPTARG
             ;;
         \?)
-            echo "Invalid option: -$OPTARG" 
+            echo "Invalid option: -$OPTARG"
+            exit 1
             ;;
     esac
 done
--- a/scripts/meta-transfer-tools.sh
+++ b/scripts/meta-transfer-tools.sh
@@ -1,0 +1,83 @@
+#! /bin/bash
+
+src=
+dst=
+metaDirName=
+configs=
+
+function usage {
+Usage="this tool is a simple wrapper to scp to copy local folder of metad to another (remote)place \n \
+    it afford 4 options:    \n \
+        1. -f (from) local metad dir \n \
+        2. -t (to) remote destination \n \
+        3. -u (update) any configs need to be changed \n \
+            different configs should be seperated by ':' \n \
+            each config has to be the form of "local_ip=172.0.0.1" \n \
+        \n \
+    for example \n \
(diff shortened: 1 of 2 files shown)


whitewum:
is it released in 2.5.0?

liuyu85cn:
> is it released in 2.5.0?

will merge into 9.30.

bright-starry-sky:
marvellously quick

bright-starry-sky:
itc

Shylock-Hg:
Could specify the space?

yixinglu (commented):
Don't forget to install this script

critical27:
Some suggestions:
1. It seems `src` is the dir where we install nebula, so how about try to read the conf under src (not only change the conf after scp is done). One of the most improvement is that, we can get the `data_path` of meta from the conf.
2. Better only change conf after scp is done
3. `meta_server_addrs` could be consider to modified as well. (If the script is designed to only from 1 to 1, not m to n, some documents need to be add later, could contact @whitewum later)

scripts/meta-transfer-tools.sh
```diff
@@ -0,0 +1,83 @@
...
+        \n \
+    for example \n \
+        ./meta-transfer-tools.sh -f /path-to-src/metad1 -t bob@172.0.0.1:/path-to-dst -u local_ip=172.0.0.1:port=10086 \n \
+    "
+
+
+echo -e $Usage
+}
+
+while getopts "hf:t:u:" opt; do
```
panda-sheep, line 25: Put a colon in front of `h`, to ignore illegal options

scripts/meta-transfer-tools.sh
```diff
@@ -0,0 +1,83 @@
...
+    case $opt in
+        h)
+            usage
```
panda-sheep, line 28: `exit 0`？

scripts/meta-transfer-tools.sh
```diff
@@ -0,0 +1,83 @@
...
+        t)
+            echo "trying to copy metad to: " $OPTARG
+            dst=$OPTARG
+            ;;
+        u)
+            echo "-u: " $OPTARG
+            configs=$OPTARG
+            ;;
+        \?)
+            echo "Invalid option: -$OPTARG"
```
panda-sheep, line 44:
```
usage
exit 1
```

panda-sheep (commented):
good job.

scripts/meta-transfer-tools.sh
liuyu85cn: OK

scripts/meta-transfer-tools.sh
liuyu85cn: ok

scripts/meta-transfer-tools.sh
liuyu85cn: good catch

panda-sheep (approved):

darionyaphet (approved):

PR #2534, opened 2021-08-18
Storage slow query

cangfengzhs:


src/interface/common.thrift
```diff
@@ -390,5 +390,7 @@ enum ErrorCode {
     E_USER_CANCEL                     = -3052,
     E_TASK_EXECUTION_FAILED           = -3053,
 
+    E_PLAN_IS_KILLED                  = -3054,
```
liuyu85cn, line 393: use 3060?

src/interface/storage.thrift
```diff
@@ -334,6 +343,7 @@ struct AddVerticesRequest {
         (cpp.template = "std::unordered_map")   prop_names,
     // if ture, when (vertexID,tagID) already exists, do nothing
     4: bool                                     if_not_exists,
+    5: optional RequestCommon                   common,
```
liuyu85cn, line 346: not used?

src/clients/meta/MetaClient.cpp
```diff
@@ -3473,6 +3474,44 @@ folly::Future<StatusOr<bool>> MetaClient::ingest(GraphSpaceID spaceId) {
...
+  folly::RWSpinLock::ReadHolder holder(sessionLock_);
+  auto it = sessionMap_.find(session_id);
+  if (it != sessionMap_.end()) {
+    return it->second;
+  }
+  return Status::SessionNotFound();
+}
+bool MetaClient::checkIsPlanKilled(SessionID sessionId, ExecutionPlanID planId) {
+  // TODO(hs.zhang): Use RCU to avoid write blocking read
+  folly::RWSpinLock::ReadHolder holder(sessionLock_);
```
critical27, line 3513: The lock itself could be a bottleneck under concurrent scenario, especially you check it in every row. Maybe need to do some profiling test here. Consider only check for every N rows.

Besides, `std::set<std::pair<SessionID, ExecutionPlanID>>` may be a hot spot too, but it won't be a big issue if the check is not very frequent.

cangfengzhs: Okay, I optimized the following points

1. Replace RWLock with RCU
2. Replace std::set with folly::F14FastSet
3. check for every N rows

Regarding the third point, there are two ways. The first is to judge whether it has reached N times outside the checkIsPlanKilled function. This method is accurate, but more complicated; the other is to define a static variable in the checkIsPlanKilled function to count uniformly. This method is simple to implement but It will not be very precise. Which do you prefer?

src/interface/common.thrift
cangfengzhs: OK, I change it.

src/interface/storage.thrift
cangfengzhs: Yes, it is not used. But I think it is better to define uniformly in all Requests

src/clients/meta/MetaClient.cpp
critical27: Sorry for late response, the latter one is ok for now.

critical27 (dismissed):

critical27:
Well done!

liuyu85cn (approved):

critical27 (dismissed):
Good job

critical27 (approved):

liuyu85cn (approved):

src/clients/meta/MetaClient.cpp
```diff
@@ -176,6 +182,11 @@ bool MetaClient::loadData() {
     return false;
   }
 
+  if (!loadSessions()) {
+    LOG(ERROR) << "Load roles Failed";
```
yixinglu, line 186: Load sessions failed

src/clients/meta/MetaClient.cpp
```diff
@@ -3474,5 +3481,55 @@ folly::Future<StatusOr<bool>> MetaClient::ingest(GraphSpaceID spaceId) {
...
 }
 
+bool MetaClient::loadSessions() {
+  auto session_list = listSessions().get();
+  if (!session_list.ok()) {
+    LOG(ERROR) << "List sessions failed, status:" << session_list.status();
+    return false;
+  }
+  SessionMap* old_session_map = sessionMap_.load();
+  SessionMap* new_session_map = new SessionMap(*old_session_map);
```
yixinglu, line 3491: Better to use camel case naming rules

src/clients/meta/MetaClient.cpp
```diff
@@ -3474,5 +3481,55 @@ folly::Future<StatusOr<bool>> MetaClient::ingest(GraphSpaceID spaceId) {
...
+  folly::F14FastSet<std::pair<SessionID, ExecutionPlanID>>* old_killed_plan = killedPlans_.load();
+  folly::F14FastSet<std::pair<SessionID, ExecutionPlanID>>* new_killed_plan =
```
yixinglu, line 3493: use `auto` to simplify the assignment.

src/storage/exec/GetNeighborsNode.h
```diff
@@ -42,7 +42,11 @@ class GetNeighborsNode : public QueryNode<VertexID> {
     if (ret != nebula::cpp2::ErrorCode::SUCCEEDED) {
       return ret;
     }
-
+    if (context_->env()->metaClient_ &&
+        context_->env()->metaClient_->checkIsPlanKilled(context_->planContext_->sessionId_,
+                                                        context_->planContext_->planId_)) {
```
yixinglu, line 47: Why not provide a function to check whether current plan is killed in BaseNode? such as:

```cpp
bool isPlanKilled() {
  auto client = context_->env()->metaClient_ ;
  auto ctx = context_->planContext_;
  return client && client->checkIsPlanKilled(ctx->sessionId_, ctx->planId_)
}
```

src/storage/mutate/UpdateEdgeProcessor.cpp
```diff
@@ -51,8 +51,16 @@ void UpdateEdgeProcessor::doProcess(const cpp2::UpdateEdgeRequest& req) {
...
+  auto session_id =
+      req.common_ref().has_value() && req.common_ref().value().session_id_ref().has_value()
+          ? *req.get_common()->get_session_id()
+          : 0;
+
+  auto plan_id = req.common_ref().has_value() && req.common_ref().value().plan_id_ref().has_value()
+                     ? *req.get_common()->get_plan_id()
+                     : 0;
+  planContext_ =
+      std::make_unique<PlanContext>(env_, spaceId_, session_id, plan_id, spaceVidLen_, isIntId_);
```
yixinglu, line 63: Why not provide a PlanContext constructor for `req.common_ref()`? In the constructor, you can check whether plan id and session id is valid? and The caller is also easier to use, such as:

```cpp
planContext_ =
      std::make_unique<PlanContext>(env_, spaceId_, spaceVidLen_, isIntId_, req.common_ref());


PlanContext(StorageEnv* env,
              GraphSpaceID spaceId,
              size_t vIdLen,
              bool isIntId,
              cpp2::ReqCommon req)
      : env_(env),
        spaceId_(spaceId),
        vIdLen_(vIdLen),
        isIntId_(isIntId),
        sessionId_(0),
        planId_(0) {
  if (req.has_value()) {
    // handle session id and plan id
  }
}
```

panda-sheep (commented):
Good job!

liuyu85cn (approved):

PR #2536, opened 2021-08-18
Breakpad.

CPWstatic:
Do not support arm for now. Because the breakpad has some compiling problems at arm platform.

Release with symbols
```
package.sh -t RelWithDebInfo -r OFF -p ON -s TRUE
```


src/daemons/MetaDaemon.cpp
```diff
@@ -56,9 +60,16 @@ DEFINE_bool(daemonize, true, "Whether run as a daemon process");
 
 static std::unique_ptr<apache::thrift::ThriftServer> gServer;
 static std::unique_ptr<nebula::kvstore::KVStore> gKVStore;
+#if defined(__x86_64__)
+static std::unique_ptr<google_breakpad::ExceptionHandler> gExceptionHandler;
```
Shylock-Hg, line 64: Define the static variable in `SetupBreakpad.cpp`

cmake/nebula/ThirdPartyConfig.cmake
```diff
@@ -106,6 +106,9 @@ find_package(FLEX REQUIRED)
 find_package(LibLZMA REQUIRED)
 find_package(Fizz REQUIRED)
 find_package(Sodium REQUIRED)
+if (${CMAKE_HOST_SYSTEM_PROCESSOR} MATCHES "x86_64")
+find_package(Breakpad REQUIRED)
```
yixinglu, line 110: fix the indent

src/daemons/MetaDaemon.cpp
```diff
@@ -198,6 +209,15 @@ int main(int argc, char* argv[]) {
     return EXIT_FAILURE;
   }
 
+#if defined(__x86_64__)
+  auto expHandler = setupBreakpad();
+  if (!expHandler.ok()) {
+    LOG(ERROR) << expHandler.status();
+    return EXIT_FAILURE;
+  }
+  gExceptionHandler = std::move(expHandler).value();
```
yixinglu, line 218: These logics could also do in the SetupBreakpad.cc

src/daemons/MetaDaemon.cpp
CPWstatic: Emmm, i thought the global one should be declared in the same scope, anyway i'm ok with your idea.

src/daemons/StorageDaemon.cpp
```diff
@@ -62,6 +66,14 @@ int main(int argc, char *argv[]) {
     return EXIT_FAILURE;
   }
 
+#if defined(__x86_64__)
+  status = setupBreakpad();
```
Shylock-Hg, line 70: Do we need this in Debug mode?

package/package.sh
```diff
@@ -10,6 +10,8 @@
 #   -d: Whether to enable sanitizer, default OFF
 #   -t: Build type, default Release
 #   -j: Number of threads, default $(nproc)
+#   -r: Whether enable compressed debug info, defaulr ON
```
jievince, line 13: typo error: defaulr

src/daemons/StorageDaemon.cpp
CPWstatic: It's useful too, because gdb is not always available.

.github/workflows/release.yml
```diff
@@ -12,28 +12,28 @@ defaults:
...
     strategy:
       fail-fast: false
       matrix:
         os:
-          - ubuntu1604
-          - ubuntu1804
-          - ubuntu2004
-          # - centos6
-          - centos7
+          #- ubuntu1604
```
Shylock-Hg, line 20: Why comment these?

CPWstatic: Just for testing.

nevermore3 (approved):

yixinglu (approved):

PR #2544, opened 2021-08-20
[DRAFT] thrift interface for push down some executors 

bright-starry-sky:
Just a draft for push down some executors as below:
 count | sum | max | min | group | de-dup | sort | limit | top | offset-paging

### DIFF
--- a/src/interface/storage.thrift
+++ b/src/interface/storage.thrift
@@ -83,6 +83,11 @@ struct EdgeProp {
 }
 
 
+struct PropItem {
+    1: common.SchemaID  schema_id,
+    2: binary           prop_name,
+}
+
 // Define a vertex property
 struct VertexProp {
     // A valid tag id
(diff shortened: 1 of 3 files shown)


src/interface/storage.thrift
```diff
@@ -129,29 +154,33 @@ struct TraverseSpec {
...
     // A list of expressions which are evaluated on each edge
-    7: optional list<Expr>                      expressions,
-    // A list of expressions used to sort the result
-    8: optional list<OrderBy>                   order_by,
-    // Combined with "limit", the random flag makes the result of each query different
-    9: optional bool                            random,
-    // Return the top/bottom N rows for each given vertex
-    10: optional i64                            limit,
+    6: optional list<Expr>                      expressions,
     // If provided, only the rows satified the given expression will be returned
```
darionyaphet, line 169: It seems that the order has been modified. What is the intention ? thanks

src/interface/storage.thrift
```diff
@@ -466,6 +500,11 @@ struct GetUUIDResp {
 /*
  * Start of Index section
  */
+struct LookupIndexIter {
```
darionyaphet, line 503: so it is the cursor ?

Sophie-Xie:
Maybe need to rewrite it, not included in v2.6.0

PR #2546, opened 2021-08-20
Move the test objects definition to test source tree.

Shylock-Hg:


jievince (approved):

yixinglu (commented):
this target `meta_test_deps` is not used in test directory.

PR #2550, opened 2021-08-23
use local id in the space range for tag edge index

panda-sheep:
As title

1） Tag, edge, tag index and edge index use space range id.
2） Compatible with the existing old schema creating previously.

Synchronize nebula-storage repo pr https://github.com/vesoft-inc/nebula-storage/pull/501

### DIFF
--- a/src/meta/processors/BaseProcessor-inl.h
+++ b/src/meta/processors/BaseProcessor-inl.h
@@ -171,6 +171,69 @@ ErrorOr<nebula::cpp2::ErrorCode, int32_t> BaseProcessor<RESP>::autoIncrementId()
 }
 
 template <typename RESP>
+ErrorOr<nebula::cpp2::ErrorCode, int32_t> BaseProcessor<RESP>::getAvailableGolbalId() {
+  // A read lock has been added before call
+  static const std::string kIdKey = "__id__";
+  int32_t id;
+  std::string val;
+  auto ret = kvstore_->get(kDefaultSpaceId, kDefaultPartId, kIdKey, &val);
+  if (ret != nebula::cpp2::ErrorCode::SUCCEEDED) {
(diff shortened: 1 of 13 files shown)


Shylock-Hg:
Why need this?

panda-sheep:
> Why need this?
Data synchronization between clusters

critical27 (approved):
Reviewed before, LGTM.

bright-starry-sky (approved):

panda-sheep:
Please see the second, compatible with old data

src/meta/processors/BaseProcessor-inl.h
```diff
@@ -170,6 +170,69 @@ ErrorOr<nebula::cpp2::ErrorCode, int32_t> BaseProcessor<RESP>::autoIncrementId()
...
+  auto ret = kvstore_->get(kDefaultSpaceId, kDefaultPartId, localIdkey, &val);
+  if (ret != nebula::cpp2::ErrorCode::SUCCEEDED) {
+    if (ret != nebula::cpp2::ErrorCode::E_KEY_NOT_FOUND) {
+      return ret;
+    }
+
+    // In order to be compatible with the existing old schema, and simple to implement,
+    // when the local_id record does not exist in space, directly use the smallest
+    // id available globally.
+    id = globalId;
```
Shylock-Hg, line 215: I think you could start local id  from 1 when the key don't exists.

panda-sheep: You can refer to the comments in vesoft-inc/nebula-storage#501

liuyu85cn (approved):

Shylock-Hg:
So now there are independent incremental id in separate spaces, the schema id will be same for different spaces, and this behavior different with before. For more compatible I think you could reuse the Global ID and just store them in the local id key to distinguish the owned space, and in this case the schema id in different space won't be same

yixinglu (approved):

Issue #2548, opened 2021-08-23
Crush when using `LIMIT` after `;`


Aiee:
**Describe the bug (__must be provided__)**

Crush when using `LIMIT` after `;`

**Your Environments (__must be provided__)**

* Commit id: a8a4e84d

**How To Reproduce(__must be provided__)**
```
(root@nebula) [nba]> YIELD 1 | limit 1
+---+
| 1 |
+---+
| 1 |
+---+
Got 1 rows (time spent 770/1127 us)

Mon, 23 Aug 2021 10:59:51 CST

(root@nebula) [nba]> YIELD 1 ; limit 1
2021/08/23 10:59:54 [ERROR] Failed to reconnect, failed to open transport, error: dial tcp 192.168.8.6:29562: connect: connection refused
2021/08/23 10:59:54 Loop error, failed to open transport, error: dial tcp 192.168.8.6:29562: connect: connection refused

Bye root!
Mon, 23 Aug 2021 10:59:54 CST

panic: Loop error, failed to open transport, error: dial tcp 192.168.8.6:29562: connect: connection refused
```

**Additional context**

```
Log file created at: 2021/08/23 11:01:53
Running on machine: nebula-dev-1
Log line format: [IWEF]mmdd hh:mm:ss.uuuuuu threadid file:line] msg
I0823 11:01:53.721838 52142 GraphDaemon.cpp:108] Starting Graph HTTP Service
I0823 11:01:53.724076 52143 WebService.cpp:125] Web service started on HTTP[21374], HTTP2[16910]
I0823 11:01:53.724114 52142 GraphDaemon.cpp:122] Number of networking IO threads: 56
I0823 11:01:53.724133 52142 GraphDaemon.cpp:131] Number of worker threads: 56
I0823 11:01:53.724383 52142 MetaClient.cpp:47] Create meta client to "127.0.0.1":38295
I0823 11:01:55.751533 52142 MetaClient.cpp:2951] Load leader of "127.0.0.1":39535 in 1 space
I0823 11:01:55.751559 52142 MetaClient.cpp:2958] Load leader ok
I0823 11:01:55.751652 52142 MetaClient.cpp:94] Register time task for heartbeat!
I0823 11:01:55.780973 52142 GraphDaemon.cpp:161] Starting nebula-graphd on 127.0.0.1:29562
I0823 11:01:55.843283 52550 GraphService.cpp:63] Authenticating user root from [::ffff:192.168.8.6]:44780
[...]

PR #2558, opened 2021-08-24
Fix the dangle limit crash.

Shylock-Hg:
Close https://github.com/vesoft-inc/nebula/issues/2548

### DIFF
--- a/src/graph/validator/SequentialValidator.cpp
+++ b/src/graph/validator/SequentialValidator.cpp
@@ -43,6 +43,16 @@ Status SequentialValidator::validateImpl() {
       break;
   }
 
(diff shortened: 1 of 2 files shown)


Aiee (approved):

jievince:
I think it has been fixed in https://github.com/vesoft-inc/nebula-graph/pull/1314, 
I will move it to current repository

Shylock-Hg:
> I think it has been fixed in [vesoft-inc/nebula-graph#1314](https://github.com/vesoft-inc/nebula-graph/pull/1314),
> I will move it to current repository

Ok, please add this test case.

PR #2572, opened 2021-08-25
format fetchvertices

nevermore3:
1、 change  QueryUtil.h QuerytUtil.cpp to PlannerUtil.h PlannerUtil.cpp
2、 add   ValidateUtil.h ValidateUtil.cpp
3、 refactor fetch vertices  
4、 support  YIELD vertex as name, id(vertex)


src/parser/parser.yy
```diff
@@ -658,6 +658,9 @@ expression
     | reduce_expression {
         $$ = $1;
     }
+    | KW_VERTEX {
```
Shylock-Hg, line 661: `vertex` is a plain identifier means the default output of fetch vertex instead of keyword.

nevermore3:
subjob of https://github.com/vesoft-inc/nebula/issues/2594

yixinglu (dismissed):
LGTM

tests/tck/features/fetch/FetchVertices.strVid.feature
```diff
@@ -444,3 +460,56 @@ Feature: Fetch String Vertices
       GO FROM 'NON EXIST VERTEX ID' OVER serve | FETCH PROP ON team $-
       """
     Then a SyntaxError should be raised at runtime:
+
+  Scenario: format yield
+    When executing query:
+      """
+      FETCH PROP ON * 'Boris Diaw' YIELD id(vertex)
```
CPWstatic, line 467: Why does the `fetch` use vertex/edge while the `subgraph` use vetices/edges? Can we unify these keywords into one so that could reduce the learning cost and confusion for users?

nevermore3: The result returned by subgraph is a list type while the result returned by fetch is a vertex type 。So the column names are different。
This can be discussed later, in fact, I think that both are okay

Shylock-Hg: I think it's ok now.

Shylock-Hg: Change to `edge` and `vertex` for subgraph is more confused.

CPWstatic (approved):

src/graph/validator/FetchVerticesValidator.cpp
```diff
@@ -16,238 +16,137 @@ namespace graph {
...
-    srcRef_ = sentence->vertices()->ref();
-    auto result = checkRef(srcRef_, vidType_);
-    NG_RETURN_IF_ERROR(result);
-    inputVar_ = std::move(result).value();
-    return Status::OK();
+Status FetchVerticesValidator::validateYield(YieldClause *yield) {
+  auto pool = qctx_->objPool();
+  bool noYield = false;
+  if (yield == nullptr) {
+    // version 3.0: return Status::SemanticError("No YIELD Clause");
```
CPWstatic, line 73: Usually, we will write some todos for comment like this:
```
// TODO: Just compatible with previous version, this will be deprecated in version 3.0.
```

src/parser/parser.yy
```diff
@@ -1048,7 +1048,17 @@ opt_argument_list
     ;
 
 argument_list
-    : expression {
+    : KW_VERTEX {
```
CPWstatic, line 1051: Do we have some exception tests for this syntax? like:
```
Go from id(vertex) over *;

Match (n:player) return id(vertex);
```
We should report error in these scenarios.

src/graph/validator/FetchVerticesValidator.cpp
nevermore3: done

src/parser/parser.yy
nevermore3: done

src/graph/validator/MatchValidator.cpp
```diff
@@ -353,6 +353,10 @@ Status MatchValidator::validateReturn(MatchReturn *ret,
   }
   if (ret->returnItems()->columns()) {
     for (auto *column : ret->returnItems()->columns()->columns()) {
+      if (ExpressionUtils::hasAny(column->expr(),
+                                  {Expression::Kind::kVertex, Expression::Kind::kEdge})) {
+        return Status::SemanticError("illegal return clause `%s'", column->toString().c_str());
```
CPWstatic, line 358:
```
The VERTEX and EDGE is not supported in return clause.
```

src/graph/validator/test/MatchValidatorTest.cpp
```diff
@@ -611,6 +611,34 @@ TEST_F(MatchValidatorTest, validateAlias) {
     EXPECT_EQ(std::string(result.message()),
               "SemanticError: Path `p' does not have the type attribute");
   }
+  {
+    std::string query = "MATCH  (v:person) return id(vertex)";
+    auto result = checkResult(query);
+    EXPECT_EQ(
+        std::string(result.message()),
+        "SemanticError: keyword: vertex | edge is not supported in return clause `id(VERTEX)'");
```
Shylock-Hg, line 619: ``vertex' and `edge'`

CPWstatic (approved):

src/graph/validator/CMakeLists.txt
```diff
@@ -24,7 +24,6 @@ nebula_add_library(
     LimitValidator.cpp
     OrderByValidator.cpp
     YieldValidator.cpp
-    TraversalValidator.cpp
```
Shylock-Hg, line 27: Why remove this?

Shylock-Hg: I think you could rename it to QueryValidator.

yixinglu (approved):

PR #2757, opened 2021-08-30
Avoid the extra expression encode decode.

Shylock-Hg:
Avoid inner decode/encode, we only need encode expression when transfer it.


tests/Makefile
```diff
@@ -70,7 +70,7 @@ slow-query: currdir
 	python3 -m pytest -n$(J) -m "not skip" tck/steps/test_kill_slow_query_via_same_service.py && \
 	python3 -m pytest -n$(J) -m "not skip" tck/steps/test_kill_slow_query_via_different_service.py
 
-tck: slow-query
+tck:
```
yixinglu, line 73: revert here

Shylock-Hg: Good catch.

yixinglu (approved):

Aiee (approved):

PR #2762, opened 2021-08-31
fixed fixed_string expression eval for go

bright-starry-sky:
fixed bug for expression eval when prop data type is fixed_string
Repetition steps：
```
CREATE SPACE test_space (partition_num=1,replica_factor=1, vid_type=fixed_string(30));
USE test_space
create tag t1(c1 fixed_string(40))
create edge e1(c1 fixed_string(40))
insert vertex t1(c1) values "1":("row"), "2":("row"), "3":("row")
insert edge e1(c1) values "1" -> "2"@0:("row"), "1" -> "3"@0:("row")
go from "1" over e1 where e1.c1 == "row"
go from "1" over e1 where $$.t1.c1 == "row"
go from "1" over e1 where $^.t1.c1 == "row"
```

### DIFF
--- a/src/storage/exec/QueryUtils.h
+++ b/src/storage/exec/QueryUtils.h
@@ -71,6 +71,10 @@ class QueryUtils final {
       }
       return Status::Error(folly::stringPrintf("Fail to read prop %s ", propName.c_str()));
     }
+    if (field->type() == meta::cpp2::PropertyType::FIXED_STRING) {
+      const auto& fixedStr = value.getStr();
+      return fixedStr.substr(0, fixedStr.find_first_of('\0'));
+    }
     return value;
   }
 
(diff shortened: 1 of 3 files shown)


src/storage/exec/QueryUtils.h
```diff
@@ -71,6 +71,10 @@ class QueryUtils final {
       }
       return Status::Error(folly::stringPrintf("Fail to read prop %s ", propName.c_str()));
     }
+    if (field->type() == meta::cpp2::PropertyType::FIXED_STRING) {
```
Shylock-Hg, line 74: I think you should remove the extra zero in the RowReader.

bright-starry-sky: > I think you should remove the extra zero in the RowReader.

No, there are other effects for reader. such as upgrade tool.

Shylock-Hg: Why upgrade tool need the raw value.

yixinglu (approved):

critical27 (approved):
LGTM

Issue #2772, opened 2021-09-01
The tck slow query case always failed and blocked


Shylock-Hg:
**Please check the FAQ documentation before raising an issue**

Please check the [FAQ](https://github.com/vesoft-inc/nebula/blob/master/docs/manual-EN/1.overview/2.quick-start/2.FAQ.md) documentation and old issues before raising an issue in case someone has asked the same question that you are asking.

**Describe the bug (__must be provided__)**

The tck slow query case always failed and blocked.
```
tck/steps/test_kill_slow_query_via_different_service.py::test_set_up_slow_query_at_first_graph_service 
tck/steps/test_kill_slow_query_via_different_service.py::test_show_all_queries_and_kill_all_slow_queries_at_second_graph_service 
[gw1] [ 50%] FAILED tck/steps/test_kill_slow_query_via_different_service.py::test_show_all_queries_and_kill_all_slow_queries_at_second_graph_service ^C

======================================================================================================================================================== FAILURES ========================================================================================================================================================
________________________________________________________________________________________________________________________ test_show_all_queries_and_kill_all_slow_queries_at_second_graph_service _________________________________________________________________________________________________________________________
[gw1] linux -- Python 3.7.0 /usr/bin/python3
/usr/local/lib/python3.7/site-packages/pytest_bdd/scenario.py:179: in scenario_wrapper
    _execute_scenario(feature, scenario, request, encoding)
/usr/local/lib/python3.7/site-packages/pytest_bdd/scenario.py:145: in _execute_scenario
    _execute_step_function(request, scenario, step, step_func)
/usr/local/lib/python3.7/site-packages/pytest_bdd/scenario.py:115: in _execute_step_function
    return_value = step_func(**kwargs)
tck/conftest.py:335: in result_should_be_in_order
    cmp_dataset(request, graph_spaces, result, order=True, strict=True)
tck/conftest.py:321: in cmp_dataset
    assert res, "\n".join(msg)
E   AssertionError: Fail to exec: SHOW ALL QUERIES | YIELD $-.SessionID AS sid, $-.ExecutionPlanID AS eid, $-.DurationInUSec AS dur WHERE $-.DurationInUSec > 1000000 AND $-.`Query` CONTAINS "GO 100000 STEPS";
E     Response: |sid|eid|dur|
E     
E     Expected: |sid|eid|dur|
E     0: |re.compile('\\d+')|re.compile('\\d+')|re.compile('\\d+')|
E     NotFoundRow: 
[...]

PR #2788, opened 2021-09-02
Fix slow query test.

CPWstatic:
The rebuild_index job is rate limited now. This makes the test has to wait for much time.

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -140,7 +140,7 @@ jobs:
         run: |
           make RM_DIR=false DEBUG=false J=${{ steps.cmake.outputs.j }} tck
         working-directory: tests/
-        timeout-minutes: 60
+        timeout-minutes: 120
       - name: Down cluster
         run: |
           make down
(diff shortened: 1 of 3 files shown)


bright-starry-sky (dismissed):
The slow query test case passed. but another test failed. Let's fix them one by one. 

newly failed:
test_delete_int_vid_one_vertex_all_tag

CPWstatic:
close #2772

critical27:
> The slow query test case passed. but another test failed. Let's fix them one by one.
> 
> newly failed:
> test_delete_int_vid_one_vertex_all_tag

There must be something wrong with the tck test framework when rebuild index speed is slower than expected. I guess it won't appear after #2795 merged.

yixinglu:
> > The slow query test case passed. but another test failed. Let's fix them one by one.
> > newly failed:
> > test_delete_int_vid_one_vertex_all_tag
> 
> There must be something wrong with the tck test framework when rebuild index speed is slower than expected. I guess it won't appear after #2795 merged.

Event if index rebuilding slow down the whole tests, delete tag should be ok. I think we should find out the cause of the failure

CPWstatic:
Fixed by #2795

Issue #2794, opened 2021-09-06
Core in rebuild index


critical27:
There must be something wrong with RebuildIndex or Job/Task Mananger. Perhaps:

`JobId + TaskId` is not unique or the tasks schedule is out of orders.

```
#0  0x00000000025cc119 in std::__detail::_Equal_helper<std::pair<int, int>, std::pair<std::pair<int, int> const, folly::BasicTokenBucket<std::chrono::_V2::steady_clock> >, std::__detail::_Select1st, std::equal_to<std::pair<int, int> >, unsigned long, true>::_S_equals (
    __eq=..., __extract=..., __k=..., __c=5426965902154393287, __n=0x0)
    at /opt/vesoft/toolset/clang/10.0.0/lib/gcc/x86_64-vesoft-linux/9.2.0/../../../../include/c++/9.2.0/bits/hashtable_policy.h:1450
#1  0x00000000025cc098 in std::__detail::_Hashtable_base<std::pair<int, int>, std::pair<std::pair<int, int> const, folly::BasicTokenBucket<std::chrono::_V2::steady_clock> >, std::__detail::_Select1st, std::equal_to<std::pair<int, int> >, std::hash<std::pair<int, int> >, std::__detail::_Mod_range_hashing, std::__detail::_Default_ranged_hash, std::__detail::_Hashtable_traits<true, false, true> >::_M_equals (
    this=0x7f28d924dd00, __k=..., __c=5426965902154393287, __n=0x0)
    at /opt/vesoft/toolset/clang/10.0.0/lib/gcc/x86_64-vesoft-linux/9.2.0/../../../../include/c++/9.2.0/bits/hashtable_policy.h:1834
#2  0x00000000025cbfb6 in std::_Hashtable<std::pair<int, int>, std::pair<std::pair<int, int> const, folly::BasicTokenBucket<std::chrono::_V2::steady_clock> >, std::allocator<std::pair<std::pair<int, int> const, folly::BasicTokenBucket<std::chrono::_V2::steady_clock> > >, std::__detail::_Select1st, std::equal_to<std::pair<int, int> >, std::hash<std::pair<int, int> >, std::__detail::_Mod_range_hashing, std::__detail::_Default_ranged_hash, std::__detail::_Prime_rehash_policy, std::__detail::_Hashtable_traits<true, false, true> >::_M_find_before_node (this=0x7f28d924dd00, __n=2, __k=..., __code=5426965902154393287)
[...]

Issue #2783, opened 2021-09-02
Storage crash when running tck


Shylock-Hg:
**Please check the FAQ documentation before raising an issue**

Please check the [FAQ](https://github.com/vesoft-inc/nebula/blob/master/docs/manual-EN/1.overview/2.quick-start/2.FAQ.md) documentation and old issues before raising an issue in case someone has asked the same question that you are asking.

**Describe the bug (__must be provided__)**

```
(lldb) bt
* thread #1, name = 'nebula-storaged', stop reason = signal SIGABRT
  * frame #0: 0x00007f38315bf53f libc.so.6`__GI_raise(sig=6) at raise.c:51
    frame #1: 0x00007f38315a9895 libc.so.6`__GI_abort at abort.c:79
    frame #2: 0x0000000003d16892 nebula-storaged`rocksdb::port::PthreadCall(char const*, int) (.part.0) + 50
    frame #3: 0x0000000003d16990 nebula-storaged`rocksdb::port::Mutex::Lock() + 32
    frame #4: 0x0000000003ab8ecd nebula-storaged`rocksdb::(anonymous namespace)::CleanupIteratorState(void*, void*) + 61
    frame #5: 0x0000000003c9c82e nebula-storaged`rocksdb::Cleanable::~Cleanable() + 30
    frame #6: 0x0000000003b3cc2a nebula-storaged`rocksdb::DBIter::~DBIter() + 1210
    frame #7: 0x0000000003d323ad nebula-storaged`rocksdb::ArenaWrappedDBIter::~ArenaWrappedDBIter() + 29
    frame #8: 0x00000000038d063e nebula-storaged`std::default_delete<rocksdb::Iterator>::operator(this=0x00007f3801c00068, __ptr=0x00007f37ed01fa00)(rocksdb::Iterator*) const at unique_ptr.h:81
    frame #9: 0x00000000038d05a4 nebula-storaged`std::unique_ptr<rocksdb::Iterator, std::default_delete<rocksdb::Iterator> >::~unique_ptr(this=0x00007f3801c00068) at unique_ptr.h:284
    frame #10: 0x00000000038d0947 nebula-storaged`nebula::kvstore::RocksPrefixIter::~RocksPrefixIter(this=0x00007f3801c00060) at RocksEngine.h:55
[...]

PR #2795, opened 2021-09-06
skip the wait in rate limiter in CI

critical27:
Skip the first second wait in rate limiter, which will take 1 second in every case in TCK tests. And then, another core occur, see #2794, remove the DCHECK temporally. 

Duration of TCK should back to normal.

TCK with ASAN still time out, probably because of spin lock. Still working on it.

image

### DIFF
--- a/src/kvstore/CMakeLists.txt
+++ b/src/kvstore/CMakeLists.txt
@@ -8,6 +8,7 @@ nebula_add_library(
     RocksEngineConfig.cpp
     LogEncoder.cpp
     NebulaSnapshotManager.cpp
+    RateLimiter.cpp
     plugins/elasticsearch/ESListener.cpp
 )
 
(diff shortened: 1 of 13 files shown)


yixinglu:
Good catch. But it's still spend too much time in CI. Maybe what else is still hurting our CI

bright-starry-sky (dismissed):
LGTM .

critical27:
> Good catch. But it's still spend too much time in CI. Maybe what else is still hurting our CI

Yeah, CentOS takes about 6 minutes to run task, fix them one by one.

yixinglu (dismissed):

critical27:
Hold on for on sec, let me fix the DCHECK

critical27:
Close #2794, #2799, #2783

CPWstatic (approved):

bright-starry-sky:
LGTM

bright-starry-sky (approved):

Issue #2602, opened 2021-08-26
[STO/QE]Push down the computation


Sophie-Xie:

critical27:
Some of the most import computation need to push down:
1. limit
2. sample
3. existed bug in filter
Three above should be complete in 2.6.0.

bright-starry-sky:
fixed bug #2762 for GO statement when prop value type is fixed_string.

bright-starry-sky:
for scan vertices via GO, filter push down is work good now. 
`CREATE SPACE test (partition_num=1,replica_factor=1, vid_type=fixed_string(30));
USE test;
create tag st1(c1 string)
create edge se1(c1 string)
create edge se2(c1 string)

insert edge se1(c1) values "1" -> "2"@0:("row"), "1" -> "3"@0:("row")
insert vertex st1(c1) values "1":("row"), "2":("row"), "3":("row")


(root@nebula) [test]> go from "1" over se2 where $$.st1.c1 == "row"
+----------+
| se2._dst |
+----------+
+----------+
Empty set (time spent 7382/7947 us)

Tue, 31 Aug 2021 13:27:15 CST

(root@nebula) [test]> 

(root@nebula) [test]> go from "1" over se1 where $$.st1.c1 == "row"
+----------+
| se1._dst |
+----------+
| "2"      |
+----------+
| "3"      |
+----------+
Got 2 rows (time spent 12583/13348 us)

Tue, 31 Aug 2021 13:27:35 CST

(root@nebula) [test]> go from "1" over * where $$.st1.c1 == "row"
+----------+----------+
| se1._dst | se2._dst |
+----------+----------+
| "2"      |          |
+----------+----------+
| "3"      |          |
+----------+----------+
Got 2 rows (time spent 10977/11707 us)

Tue, 31 Aug 2021 13:27:44 CST

(root@nebula) [test]> `

bright-starry-sky:
This issue is too complicated. Let me split it according to the following features：

1. sample push down  #2946 
2. existed bug in filter (done with #2762)
3. group by push down #2947 
4. order by push down #2948 
5. de-dup push down #2949 
6. limit push down #2950 
7. index paging scan #2951 
8. aggregation (sum/max/min/count) push down #2952

PR #2796, opened 2021-09-06
Lookup limit push down

bright-starry-sky:


src/interface/storage.thrift
```diff
@@ -521,6 +520,8 @@ struct LookupIndexRequest {
     // The list of property names. Should not be empty.
     // Support kVid and kTag for vertex, kSrc, kType, kRank and kDst for edge.
     4: optional list<binary>                return_columns,
+    // max row count of each partition in this response, 0 means no limit.
+    5: i64                                  limit = 0,
```
Shylock-Hg, line 524: Try `optional` and `NULL` means no limit instead zero.

Shylock-Hg: Or you could try max of i64

Shylock-Hg: Or negative value for this.

Shylock-Hg:
And you should add a test case.

src/interface/storage.thrift
bright-starry-sky: I don't see any advantage in your approach. your approach may make the code more complex.

Shylock-Hg: The limit 0 is allowed.

Shylock-Hg: And the other limit field in request is optional too.

bright-starry-sky: > And the other limit field in request is optional too.

good point.

bright-starry-sky:
Addressed comments.

bright-starry-sky:
Related #2602

src/storage/index/LookupBaseProcessor.h
```diff
@@ -81,6 +81,7 @@ class LookupBaseProcessor : public BaseProcessor<RESP> {
   // Save schemas when column is out of index, need to read from data
   std::vector<std::shared_ptr<const meta::NebulaSchemaProvider>> schemas_;
   std::vector<size_t> deDupColPos_;
+  int64_t limit_ = -1;
```
panda-sheep, line 84: maybe the default value of `limit `is 0 (0 means no limit) better?
This is not important

panda-sheep (approved):
Good job！

bright-starry-sky:
Corrected comments. thanks @panda-sheep

panda-sheep (dismissed):

critical27 (approved):
Generally LGTM, wait graph PR is ready.

panda-sheep (approved):

bright-starry-sky:
related #2823

cangfengzhs:
The `limit` must be the top node of the execution plan, otherwise the size of the final dataset will be smaller than the limit count due to the filter (and possibly dedup) discarding part of the data.

Shylock-Hg:
> The `limit` must be the top node of the execution plan, otherwise the size of the final dataset will be smaller than the limit count due to the filter (and possibly dedup) discarding part of the data.

Yes the limit push down can't over the filter operation.

bright-starry-sky:
> The `limit` must be the top node of the execution plan, otherwise the size of the final dataset will be smaller than the limit count due to the filter (and possibly dedup) discarding part of the data.

Good point. Thanks for your remind. 
The limit can't cover filter, ttl and dedup.  I'll deal with it later.

PR #2797, opened 2021-09-06
tck testcases strengthening

bright-starry-sky:
Remove unnecessary index rebuilding.
The index has been created before the data import, so instead of rebuilding.
Let's verify the stability of CI.

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -118,7 +118,7 @@ jobs:
       - name: Make
         run: |
           ccache -z
-          cmake --build build/ -j $(nproc)
+          cmake --build build/ -j $(($(nproc)/2+1))
           ccache -s
       - name: CTest
         env:
--- a/tests/common/utils.py
+++ b/tests/common/utils.py
@@ -426,6 +426,8 @@ def load_csv_data(
         for line in schemas.splitlines():
             resp_ok(sess, line.strip(), True)
 
+        time.sleep(1)
+
         for fd in config["files"]:
             _load_data_from_file(sess, data_dir, fd)
 
--- a/tests/tck/conftest.py
+++ b/tests/tck/conftest.py
@@ -72,21 +72,6 @@ def job_id(resp):
             return job.as_int()
 
 
-def wait_tag_or_edge_indexes_ready(sess, schema: str = "TAG"):
-    resp = resp_ok(sess, f"SHOW {schema} INDEXES")
-    jobs = []
-    for val in resp.column_values("Index Name"):
-        job = val.as_string()
-        resp = resp_ok(sess, f"REBUILD {schema} INDEX {job}", True)
-        jobs.append(job_id(resp))
-    wait_all_jobs_finished(sess, jobs)
-
-
-def wait_indexes_ready(sess):
-    wait_tag_or_edge_indexes_ready(sess, "TAG")
-    wait_tag_or_edge_indexes_ready(sess, "EDGE")
-
-
 @pytest.fixture
 def graph_spaces():
     return dict(result_set=None)
@@ -112,7 +97,6 @@ def preload_space(
     else:
         raise ValueError(f"Invalid space name given: {space}")
     resp_ok(session, f'USE {space};', True)
-    wait_indexes_ready(session)
 
 
 @given("an empty graph")
@@ -171,7 +155,6 @@ def import_csv_data(request, data, graph_spaces, session, pytestconfig):
         data_dir,
         "I" + space_generator(),
     )
-    wait_indexes_ready(session)
     assert space_desc is not None
     graph_spaces["space_desc"] = space_desc
     graph_spaces["drop_space"] = True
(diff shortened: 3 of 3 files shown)


bright-starry-sky:
build failed of ubuntu2004, gcc-9.2
`[ 81%] Building CXX object src/meta/CMakeFiles/meta_service_handler.dir/ActiveHostsMan.cpp.o
[ 81%] Building CXX object src/meta/CMakeFiles/meta_service_handler.dir/MetaServiceUtils.cpp.o
[ 81%] Building CXX object src/meta/CMakeFiles/meta_service_handler.dir/processors/parts/ListPartsProcessor.cpp.o
[ 81%] Building CXX object src/meta/CMakeFiles/meta_service_handler.dir/processors/parts/ListHostsProcessor.cpp.o
[ 81%] Building CXX object src/meta/CMakeFiles/meta_service_handler.dir/processors/parts/CreateSpaceProcessor.cpp.o
Error: Process completed with exit code 137.`

yixinglu:
Disagree to delete all wait steps. If the index has been built, `wait` will not consume much time, because the checks is based on the job status instead of directly waiting for a fixed time

bright-starry-sky:
> Disagree to delete all wait steps. If the index has been built, `wait` will not consume much time, because the checks is based on the job status instead of directly waiting for a fixed time

I see this wait only applies to the data import phase, just wait for the rebuild job, which is unnecessary. right?

yixinglu:
> > Disagree to delete all wait steps. If the index has been built, `wait` will not consume much time, because the checks is based on the job status instead of directly waiting for a fixed time
> 
> I see this wait only applies to the data import phase, just wait for the rebuild job, which is unnecessary. right?

No, this step checks all jobs, not just rebuild jobs

bright-starry-sky:
> Disagree to delete all wait steps. If the index has been built, `wait` will not consume much time, because the check is based on the job status instead of directly waiting for a fixed time.
> 
> And we still have many non-default index scenarios that need to be rebuild. This step is still useful.

yixinglu:
> > Disagree to delete all wait steps. If the index has been built, `wait` will not consume much time, because the checks is based on the job status instead of directly waiting for a fixed time
> 
> I see this wait only applies to the data import phase, just wait for the rebuild job, which is unnecessary. right?

I just don’t want to use `sleep` to check the job status, because `sleep` is sometimes unreliable

bright-starry-sky:
> > > Disagree to delete all wait steps. If the index has been built, `wait` will not consume much time, because the checks is based on the job status instead of directly waiting for a fixed time
> > 
> > 
> > I see this wait only applies to the data import phase, just wait for the rebuild job, which is unnecessary. right?
> 
> I just don’t want to use `sleep` to check the job status, because `sleep` is sometimes unreliable

In this PR, The sleep is not for the state of the job, it is just waiting for the MetaClient, same as sleep(heartbeat_interval_secs)

critical27:
> > > Disagree to delete all wait steps. If the index has been built, `wait` will not consume much time, because the checks is based on the job status instead of directly waiting for a fixed time
> > 
> > 
> > I see this wait only applies to the data import phase, just wait for the rebuild job, which is unnecessary. right?
> 
> I just don’t want to use `sleep` to check the job status, because `sleep` is sometimes unreliable

Why not just insert data with index? We don't use rebuild for all cases.

yixinglu:
> > > > Disagree to delete all wait steps. If the index has been built, `wait` will not consume much time, because the checks is based on the job status instead of directly waiting for a fixed time
> > > 
> > > 
> > > I see this wait only applies to the data import phase, just wait for the rebuild job, which is unnecessary. right?
> > 
> > 
> > I just don’t want to use `sleep` to check the job status, because `sleep` is sometimes unreliable
> 
> Why not just insert data with index? We don't use rebuild for all cases.

This is true for scenarios that are just query nebula, but there are also some DDL/DML scenarios

yixinglu:
> > > > Disagree to delete all wait steps. If the index has been built, `wait` will not consume much time, because the checks is based on the job status instead of directly waiting for a fixed time
> > > 
> > > 
> > > I see this wait only applies to the data import phase, just wait for the rebuild job, which is unnecessary. right?
> > 
> > 
> > I just don’t want to use `sleep` to check the job status, because `sleep` is sometimes unreliable
> 
> In this PR, The sleep is not for the state of the job, it is just waiting for the MetaClient, same as sleep(heartbeat_interval_secs)

Even so, i also don't agree to delete all wait job related steps. at present, it's just for rebuild index, but there are some other jobs we could test them by these steps. And i prefer to use these steps in lookup test cases to replace the `wait x seconds` steps, but i could not do that since the cause of cache updating.

if you don't want to wait in import data, you can only delete the line of waiting job finished. please don't delete these wait implementations.

critical27:
> > > > > Disagree to delete all wait steps. If the index has been built, `wait` will not consume much time, because the checks is based on the job status instead of directly waiting for a fixed time
> > > > 
> > > > 
> > > > I see this wait only applies to the data import phase, just wait for the rebuild job, which is unnecessary. right?
> > > 
> > > 
> > > I just don’t want to use `sleep` to check the job status, because `sleep` is sometimes unreliable
> > 
> > 
> > Why not just insert data with index? We don't use rebuild for all cases.
> 
> This is true for scenarios that are just query nebula, but there are also some DDL/DML scenarios

If the problem is to reduce wait time, still could remove the `REBUILD INDEX` even when DDL/DML. Would you explain why we need to REBUILD for each case?

HarrisChu:
how about just delete code for import?
https://github.com/vesoft-inc/nebula/blob/master/tests/tck/conftest.py#L115 and https://github.com/vesoft-inc/nebula/blob/master/tests/tck/conftest.py#L174

image

liuyu85cn (approved):

panda-sheep (approved):

PR #2803, opened 2021-09-07
refactor insertVertices & insertEdges

nevermore3:


src/graph/validator/MutateValidator.cpp
```diff
@@ -15,108 +15,103 @@ namespace nebula {
...
 
-Status InsertVerticesValidator::prepareVertices() {
-  vertices_.reserve(rows_.size());
-  for (auto i = 0u; i < rows_.size(); i++) {
-    auto *row = rows_[i];
-    if (propSize_ != row->values().size()) {
-      return Status::SemanticError("Column count doesn't match value count.");
-    }
+Status InsertVerticesValidator::validateValues(const std::vector<VertexRowItem *> &rows) {
+  if (rows.empty()) {
```
Shylock-Hg, line 74: DCHECK is enough, the parser won't accept empty rows.

Issue #2818, opened 2021-09-08
case expression bug


czpmango:
nebula test:
```
(czp@nebula) [nba]> RETURN case when null then true else false end
+-----------------------------------------+
| CASE WHEN NULL THEN true ELSE false END |
+-----------------------------------------+
| BAD_TYPE                                |
+-----------------------------------------+
```
neo4j test:
```
neo4j@neo4j> RETURN case when null then true else false end;
+-----------------------------------------+
| case when null then true else false end |
+-----------------------------------------+
| FALSE                                   |
+-----------------------------------------+
```

The result of `yield count(BAD_TYPE)` is `BAD_DATA` in version v2.0.1 but `0` in version v2.5.0.
IMO, aggregate function behavior in version v2.0.1 is correct, bcz you can't just treat `BadNull` as `Null`.

Related questions in the Forum: [https://discuss.nebula-graph.com.cn/t/topic/5658/31](https://discuss.nebula-graph.com.cn/t/topic/5658/31)

PR #2819, opened 2021-09-08
fix case expression

czpmango:
Fixes [https://github.com/vesoft-inc/nebula/issues/2818](https://github.com/vesoft-inc/nebula/issues/2818)

### DIFF
--- a/tests/tck/features/aggregate/Agg.feature
+++ b/tests/tck/features/aggregate/Agg.feature
@@ -13,6 +13,13 @@ Feature: Basic Aggregate and GroupBy
       | 1        | 2     |
     When executing query:
       """
+      YIELD COUNT(CASE WHEN null THEN null ELSE 1 END) AS nulls
+      """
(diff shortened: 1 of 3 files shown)


tests/tck/features/aggregate/Agg.feature
```diff
@@ -11,6 +11,13 @@ Feature: Basic Aggregate and GroupBy
     Then the result should be, in any order, with relax comparison:
       | COUNT(*) | (1+1) |
       | 1        | 2     |
+    When executing query:
+      """
+      YIELD COUNT(CASE WHEN null THEN null ELSE 1 END) AS nulls
```
jievince, line 16: Add `YIELD COUNT(CASE WHEN 1 THEN null ELSE null END) AS nulls`

yixinglu (approved):

jievince (approved):
LGTM

Issue #2566, opened 2021-08-25
Exception in visitor results in an infinite loop


czpmango:
Version: nebula:v2.5.0

```
RETURN time("19:42")+13
```
infinite loop...

PR #2820, opened 2021-09-08
fix time() infinite loop

czpmango:
Fixes [https://github.com/vesoft-inc/nebula/issues/2566](https://github.com/vesoft-inc/nebula/issues/2566)

### DIFF
--- a/src/graph/visitor/DeduceTypeVisitor.cpp
+++ b/src/graph/visitor/DeduceTypeVisitor.cpp
@@ -34,6 +34,7 @@ static const std::unordered_map<Value::Type, Value> kConstantValues = {
     {Value::Type::STRING, Value("123")},
     {Value::Type::DATE, Value(Date())},
     {Value::Type::DATETIME, Value(DateTime())},
+    {Value::Type::TIME, Value(Time())},
     {Value::Type::VERTEX, Value(Vertex())},
     {Value::Type::EDGE, Value(Edge())},
     {Value::Type::PATH, Value(Path())},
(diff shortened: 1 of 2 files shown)


jievince (approved):

Shylock-Hg:
Why lead to inifinite loop?

czpmango:
> Why lead to inifinite loop?

The future's return was not triggered because of the exception, and the current request remained in the asynchronous queue.
When an infinite loop occurs, the log after graphd exits is as follows:
```
E0908 17:45:09.338558 13136 Cpp2Worker.cpp:369] Failed to join outstanding requests.
F0908 17:45:09.339707 13136 ThriftServer.cpp:585] Could not drain active requests within allotted deadline. Deadline value: 30 secs. Abort because undefined behavior is possible. Underlying reasons could be either requests that have never terminated, long running requests, or long queues that could not be fully processed.
```

Shylock-Hg:
> > Why lead to inifinite loop?
> 
> The future's return was not triggered because of the exception, and the current request remained in the asynchronous queue.
> When an infinite loop occurs, the log after graphd exits is as follows:
> 
> ```
> E0908 17:45:09.338558 13136 Cpp2Worker.cpp:369] Failed to join outstanding requests.
> F0908 17:45:09.339707 13136 ThriftServer.cpp:585] Could not drain active requests within allotted deadline. 
> Deadline value: 30 secs. Abort because undefined behavior is possible. 
> Underlying reasons could be either requests that have never terminated, long running requests, or long queues that could not be fully processed.
> ```

Ok, maybe we could add `dcheck` to help avoid this issue.

yixinglu (approved):

jievince (approved):

PR #2828, opened 2021-09-09
Fix storage index

cangfengzhs:
1. modify plan that contain filter.Filter node's input must be vertex/edge node install of scan node.
2. Fix some error in index unit test.

This can only solve the problem of storing and returning truncated data. It cannot deal with the problem of incorrect data set returned due to index truncation. Later, a detailed design for the processing of truncated indexes will be given for everyone to discuss


src/storage/exec/IndexOutputNode.h
```diff
@@ -144,26 +150,22 @@ class IndexOutputNode final : public RelNode<T> {
         ret = vertexRowsFromData(data);
         break;
       }
+      default:
+        assert(false);
```
yixinglu, line 154: replace the `assert` with `LOG(FATAL)`

src/storage/exec/IndexOutputNode.h
```diff
@@ -329,9 +334,40 @@ class IndexOutputNode final : public RelNode<T> {
       default: {
         auto v = IndexKeyUtils::getValueFromIndexKey(
             context_->vIdLen(), data.first, col, fields_, context_->isEdge(), hasNullableCol_);
+        if (v.isStr()) {
+          auto iter = std::find_if(fields_.begin(), fields_.end(), [&col](const auto& field) {
+            return col == field.get_name();
+          });
```
yixinglu, line 340: CHECK whether the iter is valid: `DCHECK(iter != xx.end());`

cangfengzhs: > CHECK whether the iter is valid: `DCHECK(iter != xx.end());`

Just share the discussion details, it is not finished

cangfengzhs:
This bug has been fixed in https://github.com/vesoft-inc/nebula/pull/3196

PR #2831, opened 2021-09-09
Add sample executor.

Shylock-Hg:
Sub job of #2602 

The storage only sample the data in one storage instance, so when collect the data of many storages we should sample them again to keep result even


yixinglu (commented):
When to use the sample plan node? Can you provide a complete implementation? Let's see the necessity of introducing the sample plan node.

Shylock-Hg:
> When to use the sample plan node? Can you provide a complete implementation? Let's see the necessity of introducing the sample plan node.

Sample the result of GN now.

PR #2841, opened 2021-09-10
init test zone

HarrisChu:
hold on merge, just submit example for review, I would any more test cases.

some pending tasks:
1. the cases depend on k8s namespace, currently use `default` namespace. e.g. ADD ZONE z2 "{}-storaged-1.{}-storaged-headless.default.svc.cluster.local":9779
2. need use chaos-mesh to simulate storage is crashed.
3. use nebula-operator to setup the nebula cluster, but it would wait for the first 20 seconds for readiness probe. https://github.com/vesoft-inc/nebula-operator/issues/61


HarrisChu:
result 
image

HarrisChu:
@kikimo CC

HarrisChu:
update, run test case via github action.
https://github.com/HarrisChu/nebula/runs/3730960719?check_suite_focus=true

Issue #2739, opened 2020-10-21
The default value stored as an expression existed one problem.


laura-ding:
The default value stored as an expression, when the expression is a system call, the in edge and out edge properties may not be the same.

PR #2854, opened 2021-09-13
fix default value bug

czpmango:
Fixes [https://github.com/vesoft-inc/nebula/issues/2739](https://github.com/vesoft-inc/nebula/issues/2739)

### DIFF
--- a/src/graph/validator/MaintainValidator.cpp
+++ b/src/graph/validator/MaintainValidator.cpp
@@ -39,19 +39,17 @@ Status SchemaValidator::validateColumns(const std::vector<ColumnSpecification *>
       column.type.set_type_length(spec->typeLen());
     }
     for (const auto &property : spec->properties()->properties()) {
+      auto *defaultExpr = const_cast<Expression *>(property->defaultValue());
       if (property->isNullable()) {
         column.set_nullable(property->nullable());
-      } else if (property->isDefaultValue()) {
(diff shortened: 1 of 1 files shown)


src/graph/validator/MaintainValidator.cpp
```diff
@@ -39,19 +39,17 @@ Status SchemaValidator::validateColumns(const std::vector<ColumnSpecification *>
...
           return Status::SemanticError("Wrong default value experssion `%s'",
                                        property->defaultValue()->toString().c_str());
         }
-        auto *defaultValueExpr = property->defaultValue();
-        // some expression is evaluable but not pure so only fold instead of
-        // eval here
-        auto foldRes = ExpressionUtils::foldConstantExpr(defaultValueExpr);
-        NG_RETURN_IF_ERROR(foldRes);
-        column.set_default_value(foldRes.value()->encode());
+        auto val = defaultExpr->eval(QueryExpressionContext()(nullptr));
```
Shylock-Hg, line 50: Can't evaluate here when create schema. If user give `time()` as default value expression, he want the time when insert data instead of create schema.

Shylock-Hg: The correct way is evaluate when insert edges in #2739 , or introduce transaction.

czpmango: ACK.

PR #2862, opened 2021-09-14
fix insert edge

czpmango:
Fixes [https://github.com/vesoft-inc/nebula/issues/2739](https://github.com/vesoft-inc/nebula/issues/2739)

### DIFF
--- a/src/graph/validator/MutateValidator.cpp
+++ b/src/graph/validator/MutateValidator.cpp
@@ -148,7 +148,7 @@ Status InsertEdgesValidator::toPlan() {
                                   nullptr,
                                   spaceId_,
                                   std::move(edges_),
-                                  std::move(propNames_),
+                                  std::move(entirePropNames_),
                                   ifNotExists_,
                                   useChainInsert);
   root_ = doNode;
@@ -196,6 +196,11 @@ Status InsertEdgesValidator::prepareEdges() {
   auto useToss = isoLevel == IsoLevel::TOSS;
   auto size = useToss ? rows_.size() : rows_.size() * 2;
   edges_.reserve(size);
+
+  size_t fieldNum = schema_->getNumFields();
+  for (size_t j = 0; j < fieldNum; ++j) {
+    entirePropNames_.emplace_back(schema_->field(j)->name());
+  }
   for (auto i = 0u; i < rows_.size(); i++) {
     auto *row = rows_[i];
     if (propNames_.size() != row->values().size()) {
@@ -233,6 +238,34 @@ Status InsertEdgesValidator::prepareEdges() {
     auto valsRet = SchemaUtil::toValueVec(row->values());
     NG_RETURN_IF_ERROR(valsRet);
     auto props = std::move(valsRet).value();
+
+    std::vector<Value> entirePropValues;
+    for (size_t j = 0; j < fieldNum; ++j) {
+      auto *field = schema_->field(j);
+      auto propName = entirePropNames_[j];
+      auto iter = std::find(propNames_.begin(), propNames_.end(), propName);
+      if (iter == propNames_.end()) {
+        if (field->hasDefault()) {
+          auto *defaultValue = field->defaultValue();
+          DCHECK(!!defaultValue);
+          auto v = defaultValue->eval(QueryExpressionContext()(nullptr));
+          entirePropValues.emplace_back(v);
+        } else {
+          if (!field->nullable()) {
+            return Status::SemanticError(
+                "The property `%s' is not nullable and has no default value.", field->name());
+          }
+          entirePropValues.emplace_back(Value(NullType::__NULL__));
+        }
+      } else {
+        auto v = props[std::distance(propNames_.begin(), iter)];
+        if (!field->nullable() && v.isNull()) {
+          return Status::SemanticError("The non-nullable property `%s' could not be NULL.",
+                                       field->name());
+        }
+        entirePropValues.emplace_back(v);
+      }
+    }
     storage::cpp2::NewEdge edge;
     storage::cpp2::EdgeKey key;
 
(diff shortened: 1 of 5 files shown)


yixinglu (commented):
Fix what problem? could you give some cases ?

czpmango:
> Fix what problem? could you give some cases ?

In the case of multiple replicas, insert statements may cause inconsistent attribute values ​​between replicas.
```
create edge E(prop timestamp default timestamp())
insert edge E() values "1"->"2":()
```

Shylock-Hg:
> > Fix what problem? could you give some cases ?
> 
> In the case of multiple replicas, insert statements may cause inconsistent attribute values ​​between replicas.
> 
> ```
> create edge E(prop timestamp default timestamp())
> insert edge E() values "1"->"2":()
> ```

Between partitions instead of replicas.

czpmango:
> > > Fix what problem? could you give some cases ?
> > 
> > 
> > In the case of multiple replicas, insert statements may cause inconsistent attribute values ​​between replicas.
> > ```
> > create edge E(prop timestamp default timestamp())
> > insert edge E() values "1"->"2":()
> > ```
> 
> Between partitions instead of replicas.

You are right.

Shylock-Hg:
Add case to compare properties of in/out edges

Shylock-Hg:
Close #2867

kikimo:
> Close #2867
@Shylock-Hg 

We are testing TOSS feature, not sure this is the same problem you are confronting.

Shylock-Hg:
> > Close #2867
> > @Shylock-Hg
> 
> We are testing TOSS feature, not sure this is the same problem you are confronting.
> If it's timestamp() mismatch, then #2867 is a whold different problem.

TOSS is another way could resolve this problem.

kikimo:
> > > Close #2867
> > > @Shylock-Hg
> > 
> > 
> > We are testing TOSS feature, not sure this is the same problem you are confronting.
> > If it's timestamp() mismatch, then #2867 is a whold different problem.
> 
> TOSS is another way could resolve this problem.

Sadly TOSS cannot fix too. The problem issue #2867 mentioning is the mismatch of int value of edge props without default. BTW @liuyu85cn has already did some fix for the default timestamp() value mismatch at the storage level.

Shylock-Hg:
> > > > Close #2867
> > > > @Shylock-Hg
> > > 
> > > 
> > > We are testing TOSS feature, not sure this is the same problem you are confronting.
> > > If it's timestamp() mismatch, then #2867 is a whold different problem.
> > 
> > 
> > TOSS is another way could resolve this problem.
> 
> Sadly TOSS cannot fix too. The problem issue #2867 mentioning is the mismatch of int value of edge props without default. BTW @liuyu85cn has already did some fix for the default timestamp() value mismatch at the storage level.

Could resolve the default value issue.

czpmango:
> > Close #2867
> > @Shylock-Hg
> 
> We are testing TOSS feature, not sure this is the same problem you are confronting.
> If it's timestamp() mismatch, then #2867 is a whold different problem.

This pr and issue#2739 has nothing to do with toss, it is just to fix `Default Value` problem.

src/graph/validator/MutateValidator.cpp
```diff
@@ -233,6 +238,32 @@ Status InsertEdgesValidator::prepareEdges() {
...
+      auto iter = std::find(propNames_.begin(), propNames_.end(), propName);
+      if (iter == propNames_.end()) {
+        if (field->hasDefault()) {
+          auto *defaultValue = field->defaultValue();
+          DCHECK(!!defaultValue);
+          auto v = defaultValue->eval(QueryExpressionContext()(nullptr));
+          entirePropValues.emplace_back(v);
+        } else {
+          if (!field->nullable()) {
+            return Status::SemanticError("The not null field doesn't have a default value.");
```
yixinglu, line 255: refine the error message: 
```
The property `propName' is not nullable and has no default value.
```

src/graph/validator/MutateValidator.cpp
```diff
@@ -233,6 +238,32 @@ Status InsertEdgesValidator::prepareEdges() {
...
+          }
+          entirePropValues.emplace_back(Value(NullType::__NULL__));
+        }
+      } else {
+        auto v = props[std::distance(propNames_.begin(), iter)];
+        if (!field->nullable() && v.isNull()) {
+          return Status::SemanticError("The not null field cannot be null");
```
yixinglu, line 262: ditto:

```
The non-nullable property `propName' could not be NULL.
```

src/graph/validator/MutateValidator.cpp
```diff
@@ -233,6 +238,32 @@ Status InsertEdgesValidator::prepareEdges() {
     auto valsRet = SchemaUtil::toValueVec(row->values());
     NG_RETURN_IF_ERROR(valsRet);
     auto props = std::move(valsRet).value();
+
+    std::vector<Value> entirePropValues;
+    for (size_t j = 0; j < fieldNum; ++j) {
+      auto *field = schema_->field(j);
+      auto propName = entirePropNames_[j];
```
yixinglu, line 245: It seems that the `entirePropNames_` is not necessary, why not use `schema_->field(j)->name()` here directly. And you needs not to save it as the field of `MutateValidator`

src/graph/validator/MutateValidator.cpp
czpmango: ok. I copied previous error messege in storage..

src/graph/validator/MutateValidator.cpp
czpmango: ACK.

src/graph/validator/MutateValidator.cpp
czpmango: ACK.

src/graph/validator/MutateValidator.cpp
czpmango: Also used in `InsertEdgesValidator::toPlan()`.

yixinglu (approved):

Aiee (approved):

PR #2853, opened 2021-09-13
Feature/sample go

Shylock-Hg:
Sub job of #2602

PR #2878, opened 2021-09-16
Change limit to expression.

Shylock-Hg:
For the limit push down in Loop

Subjob of #2602


src/graph/planner/plan/Query.h
```diff
@@ -663,14 +719,20 @@ class Limit final : public SingleInputNode {
...
+  Limit(QueryContext* qctx, PlanNode* input, int64_t offset, Expression* count)
+      : SingleInputNode(qctx, Kind::kLimit, input) {
+    offset_ = offset;
     count_ = count;
   }
 
   void cloneMembers(const Limit&);
 
  private:
   int64_t offset_{-1};
```
czpmango, line 734: Need change `offset` into an expression?

Shylock-Hg: offset won't push down

src/graph/planner/plan/Query.h
```diff
@@ -652,9 +683,34 @@ class Limit final : public SingleInputNode {
...
-  int64_t count() const { return count_; }
+  // Get constant count value
+  int64_t count() const {
+    if (count_ == nullptr) {
+      return -1;
+    }
+    DCHECK(ExpressionUtils::isEvaluableExpr(count_));
+    QueryExpressionContext ctx;
+    return count_->eval(ctx).getInt();
+  }
```
czpmango, line 703: Is this interface necessary?

src/graph/planner/plan/Query.h
czpmango: It is better to ensure consistency in use.
```
go from ....| offset abs(1)+3 limit abs(3)+1
go from ....| limit abs(3)+1,abs(3)
```

Shylock-Hg: This is used for new syntax for `go ... limit [..]`

src/graph/planner/plan/Query.h
Shylock-Hg: Comptible.

Shylock-Hg:
Subjob of #2602

src/graph/optimizer/rule/LimitPushDownRule.cpp
```diff
@@ -50,6 +50,10 @@ StatusOr<OptRule::TransformResult> LimitPushDownRule::transform(
   const auto proj = static_cast<const Project *>(projGroupNode->node());
   const auto gn = static_cast<const GetNeighbors *>(gnGroupNode->node());
 
+  DCHECK(graph::ExpressionUtils::isEvaluableExpr(limit->countExpr()));
```
yixinglu, line 53: Don't use DCHECK to debug your code if it's the regular branch you need to handle.

yixinglu (commented):
I don't know what this PR is for? could you give some description?

src/graph/optimizer/rule/LimitPushDownRule.cpp
Shylock-Hg: This won't happen in correct state.

Shylock-Hg:
> I don't know what this PR is for? could you give some description?

Push down limit to GN in Loop for each step.

nevermore3 (dismissed):

src/graph/optimizer/rule/LimitPushDownRule.cpp
yixinglu: So why did you do the same check in line 54?

Shylock-Hg: Handle exception scenes

Sophie-Xie:
Include in #2853

PR #2885, opened 2021-09-16
Improve memory watermark detection

### LINKED ISSUE #2497: memory_high_watermark use freeram as available memory
NEED DISCUSS.

we use freeram as available memory, it would prevent queries if there are large buffer/cache.

image

image

especially when deploy graph and storage in same machine, storage copy data into buffer cache

### LINKED ISSUE #2591: [QE]Memory usage threshold refinery
fixes vesoft-inc/nebula-docker-compose#93

yixinglu:
fix #2497 and fix #2591


HarrisChu:
cc @MegaByte875 

Watermark would check `/sys/fs/cgroup/memory.soft_limit_in_bytes` in container.
Means if the pod limit is 2G, it would prevent queries if the memory is 1.6G by default.

czpmango (approved):
LGTM.

src/graph/service/QueryEngine.cpp
```diff
@@ -39,19 +41,45 @@ Status QueryEngine::init(std::shared_ptr<folly::IOThreadPoolExecutor> ioExecutor
...
                                              indexManager_.get(),
                                              storage_.get(),
                                              metaClient_,
                                              charsetInfo_);
-  auto* instance = new QueryInstance(std::move(ectx), optimizer_.get());
+  auto* instance = new QueryInstance(std::move(qctx), optimizer_.get());
   instance->execute();
 }
 
+Status QueryEngine::setupBackgroundThread() {
```
CPWstatic, line 58: We had many background thread now, better name it properly.

CPWstatic: Such as MemoryMonitor.

critical27 (commented):
Generally LGTM, wait conflicts resolved.

src/graph/context/Iterator.cpp
```diff
@@ -8,12 +8,34 @@
...
+bool Iterator::hitsSysMemoryHighWatermark() const {
+  if (checkMemory_) {
+    if (numRowsModN_ >= FLAGS_num_rows_to_check_memory) {
+      numRowsModN_ -= FLAGS_num_rows_to_check_memory;
+    }
+    if (UNLIKELY(numRowsModN_ == 0)) {
+      if (MemoryUtils::kHitMemoryHighWatermark.load()) {
+        throw std::runtime_error(
+            folly::sformat("Used memory hits the high watermark({}) of total system memory.",
+                           FLAGS_system_memory_high_watermark_ratio));
```
bright-starry-sky, line 30: Where is handler the runtime error?

bright-starry-sky: > Where is handler the runtime error?

What's about as below ?
```
if (MemoryUtils::kHitMemoryHighWatermark.load()) {
LOG(ERROR);
return true;
}
```

critical27: There is a thenError in QueryInstance

bright-starry-sky: > There is a thenError in QueryInstance

got . LGTM now. Please resolved conflicts.

PR #2886, opened 2021-09-16
[DRAFT] index paging scan

bright-starry-sky:


src/interface/storage.thrift
```diff
@@ -481,6 +481,11 @@ struct GetUUIDResp {
 /*
  * Start of Index section
  */
+struct PagingScanPair {
```
Shylock-Hg, line 484: Maybe should change `end` to `count`, the graph don't know the end key.

Issue #2594, opened 2021-08-26
[QE]Better yield data format (in vertices and edges) 


Sophie-Xie:

critical27:
Mark: Storage need to check if the vertex exists when fetch tags.

critical27:
The time to merge need to discuss later, IIRC, not 2.6.0 for all of related feature.

PR #2895, opened 2021-09-17
format lookup

nevermore3:
support  YIELD vertex OR edge


nevermore3:
subjob of #2594

src/graph/validator/LookupValidator.cpp
```diff
@@ -54,51 +54,148 @@ Status LookupValidator::prepareFrom() {
...
+}
+
+Status LookupValidator::validateYield() {
+  auto pool = qctx_->objPool();
+  auto* newCols = pool->add(new YieldColumns());
+  lookupCtx_->yieldExpr = newCols;
   if (lookupCtx_->isEdge) {
+    idxReturnCols_.emplace_back(kSrc);
+    idxReturnCols_.emplace_back(kDst);
+    idxReturnCols_.emplace_back(kType);
```
bright-starry-sky, line 162: Do we need to return EdgeType ?

nevermore3: need EdgeType when  YIELD edge

bright-starry-sky: > need EdgeType when YIELD edge

I think you can get the EdgeType from graph layer. you can find edgeType by indexID. 
this will cause additional net|memory load if pushing down to the storage layer.

nevermore3: we use iterator to collect edge, so the returned result needs to include edgetype,

yixinglu (approved):
LGTM

yixinglu (approved):
LGTM

bright-starry-sky (approved):
Good job.

Issue #2583, opened 2021-08-26
Limit push down over filter


Shylock-Hg:
**Please check the FAQ documentation before raising an issue**

**Describe the bug (__must be provided__)**

See detail in https://discuss.nebula-graph.com.cn/t/topic/5474

And the limit shouldn't perform before filter operation (e.g. filter, dedup and etc.)

**Your Environments (__must be provided__)**

* OS: `uname -a`
* Compliler: `g++ --version` or `clang++ --version`
* CPU: `lscpu`
* Commit id (e.g. `a3ffc7d8`)

**How To Reproduce(__must be provided__)**

1. Step 1
2. Step 2
3. Step 3

**Expected behavior**

**Additional context**

Provide logs and configs, or any other context to trace the problem.

critical27:
Same with #2602, this is closed

Shylock-Hg:
Sub job of #2602

PR #2905, opened 2021-09-18
fixed bug limit push down of index scan.

bright-starry-sky:
related PR #2796 
The logic of the previous PR is simple, handles limits during the index scan phase. Scanning filter, data validity, etc., are not considered.

for this PR , The items below will be covered by：

- index filter
- index ttl
- Invalid data for index
- Invalid data for prop value
- skip the execution plan when limit = 0

Temporarily unsupported

- de-dup execute node. currently, de-Dup does not work on index scans. We can ignore it for now. Handler it when `Limit + Paging Scan` is integrated

### DIFF
--- a/src/storage/exec/IndexEdgeNode.h
+++ b/src/storage/exec/IndexEdgeNode.h
@@ -22,7 +22,7 @@ class IndexEdgeNode final : public RelNode<T> {
                 IndexScanNode<T>* indexScanNode,
                 const std::vector<std::shared_ptr<const meta::NebulaSchemaProvider>>& schemas,
                 const std::string& schemaName,
-                int64_t limit)
+                int64_t limit = -1)
       : context_(context),
         indexScanNode_(indexScanNode),
         schemas_(schemas),
@@ -42,11 +42,7 @@ class IndexEdgeNode final : public RelNode<T> {
     data_.clear();
     std::vector<storage::cpp2::EdgeKey> edges;
     auto* iter = static_cast<EdgeIndexIterator*>(indexScanNode_->iterator());
-    int64_t count = 0;
     while (iter && iter->valid()) {
-      if (limit_ > -1 && count++ == limit_) {
-        break;
-      }
       if (context_->isPlanKilled()) {
         return nebula::cpp2::ErrorCode::E_PLAN_IS_KILLED;
       }
@@ -66,6 +62,7 @@ class IndexEdgeNode final : public RelNode<T> {
       edges.emplace_back(std::move(edge));
       iter->next();
     }
+    int64_t count = 1;
     for (const auto& edge : edges) {
       auto key = NebulaKeyUtils::edgeKey(context_->vIdLen(),
                                          partId,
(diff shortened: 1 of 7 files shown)


src/storage/exec/IndexEdgeNode.h
```diff
@@ -66,6 +62,7 @@ class IndexEdgeNode final : public RelNode<T> {
       edges.emplace_back(std::move(edge));
       iter->next();
     }
+    int64_t count = 1;
```
critical27, line 65: Why `count` is initialized with 1?

bright-starry-sky: > Why `count` is initialized with 1?

 limit has three types of boundary value:
limit == -1, This means no limits and is a default value. 
limit == 0, This means don't need to return any data. this one don't need to any data scan. handler in LookupProcessor.cpp:26
limit > 0, true used for limit. the count number begin with 1.

Shylock-Hg:
Close #2583

src/storage/exec/IndexEdgeNode.h
critical27: I mean you should initialize with 0? See line 82 `limit_ > 0 && ++count > limit_`

panda-sheep: It looks ok
Because here put the data in first, and then judge
If `count` is initialized to 0, the code on line 82 is as follows:
```
if (limit_> 0 && ++count >= limit_) {
    break;
}
```

panda-sheep: But I prefer to initialize `count` to 0, wdyt?

src/storage/index/LookupBaseProcessor-inl.h
```diff
@@ -319,8 +321,8 @@ std::unique_ptr<IndexOutputNode<IndexID>> LookupBaseProcessor<REQ, RESP>::buildP
   auto indexId = ctx.get_index_id();
   auto colHints = ctx.get_column_hints();
 
-  auto indexScan = std::make_unique<IndexScanNode<IndexID>>(
-      context_.get(), indexId, std::move(colHints), limit_);
+  auto indexScan =
+      std::make_unique<IndexScanNode<IndexID>>(context_.get(), indexId, std::move(colHints));
```
panda-sheep, line 325: Why doesn't this plan need `limit`?

src/storage/exec/IndexEdgeNode.h
bright-starry-sky: > But I prefer to initialize `count` to 0, wdyt?

Aha, the point of ambiguity is here, they are all right. the code readability is a bit low.
Changed the init count to 0.

src/storage/index/LookupBaseProcessor-inl.h
bright-starry-sky: > Why doesn't this plan need `limit`?

if FilterNode exists. will check the limit in FilterNode. skip the IndexScanNode.

panda-sheep (approved):
Good job！

critical27 (approved):

PR #2913, opened 2021-09-23
Add the basic cached paging context.

Shylock-Hg:
Subjob of #2602

### DIFF
--- a/src/graph/session/PagingContext.h
+++ b/src/graph/session/PagingContext.h
@@ -6,8 +6,6 @@
 
 #pragma once
 
-#include <bits/stdint-intn.h>
-
 #include <tuple>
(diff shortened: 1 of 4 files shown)


src/graph/session/PagingContext.h
```diff
@@ -0,0 +1,40 @@
...
+                                                                    int64_t offset) const;
+
+  void update(GraphSpaceID spaceId,
+              int32_t schemaId,
+              int64_t offset,
+              std::vector<std::tuple<PartitionID, std::string>> keys);
+
+ private:
+  // Cached paging key
+  std::unordered_map<
```
Shylock-Hg, line 32: Should cache the filter too.

Issue #2911, opened 2021-09-23
Support direct connection with native account when graph auth_type = cloud


kqzh:
**Is your feature request related to a problem? Please describe.**
when I try to use nebula native account connect to graph which auth_type=cloud, it will be refused, because login request will send to DBaaS whether it is a native account or not.

**Describe the solution you'd like**
I think we can support user use native account to login when graph auth_type=cloud, such as checked metad first, if not exist, then reqeust to cloud

**Describe alternatives you've considered**
n/a

**Additional context**
n/a

bright-starry-sky:
we can test the method of `auth(user, pwd)` via PR #2914.

PR #2914, opened 2021-09-23
Support cloud user and nebula native user can be mixed

bright-starry-sky:
related the new feature for cloud user and nebula native user. issue #2911 .

Please hold this PR. we need both front-end and back-end debugging.

### DIFF
--- a/src/graph/service/GraphService.cpp
+++ b/src/graph/service/GraphService.cpp
@@ -164,12 +164,21 @@ bool GraphService::auth(const std::string& username, const std::string& password
   if (!FLAGS_enable_authorize) {
     return true;
   }
+
   if (FLAGS_auth_type == "password") {
     auto authenticator = std::make_unique<PasswordAuthenticator>(queryEngine_->metaClient());
     return authenticator->auth(username, encryption::MD5Utils::md5Encode(password));
   } else if (FLAGS_auth_type == "cloud") {
-    auto authenticator = std::make_unique<CloudAuthenticator>(queryEngine_->metaClient());
-    return authenticator->auth(username, password);
+    // Cloud user and native user will be mixed.
+    // Since cloud user and native user has the same transport protocol,
+    // There is no way to identify which one is in the graph layer，
+    // let's check the native user's password first, then cloud user.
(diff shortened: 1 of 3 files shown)


src/graph/service/GraphService.cpp
```diff
@@ -164,12 +164,21 @@ bool GraphService::auth(const std::string& username, const std::string& password
...
+    // Cloud user and native user will be mixed.
+    // Since cloud user and native user has the same transport protocol,
+    // There is no way to identify which one is in the graph layer，
+    // let's check the native user's password first, then cloud user.
+    auto pwdAuth = std::make_unique<PasswordAuthenticator>(queryEngine_->metaClient());
+    if (pwdAuth->auth(username, encryption::MD5Utils::md5Encode(password))) {
+      return true;
+    }
+    auto cloudAuth = std::make_unique<CloudAuthenticator>(queryEngine_->metaClient());
+    return cloudAuth->auth(username, password);
```
kqzh, line 181: so in cloudAuth->auth function, there is no need to checkShadowAccount now?

Shylock-Hg:
I think better to add one `field` in request to identify them

bright-starry-sky:
And you need to delete the god user "root" first before create cloud root user.

veezhang:
Would it be better to authorize explicitly rather than implicitly?

For example, use a code code, or add a field to indicate explicitly that the cloud allows password authentication.

```c++

bool CloudAuthenticator::auth(const std::string& user, const std::string& password) {
  // ...
  try {
    auto json = folly::parseJson(result.value());
    if (json["code"].asString().compare("0") != 0) {
      LOG(ERROR) << "Cloud authentication failed, user: " << user;
      return false;
    }

    // password_authentication eq true indicates that cloud allows use the password authentication
    if (json["password_authentication"].asString().compare("true") == 0) {
      auto authenticator = std::make_unique<PasswordAuthenticator>(metaClient_);
      return authenticator->auth(user, encryption::MD5Utils::md5Encode(password));
    }
  } catch (std::exception& e) {
    LOG(ERROR) << "Invalid json: " << e.what();
    return false;
  }
  return true;
}
```

panda-sheep (approved):
Good job

kqzh (approved):
LGTM

PR #2921, opened 2021-09-24
Add docker major version tag when releasing

yixinglu:
For releasing each version, action will update the major version tag for each docker image, such as:

```
v3.0.0 -> v3, latest
v2.6.0 -> v2, latest
v2.5.1 -> v2, latest
```


.github/workflows/release.yml
```diff
@@ -71,7 +71,10 @@ jobs:
         with:
           context: .
           file: ./docker/Dockerfile.${{ matrix.service }}
-          tags: vesoft/nebula-${{ matrix.service }}:${{ steps.tagname.outputs.tag }},vesoft/nebula-${{ matrix.service }}:latest
+          tags: |
+            vesoft/nebula-${{ matrix.service }}:${{ steps.tagname.outputs.tag }}
+            vesoft/nebula-${{ matrix.service }}:${{ steps.tagname.outputs.majorver }}
+            vesoft/nebula-${{ matrix.service }}:latest
```
HarrisChu, line 77: suppose we release v3.1.0 and v2.7.1 at the same, it may be confused for `latest` version.
how about just mark the highest version as latest?

in this case, v3 should be the latest, but v2 should not.

yixinglu: Fixed

HarrisChu (approved):

Aiee (approved):

PR #2957, opened 2021-09-28
format find path

nevermore3:
must add YIELD path in findpath sentence

incompatible
KW_PATH changed from non-keyword to keyword, so disallow users  to define path as a variable .
some test case :
find path from a to b over like |order by $-.path  or. match path = (a)-[]->(b) return path.   are not allowed

### DIFF
--- a/.linters/cpp/checkKeyword.py
+++ b/.linters/cpp/checkKeyword.py
@@ -37,6 +37,7 @@ reserved_key_words = [
     'KW_WHEN',
     'KW_DELETE',
     'KW_FIND',
+    'KW_PATH',
     'KW_LOOKUP',
     'KW_ALTER',
     'KW_STEPS',
--- a/src/graph/validator/FindPathValidator.cpp
+++ b/src/graph/validator/FindPathValidator.cpp
@@ -72,23 +72,14 @@ Status FindPathValidator::validateYield(YieldClause* yield) {
     yield = pool->add(new YieldClause(yieldColumns));
   }
 
-  YieldColumns* newCols = pool->add(new YieldColumns());
   for (auto& col : yield->columns()) {
-    if (!ExpressionUtils::hasAny(col->expr(), {Expression::Kind::kPathBuild})) {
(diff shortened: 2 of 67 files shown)


nevermore3:
subjob of #2594

.linters/cpp/checkKeyword.py
```diff
@@ -38,6 +38,7 @@
     'KW_WHEN',
     'KW_DELETE',
     'KW_FIND',
+    'KW_PATH',
```
Shylock-Hg, line 41: Add key work means incompatible. The user can't use `path` as label again.

nevermore3: yes

src/parser/parser.yy
```diff
@@ -1068,6 +1067,11 @@ argument_list
         Expression *arg = EdgeExpression::make(qctx->objPool());
         $$->addArgument(arg);
     }
+    | KW_PATH {
+        $$ = ArgumentList::make(qctx->objPool());
+        Expression *arg = PathBuildExpression::make(qctx->objPool());
+        $$->addArgument(arg);
+    }
```
yixinglu, line 1074: Delete it if not support the `endnotes(PATH)` syntax.

yixinglu (approved):

src/graph/validator/FindPathValidator.cpp
```diff
@@ -64,5 +63,22 @@ Status FindPathValidator::validateWhere(WhereClause* where) {
   return Status::OK();
 }
 
+Status FindPathValidator::validateYield(YieldClause* yield) {
+  if (yield == nullptr) {
+    return Status::SemanticError("Missing yield clause.");
+  }
+  if (yield->columns().size() != 1) {
+    return Status::SemanticError("Only support yield path");
```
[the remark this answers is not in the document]
nevermore3: lower case is also ok

CPWstatic (approved):

Aiee (approved):

CPWstatic (approved):

PR #2958, opened 2021-09-28
optimize index select

MMyheart:


MMyheart:
> Thanks for your contribution @MMyheart , Could you use `clang-format-10` to format this PR?

Done

src/graph/optimizer/OptimizerUtils.cpp
```diff
@@ -850,10 +850,30 @@ void OptimizerUtils::eraseInvalidIndexItems(
   }
 }
 
+bool OptimizerUtils::isBestIndex(const nebula::meta::cpp2::IndexItem& indexItem,
+                                 const std::vector<std::string>& returnCols) {
+  auto indexFields = indexItem.get_fields();
```
yixinglu, line 855: `const auto&`

yixinglu (commented):
According to my understanding, are you optimizing the following case?
```
LOOKUP ON player WHERE player.name=="Tim" AND player.age==40 YIELD player.age
```
Although both `player.name` and `player.age` have their own indexes, the optimizer should choose to use `player.age` as much as possible based on the returned columns.

Make sense. Very excellent!

src/graph/optimizer/OptimizerUtils.cpp
MMyheart: In this scenario, using `player.name` or `player.age` has the same result that need to scan with data in the storaged. 
So I don't think there is much difference between the two choices though the number of lines scanned may be vary.

src/graph/optimizer/OptimizerUtils.cpp
```diff
@@ -850,10 +850,30 @@ void OptimizerUtils::eraseInvalidIndexItems(
   }
 }
 
+bool OptimizerUtils::isBestIndex(const nebula::meta::cpp2::IndexItem& indexItem,
```
jievince, line 853: Is it intended to reduce the overhead of lookup the data?

MMyheart: yes, it can avoid `IndexEdgeNode` or `IndexTagNode` of execution plan nodes in the storaged, thus reducing the overhead of lookup the data

yixinglu (approved):

jievince (approved):

czpmango:
It would be more reasonable to put the logic of selecting a better index based on `returnCol` into the `OptimizerUtils::selectIndex` to participate in the scoring mechanism(Like give a higher score), rather than adding another pass after the index  scoring, which could lead to a conflict between the two index selection logic.

czpmango:
This idea proposed by PR is worth considering. However, due to the complexity of statements (such as complex pattern index selection) and some implementation reasons (such as the possible problems caused by string not returning to the table), we cannot comprehensively evaluate its impact on performance at present. 
More scenarios may need to be tested to avoid potential performance degradation. So this PR is going to hold for now.

src/graph/optimizer/OptimizerUtils.cpp
```diff
@@ -873,6 +893,17 @@ bool OptimizerUtils::findOptimalIndex(const Expression* condition,
 
   std::sort(results.begin(), results.end());
 
+  if (!returnCols.empty()) {
+    // find best index which has all of returnCol
+    for (auto resultIter = results.rbegin(); resultIter < results.rend(); resultIter++) {
+      if (isBestIndex(*resultIter->index, returnCols)) {
+        std::vector<IndexResult> bestIndexResult{*resultIter};
+        results.swap(bestIndexResult);
```
czpmango, line 901: Make sure the scores are the same here.

yixinglu:
@MMyheart  Could you try to solve the conflicts recently?

xtcyclist:
This PR has been idle for too long without a conclusion. Close it for now.

PR #2964, opened 2021-09-28
If an edge prepareLocal(), it should not go into resume process. 

liuyu85cn:
prepareLocal() failed will also lock edge in transaction manager, which is not necessary. 

(because there isn't any prime)


critical27 (approved):
LGTM

panda-sheep (approved):

critical27:
@kikimo you could try more signal.

PR #2970, opened 2021-09-29
improve storage profiling

cangfengzhs:
Use profile flag in RPC request instead of gflagsimage


critical27:
Post a screenshot here.

yixinglu (approved):
LGTM

liuyu85cn (approved):

src/clients/storage/GraphStorageClient.cpp
```diff
@@ -15,6 +15,7 @@ folly::SemiFuture<StorageRpcResponse<cpp2::GetNeighborsResponse>> GraphStorageCl
     GraphSpaceID space,
     SessionID session,
     ExecutionPlanID plan,
+    bool profile,
```
SuperYoko, line 18: Maybe use some flag to show it's a boolean like enable/is?

Issue #2873, opened 2021-09-15
alter schema change property to a different type (string->int)


Sophie-Xie:
Prepare the space

```
(root@nebula) [(none)]> create space test1(vid_type=fixed_string(30))
Execution succeeded (time spent 39512/41812 us)

Tue, 14 Sep 2021 10:27:53 CST

(root@nebula) [(none)]> use test1
[ERROR (-1005)]: SpaceNotFound:

Tue, 14 Sep 2021 10:27:56 CST

(root@nebula) [(none)]> use test1
Execution succeeded (time spent 48813/51626 us)

Tue, 14 Sep 2021 10:27:58 CST
```

```
(root@nebula) [test1]> create tag test(col1 string , col2 string);
Execution succeeded (time spent 83144/85189 us)

Tue, 14 Sep 2021 10:38:20 CST

(root@nebula) [test1]> insert vertex test(col1,col2) values "11":("aa","bb");
Execution succeeded (time spent 6843/9300 us)

Tue, 14 Sep 2021 10:38:42 CST

(root@nebula) [test1]> fetch prop on test "11"
+--------------------------------------+
| vertices_                            |
+--------------------------------------+
| ("11" :test{col1: "aa", col2: "bb"}) |
+--------------------------------------+
Got 1 rows (time spent 11830/14468 us)

Tue, 14 Sep 2021 10:38:58 CST

(root@nebula) [test1]> alter tag test change (col2 int)
Execution succeeded (time spent 40178/44064 us)

Tue, 14 Sep 2021 10:41:12 CST

(root@nebula) [test1]> fetch prop on test "11"
+--------------------------------------+
| vertices_                            |
+--------------------------------------+
| ("11" :test{col1: "aa", col2: "bb"}) |
+--------------------------------------+
Got 1 rows (time spent 9940/13559 us)

Tue, 14 Sep 2021 10:41:18 CST

(root@nebula) [test1]> insert vertex test(col1,col2) values "22":("aa",99)
Execution succeeded (time spent 15569/18790 us)

Tue, 14 Sep 2021 10:42:04 CST

(root@nebula) [test1]> create tag index test_index on test(col2);
Execution succeeded (time spent 68325/71791 us)

Tue, 14 Sep 2021 10:42:08 CST

(root@nebula) [test1]> rebuild tag index test_index
+------------+
| New Job Id |
+------------+
| 30         |
+------------+
Got 1 rows (time spent 61606/66416 us)

[...]

PR #3001, opened 2021-10-01
Check the property value is DEFAULT | NULLABLE when ALTER TAG|EDGE

bright-starry-sky:
for issue #2873 
TODO : 

- Set default values during rebuild index
- Add test cases


critical27 (approved):

critical27:
See #2873

PR #3003, opened 2021-10-05
split onElected and onLeaderReady

liuyu85cn:
previously transaction register a callback function to Raft,

called at onElected(), to do a partition scan, to check if there is a (double) prime key.

but some raft internal status is not set until leader send its first log. 

which will let the transaction manager scan failed. 

so split the origin onElected to two part(1st still named onElected(), and 2nd named onLeaderReady()). 

and the scan will be called at onLeaderReady.


bright-starry-sky (approved):
Well done.

critical27 (approved):

PR #3007, opened 2021-10-07
Track memory consumption when finishing graph operator execution

yixinglu:
It’s so sad to encounter the following problems：

1. If I try to collect every value size, and want to check every query’s memory limit each time, it still blocks every thread even if I use ThreadLocalPtr in QueryConext since it will access all threads counters by `accessAllThreads`. Unless I'm not checking memory limit when counting the size. The main reason I didn't do this was that it affected all the places where value and expression are used, but it does not solve the problem very well. Maybe the method to count a single value size is not a good idea.
2. I could not find a workable way to count the Value size in query level rather than process level in serialization and deserialization stage since I couldn’t pass the QueryContext to fbthrift to complete that.

At present, I only update the memory usage in every graph executor finish function. Because computing each Value size is a so heavy operation, especially DataSet. I try to collect each value size in every graph executor and then i give up it since it's too noisy for executors to update the memory consumption every time. I need to rethink how to do that by a mini-batch(or dataset) way.

This is not a satisfactory solution, I will continue to optimize it later and do not rule out the introduction of memory pool.

Performance effects of this PR for very big size dataset need to be tested.


yixinglu:
> We could separate this feature into smaller tasks:
> 
> 1. Process level stats could use thread local counters. We only need to add each thread's counter to get an approximate consumption when we receive a request  (no more atomic variable).
> 2. Query level stats just check the consumption in `QueryContext`.
> 
> But all tasks above need a fundamental component: custom new/delete for `Value`. We could check if this is feasible later.
> 
> PS: `accessAllThreads` will only block creating/removing `ThreadLocal`? It won't block all threads if I remember correctly.

for task 1:

We have checked memory watermark periodically, does it really need to do this thing in process level?

for task 2:

here the key point is when to check the memory consumption?

I have considered to use operator new/delete for Value to count the size of it. But it's still hard to count the query level's stats. So i don't do it in this PR.

Sophie-Xie:
fixed in https://github.com/vesoft-inc/nebula/pull/3530

PR #3026, opened 2021-10-09
Pass graph profile param into storage and cleanup graph storage client interfaces

yixinglu:
as title.


bright-starry-sky:
Nice job , That's what I've always wanted to do. 
In addition, I feel that our thrift code should be repackaged to solve the problem that once we modify thrift files, a small number of files will be modified. 
But it's a big project and it's not going to be dealt with this PR.
LGTM.

PR #2903, opened 2021-09-18
[Raft] rollback related optimize

critical27:
1. Apply #2483 to v2.x
2. Some rollback is useless now
3. Add a rpc backoff when timeout, we should not retry often to make service worse. Accordingly, optimize not to invalid leader when storage client timeout.

I am still doing some test, don't merge to 2.6.0 plz.

PR #3031, opened 2021-10-11
separate thenValue/thenError when rpc fail to handle timeout

critical27:
Separate from #2903, we could recognize time out from exception type:
1. For storage client, don't call `invalidLeader`, it will make further request fail again.
2. For raft client, we should add some back-off time for retry, this will be handled in another PR.


Shylock-Hg:
How about meta client?

critical27:
> How about meta client?

Meta client already has a retry mechanism, and all rpc in meta should be small enough, there should not be any timeout. I'll add it if necessary. WDYT?

Shylock-Hg:
> > How about meta client?
> 
> Meta client already has a retry mechanism, and all rpc in meta should be small enough, there should not be any timeout. I'll add it if necessary. WDYT?

Got.

src/kvstore/raftex/Host.cpp
```diff
@@ -358,6 +344,43 @@ void Host::appendLogsInternal(folly::EventBase* eb, std::shared_ptr<cpp2::Append
...
+                     }
+                     self->setResponse(r);
+                     self->lastLogIdSent_ = self->logIdToSend_ - 1;
+                   }
+                   // a new raft log or heartbeat will trigger another appendLogs in Host
+                   self->noMoreRequestCV_.notify_all();
+                   return;
+                 })
+      .thenError(folly::tag_t<std::exception>{}, [self = shared_from_this()](std::exception&& ex) {
+        VLOG(2) << self->idStr_ << ex.what();
```
darionyaphet, line 373: The error message should use LOG(ERROR)?

src/kvstore/raftex/Host.cpp
```diff
@@ -358,6 +344,43 @@ void Host::appendLogsInternal(folly::EventBase* eb, std::shared_ptr<cpp2::Append
             return;
           }
         }
+      })
+      .thenError(folly::tag_t<TransportException>{},
+                 [self = shared_from_this(), req](TransportException&& ex) {
+                   VLOG(2) << self->idStr_ << ex.what();
```
darionyaphet, line 350: ditto

panda-sheep (approved):
Good job！👍

src/clients/storage/StorageClientBase-inl.h
```diff
@@ -138,49 +139,64 @@ folly::SemiFuture<StorageRpcResponse<Response>> StorageClientBase<ClientType>::c
...
-                  auto* leader = code.get_leader();
-                  if (isValidHostPtr(leader)) {
-                    updateLeader(spaceId, code.get_part_id(), *leader);
-                  } else {
-                    invalidLeader(spaceId, code.get_part_id());
-                  }
-                } else if (code.get_code() == nebula::cpp2::ErrorCode::E_PART_NOT_FOUND ||
-                           code.get_code() == nebula::cpp2::ErrorCode::E_SPACE_NOT_FOUND) {
-                  invalidLeader(spaceId, code.get_part_id());
+          .thenValue([this, context, host, spaceId, start](Response&& resp) {
```
darionyaphet, line 142: the difference between `then` and `thenValue` is separate error handling?

critical27: In general, then = thenTry = thenValue + thenError

liuyu85cn (approved):
LGTM

darionyaphet (approved):

PR #3041, opened 2021-10-12
Expression test modify

zhaohaifei:
Background：
Since the original repos are separate, the nebula/src/common/expression/test module is written based on the original repo：nebula-common. The parser module in nebual-graph cannot be called directly. Now that the repos are merged, we want to directly call the parser module to do the test.

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,0 +1,14 @@
+# Copyright (c) 2021 vesoft inc. All rights reserved.
+#
+# This source code is licensed under Apache 2.0 License,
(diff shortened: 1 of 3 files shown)


zhaohaifei:
> clearly

Not to improve, but to verify the correctness of the original expression。

zhaohaifei:
> Where to use the graph parser in the following tests? I don't think this changes improve these tests clearly.

Not to improve, but to verify the correctness of the original expression。

PR #3043, opened 2021-10-12
Geo spatial: 3. geography schema, data, index and optimization

jievince:
Close https://github.com/vesoft-inc/nebula/issues/2595

TODO:
- Support to specify s2 region covering options when create index on geo column:
`CREATE TAG INDEX geo_index on shape(geo) s2_min_level = 0, s2_max_level=30, s2_max_cells=8`
- Support joint index for geography column with other columns.
The joint index could only have one geography field and it must be at the end.
(For the time reason, we only support create a single index on a geography column)
- Support geo predicate index optimization for `MATCH`.  I have only done it for `LOOKUP` currently.
This could be done after the index selection and index optimization are done.
(For `MATCH`, geo predicate index will not apply to its where clause currently, it will degenerate to `IndexFullScan` + `Filter`;)
- Try to find the best default value for s2 region cover options.
- Improve `GeoIndex::coveredBy`

### DIFF
--- a/tests/tck/features/geo/GeoBase.feature
+++ b/tests/tck/features/geo/GeoBase.feature
@@ -67,7 +67,6 @@ Feature: Geo base
     Then the result should be, in any order:
       | Tag          | Create Tag                                                                                  |
       | "only_point" | 'CREATE TAG `only_point` (\n `geo` geography(point) NULL\n) ttl_duration = 0, ttl_col = ""' |
-    Then drop the used space
 
   Scenario: test geo CURD
     # Any geo shape(point/linestring/polygon) is allowed to insert to the column geography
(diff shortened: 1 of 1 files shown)


ZOUG:
Are multipart geometrical datatypes (MultiPoint, MultiLineString, MultiPolygon) supported at the moment? If not, are there any plans for them?

jievince:
> Are multipart geometrical datatypes (MultiPoint, MultiLineString, MultiPolygon) supported at the moment? If not, are there any plans for them?
Hi @ZOUG , Sorry for the late reply. 
Which scenarios do you want to use these Multi-shape? If there is a real need, we will consider implementing them.
By the way, 
Geo involves a lot of things. At present, we have only implemented part of the core feature.  If everybody needs other things, such as more complex shapes, or wants to import geo objects from GeoJson, etc. You are welcome to open an issue to put forward your requirements.

src/graph/optimizer/rule/OptimizeEdgeIndexScanByFilterRule.cpp
```diff
@@ -103,13 +110,43 @@ StatusOr<TransformResult> OptimizeEdgeIndexScanByFilterRule::transform(
...
+  if (condition->kind() == ExprKind::kLogicalAnd) {
+    for (auto& operand : static_cast<const LogicalExpression*>(condition)->operands()) {
+      if (operand->kind() == ExprKind::kRelIn) {
+        auto inExpr = static_cast<RelationalExpression*>(operand);
+        // Do not apply this rule if the IN expr has a valid index or it has only 1 element in the
+        // list
+        if (static_cast<ListExpression*>(inExpr->right())->size() > 1) {
+          return TransformResult::noTransform();
+        } else {
+          transformedExpr = graph::ExpressionUtils::rewriteInExpr(condition);
```
czpmango, line 133: Line 133 is a bug.

PR #3056, opened 2021-10-13
disable fetch & go missing yield clause

nevermore3:
1、 Must add YIELD clause in the FETCH vertex、FETCH edge & GO sentence
2、Remove the default output columns 【src, dst, rank】in fetch edge
3、Remove the default output column 【VertexID】in fetch vertex

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -48,7 +48,7 @@ include(NebulaCustomTargets)
 
 add_custom_target(
     install-all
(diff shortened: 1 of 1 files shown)


nevermore3:
subjob of https://github.com/vesoft-inc/nebula/issues/3135

Shylock-Hg:
Good job! Agree with removing the default return column. But shall we give default result without `YIELD` clause?

PR #3073, opened 2021-10-14
Support Disk Balance

darionyaphet:
issue: (vesoft-inc/nebula-ent#604)


Sophie-Xie:
fixed in https://github.com/vesoft-inc/nebula/pull/3394

wenhaocs:
Was trying to understand the logic and the usage of "path". After reading for a while, seems the logic looks good. Could you add a detailed description about how it works? Sorry if it's not ready for review yet.

PR #3145, opened 2021-10-19
[Enhance]Refining balancer remove some duplicate variables

darionyaphet:


Sophie-Xie:
fix in  https://github.com/vesoft-inc/nebula/pull/3394

Issue #3018, opened 2021-10-09
raft block on accquiring lock


kikimo:
**Please check the FAQ documentation before raising an issue**

raft sometimes block on accquiring lock when exiting, and hand there forever, only `kill -9` can stop the instance.

**Describe the bug (__must be provided__)**

**Your Environments (__must be provided__)**

* OS: `uname -a`
* Compliler: `g++ --version` or `clang++ --version`
* CPU: `lscpu`
* Commit id (e.g. `a3ffc7d8`)

**How To Reproduce(__must be provided__)**

1. Step 1
2. Step 2
3. Step 3

**Expected behavior**

**Additional context**

Provide logs and configs, or any other context to trace the problem.

Issue #3147, opened 2021-10-19
Raft deadlock


kikimo:
**Describe the bug (__must be provided__)**

Raft deadlock after stress test.

**Your Environments (__must be provided__)**

* OS: `uname -a`
* Compliler: `g++ --version` or `clang++ --version`
* CPU: `lscpu`
* Commit id 650776b8d54b6b1045866e83c9f2f85087623e1b (v2.6.0)

**How To Reproduce(__must be provided__)**

1. 3store + 1meta + 1graph
2. keep triggering leader change and start 1024 concurrent clients sending add edge request
3. run for about two minutes

stop the stresser, we see one storage instance offline:

```txt
(root@nebula) [(none)]> show hosts;
+-----------------+-------+-----------+--------------+----------------------+------------------------+
| Host            | Port  | Status    | Leader count | Leader distribution  | Partition distribution |
+-----------------+-------+-----------+--------------+----------------------+------------------------+
| "192.168.15.11" | 59007 | "ONLINE"  | 1            | "ttos_3p3r:1"        | "ttos_3p3r:1"          |
+-----------------+-------+-----------+--------------+----------------------+------------------------+
| "192.168.15.11" | 37549 | "ONLINE"  | 0            | "No valid partition" | "ttos_3p3r:1"          |
+-----------------+-------+-----------+--------------+----------------------+------------------------+
| "192.168.15.11" | 56039 | "OFFLINE" | 0            | "No valid partition" | "ttos_3p3r:1"          |
+-----------------+-------+-----------+--------------+----------------------+------------------------+
| "Total"         |       |           | 1            | "ttos_3p3r:1"        | "ttos_3p3r:3"          |
+-----------------+-------+-----------+--------------+----------------------+------------------------+
Got 4 rows (time spent 1249/11228 us)

Tue, 19 Oct 2021 16:23:00 CST
```

pstack shows that raft of that instance is blocked on `raftLock_`:

```txt
Thread 61 (Thread 0x7f4d317ff700 (LWP 3413758) "IOThreadPool11"):
#0  0x00007f4d531c354d in __lll_lock_wait () from /lib64/libpthread.so.0
#1  0x00007f4d531bee9b in _L_lock_883 () from /lib64/libpthread.so.0
#2  0x00007f4d531bed68 in pthread_mutex_lock () from /lib64/libpthread.so.0
[...]

PR #3172, opened 2021-10-21
[Raft] make me crazy

critical27:
Origins from and rely on #3031, some code may be conflicted as well, wait that merge first.

Change notes:
1. Cherry-pick #2483 and apply to 2.0
2. Since #2483, we could remove the weird code in Host (`magic code`), for example:
```
    if (logId <= lastLogIdSent_) {
      LOG(INFO) << idStr_ << "The log " << logId << " has been sended"
                << ", lastLogIdSent " << lastLogIdSent_;
      cpp2::AppendLogResponse r;
      r.set_error_code(cpp2::ErrorCode::SUCCEEDED);
      return r;
    }
```
and many
```
              } else if (self->lastLogIdSent_ == resp.get_last_log_id()) {
                VLOG(2) << self->idStr_ << "We send nothing in the last request"
                        << ", so we don't send the same logs again";
                self->lastLogIdSent_ = resp.get_last_log_id();
                self->lastLogTermSent_ = resp.get_last_log_term();
                self->followerCommittedLogId_ = resp.get_committed_log_id();
                cpp2::AppendLogResponse r;
                r.set_error_code(cpp2::ErrorCode::SUCCEEDED);
                self->setResponse(r);
```
3. We can't send the wals after snapshot, because if all the wals after snapshot has been sent, and we reboot during the snapshot is sending, the partition could be elected as leader (only check lastLogId). Just delete these logic in the code, follower only can append more logs after snapshot is received.
4. Fix a partition is reset way more than expected times during waiting for snapshot.
5. Many logic in `Host` has been unified. I will describe the invariant later.

Todo: 
1. Follow-up of #3031, add a back-off time when leader send logs timeout, to avoid make thing worse.
2. I tried to use different eventbase for each `Host` when appendLogs, but it seems failed (crashed a lot), still working on it.


src/kvstore/raftex/RaftPart.cpp
```diff
@@ -1470,64 +1429,101 @@ void RaftPart::processAppendLogRequest(const cpp2::AppendLogRequest& req,
...
+                << req.get_last_log_id_sent() << ". So ask leader to send logs from committedLogId "
+                << committedLogId_;
+      TermID committedLogTerm = wal_->getLogTerm(committedLogId_);
+      if (committedLogTerm > 0) {
+        resp.set_last_log_id(committedLogId_);
+        resp.set_last_log_term(committedLogTerm);
+      }
+      resp.set_error_code(cpp2::ErrorCode::E_LOG_GAP);
+      return;
+    }
```
kikimo, line 1456: emm...reseting last_log_id directly to committedLogId_ is kind of..brutal, but it's right and work.
A possible optimization when mismatch happends:

From follower side:

1. when Term(leader) < Term(follower), set resp.last_log_id to the index of first occurance of Term(follower) in follower's log entries.
2. when Term(leader) > Term(follower), just decrement last_log_id

apply the similar process to update follower's `nextIndex`(term from raft thesis, think it's similar to our last_log_id) when leader receive a append entries response. 

The above algorithm will make raft quickly find the match log without retransfering entreis already existed in follower's log, but this would further complex raft algorithm(indeed we can still further optimize this algorithm by using a updated version of binary search), so I'm not sure whether it's worth doing so.

critical27: > reseting last_log_id directly to committedLogId_ is kind of..brutal, but it's right and work.

Totally agreed. I will think more about it. It need some complex process to make the `last_log_id` right, as for your suggestion: 
> 1. when Term(leader) < Term(follower), set resp.last_log_id to the index of first occurence of Term(follower) in follower's log entries. 
But once leader has receive the `resp.last_log_id` from follower, maybe the leader can't do anything, because the leader may lack some log in our implementation.
> when Term(leader) > Term(follower), just decrement last_log_id
This is more like the process in origin paper, the `decrement` will last for quite a long time to detect `nextIndex`.

critical27: We could do it for further optimization, leave it alone in this PR.

critical27:
Close #3018, #3019, #3147

kikimo:
LGTM

kikimo (approved):
LGTM

src/kvstore/raftex/RaftPart.cpp
```diff
@@ -973,21 +976,6 @@ bool RaftPart::prepareElectionRequest(cpp2::AskForVoteRequest& req,
...
-  if (UNLIKELY(status_ == Status::STARTING)) {
-    VLOG(2) << idStr_ << "The partition is still starting";
-    return false;
-  }
-
-  if (UNLIKELY(status_ == Status::WAITING_SNAPSHOT)) {
-    VLOG(2) << idStr_ << "The partition is still waiting snapshot";
-    return false;
-  }
-
```
darionyaphet, line 990: Why not check?

darionyaphet (approved):

PR #3191, opened 2021-10-22
return __NULL__ when try to getProp a non-exist vid

liuyu85cn:
```sql
create space ttt(partition_num=1, replica_factor=1, vid_type = FIXED_STRING(30));
CREATE TAG t1(a string, b int);
INSERT VERTEX t1(a, b) VALUE "player100":("Hello", 100);
fetch prop on t1 "player1000";
```

previously, if we try to fetch a non-exist vid's props. we got an empty output.
```
(root@nebula) [ttt]> fetch prop on t1 "player888";
+-----------+
| vertices_ |
+-----------+
+-----------+
Got 1 rows (time spent 2207/2657 us)
```

now as discussed, return a row has a value NULL.
```
(root@nebula) [ttt]> fetch prop on t1 "player888";
+-----------+
| vertices_ |
+-----------+
| __NULL__  |
+-----------+
Got 1 rows (time spent 2207/2657 us)
```

### DIFF
--- a/src/storage/exec/GetPropNode.h
+++ b/src/storage/exec/GetPropNode.h
@@ -26,21 +26,16 @@ class GetTagPropNode : public QueryNode<VertexID> {
   }
 
   nebula::cpp2::ErrorCode doExecute(PartitionID partId, const VertexID& vId) override {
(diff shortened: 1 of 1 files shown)


liuyu85cn:
Talked  to nevermore3 offline, this PR won't merge. nevermore3 will create a new PR, include the graph code, based on this

PR #3213, opened 2021-10-26
Fix mixed usage of default and optional value definition in thrift

Aiee:
#### What type of PR is this?
- [x] bug
- [ ] feature

This PR changes the values defined in the thrift file to be either optional or have an optional value.

The mixed usage of default and optional values in thrift files produces incorrect code.
For example, in the Golang client, the generated code cannot be compiled, and manual modification to the auto-generated code is required, which is a pain.

### DIFF
--- a/src/clients/meta/MetaClient.cpp
+++ b/src/clients/meta/MetaClient.cpp
@@ -345,7 +345,7 @@ bool MetaClient::loadSchemas(GraphSpaceID spaceId,
                                           const cpp2::ColumnDef& col) {
     bool hasDef = col.default_value_ref().has_value();
     auto& colType = col.get_type();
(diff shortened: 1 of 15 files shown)


jievince:
> #### What type of PR is this?
> * [x]  bug
> * [ ]  feature
> 
> #### What this PR does / why we need it?
> This PR changes the values defined in the thrift file to be either optional or have an optional value.
> 
> The mixed usage of default and optional values in thrift files produces incorrect code. For example, in the Golang client, the generated code cannot be compiled, and manual modification to the auto-generated code is required, which is a pain.

Is this PR still reviewable?

Aiee:
Too many conflicts to solve. Also, the approach should be reconsidered so that current test cases can pass. Closing.

PR #3226, opened 2021-10-28
Delete the useless target `install-all`

yixinglu:
This target will result in executable error when running `sudo make install-all` and could be replaced with `make install`.

image

introduced in #3195


jievince (approved):

jievince:
We may need the targe `graphd`, `metad` and `storaged.

PR #3227, opened 2021-10-28
Improve test frame

jackwener:
#### What type of PR is this?
- [x] enhancement

related issue: None

There are so many redundant code in test of `src/common/function/test/FunctionManagerTest.cpp`.

To simplify the code and make it easy to add test case.                                                             `

### DIFF
--- a/src/common/function/test/FunctionManagerTest.cpp
+++ b/src/common/function/test/FunctionManagerTest.cpp
@@ -6,6 +6,9 @@
 
 #include <gtest/gtest.h>
 
+#include <string>
+#include <vector>
+
 #include "common/base/Base.h"
 #include "common/datatypes/DataSet.h"
 #include "common/datatypes/Edge.h"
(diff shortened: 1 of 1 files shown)


src/common/function/test/FunctionManagerTest.cpp
```diff
@@ -818,657 +821,175 @@ TEST_F(FunctionManagerTest, time) {
...
-    EXPECT_EQ(Value::Type::LIST, result.value());
+    struct TestReurnType {
+      std::string func_name;          // function being tested
+      std::vector<Value::Type> args;  // parameters for test
+      Value::Type returnType;         // the correct return type expected
+    };
+
+    // add test case
+    std::vector<TestReurnType> test_cases = {
+        TestReurnType{"abs", {Value::Type::INT}, Value::Type::INT},
```
CPWstatic, line 832: It would be much more helpful if TestReturnType run the test directly. The `TestReturnType` could be a macro or util function. And we need to easily know which one case failed, that means something like line numbers.

jackwener: It can useful that adding info about func for locating

src/common/function/test/FunctionManagerTest.cpp
```diff
@@ -6,13 +6,17 @@
 
 #include <gtest/gtest.h>
 
+#include <string>
```
nevermore3, line 9: redundant header file

jackwener: fixed it

jievince:
Please resolve the conflicts

PR #3229, opened 2021-10-29
Push nightly at begin of day.

Shylock-Hg:
#### What type of PR is this?
- [ ] bug
- [x] feature

close #xxx   
（If it is requirement, issue(s) number must be listed.）

The nebula-cpp will download the nightly package `$(date).deb/rpm` in ci pipeline，but the nightly build at 18:00 , so it will failed for the not found package.
  
#### Checklist：
- [ ] Documentation affected （If need to modify document, please label it.)
- [ ] Incompatible （If it is incompatile, please describle it and label it.）
- [ ] Need to cherry pick （If need to cherry pick to some branchs, please label the destination version(s).)
- [ ] Performance regression: Consumes more CPU
- [ ] Performance regression: Consumes more Memory

#### Release notes：

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -2,7 +2,7 @@ name: nightly
 
 on:
   schedule:
-    - cron: '0 18 * * *'
+    - cron: '30 0 * * *'
 
 defaults:
   run:


Shylock-Hg:
> We want to build nebula nightly package at night, because we can reuse our idle runners. Otherwise this nightly jobs will affect our daily development. You should notice the schedule time is UTC. Why not use nightly docker image to test the client?

The cpp client test run in docker already. So can't run docker again.

Shylock-Hg:
> We want to build nebula nightly package at night, because we can reuse our idle runners. Otherwise this nightly jobs will affect our daily development. You should notice the schedule time is UTC. Why not use nightly docker image to test the client?

Or we could provide the nightly package without datetime in name.

nevermore3 (approved):

Issue #3106, opened 2021-10-15
View sample data for tags & edges 


Sophie-Xie:
**Describe the background of the requirement**
When customers get a new Nebula Graph and don't know the data details, customers always don't know how to start using our products(Nebula Graph、explorer、studio)

**Describe the detail of the requirement**
For relational database(mysql):
When we get a new database, we can know it with below:
1.show databases;
2.show tables;
3.desc <table_name> or show create <table_name>;
4.select * from <table_name> limit 10;

For nebula graph:
when we get a new database:
1.show spaces;
2.show tags & show edges;
3.show create {TAG <tag_name> | EDGE <edge_name>};
No match (v:<tag_name>) return v limit 10; (need to create index)

Can we supply a way to view sample data for tags & edges without index? then at least we can get a "vid" for query or any other operations.

**Describe the scenarios of the requirement**
1.First use of our products，don't know data details, want to understand the data structure quickly.
2.Don't know data details, want to test expand through explorer.(Need VID)

**Describe the values of the requirement**
General: nebula graph usability
Values for customer: Facilitate customers to better use our products(Nebula Graph,explorer,studio)，make it easier for customers to understand the data structure
Importance of the business project:it's common request

from @Jerry319609l

PR #3262, opened 2021-11-03
Feature/scan multiple parts

Shylock-Hg:
#### What type of PR is this?
- [ ] bug
- [x] feature

Subjob of #3106 
（If it is requirement, issue(s) number must be listed.）

Extend the Scan storage interface to support scan multiple partitions. And ScanVertex will return kVid as first column to keep same with GetProp.
  
#### Additional context:
@jievince @Nicole00 The scan interface need update.

#### Checklist：
- [ ] Documentation affected （If need to modify document, please label it.)
- [ ] Incompatible （If it is incompatile, please describle it and label it.）
- [ ] Need to cherry pick （If need to cherry pick to some branchs, please label the destination version(s).)
- [ ] Performance regression: Consumes more CPU
- [ ] Performance regression: Consumes more Memory

#### Release notes：


src/storage/exec/ScanNode.h
```diff
@@ -0,0 +1,208 @@
...
+                              std::unordered_map<PartitionID, cpp2::ScanCursor>* cursors,
+                              nebula::DataSet* resultDataSet)
+      : context_(context),
+        tagNodes_(std::move(tagNodes)),
+        enableReadFollower_(enableReadFollower),
+        limit_(limit),
+        cursors_(cursors),
+        resultDataSet_(resultDataSet) {
+    name_ = "ScanVertexPropNode";
+    std::vector<TagNode*> tags;
```
critical27, line 35: I don't quite understand, would you explain more about why we need the `GetTagPropNode` here?
Since you have already scan data by rocksdb iterator (line 60), why we use get again(line 80)?

critical27 (commented):
Good job. Only one question, others LGTM.

src/interface/storage.thrift
```diff
@@ -557,24 +557,29 @@ struct LookupAndTraverseRequest {
  * End of Index section
  */
 
+struct ScanCursor {
+    3: bool                                 has_next,
+    // next start key of scan, only valid when has_next is true
+    4: optional binary                      next_cursor,
+}
+
 struct ScanVertexRequest {
```
critical27, line 566: @jievince @Nicole00 PTAL, do we need to keep the old interface? It depends on how we use them. Do we scan in  parallel(code level) or scan for each part?

src/storage/exec/ScanNode.h
Shylock-Hg: 1. line 60 just get one tag key, but we support scan multiple tags.
2. behave same with GetProp interface, so I implment ScanVertexPropNode by wrap the GetTagPropNode.

Shylock-Hg: For second, the graph will process the result of ScanVertex and GetProp in same code.

src/interface/storage.thrift
Nicole00: I suggest to keep the old interface, StorageClient provides interface to scan just one part. And Connectors depends on this kind of interface.

Shylock-Hg: Youd can't just scan one partition, just wrap the origin partition ID to list

src/storage/exec/ScanNode.h
critical27: I got your point, some of them looks weird:
1. We don't support multiple tags/edges for now, see `VertexProp` and `EdgeProp` in the `ScanVertexRequest` and `ScanEdgeRequest`. We can't support them unless we separate data into different DataSet according to tag/edgeType.
2. The return format could be unified, but I don't suggest to implement like now, because the extra `get` in rocksdb is really a waste.

src/interface/storage.thrift
```diff
@@ -557,24 +557,29 @@ struct LookupAndTraverseRequest {
  * End of Index section
  */
 
+struct ScanCursor {
+    3: bool                                 has_next,
```
critical27, line 561: The `has_next` seems useless, just judge by `next_cursor` is enough?

src/interface/storage.thrift
critical27: > Youd could just scan one partition, just wrap the origin partition ID to list

The hard part is the cursor, some part may have scanned all data, while others still has more. I guess this need some extra code to do it.

How about we only change the code from
```
2: common.PartitionID part_id,
3: optional binary cursor,
```
to
```
2: list<common.PartitionID> part_id,
3: optional list<binary> cursor,
```

In that way, the java/python client could still only support the old interface `scan(partId)` to users, but only convert to a `list<partId>` when sending RPC. I think this change is the minimum one?

jievince: > @jievince @Nicole00 PTAL, do we need to keep the old interface? It depends on how you use them. Do we scan in parallel(code level) or scan for each part one by one?

I use the old scan interface to scan in parallel: I start multiple threads, each thread to scan a part

src/interface/storage.thrift
Shylock-Hg: Yes, I just keep the origin fields.

src/interface/storage.thrift
Shylock-Hg: @critical27 When partition scan to end, its cursor will set to null.

src/storage/exec/ScanNode.h
Shylock-Hg: 1. Sorry it's my fault, #3106 require scan multiple tags, I will add it later like `GetProp`.
2. Yes, I will try to optimize it.

src/interface/storage.thrift
Nicole00: @Shylock-Hg @critical27  There's one `incompatibility` problem if we don't keep the origin interfaces.  Change the partition id from int32 to list when sending RPC request will cause the incompatibility, and Nebula Server will not support **Backward Compatible**.

src/storage/exec/ScanNode.h
```diff
@@ -0,0 +1,293 @@
...
+      auto tagIdIndex = tagNodesIndex_.find(tagId);
+      if (tagIdIndex != tagNodesIndex_.end()) {
+        break;
+      }
+    }
+    if (iter->valid()) {
+      auto vIdSlice = NebulaKeyUtils::getVertexId(vIdLen, iter->key());
+      currentVertexId = vIdSlice.subpiece(0, vIdSlice.find_first_of('\0')).toString();
+    }
+    for (; iter->valid() && static_cast<int64_t>(resultDataSet_->rowSize()) < rowLimit;
```
critical27, line 78: May be just use one for loop is enough?
```
std::string prev;
for (; iter->valid() && static_cast<int64_t>(resultDataSet_->rowSize()) < rowLimit; iter->next()) {
    auto vId = NebulaKeyUtils::getVertexId(...);
    if (vId != prev && !prev.empty()) {
        collectOneRow(isIntId, vIdLen, prev);
    }
    prev = vId;
    ...
}
```

src/storage/exec/ScanNode.h
```diff
@@ -0,0 +1,293 @@
...
+      if (static_cast<int64_t>(resultDataSet_->rowSize()) >= rowLimit) {
+        break;
+      }
+      auto value = iter->val();
+      tagNodes_[tagIdIndex->second]->doExecute(key.toString(), value.toString());
+    }  // iterate key
+    if (static_cast<int64_t>(resultDataSet_->rowSize()) < rowLimit) {
+      collectOneRow(isIntId, vIdLen, currentVertexId);
+    } else {
+      for (auto& tagNode : tagNodes_) {
```
critical27, line 101: The clear here seems useless? Because `collectOneRow` has invoke the function. We need to clear the key/value in `TagNode` once we have called `collectOneRow`.

src/interface/storage.thrift
critical27: > @Shylock-Hg @critical27 There's one `incompatibility` problem if we don't keep the origin interfaces. Change the partition id from int32 to list when sending RPC request will cause the incompatibility, and Nebula Server will not support **Backward Compatible**.

I mean the client could still use the origin `user interface` (not the rpc interface, I mean sth like `scan(spaceId, partId)` in java client), but wrap into the new rpc interface using old user interface. It is ok for incompatibility in rpc layer for 1230 version. But the user interface better keep the same.

Shylock-Hg: Yes, we had agree to this after offline discussion

src/storage/exec/ScanNode.h
Shylock-Hg: Good catch! Fixed.

src/storage/exec/ScanNode.h
Shylock-Hg: Yes, we had avoid it in previous code.

src/interface/storage.thrift
critical27: 👍

critical27 (approved):
Good job, thx!

cangfengzhs:
Overall it's ok. There is a small problem. I think from the perspective of a volcano model, only the leaf nodes TagNode/EdgeNode need to access `kvstore`, so cursor should be directly assigned to TagNode/EdgeNode instead of directly accessing `kvstore` in ScanNode and then key/value is given to TagNode /EdgeNode wraps once. Of course, this has something to do with the inadequate standards of our current implementation plan, which can be modified later.

cangfengzhs (approved):

Shylock-Hg:
> Overall it's ok. There is a small problem. I think from the perspective of a volcano model, only the leaf nodes TagNode/EdgeNode need to access `kvstore`, so cursor should be directly assigned to TagNode/EdgeNode instead of directly accessing `kvstore` in ScanNode and then key/value is given to TagNode /EdgeNode wraps once. Of course, this has something to do with the inadequate standards of our current implementation plan, which can be modified later.

Yes, the ScanNode is too heavy, refactor in volcano model maybe is better.

Issue #3270, opened 2021-11-04
MATCH Attribute filtering does not support negative numbers


nevermore3:
```
(root@nebula) [nba]> match (v:player {age:30}) return v
+-------------------------------------------------------------------+
| v                                                                 |
+-------------------------------------------------------------------+
| ("DeAndre Jordan" :player{age: 30, name: "DeAndre Jordan"})       |
| ("Blake Griffin" :player{age: 30, name: "Blake Griffin"})         |
| ("Kevin Durant" :player{age: 30, name: "Kevin Durant"})           |
| ("Russell Westbrook" :player{age: 30, name: "Russell Westbrook"}) |
+-------------------------------------------------------------------+
```
Got 4 rows (time spent 17979/19717 us)

Thu, 04 Nov 2021 11:17:39 CST

```
(root@nebula) [nba]> match (v:player {age:-10}) return v
[ERROR (-1009)]: SemanticError: Props must be constant: `-(10)'



(root@nebula) [nba]> match (v:player) where v.age == -1 return v
+--------------------------------------------+
| v                                          |
+--------------------------------------------+
| ("Null1" :player{age: -1, name: __NULL__}) |
+--------------------------------------------+
```

(root@nebula) [nba]> match (v:player {age:+10}) return v
[ERROR (-1009)]: SemanticError: Props must be constant: `+(10)'

Shylock-Hg:
In fact, it's evaluable expresion instead of contant. e.g. `match (v {age: 2 + 23}) return v` is valid query too.

PR #3272, opened 2021-11-04
enhance match attribute filter

nevermore3:
#### What type of PR is this?
enhancement

#### What does this PR do?
close https://github.com/vesoft-inc/nebula/issues/3270

### DIFF
--- a/src/graph/validator/MatchValidator.cpp
+++ b/src/graph/validator/MatchValidator.cpp
@@ -180,7 +180,7 @@ Status MatchValidator::buildNodeInfo(const MatchPath *path,
     }
     Expression *filter = nullptr;
     if (props != nullptr) {
-      auto result = makeNodeSubFilter(props, "*");
+      auto result = makeNodeSubFilter(const_cast<MapExpression *>(props), "*");
       NG_RETURN_IF_ERROR(result);
       filter = result.value();
     } else if (node->labels() != nullptr && !node->labels()->labels().empty()) {
@@ -248,7 +248,7 @@ Status MatchValidator::buildEdgeInfo(const MatchPath *path,
     }
     Expression *filter = nullptr;
     if (props != nullptr) {
-      auto result = makeEdgeSubFilter(props);
+      auto result = makeEdgeSubFilter(const_cast<MapExpression *>(props));
       NG_RETURN_IF_ERROR(result);
       filter = result.value();
     }
@@ -519,7 +519,7 @@ Status MatchValidator::validateUnwind(const UnwindClause *unwindClause,
   return Status::OK();
 }
 
-StatusOr<Expression *> MatchValidator::makeEdgeSubFilter(const MapExpression *map) const {
+StatusOr<Expression *> MatchValidator::makeEdgeSubFilter(MapExpression *map) const {
   auto *pool = qctx_->objPool();
   DCHECK(map != nullptr);
   auto &items = map->items();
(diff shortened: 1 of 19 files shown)


Shylock-Hg:
Add test cases.

tests/tck/features/match/Base.feature
```diff
@@ -518,6 +518,71 @@ Feature: Basic match
...
+      | [:like "Tony Parker"->"Tim Duncan" @0 {likeness: 95}]        |
+    When executing query:
+      """
+      match (v:player)-[e:like{likeness:"99"}]->()  return e
+      """
+    Then the result should be, in any order, with relax comparison:
+      | e |
+    When executing query:
+      """
+      match (v:player{age:"24"-1})  return v
```
Shylock-Hg, line 581: This will report error in cypher.

nevermore3: done

src/parser/MatchSentence.h
```diff
@@ -163,6 +163,8 @@ class MatchNode final {
 
   const MapExpression* props() const { return props_; }
 
+  MapExpression* props() { return props_; }
+
```
czpmango, line 167: Why delete const?

src/graph/validator/MatchValidator.cpp
```diff
@@ -527,8 +525,8 @@ Status MatchValidator::validateUnwind(const UnwindClause *unwindClause,
 }
 
 StatusOr<Expression *> MatchValidator::makeSubFilter(const std::string &alias,
-                                                     const MapExpression *map,
-                                                     const std::string &label) const {
+                                                     MapExpression *map,
+                                                     const std::string &label) {
```
czpmango, line 529: Expressions related implementation changes should not change the Interface of the Validator.

nevermore3: need to rewrite the expression after fold  into the mapexpression

tests/tck/features/expression/function/TypeConversion.feature
```diff
@@ -9,72 +9,81 @@ Feature: TypeConversion Expression
...
+    When executing query:
+      """
+      YIELD [toFloat(true), toFloat(false)] AS yield_toFloat
+      """
+    Then a SemanticError should be raised at runtime: Type error `toFloat(true)'
 
   Scenario: toInteger
     When executing query:
       """
-      YIELD [toInteger(true), toInteger(false), toInteger(1), toInteger(3.14),
```
Shylock-Hg, line 60: Why change this case?

nevermore3: toInteger(false), toInteger(true) will report error in opencypher,   other modified cases too

Shylock-Hg: I think you should keep the origin case, and add new one to test the origin function.

nevermore3: i know, but the result of origin case is wrong, and the modified logic test has been covered

CPWstatic (dismissed):
LGTM. Please fix the conflicts.

yixinglu (approved):
LGTM

cangfengzhs (approved):
LGTM. But, I have a small question. Is it reasonable to expose `BAT_TYPE` to users? I haven't learned about GQL and opencypher. If they are like this, that's OK. But if they are not like this, is it necessary for us to standardize the `null type` we return to user? After all, our `Value` has many kinds of `null type`.

jievince:
> LGTM. But, I have a small question. Is it reasonable to expose `BAT_TYPE` to users? I haven't learned about GQL and opencypher. If they are like this, that's OK. But if they are not like this, is it necessary for us to standardize the `null type` we return to user? After all, our `Value` has many kinds of `null type`.
Neither of them return `BAD_TYPE`, so the `NULL`, `BAD_TYPE` should be reconsidered

czpmango:
> LGTM. But, I have a small question. Is it reasonable to expose `BAT_TYPE` to users? I haven't learned about GQL and opencypher. If they are like this, that's OK. But if they are not like this, is it necessary for us to standardize the `null type` we return to user? After all, our `Value` has many kinds of `null type`.

All bad nulls is actually just an error with an error message. It may be more reasonable to provide a Value overload with customizable error messages, bcz current expressions’ ability to report and transmit errors is very limited.

PR #3279, opened 2021-11-05
Add cpu binding

critical27:
I found lots of context switch during other test. So I try to set cpu affinity, but the sad conclusion is that whether binding cpu or not has very little impact on performance at least for now. The only good news is that, it do help a bit to reduce the context switch.

As for graphd, the bottleneck clearly is not here. I will do other test later. We need to find the bottleneck for storage using other stress tool.


bright-starry-sky:
itc

src/common/cpu/CMakeLists.txt
```diff
@@ -0,0 +1,11 @@
+# Copyright (c) 2020 vesoft inc. All rights reserved.
```
darionyaphet, line 1: 2021 ?

cangfengzhs:
I have also encountered the bottleneck of the stress tool before, which looks like lock contention caused by the graphd session mechanism.

src/common/cpu/CMakeLists.txt
```diff
@@ -0,0 +1,11 @@
...
+#
+# This source code is licensed under Apache 2.0 License.
+
+nebula_add_library(
+    cpu_obj OBJECT
+    CpuInfo.cpp
+    CpuBinder.cpp
+)
+
+nebula_add_subdirectory(test)
```
SuperYoko, line 11: New line at EOF.

critical27:
Add later if necessary, close for now

PR #3280, opened 2021-11-05
Add key value separation feature to Nebula storage engine

wenhaocs:
This PR exposes the key-value separation support of RocksDB in Nebula. Based on the experiment with internal dataset and testing environment, we can achieve as much as 2.67x latency improvement.


wenhaocs:
rebase

PR #3281, opened 2021-11-08
Add key value separation feature to Nebula storage engine

wenhaocs:
This PR exposes the key-value separation support of RocksDB in Nebula. Based on the experiment with internal dataset and testing environment, we can reduce query latency as much as 80%.


critical27 (approved):
Well done, LGTM

panda-sheep (approved):
Good job, 👍

conf/nebula-storaged.conf.default
```diff
@@ -95,6 +95,16 @@
 # Whether or not to enable rocksdb's whole key bloom filter, disabled by default.
 --enable_rocksdb_whole_key_filtering=false
 
+############## Key-Value separation ##############
+# Whether or not to enable BlobDB (RocksDB key-value separation support)
+--rocksdb_enable_kv_separation=false
+# RocksDB key value separation threshold. Values at or above this threshold will be written to blob files during flush or compaction.
+--rocksdb_kv_separation_threshold=0
```
darionyaphet, line 102: if the threshold is set to zero, any writing will move to the blob file?

PR #3282, opened 2021-11-08
enable kv interface in 2.0

critical27:
As title. For later kv test or jepsen, @HarrisChu @kikimo, you guys could try it. Perhaps the java client need some update for jepsen.


HarrisChu:
OK for me, 2 type testings:
1. stress testing with k6, will use golang to invoke via thrift interface.
2. jepsen

kikimo:
@critical27 is'it possible to add an append api?

critical27:
> @critical27 is'it possible to add an append api?

What is the purpose? Could, but the performance will be very poor.

yixinglu:
Hi @critical27 plz solve the conflicts.

liuyu85cn (approved):
do we support getPrefix or removePrefix ? generally LGTM

pengweisong:
I think these pure kv interfaces are more suitable in separate client semantically?

SuperYoko (commented):
LGTM. but why do we need to remove this GeneralStorageClient&Server? Maybe I need to modify my rpc modification to specific interface and keep a rpc server for those request not from graph client, so that we can keep tools in different process.

critical27:
> do we support getPrefix or removePrefix ? generally LGTM

We can't do it just like a put/get, because we need to fan out to do prefix/removePrefix from all partition, think them as a `lookup`. But we could do it if really necessary.

critical27:
> I think these pure kv interfaces are more suitable in separate client semantically?

👌 , I'll move it back.

critical27:
> LGTM. but why do we need to remove this GeneralStorageClient&Server? Maybe I need to modify my rpc modification to specific interface and keep a rpc server for those requests not from graph client, so that we can keep tools in different process.

I'll keep the GeneralStorageClient. But the server do need to be removed, because we bind a service to a port separately, we have already used 4 ports. We definitely don't want to add a port more, the kv interface is mainly for test only.

critical27:
> > I think these pure kv interfaces are more suitable in separate client semantically?
> 
> Ah... The reason is we don't want to add an extra service here. (we need one more extra port to bind), so both the service and client will be removed. In 1.0, they are in the same service as well.

src/clients/storage/GraphStorageClient.cpp
```diff
@@ -623,6 +623,82 @@ StorageRpcRespFuture<cpp2::ScanVertexResponse> GraphStorageClient::scanVertex(
       });
 }
 
+folly::SemiFuture<StorageRpcResponse<cpp2::KVGetResponse>> GraphStorageClient::get(
```
yixinglu, line 626: Could the `GraphStorageClient` be renamed to `StorageClient`?

critical27: 👌

yixinglu (approved):
LGTM

panda-sheep (approved):
Good job!

PR #3285, opened 2021-11-08
Modify index bound value

cangfengzhs:
#### What type of PR is this?
- [x] bug
- [ ] feature
- [ ] enhancement

#### What does this PR do?
close issue 2745

Fix the problem of index query interval error
  
#### Checklist：

#### Release notes：


src/graph/optimizer/rule/IndexScanRule.cpp
```diff
@@ -199,46 +199,101 @@ Status IndexScanRule::appendIQCtx(const IndexItem& index, IndexQueryCtx& iqctx)
...
-      begin = OptimizerUtils::normalizeValue(col, item.value_);
+      begin = {item.value_, true};
       break;
     }
     // because only type for bool is true/false, which can not satisify [start,
     // end)
     if (col.get_type().get_type() == nebula::cpp2::PropertyType::BOOL) {
       return Status::SemanticError("Range scan for bool type is illegal");
     }
-    NG_RETURN_IF_ERROR(OptimizerUtils::boundValue(item.relOP_, item.value_, col, begin, end));
```
czpmango, line 224: Delete all boundValue related functions.

src/graph/optimizer/OptimizerUtils.cpp
```diff
@@ -949,5 +949,15 @@ void OptimizerUtils::copyIndexScanData(const nebula::graph::IndexScan* from,
...
+Status OptimizerUtils::compareAndSwapBound(std::pair<Value, bool>& a, std::pair<Value, bool>& b) {
+  if (a.first > b.first) {
+    std::swap(a, b);
+  } else if (a.first < b.first) {  // do nothing
+  } else if (a.second > b.second) {
+    std::swap(a, b);
+  }
+  return Status::OK();
+}
+
```
czpmango, line 961: This is so weird as a utility function.

tests/tck/features/optimizer/IndexScanRule.feature
```diff
@@ -20,25 +20,25 @@ Feature: Match index selection
...
       MATCH (v:player)
       WHERE
-        v.name<="Aron Baynes"
-        or v.name>"Yao Ming"
-        or v.name=="Kobe Bryant"
-        or v.age>40
+      v.name<="Aron Baynes"
+      or v.name>"Yao Ming"
+      or v.name=="Kobe Bryant"
+      or v.age>40
```
czpmango, line 41: Is this a better format?

src/graph/optimizer/rule/IndexScanRule.cpp
```diff
@@ -199,46 +199,101 @@ Status IndexScanRule::appendIQCtx(const IndexItem& index, IndexQueryCtx& iqctx)
...
+        include = true;
+        [[fallthrough]];
+      case Expression::Kind::kRelLT: {
+        if (end.first.empty()) {
+          end.first = item.value_;
+          end.second = include;
+        } else {
+          auto tmp = std::make_pair(item.value_, include);
+          OptimizerUtils::compareAndSwapBound(end, tmp);
+        }
```
czpmango, line 264: What does this code mean？

src/graph/optimizer/rule/IndexScanRule.cpp
cangfengzhs: There are some other functions that still use them, such as `mergeRangeColumnHints`. The logic of these functions is very complicated, so I didn't modify them.

czpmango: If this function is called, isn't the value it changes illegal？
If we add an interval open and close flag, we shouldn't do boundValue anymore.

tests/tck/features/optimizer/IndexScanRule.feature
```diff
@@ -20,25 +20,25 @@ Feature: Match index selection
...
-      | 6  | Project     | 5            |                                                                                                                        |
-      | 5  | Filter      | 15           |                                                                                                                        |
-      | 15 | GetVertices | 11           |                                                                                                                        |
-      | 11 | IndexScan   | 0            | {"indexCtx": {"columnHints":{"scanType":"RANGE","column":"name","beginValue":"\"Tim Duncan","endValue":"\"Yao Ming"}}} |
-      | 0  | Start       |              |                                                                                                                        |
+      | id | name        | dependencies | operator info                                                                                                              |
+      | 10 | Project     | 13           |                                                                                                                            |
```
czpmango, line 24: Add the following case and profile the `begin_value` and `end_value` in `ColumnHints`
```
profile match (v:player) where v.age>30 and v.age<=40 return v
```

tests/tck/features/optimizer/IndexScanRule.feature
cangfengzhs: This is automatically formatted by vscode. It was not in this format before, but lint would report an error, so I formatted it with vscode.

src/graph/optimizer/rule/IndexScanRule.cpp
cangfengzhs: OK, I have deleted `boundValue` just now, but the logic of `selectIndex` is very complicated, and there are a lot of code repetitions. I am more inclined to reorganize this part of the logic when you reconstruct the logic of selecting indexes.

tests/tck/features/optimizer/IndexScanRule.feature
cangfengzhs: OK

src/graph/optimizer/OptimizerUtils.cpp
cangfengzhs: I add some comment to this function.

src/graph/optimizer/rule/IndexScanRule.cpp
czpmango: Ok.

src/graph/optimizer/rule/IndexScanRule.cpp
cangfengzhs: assign min(end,tmp) to end

src/graph/optimizer/OptimizerUtils.cpp
```diff
@@ -724,7 +208,15 @@ bool mergeRangeColumnHints(const ColumnDef& field,
       }
     }
   }
-  return !(*begin >= *end);
+  bool ret = true;
+  if (begin->first > end->first) {
+    ret = false;
+  } else if (begin->first == end->first) {
+    if (begin->second == false && end->second == false) {
```
critical27, line 215: What will happen we pass interval of (x, x)? Report error?

src/graph/optimizer/OptimizerUtils.cpp
```diff
@@ -748,34 +240,33 @@ bool getIndexColumnHintInExpr(const ColumnDef& field,
...
-  } else if (newBegin == end) {
-    // end == neBegin == begin + 1
-    h.hint.set_scan_type(storage::cpp2::ScanType::PREFIX);
-    h.hint.set_begin_value(std::move(begin));
-    h.score = IndexScore::kPrefix;
+  } else if (begin.first == end.first) {
+    if (begin.second == false && end.second == false) {
+      return false;
+    } else {
+      h.hint.set_scan_type(storage::cpp2::ScanType::RANGE);
```
critical27, line 260: All PREFIX has convert to [value, value]?

src/graph/optimizer/OptimizerUtils.cpp
```diff
@@ -748,34 +240,33 @@ bool getIndexColumnHintInExpr(const ColumnDef& field,
...
+    h.hint.set_include_begin(begin.second);
+    h.hint.set_include_end(end.second);
     h.score = IndexScore::kRange;
-  } else if (newBegin == end) {
-    // end == neBegin == begin + 1
-    h.hint.set_scan_type(storage::cpp2::ScanType::PREFIX);
-    h.hint.set_begin_value(std::move(begin));
-    h.score = IndexScore::kPrefix;
+  } else if (begin.first == end.first) {
+    if (begin.second == false && end.second == false) {
```
critical27, line 257: Why we don't need to return false for case (x, x], [x, x)?
> (x, x) has return false in line 244

src/graph/optimizer/OptimizerUtils.cpp
cangfengzhs: No errors will be reported and will be executed normally. But no data will be returned.

src/graph/optimizer/OptimizerUtils.cpp
cangfengzhs: No. This is only used to calculate whether the union of multiple IndexItems is an empty set (begin>end)

src/graph/optimizer/OptimizerUtils.cpp
cangfengzhs: This is a good suggestion. I was not sure if (x,x] or [x,x) represented the empty set.

src/graph/optimizer/OptimizerUtils.cpp
critical27: 👌

critical27 (approved):

czpmango (approved):

PR #3321, opened 2021-11-16
Replace group when create space

darionyaphet:
#### What type of PR is this?
- [ ] bug
- [ ] feature
- [x] enhancement

#### Checklist：

#### Release notes：

### DIFF
--- a/src/graph/validator/AdminValidator.cpp
+++ b/src/graph/validator/AdminValidator.cpp
@@ -20,8 +20,8 @@ Status CreateSpaceValidator::validateImpl() {
   ifNotExist_ = sentence->isIfNotExist();
   auto status = Status::OK();
   spaceDesc_.set_space_name(std::move(*(sentence->spaceName())));
-  if (sentence->groupName()) {
-    spaceDesc_.set_group_name(std::move(*(sentence->groupName())));
+  if (sentence->zoneNames()) {
+    spaceDesc_.set_zone_names(sentence->zoneNames()->zoneNames());
   }
   StatusOr<std::string> retStatusOr;
   std::string result;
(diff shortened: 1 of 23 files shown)


SuperYoko:
use this to mark checkbox~
```
- [x] enhancement
```

src/graph/validator/AdminValidator.cpp
```diff
@@ -20,8 +20,8 @@ Status CreateSpaceValidator::validateImpl() {
   ifNotExist_ = sentence->isIfNotExist();
   auto status = Status::OK();
   spaceDesc_.set_space_name(std::move(*(sentence->spaceName())));
-  if (sentence->groupName()) {
-    spaceDesc_.set_group_name(std::move(*(sentence->groupName())));
+  if (sentence->zoneNames()) {
+    spaceDesc_.set_zone_names(std::move(sentence->zoneNames()->zoneNames()));
```
SuperYoko, line 24: why  `->zoneNames()->zoneNames()`? just curious.

src/meta/processors/zone/UpdateZoneProcessor.cpp
```diff
@@ -119,41 +119,52 @@ void DropHostFromZoneProcessor::process(const cpp2::DropHostFromZoneReq& req) {
...
-    spaceIter->next();
+
+    if (code != nebula::cpp2::ErrorCode::SUCCEEDED) {
+      break;
+    } else {
+      spaceIter->next();
+    }
+  }
+
+  if (code != nebula::cpp2::ErrorCode::SUCCEEDED) {
```
SuperYoko, line 164: maybe scope guard is okey to simplfy this logic?

src/graph/validator/AdminValidator.cpp
darionyaphet: It's just typed conversion and return the std::vector<std::string>

liwenhui-soul: compile error by clang here. no need std::move

darionyaphet:
rely on add host to fix describe space

src/meta/processors/zone/UpdateZoneProcessor.cpp
darionyaphet: after talk offline we think this is fine

darionyaphet:
This PR have merged into [White list](https://github.com/vesoft-inc/nebula/pull/3343) so close it

PR #3322, opened 2021-11-16
[Raft] Add prevote to make it better

### LINKED ISSUE #3108: Implement raft prevote to prevent unnecessary leader election/change.
**Is your feature request related to a problem? Please describe.**

There is no prevote in current raft implementation, when network partition happends and recovers later, a rejoin server may disrupt the  cluster by starting a new election.

**Describe the solution you'd like**

On the paper [CONSENSUS: BRIDGING THEORY AND PRACTICE 9.6](https://web.stanford.edu/~ouster/cgi-bin/papers/OngaroPhD.pdf)(fulll version of raft paper), there describe a way called pre-vote to avoid this situation.

**Describe alternatives you've considered**

**Additional context**

critical27:
### Reference
See $9.6 in https://web.stanford.edu/~ouster/cgi-bin/papers/OngaroPhD.pdf, my implementation is almost same in https://www.openlife.cc/sites/default/files/4-modifications-for-Raft-consensus.pdf

### Change notes:
* Prevote is used to prevent some unstable case when a network separated node rejoin (see the reference for details). Use `current_term + 1` to start a prevote, if I get majority votes, then start a formal election with `++current_term` (which is the same term in prevote). I believe most of the election logic is same as paper after this PR. Many scenario when we failed to get a leader will be eliminated.
* Some `wal_->rollbackToLog(lastLogId_);` is added, because if term changed during sending logs, the `lastLogId_` won't be updated. However, since we have write some logs into wal, if we don't rollback those logs, we can't write any logs anymore because the logId inconsistency in wal.

Close #3108

@kikimo and I will do some test first, but ready to review.


darionyaphet:
please fix the conflict thanks ~

src/kvstore/raftex/RaftPart.h
```diff
@@ -344,25 +344,27 @@ class RaftPart : public std::enable_shared_from_this<RaftPart> {
...
+  // Return true if I have been granted majority votes on proposedTerm, no matter isPreVote or not
   bool handleElectionResponses(const ElectionResponses& resps,
                                const std::vector<std::shared_ptr<Host>>& hosts,
-                               TermID proposedTerm);
+                               TermID proposedTerm,
+                               bool isPreVote);
 
-  // The method returns the partition's role after the election
-  Role processElectionResponses(const ElectionResponses& results,
+  // Return true if I have been granted majority votes on proposedTerm, no matter isPreVote or not
```
liuyu85cn, line 363: comments of this method is exact the same as the above one...

liuyu85cn (approved):

src/kvstore/raftex/RaftPart.cpp
```diff
@@ -1184,15 +1184,15 @@ void RaftPart::statusPolling(int64_t startTime) {
       return;
     }
   }
-  size_t delay = FLAGS_raft_heartbeat_interval_secs * 1000 / 3;
+  size_t delay = FLAGS_raft_heartbeat_interval_secs * 1000 / 3 + +folly::Random::rand32(500);
   if (needToStartElection()) {
-    if (leaderElection().get()) {
-      VLOG(2) << idStr_ << "Stop the election";
+    if (leaderElection(true).get() && leaderElection(false).get()) {
```
darionyaphet, line 1189: !leaderElection(true).get() || !leaderElection(false).get() would be better?

darionyaphet, line 1189: call `leaderElection` twice in a row 👍

darionyaphet (approved):
It is looks simpler

critical27:
Wait @kikimo to do a final test.

kikimo (approved):
LGTM , it's time to merge and I'll retest all these raft PRs later.

Issue #2725, opened 2021-06-15
Unexpected result of getProps


Shylock-Hg:
The getProps won't return row of vertex which don't contains the request tag.
e.g. in space nba
If we have 
```
'Spurs'
```
as the request vids.
And 
```
player.age
```
as the request properties.
We will get empty result even if the vid 'Spurs' is existed in space.

critical27:
We will convert to getVertices instead of getTags in 1230

Sophie-Xie:
resolve in：https://github.com/vesoft-inc/nebula/issues/2535

Issue #3123, opened 2021-10-18
support vertices without tags


Sophie-Xie:
from roadmap

PR #3328, opened 2021-11-18
add verteKey

cangfengzhs:
check vertex key in fetch

#### What type of PR is this?
- [ ] bug
- [x] feature
- [ ] enhancement

#### What does this PR do?
part of support vertex without tag

https://github.com/vesoft-inc/nebula/issues/3123
Close https://github.com/vesoft-inc/nebula/issues/2725

There are two remaining issues
1. Fetch returns `EMPTY`. At first I wanted to change it to return `NULL`, but I found that graph would change the execution flow based on the result returned by storage. If storage returns `EMPTY`, graph will think that the corresponding property has not been obtained, and will make a request again in the subsequent execution process; if storage returns `NULL`, graph will think that the value of the property has been obtained, which is `NULL`.

If it is acceptable to return to EMPTY, there is no need to make changes. Otherwise I need some help from someone who is familiar with graph service.

2. The issue of consistency between vertex and tag. For example, suppose there is a vertex and two tags, v1(t1,t2).`Update v1(t1)` and `Delete v1` are executed concurrently in any order, the final result should be that v1 does not exist (of course, t1, t2 also does not exist), but in the current scheme, the data may be v1(t1) after the execution is completed.

In the original design, we considered using locks to maintain consistency, but because the implementation of locks is more complicated, and other people have proposed a simpler solution, we did not adopt the lock solution. But at present, there is still a situation of breaking consistency without using a lock. This problem will be fixed in the future.

#### Checklist：

#### Release notes：


src/common/utils/Types.h
```diff
@@ -18,7 +18,7 @@ enum class NebulaKeyType : uint32_t {
   kSystem = 0x00000004,
   kOperation = 0x00000005,
   kKeyValue = 0x00000006,
-  // kVertex = 0x00000007,
+  kVertex = 0x00000007,
```
Shylock-Hg, line 21: Will storing tag/vertex in separate places decrease performance of getting properties of vertex?

cangfengzhs: There will be a decline, but it is negligible for the following reasons:
1. Only when the tag to be read does not exist, will the vertex be judged based on the vertexKey. The probability of this situation is very low.
2. Even if the vertexKey is read, it is only a `get()` request, and the overhead is very small.
3. Someone previously proposed a solution to put `vertexKey` and `tagKey` together, and use the method of `prefix` for query. But the performance of rocksdb's `seek()` and `iterate()` is very poor, which may be dozens of times that of `get()`. This approach will bring performance degradation instead. Of course this is just a theoretical analysis, I will do a simple experiment if necessary.

Shylock-Hg: ok

critical27 (approved):
LGTM

critical27:
I don't quite remember the design, it seems we need to write vertexKey when `delete tag`

cangfengzhs:
> 1. I don't quite remember the design, it seems we need to write vertexKey when `delete tag`
> 2. Where do we delete the vertexKey? Not in this PR? Just delete them when we `delete vertex`

1. We don't need write vertexKey when `delete tag`, because this will not generate a separate（without vertexKey） tag key.
2. `delete vertex` is already in this PR, just in `DeleteVerticesProcessor`.

critical27:
> > 1. I don't quite remember the design, it seems we need to write vertexKey when `delete tag`
> > 2. Where do we delete the vertexKey? Not in this PR? Just delete them when we `delete vertex`
> 
> 1. We don't need write vertexKey when `delete tag`, because this will not generate a separate（without vertexKey） tag key.
> 2. `delete vertex` is already in this PR, just in `DeleteVerticesProcessor`.

> > 1. I don't quite remember the design, it seems we need to write vertexKey when `delete tag`
> > 2. Where do we delete the vertexKey? Not in this PR? Just delete them when we `delete vertex`
> 
> 1. We don't need write vertexKey when `delete tag`, because this will not generate a separate（without vertexKey） tag key.
> 2. `delete vertex` is already in this PR, just in `DeleteVerticesProcessor`.

I see it, my bad.

The reason I suggest to write vertexKey when `delete tag` is that user only want to delete a tag, in this case, user believe the vertex do exists. In other words, we can regard `delete tag` is special type of `update`.

cangfengzhs:
> 

But if user delete a tag whose vertex is not exist, this will insert a vertexKey. This is very abnormal.

cangfengzhs:
> 
> I see it, my bad.
> 
> The reason I suggest to write vertexKey when `delete tag` is that user only want to delete a tag, in this case, user believe the vertex do exists. In other words, we can regard `delete tag` is special type of `update`.

But if user delete a tag whose vertex is not exist, this will insert a vertexKey. This is very abnormal.

tests/tck/features/delete/DeleteTag.IntVid.feature
```diff
@@ -41,6 +41,7 @@ Feature: Delete int vid of tag
       """
     Then the result should be, in any order:
       | player.name | player.age |
+      | EMPTY       | EMPTY      |
```
yixinglu, line 44: We shouldn't return a result with all properties empty value when the vertex doesn't exist. The empty response is what we expect.

cangfengzhs: Because we add a `vertexKey`, `Delete Tag` will not delete vertex at same time. Vertex is still existed

czpmango: Do we plan to support `fetch tags on hash("Tim Duncan")` later?
This seems like a natural user demand scenario bcz TagKey and VertexKey are abstracted out.

Shylock-Hg (approved):
LGTM

Shylock-Hg:
Close https://github.com/vesoft-inc/nebula/issues/2725

PR #3297, opened 2021-11-10
Spelling

jsoref:
#### What type of PR is this?
- [x] bug
- [ ] feature
- [ ] enhancement

#### What does this PR do?
This PR corrects misspellings identified by the [check-spelling action](https://github.com/marketplace/actions/check-spelling).

The misspellings have been reported at https://github.com/jsoref/nebula/commit/a65eeee9438fd920b98d9bec766168de51f40b9c#commitcomment-59912745

#### Special notes for your reviewer, ex. impact of this fix, etc:
I expect to adjust the PR based on project preferences.

Also, CI is currently broken in various ways.

#### Additional context:
The action reports that the changes in this PR would make it happy: https://github.com/jsoref/nebula/commit/4fbd464cc83ca5c566e1611f8d13545a04948e9d

Note: this PR does not include the action. If you're interested in running a spell check on every PR and push, that can be offered separately.

#### Checklist：
- [x] Documentation affected （Please add the label if documentation needs to be modified.)
- [ ] Incompatible （If it is incompatible, please describe it and add corresponding label.）
- [ ] Need to cherry-pick （If need to cherry-pick to some branches, please label the destination version(s).)
- [ ] Performance impacted: Consumes more CPU/Memory

#### Release notes：
Please confirm whether to reflect in release notes and how to describe:
>   As is, it's quite possible that this results in behavior breaks

PR #3331, opened 2021-11-18
Spell check with spelling

jsoref:
#### What type of PR is this?
- [x] bug
- [x] feature
- [ ] enhancement

#### What does this PR do?

This PR adds the [check-spelling action](https://github.com/marketplace/actions/check-spelling) per https://github.com/vesoft-inc/nebula/pull/3297#issuecomment-965954117 and corrects the few additional misspellings that were added between when that PR was created and now.

An overview of the action is available here: https://www.check-spelling.dev/

* When contributors add words that check-spelling doesn't recognize it will complain.
* It will give recommendations for how to update its metadata in order to accept the additional item.
* If the item(s) are actually misspelled words, then someone should fix the misspelling (preferably before merging).
* If the item(s) are in files that shouldn't be checked then something should be added/adjusted in `excludes.txt`.
* If the item(s) are acceptable (e.g. someone's name), then updating `expect.txt` is the correct course of action. -- One can do that manually or by using the suggested script.

In order for this action to work well, you should add a [branch protection rule](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/defining-the-mergeability-of-pull-requests/managing-a-branch-protection-rule) to enable [Require status checks before merging](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/defining-the-mergeability-of-pull-requests/about-protected-branches#require-status-checks-before-merging). If you don't, then when someone merges a PR that adds a typo, other contributors will get complaints (in the form of comments) about the typo when they make new PRs and will be confused by the message as they aren't intentionally adding those words.

#### Additional context:

I'll watch PRs to the repository for a month or so to ensure that things go smoothly. If you have trouble at any time (either during that window or thereafter), feel free to `@` me.

#### Checklist：
- [x] Documentation affected （Please add the label if documentation needs to be modified.) -- you may need to add some documentation for how to interact with the tool. 
- [x] Performance impacted: Consumes more CPU/Memory -- this should add ~1minute of GitHub action runtime per commit / PR update.

#### Release notes：


yixinglu:
Hi @jsoref 

Could you publish your spellchecking action to the [GitHub marketplace](https://docs.github.com/en/enterprise-cloud@latest/actions/creating-actions/publishing-actions-in-github-marketplace)? If so, we can use it directly in our workflow and need not to add it in the repo. Maybe we could reuse your action in other repos. WDYT?

yixinglu:
> Hi @jsoref
> 
> Could you publish your spellchecking action to the [GitHub marketplace](https://docs.github.com/en/enterprise-cloud@latest/actions/creating-actions/publishing-actions-in-github-marketplace)? If so, we can use it directly in our workflow and need not to add it in the repo. Maybe we could reuse your action in other repos. WDYT?

My fault, I see you have used some actions in workflow to do the checking. Those configuration files placed in the action directory misleads me.

Only one thing I concerned is whether it's a good practice to maintain the huge expect.txt file in repo.

jsoref:
My original design had the expect file living in a bucket instead. The code doesn't really care where it lives, you can configure a different source either by specifying a different location using `with:`/`config:` or by having a step that puts the files in place before the action runs.

It turns out that having the data in the repository generally has worked better as it enables forks to easily get ✅ and show how. But, you don't have to do it that way if you don't want to.

yixinglu:
make sense.

yixinglu:
After our discussion offline, we don't think it's a good practice to maintain these expected words in nebula repo. I close this PR temporarily now. 

Thx for your contribution again.

PR #3335, opened 2021-11-19
Insert vertex only

cangfengzhs:
#### What type of PR is this?
- [ ] bug
- [x] feature
- [ ] enhancement

#### What does this PR do?
add `insert vertex only` statement
For example:
```
Insert vertex (1,2,3);
```
About parentheses in statement.at the beginning, I wanted to implement it in the following format, but there was ambiguity:
```
Insert vertex t();  // stmt1
Insert vertex t() Values 1:(); // stmt2
```
In stmt1, `t()` represents a function, but in stmt2, `t()` represents a tag

part of issue: https://github.com/vesoft-inc/nebula/issues/3123
need PR: https://github.com/vesoft-inc/nebula/pull/3328
  
#### Checklist：

#### Release notes：


Shylock-Hg:
How about `insert vertex values 1:(), 2:() ...`?

cangfengzhs:
> How about `insert vertex values 1:(), 2:() ...`?

This problem has been initially solved. `insert vertex ifNotExists? VALUES vid_list` syntax is used.

src/graph/validator/MutateValidator.h
```diff
@@ -38,6 +38,20 @@ class InsertVerticesValidator final : public Validator {
   std::vector<storage::cpp2::NewVertex> vertices_;
 };
 
+class InsertVerticesOnlyValidator final : public Validator {
```
Shylock-Hg, line 41: Why don't reuse InsertVerticesValidator?

src/parser/Sentence.h
```diff
@@ -52,6 +52,7 @@ class Sentence {
     kDropTag,
     kDropEdge,
     kInsertVertices,
+    kInsertVerticesOnly,
```
Shylock-Hg, line 55: ditto

src/graph/validator/MutateValidator.h
cangfengzhs: The syntax of `InsertVerticesOnly` and the syntax of `InsertVertices` are quite different in the parser process. This causes the data structure contained in the two Sentences to be different. If both Sentences use `InsertVerticesValidator`, two completely different `toPlan` logics are also required. So simply use a new `InsertVerticesOnlyValidator`. This also ensures that the impact on `InsertVertices` is minimal, and other bugs will not be introduced during the modification process.

Shylock-Hg: https://github.com/vesoft-inc/nebula/blob/df3781cb1b530d93a6aaaff0f032cf626884f2a5/src/parser/parser.yy#L2839

I think just need add a new reduce rule like `vid { $$ = new VertexRowIten($1, new ValueList())}` in `vertex_row_item` rule.

cangfengzhs: But, `InsertVertices` can not use vertexID list as "1,2,3".It needs "1:(),2:(),3:()" instead. So if add a new reduce rule in `vertex_row_item`, the parse process of `insertVertices` will be very strange.

tests/tck/features/insert/insertVertexOnly.feature
```diff
@@ -0,0 +1,32 @@
...
+      CREATE EDGE e();
+      """
+    And wait 6 seconds
+    When executing query:
+      """
+      INSERT VERTEX VALUES 1:(),2:(),3:();
+      INSERT EDGE e() VALUES 1->2:(),2->3:();
+      """
+    Then the execution should be successful
+    When executing query:
```
Shylock-Hg, line 25: Add test case like `fetch prop on * 1`

wenhaocs (commented):
In previous conversations, there were more files. Just want to make sure this PR contains the files you want. Otherwise, good to go after fixing shylock's comment.

cangfengzhs:
> In previous conversations, there were more files. Just want to make sure this PR contains the files you want. Otherwise, good to go after fixing shylock's comment.

I adjust the implement according to shylock's suggestion and use another way which minimize the changes of previous code.

src/clients/storage/StorageClient.cpp
```diff
@@ -218,11 +218,12 @@ StorageRpcRespFuture<cpp2::GetPropResponse> StorageClient::getProps(
     req.set_space_id(param.space);
     req.set_parts(std::move(c.second));
     req.set_dedup(dedup);
-    if (vertexProps != nullptr) {
```
Shylock-Hg, line 221: Why change this?

Shylock-Hg: wait fix in https://github.com/vesoft-inc/nebula/pull/3443

wenhaocs: The same. Fixing something else?

Shylock-Hg: @cangfengzhs Please check why vertexProp equals to null, maybe not caused by #3443

cangfengzhs: `FetchVerticesPlanner::buildVertexProps` return a `nullptr` when props is empty

Shylock-Hg (approved):

nevermore3 (approved):

PR #3339, opened 2021-11-22
Some changes to avoid confusion

wenhaocs:
The original naming in the storage plan tree execution is confusing: when adding dependents to a node, all its children are set to `hasDependents_`. I believe `isRoot` is more accurate.


src/storage/exec/RelNode.h
```diff
@@ -68,7 +68,7 @@ class RelNode {
 
   void addDependency(RelNode<T>* dep) {
     dependencies_.emplace_back(dep);
-    dep->hasDependents_ = true;
+    dep->isRoot_ = false;
```
panda-sheep, line 71: `hasDependents_` indicates that other nodes depend on this node.
This node must be executed first, and then the other can be executed.

panda-sheep, line 71: `hasDependents_` indicates that other nodes depend on this node.
This node must be executed first, and then the other can be executed.
I feel `hasDependents_`  is  clearer?

panda-sheep, line 71: `hasDependents_` indicates that other nodes depend on this node.
This node must be executed first, and then the other can be executed.

wenhaocs: First, this function `addDependency ` will add all `dep` to the member vector `dependencies_ `. From this alone, it is natural to understand that `this` node depends on `dep`. From the execution flow, it can also be verified. `GetNeighbor` depends on `hashJoin`, depends on `Edge`.... But it conflicts with `dep->hasDependents_ = true;` This line is saying `dep` is depending on something, rather than `this` depends on `dep`. However, let's take an example, in `GetNeighborsProcessor`, the `Edge` node does not depend on anything in the execution flow. It is the leaf node. But with the current logic, `Edge` node will be set `hasDependents_`.

panda-sheep: you're right.
The true semantics of `hasDependents_` is to be dependent.

src/storage/test/StorageDAGBenchmark.cpp
```diff
@@ -56,7 +56,7 @@ class FutureDAG {
           // add dependency of root node
           node->addDependency(input.get());
         }
-        if (!node->hasDependents_) {
+        if (!node->isDependent_) {
```
SuperYoko, line 59: Maybe a method return (or direct use) dependencies_.empty() is enough?

wenhaocs: I don't think it works. As an example, in the tree structure below, what this `if` branch achieves is it appends the `GetNeighborsNode` to `output`. If we look at the `hasDependents_ ` vector in this tree, you can find that only the leaf nodes have empty vector. All other nodes including the root have non-empty vector. It's impossible to distinguish the root from others via the vector size.
image

panda-sheep (dismissed):
Good!

jackwener:
LGTM

src/storage/exec/StoragePlan.h
```diff
@@ -36,11 +36,11 @@ template <typename T>
 class StoragePlan {
  public:
   nebula::cpp2::ErrorCode go(PartitionID partId, const T& input) {
-    // find all leaf nodes, and a dummy output node depends on all leaf node.
+    // find all nodes without dependents, and a dummy output node depends on all those nodes.
```
jackwener, line 39:
```suggestion
    // find all ending end nodes(without dependents), and a dummy output node depends on all those nodes.
```
It's more direct

src/storage/test/StorageDAGBenchmark.cpp
wenhaocs: Can you approve the change if it looks good to you? Thank you!

yixinglu (approved):
LGTM

SuperYoko (approved):
LGTM

panda-sheep (approved):

PR #3351, opened 2021-11-25
[fix] can't show edge after clone space

liuyu85cn:
#### What type of PR is this?
- [X ] bug

#### What does this PR do?

Fix a bug can't show edge after clone space. 

Discussed from forum:

https://discuss.nebula-graph.com.cn/t/topic/6565


panda-sheep (approved):
By the way, do we need to add test cases?

yixinglu (dismissed):

yixinglu:
should use following commands to format the feature files:

```
cd /path/to/nebula/tests
make fmt
```

yixinglu (approved):

panda-sheep (approved):

Issue #3305, opened 2021-11-11
change default session_idle_timeout_secs 


HarrisChu:
refer https://discuss.nebula-graph.com.cn/t/topic/6429/17
Currently, session_idle_timeout_secs is 0 by default, It means no expiration.
If user doesn't release the session, and then it would be a lot of sessions in graphd, and may report to meta with useless sessions.

Suggest change the default value with 30min or 1hour?

HarrisChu:
BTW:
`wait_timeout` in mysql is 28800 by default, it means 8 hours

PR #3357, opened 2021-11-25
Modify default session timeout

Aiee:
#### What type of PR is this?
- [ ] bug
- [x] feature
- [x] enhancement

#### What does this PR do?
1. Modify default value for `session_idle_timeout_secs` and `client_idle_timeout_secs` from 0 (never timeout) to 28800(8 hours, which is also the default timeout MySQL uses).
2. Add gflag validator for `session_idle_timeout_secs` which checks the value when program starts and updating value. The valid range for `session_idle_timeout_secs` is [1, 604800]

Close https://github.com/vesoft-inc/nebula/issues/3305
  
#### Additional context:

Incompatible: Users should modify their configuration, i.e. change 0 to a valid value after this PR is merged.

Ref: https://dev.mysql.com/doc/refman/8.0/en/server-system-variables.html#sysvar_wait_timeout

#### Checklist：
- [x] Incompatible （If it is incompatible, please describe it and add corresponding label.）

#### Release notes：

### DIFF
--- a/conf/nebula-graphd.conf.default
+++ b/conf/nebula-graphd.conf.default
@@ -51,10 +51,11 @@
 --reuse_port=false
 # Backlog of the listen socket, adjust this together with net.core.somaxconn
 --listen_backlog=1024
-# Seconds before the idle connections are closed, 0 for never closed
---client_idle_timeout_secs=0
-# Seconds before the idle sessions are expired, 0 for no expiration
---session_idle_timeout_secs=0
+# The number of seconds Nebula service waits before closing the idle connections
+--client_idle_timeout_secs=28800
+# The number of seconds before idle sessions expire
+# The range should be in [1, 604800]
+--session_idle_timeout_secs=28800
 # The number of threads to accept incoming connections
 --num_accept_threads=1
 # The number of networking IO threads, 0 for # of CPU cores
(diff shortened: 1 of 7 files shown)


yixinglu (commented):
Maybe you could cleanup some logics about no timeout handling.

conf/nebula-graphd.conf.default
```diff
@@ -51,10 +51,11 @@
 --reuse_port=false
 # Backlog of the listen socket, adjust this together with net.core.somaxconn
 --listen_backlog=1024
-# Seconds before the idle connections are closed, 0 for never closed
---client_idle_timeout_secs=0
-# Seconds before the idle sessions are expired, 0 for no expiration
---session_idle_timeout_secs=0
+# The number of seconds Nebula service waits before closing the idle connections
+--client_idle_timeout_secs=28800
```
darionyaphet, line 55: 28800 is 8 hours Is it a little long?

HarrisChu:
after this pr, users should modify their configuration, i.e. change 0 to a valid value.
so I add `incompatible` label.

conf/nebula-graphd.conf.default
Aiee: 28800 seconds is the default timeout MySQL uses.
Ref: https://dev.mysql.com/doc/refman/8.0/en/server-system-variables.html#sysvar_wait_timeout

darionyaphet: 👍

yixinglu (approved):
LGTM

darionyaphet (approved):
LGTM

nevermore3 (approved):

PR #3359, opened 2021-11-25
Subscribe meta change p1

sherman-the-tank:
#### What type of PR is this?
- [ ] bug
- [X] feature
- [ ] enhancement

#### What does this PR do?
This PR provides the first step towards the pub-sub mechanism between meta server and graph/storage servers

N/A
  
This PR is actually much simpler then it looks like. Most of the changes are mechanical

#### Checklist：
- [X] Documentation affected （Please add the label if documentation needs to be modified.)

#### Release notes：


wenhaocs (commented):
Since it's a large diff, can you point out which files contain the key logic providing streaming services, which files are using the streaming, and the remaining are just adapting to the new thrift interface, in order to facilitate the review? Thank you!

src/daemons/MetaDaemon.cpp
```diff
@@ -327,8 +371,16 @@ int main(int argc, char* argv[]) {
     return EXIT_FAILURE;
   }
 
+  // Start the MetaStreamService in a separate thread so it will not block this thread
+  auto streamSvc = std::make_shared<nebula::meta::MetaStreamService>();
+  std::thread streamThread([] {
```
critical27, line 376: The code is in another PR?

src/meta/PubManager.cpp
```diff
@@ -0,0 +1,56 @@
...
+  }
+
+  // Remove publishers that are too old
+  for (auto& iter : oldPublishers) {
+    auto pub = *iter;
+    pub->evb_->runInEventBaseThreadAlwaysEnqueue([pub] {
+      std::move(pub->publisher_).complete();
+      VLOG(3) << "Terminate an old publisher";
+    });
+    publishers_.erase(iter);
```
yixinglu, line 51: it seems that we could not erase the iterator of another container in one container

yixinglu (commented):
We may need to upgrade the third party firstly to make this PR more clear.

Sophie-Xie:
This PR will not be merged in 3.x.

PR #3363, opened 2021-11-26
Eliminte the useless row collect.

Shylock-Hg:
#### What type of PR is this?
- [ ] bug
- [ ] feature
- [x] enhancement

#### Checklist：

#### Release notes：


yixinglu (commented):
Maybe we could also revise the DataCollect Node generation rule, check whether this node is not necessary in some scenarios. WDYT?

Shylock-Hg:
> 

The plan generated by planner is `DataCollect -> Limit -> Project`, after optimization it's `DataCollect -> Project -> Limit`. So in planner, we don't know it's useless.

Shylock-Hg:
> Maybe we could also revise the DataCollect Node generation rule, check whether this node is not necessary in some scenarios. WDYT?

The plan generated by planner is `DataCollect -> Limit -> Project`, after optimization it's `DataCollect -> Project -> Limit`. So in planner, we don't know it's useless.

yixinglu:
> > Maybe we could also revise the DataCollect Node generation rule, check whether this node is not necessary in some scenarios. WDYT?
> 
> The plan generated by planner is `DataCollect -> Limit -> Project`, after optimization it's `DataCollect -> Project -> Limit`. So in planner, we don't know it's useless.

I mean you can check when we need to append the DataCollect plan node into the plan, maybe now it's not necessary to do that.

CPWstatic (approved):

jievince (approved):

czpmango:
> Maybe we could also revise the DataCollect Node generation rule, check whether this node is not necessary in some scenarios. WDYT?

IMHO, it's unnecessary in most cases, bcz `Project` may also works. Besides, introducing redundant plannode types is not optimizer friendly.
https://github.com/vesoft-inc/nebula/blob/32394f4b767c66bfd067fcfe6f928988fd58a61f/src/graph/planner/SequentialPlanner.cpp#L39-L60

src/graph/optimizer/rule/EliminateRowCollectRule.cpp
```diff
@@ -0,0 +1,73 @@
...
+}
+
+// TODO match DataCollect->(Any Node with real result)
+const Pattern& EliminateRowCollectRule::pattern() const {
+  static Pattern pattern = Pattern::create(graph::PlanNode::Kind::kDataCollect,
+                                           {Pattern::create(graph::PlanNode::Kind::kProject)});
+  return pattern;
+}
+
+bool EliminateRowCollectRule::match(OptContext* octx, const MatchedResult& matched) const {
```
czpmango, line 35: Maybe `EliminateNoopDataCollect`?

PR #3384, opened 2021-12-01
add step fail hook

HarrisChu:
https://discuss.nebula-graph.com.cn/t/topic/6707/2

#### What type of PR is this?
- [ ] bug
- [ ] feature
- [x] enhancement

#### What does this PR do?
currently, we just display error information in `cmp_dataset` function, if there's any error not in the function, we cannot find more details.
so add step error hook, and display space if using a random space.

image

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -7,6 +7,7 @@
 import json
 import os
 import pytest
+import logging
 
 from tests.common.configs import all_configs
 from tests.common.types import SpaceDesc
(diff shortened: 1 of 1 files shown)


SuperYoko (approved):
Nice

yixinglu (approved):

tests/conftest.py
```diff
@@ -64,6 +65,16 @@ def pytest_addoption(parser):
                      help="Nebula Graph workspace")
 
 
+def pytest_bdd_step_error(request, feature, scenario, step, step_func, step_func_args):
```
darionyaphet, line 68: `pytest_bdd_step_error` not by called?

HarrisChu: it registered by pytest automatically.
1. pytest scan `conftest`
2. match the same definition with hook in pytest-bdd.

https://github.com/pytest-dev/pytest-bdd/blob/master/pytest_bdd/plugin.py

darionyaphet (approved):
LGTM

tests/conftest.py
darionyaphet: 👍

PR #3390, opened 2021-12-01
add cmake git hooks options

jiayuehua:
#### What type of PR is this?
- [*] bug
- [ ] feature
- [ ] enhancement

#### What does this PR do?

add option that can toggle wether cmake generate git hooks , default ON, So that platform without sh shell can set it OFF and commit code.
#### Checklist：

#### Release notes：


cmake/nebula/GitHooksConfig.cmake
```diff
@@ -1,19 +1,21 @@
-if(EXISTS "${CMAKE_CURRENT_SOURCE_DIR}/.git/")
```
yixinglu, line 1:
```cmake
if (ENABLE_CREATE_GIT_HOOKS AND (EXISTS "xxx"))
 # ...
endif()
```

jiayuehua: fix it.

jiayuehua:
has make if condition more clear.

yixinglu (approved):
LGTM

yixinglu:
@sherman-the-tank Yuehua uses visual studio IDE to develop nebula project locally. I think it's ok to introduce this option, it will not affect current workflow of others and is convenient for him to develop under windows system. WDYT?

jiayuehua:
Add include what you use to let developer delete no use include files.

cmake/IncludeWhatYouUse.cmake
```diff
@@ -0,0 +1,11 @@
+if(ENABLE_IncludeWhatYouUse)
+  find_program(IncludeWhatYouUse include-what-you-use)
```
yixinglu, line 2: `IncludeWhatYouUse_FOUND`

CMakeLists.txt
```diff
@@ -26,6 +26,8 @@ project("Nebula Graph" C CXX)
 option(ENABLE_PACK_ONE "Whether to package into one" ON)
 option(ENABLE_VERBOSE_BISON "Enable Bison to report state" OFF)
 option(ENABLE_PACKAGE_TAR "Enable package artifacts to tar." OFF)
+option(ENABLE_CREATE_GIT_HOOKS "Enable create git hooks." ON)
+option(ENABLE_IncludeWhatYouUse "Enable Include_What_You_Use" OFF)
```
sherman-the-tank, line 30: Please use all CAPITAL, like other options, following the convention

cmake/IncludeWhatYouUse.cmake
```diff
@@ -0,0 +1,11 @@
+if(ENABLE_IncludeWhatYouUse)
```
sherman-the-tank, line 1: Suggest to place this file in cmake/nebula

cmake/IncludeWhatYouUse.cmake
yixinglu: Why not to turn on iwyu when find the program automatically? i think this option is not necessary.

sherman-the-tank (commented):
Well done!

Just couple styling comments

cmake/IncludeWhatYouUse.cmake
jiayuehua: should place in cmake. This is a general cmake tool

jiayuehua: ok I change it to CAPITAL

cmake/IncludeWhatYouUse.cmake
jiayuehua: because it run very slow,  Just let developer run it  .

yixinglu (approved):

Aiee (approved):

PR #3392, opened 2021-12-02
ignore existed index

critical27:
Mainly for POC, same as 1.0. If specified with IGNORE_EXISTED_INDEX, we won't read or delete old index when insert. It could be useful when we do the first import, or the index is rarely modified.

Pros: no more read old value or delete old index when we insert, only write will be involved, same as no index, performance will get a great improvement.
Cons: since we don't delete index, lookup will read some wrong data. **User should really be careful about this**

For example :
```
INSERT VERTEX person(id) VALUES "100":(1), "200":(1)

LOOKUP ON person WHERE person.id == 1 YIELD id(vertex) as id
// both 100 and 200 will satisfy

INSERT VERTEX IGNORE_EXISTED_INDEX person(id) VALUES "200":(2)

LOOKUP ON person WHERE person.id == 1 YIELD id(vertex) as id
// both 100 and 200 will satisfy, since old index of 200 is not deleted

LOOKUP ON person WHERE person.id == 2 YIELD id(vertex) as id
// only 200 will satisfy
```

Documents may or may not add this `IGNORE_EXISTED_INDEX` (up to you guys), if it is misused, wrong data will be looked up.


HarrisChu:
LOOKUP ON person WHERE person.id == 1 YIELD id(vertex) as id
// only 200 will satisfy

person.id == 2?

HarrisChu:
after this, when running rebuild index, what's the behavior for "200" vertext?

critical27:
> LOOKUP ON person WHERE person.id == 1 YIELD id(vertex) as id // only 200 will satisfy
> 
> person.id == 2?

My bad, 👍

critical27:
> person.id == 1

Both person.id == 1 and person.id == 2 will satisfy

critical27:
There are memory corruption, hold on one second.

panda-sheep (approved):
Good job!

wenhaocs (approved):
Nice and clean. Great job.

src/parser/parser.yy
```diff
@@ -162,7 +162,7 @@ static constexpr size_t kCommentLengthLimit = 256;
 %token KW_BOOL KW_INT8 KW_INT16 KW_INT32 KW_INT64 KW_INT KW_FLOAT KW_DOUBLE
 %token KW_STRING KW_FIXED_STRING KW_TIMESTAMP KW_DATE KW_TIME KW_DATETIME
 %token KW_GO KW_AS KW_TO KW_USE KW_SET KW_FROM KW_WHERE KW_ALTER
-%token KW_MATCH KW_INSERT KW_VALUE KW_VALUES KW_YIELD KW_RETURN KW_CREATE KW_VERTEX KW_VERTICES
+%token KW_MATCH KW_INSERT KW_VALUE KW_VALUES KW_YIELD KW_RETURN KW_CREATE KW_VERTEX KW_VERTICES KW_IGNORE_EXISTED_INDEX
```
yixinglu, line 165: emm.. This `IGNORE_EXISTED_INDEX` keyword is a little complicated to use, maybe it's better to split it into some words such as:

```
INSERT IGNORE EXISTS INDEX VERTEX ...
```

reference mysql insert ignore statement.

critical27: I understand, but this is for almost the only one poc. They have used it since 1.X... Maybe just keep it.

PR #3405, opened 2021-12-03
support specify partId mode

panda-sheep:
#### What type of PR is this?
- [ ] bug
- [X] feature
- [ ] enhancement

#### What does this PR do?
 support specify partId mode in upgrade tool

#### Checklist：

#### Release notes：

### DIFF
--- a/src/tools/db-upgrade/DbUpgrader.cpp
+++ b/src/tools/db-upgrade/DbUpgrader.cpp
@@ -151,6 +151,76 @@ Status UpgraderSpace::initSpace(const std::string& sId) {
   return Status::OK();
 }
 
+void UpgraderSpace::writeSstFile(std::vector<kvstore::KV>& data,
+                                 std::string& newPartPath,
+                                 PartitionID partId,
+                                 bool isTag) {
+  VLOG(2) << "Send record total rows " << data.size();
+  /*
+    auto code = writeEngine_->multiPut(data);
+    if (code != nebula::cpp2::ErrorCode::SUCCEEDED) {
+      LOG(FATAL) << "Write multi put in space id " << spaceId_ << " part id " << partId
+                 << " failed.";
+    }
+  */
+  rocksdb::Options options;
+  options.file_checksum_gen_factory = rocksdb::GetFileChecksumGenCrc32cFactory();
+  rocksdb::SstFileWriter sstFileWriter(rocksdb::EnvOptions(), options);
+
+  auto newPartPathTmp = getCurrentPartSstFile(partId, isTag);
+  if (newPartPathTmp == newPartPath) {
+    sleep(2);
+    newPartPath = getCurrentPartSstFile(partId, true);
+  } else {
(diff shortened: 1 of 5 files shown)


src/tools/db-upgrade/DbUpgraderTool.cpp
```diff
@@ -164,15 +174,96 @@ int main(int argc, char* argv[]) {
...
+    auto space = subDirs[0];
+    GraphSpaceID spaceId;
+    try {
+      spaceId = folly::to<GraphSpaceID>(space);
+    } catch (const std::exception& ex) {
+      LOG(ERROR) << "Cannot convert space id " << space;
+      return EXIT_FAILURE;
+    }
+
+    auto sRet = schemaMan->toGraphSpaceName(spaceId);
```
critical27, line 218: spaceName is useless?

src/tools/db-upgrade/DbUpgraderTool.cpp
```diff
@@ -164,15 +174,96 @@ int main(int argc, char* argv[]) {
...
+      }
+      iter->next();
+    }
+  }
+  if (!partsAll.empty()) {
+    std::string partstr;
+    for (auto part : partsAll) {
+      partstr += folly::stringPrintf("%d ", part);
+    }
+    LOG(FATAL) << "There are some part " << partstr << "not found";
```
critical27, line 260: Good! Just use `LOG(ERROR)` and `return EXIT_FAILURE;`, no need to core.

src/tools/db-upgrade/DbUpgrader.cpp
```diff
@@ -322,33 +403,254 @@ void UpgraderSpace::runPartV1() {
...
         lastSrcVertexId = svId;
         lastEdgeType = edgetype;
         lastRank = ranking;
         lastDstVertexId = dvId;
       }
 
-      if (data.size() >= FLAGS_write_batch_num) {
-        VLOG(2) << "Send record total rows " << data.size();
+      // tag data
+      if (tagData.size() >= FLAGS_write_batch_num) {
```
critical27, line 422: Some code could be unified, here and below when `if (tagData.size() > 0) {`, but it is ok for this tool.

critical27 (commented):
Good job, thx a lot.

src/tools/db-upgrade/DbUpgrader.cpp
panda-sheep: yeah, fixed.

src/tools/db-upgrade/DbUpgraderTool.cpp
panda-sheep: Here only check whether the space exists or not

src/tools/db-upgrade/DbUpgraderTool.cpp
panda-sheep: ok

Issue #3170, opened 2021-10-20
Drop the Space when Rebuilding Index, resulting in storaged  crash


Donald-Su:
**Describe the bug (__must be provided__)**
- a cluster run the same machine（use different port）, and work well
- create the space, and set **replica_factor 3**  
- **Drop the Space when Rebuilding Index, resulting in storaged  crash**

**Your Environments (__must be provided__)**

* OS: 
   * CentOS release 6.6
   * CentOS Linux release 7.3.1611
* nebula version： **v2.5.1**

**How To Reproduce(__must be provided__)**

1. Step 1：create the space, and then rebuild index,  the status like follow 
![image](https://user-images.githubusercontent.com/43985465/138134048-a89b819a-4eb4-4915-bf33-7b2e4d55af1c.png)

2. Step 2: drop the space, the status still have **QUEUE** and **RUNNING**
![image](https://user-images.githubusercontent.com/43985465/138137927-99a6834f-2c68-4507-91c2-ca5ad97fe039.png)

3. Step 3: later, one of the the cluster will crash, use `show hosts` display "OFFLINE", like follow
![image](https://user-images.githubusercontent.com/43985465/138134824-e3bdadf3-bad5-4779-8de2-dfe9a998883d.png)

**Additional context**
- test many times, the status same core in  **RebuildEdgeIndexTask** or **RebuildTagIndexTask** like follow.
![image](https://user-images.githubusercontent.com/43985465/138135778-ab5f8be3-defc-4856-87ce-423d29592685.png)

- the debug info of storaged process
```
#0 0x00007f9fd86c79d9 in raise () from /lib64/libc.so.6
#1 0x00007f9fd86c90e8 in abort () from /lib64/libc.so.6
#2 0x00000000028316ad in rocksdb::port::PthreadCall(char const*, int) [clone .part.1] ()
#3 0x0000000003df24b0 in rocksdb::port::Mutex::Lock() ()
#4 0x0000000003b8290b in rocksdb::(anonymous namespace)::CleanupIteratorState(void*, void*) ()
#5 0x0000000003d6c8ec in rocksdb::Cleanable::~Cleanable() ()
#6 0x0000000003bfff52 in rocksdb::DBIter::~DBIter() ()
#7 0x0000000003e107cc in rocksdb::ArenaWrappedDBIter::~ArenaWrappedDBIter() ()
[...]

PR #3406, opened 2021-12-03
issue_3170: drop space after stop all AdminTaskJobs

Nivras:
#### What type of PR is this?
- [x] bug
- [ ] feature
- [ ] enhancement

#### What does this PR do?
* When Storage use MetaClient Heartbeat find out there is a space have been dropped, before remove the space, cancel all AdminTasks in this space and wait them stopped first.

* mark AdminTask status in memory, use std::atomic, when the AdminTask in task, and status is QUEUE, we will set QUEUE to CANCELED, and do not run this task. 

issue: https://github.com/vesoft-inc/nebula/issues/3170
  
This PR is trying to stop all the AdminTasks(RebuildTagIndex, RebuildEdgeIndex, Stats) before storage delete a space. AdminTaskManager support the interface to cancel a job, but this interface only set a flag (canceled_) to true, when the job pending in the taskQueue, it won't be canceled until the job going to run.

So this PR add a atomic enum to mark the job's status, the status is `QUEUE, RUNNING, CANCELED, FINISNED`, and when we canceled a job, first if the job's status is QUEUE, change status to CANCELED, and return. else we only change the flag(canceled_), because this job is RUNNING or FINISHED.

When TaskManager take a job in taskQueue, it will check the job's status, and if the status is CANCELED, TaskManager will clean this job without running it. if the job's status is QUEUE, taskManger change it to RUNNING and run it. And a RUNNING job will be canceled if flag(canceled_) is true.

#### Checklist：

#### Release notes：


src/clients/meta/MetaClient.cpp
```diff
@@ -948,6 +948,15 @@ void MetaClient::diff(const LocalCache& oldCache, const LocalCache& newCache) {
       for (const auto& oldPart : oldParts) {
         listener_->onPartRemoved(spaceId, oldPart.first);
       }
+
+      // wait spaceId all AdminTask stopped
+      if (taskListener_ != nullptr) {
+        taskListener_->cancelSpaceJobs(spaceId);
```
critical27, line 954: Maybe only expose one interfaces like `waitSpaceJobs`, and we could do all the things in AdminTaskManager.

src/storage/admin/AdminTaskManager.cpp
```diff
@@ -249,6 +257,17 @@ void AdminTaskManager::schedule() {
     }
 
     auto task = it->second;
+    auto queueStat = AdminTaskStatus::QUEUE;
+    if (!task->taskStatus_.compare_exchange_strong(queueStat, AdminTaskStatus::RUNNING)) {
+      if (queueStat == AdminTaskStatus::CANCELED) {
+        // task has already been canceled, remove it from the queue
+        LOG(INFO) << folly::sformat("job {} has been calceled", task->getJobId());
+        task->finish(nebula::cpp2::ErrorCode::SUCCEEDED);
```
critical27, line 265: I remember there is a error code related to canceled task

src/clients/meta/MetaClient.h
```diff
@@ -795,6 +813,10 @@ class MetaClient {
   MetaChangedListener* listener_{nullptr};
   // The lock used to protect listener_
   folly::RWSpinLock listenerLock_;
+  // The taskListener_ is the AdminTaskManager
+  AdminTaskManagerListener* taskListener_{nullptr};
+  // The lock used to protect taskListerer_
+  folly::RWSpinLock taskListenerLock_;
```
critical27, line 819: If only one place use this taskListener_, we could remove the lock.

src/storage/admin/RebuildIndexTask.h
```diff
@@ -36,7 +36,11 @@ class RebuildIndexTask : public AdminTask {
                                                    const IndexItems& items,
                                                    kvstore::RateLimiter* rateLimiter) = 0;
 
-  void cancel() override { canceled_ = true; }
+  void cancel() override {
+    canceled_ = true;
+    auto queueStat = AdminTaskStatus::QUEUE;
+    taskStatus_.compare_exchange_strong(queueStat, AdminTaskStatus::CANCELED);
```
critical27, line 42: What about the RUNNING tasks?

Nivras: subtask will check the canceled_ flag in every loop(loop for rocksdb iterator), and if canceled_ is true, subtask will return, and AdminTaskManager::schedule() will finish and erase this task form taskQueue.

critical27: 🆗

critical27 (approved):
LGTM now, good job

src/storage/admin/AdminTask.h
```diff
@@ -98,6 +102,7 @@ class AdminTask {
   virtual void subTaskFinish(nebula::cpp2::ErrorCode rc) {
     auto suc = nebula::cpp2::ErrorCode::SUCCEEDED;
     rc_.compare_exchange_strong(suc, rc);
+    taskStatus_.store(AdminTaskStatus::FINISHED, std::memory_order_relaxed);
```
pengweisong, line 105: may move to task.finish()

src/storage/admin/AdminTaskManager.h
```diff
@@ -45,6 +53,10 @@ class AdminTaskManager {
   nebula::cpp2::ErrorCode cancelJob(JobID jobId);
   nebula::cpp2::ErrorCode cancelTask(JobID jobId, TaskID taskId = -1);
 
+  void cancelSpaceJobs(GraphSpaceID spaceId);
```
pengweisong, line 56: because the parameter hit that we are handle jobs for specified space, then we can name these functions for short? sth like: cancelJobs(spaceId), runningJobCnt(spaceId), waitCancelJobs(spaceId).

src/storage/admin/AdminTaskManager.cpp
```diff
@@ -353,5 +372,40 @@ bool AdminTaskManager::isFinished(JobID jobID, TaskID taskID) {
   return iter->second->unFinishedSubTask_ == 0;
 }
 
+void AdminTaskManager::cancelSpaceJobs(GraphSpaceID spaceId) {
+  auto it = tasks_.begin();
```
pengweisong, line 376: will the iterator invalidated by other thread who use `tasks_` at the same time?

src/storage/admin/AdminTaskManager.cpp
```diff
@@ -353,5 +372,40 @@ bool AdminTaskManager::isFinished(JobID jobID, TaskID taskID) {
...
+      FLOG_INFO("cancel task(%d, %d), spaceId: %d", it->first.first, it->first.second, spaceId);
+    }
+    ++it;
+  }
+}
+
+int32_t AdminTaskManager::spaceRunningJobCnt(GraphSpaceID spaceId) {
+  int32_t jobCnt = 0;
+  auto it = tasks_.begin();
+  while (it != tasks_.end()) {
```
pengweisong, line 390: missing it++? why not using for loop?

src/storage/admin/AdminTaskManager.cpp
```diff
@@ -112,6 +115,11 @@ void AdminTaskManager::handleUnreportedTasks() {
                                       jobId,
                                       taskId,
                                       fut.value().status().toString());
+          if (fut.value().status() == Status::Error("Space not existed!")) {
+            // space has been droped, remove the task status.
+            keys.emplace_back(key.data(), key.size());
+            break;
```
pengweisong, line 121: should this break be removed?

src/storage/admin/AdminTask.h
```diff
@@ -33,6 +33,8 @@ class AdminSubTask {
 
 enum class TaskPriority : int8_t { LO, MID, HI };
 
+enum class AdminTaskStatus : int8_t { QUEUE, RUNNING, FINISHED, CANCELED };
```
pengweisong, line 36: 'isRunning()' is enough, there is no need to introduce an new status, it will caused confused with the old one.

liuyu85cn:
so now, drop space may stuck in console for a while?

Nivras:
> so now, drop space may stuck in console for a while?

No, console execute "drop space" will delete space infos in meta, and return. This PR only changed the storage, storage will drop the space when it found this space is not in heartbeat info from meta.

src/storage/admin/AdminTask.h
```diff
@@ -70,13 +70,17 @@ class AdminTask {
 
   virtual int8_t getPriority() { return static_cast<int8_t>(ctx_.pri_); }
 
-  virtual void finish() { finish(rc_); }
+  virtual void finish() {
+    isRunning_ = false;
```
pengweisong, line 74: duplicated with the same sentence finish(rc_)?

src/storage/admin/AdminTask.h
```diff
@@ -102,20 +108,27 @@ class AdminTask {
...
+  virtual bool isRunning() { return isRunning_; }
+
+  virtual bool isCanceled() { return canceled_; }
+
   meta::cpp2::AdminCmd cmdType() { return ctx_.cmd_; }
 
  public:
   std::atomic<size_t> unFinishedSubTask_;
   SubTaskQueue subtasks_;
+  std::atomic<bool> isRunning_{false};
```
pengweisong, line 125: keep consistent naming with canceled? call running_ and canceled , or isRunning and isCanceled

src/storage/admin/AdminTaskManager.cpp
```diff
@@ -353,5 +369,37 @@ bool AdminTaskManager::isFinished(JobID jobID, TaskID taskID) {
   return iter->second->unFinishedSubTask_ == 0;
 }
 
+void AdminTaskManager::cancelJobs(GraphSpaceID spaceId) {
```
pengweisong, line 372: cancelTasks?

src/storage/admin/AdminTaskManager.cpp
```diff
@@ -353,5 +369,37 @@ bool AdminTaskManager::isFinished(JobID jobID, TaskID taskID) {
...
+  auto it = tasks_.begin();
+  while (it != tasks_.end()) {
+    if (it->second->getSpaceId() == spaceId) {
+      it->second->cancel();
+      removeTaskStatus(it->second->getJobId(), it->second->getTaskId());
```
pengweisong, line 377: could call cancelTask?

src/storage/admin/StatsTask.cpp
```diff
@@ -73,6 +73,11 @@ nebula::cpp2::ErrorCode StatsTask::genSubTask(GraphSpaceID spaceId,
                                               PartitionID part,
                                               std::unordered_map<TagID, std::string> tags,
                                               std::unordered_map<EdgeType, std::string> edges) {
+  if (UNLIKELY(canceled_)) {
+    LOG(ERROR) << "Stats task is canceled";
+    return nebula::cpp2::ErrorCode::SUCCEEDED;
```
pengweisong, line 78: nebula::cpp2::ErrorCode::E_USER_CANCEL？

liuyu85cn:
Mmmm, functional, this PR looks good to me. Also, I suggest not to modify meta client. 
(why meta client need to know anything about task manager?)

I think it's better to open a add (un)register function in kvstore, 
which can accept any function to callback before it delete one space.

pengweisong (dismissed):
LGTM; Perfect job!

Nivras:
> Mmmm, functional, this PR looks good to me. Also, I suggest not to modify meta client. (why meta client need to know anything about task manager?)
> 
> I think it's better to open a add (un)register function in kvstore, which can accept any function to callback before it delete one space.

Thanks~ meta client should not know the taskManger, I will change this PR for this.

liuyu85cn (dismissed):
good job, LGTM

pengweisong (approved):

liuyu85cn (approved):

src/storage/admin/AdminTask.h
```diff
@@ -102,20 +105,27 @@ class AdminTask {
...
+  virtual bool isRunning() { return running_; }
+
+  virtual bool isCanceled() { return canceled_; }
+
   meta::cpp2::AdminCmd cmdType() { return ctx_.cmd_; }
 
  public:
   std::atomic<size_t> unFinishedSubTask_;
   SubTaskQueue subtasks_;
+  std::atomic<bool> running_{false};
```
Shylock-Hg, line 122: Maybe a enum is better than bool.

src/storage/admin/AdminTask.h
```diff
@@ -102,20 +105,27 @@ class AdminTask {
...
 
  protected:
   StorageEnv* env_;
   TaskContext ctx_;
   std::atomic<nebula::cpp2::ErrorCode> rc_{nebula::cpp2::ErrorCode::SUCCEEDED};
+  std::atomic<bool> canceled_{false};
```
Shylock-Hg, line 128: And here.

PR #3408, opened 2021-12-05
Step 2: fix tests for 3rd party v3

sherman-the-tank:
#### What type of PR is this?
- [ ] bug
- [ ] feature
- [X] enhancement

#### What does this PR do?
Fixed some bugs

#### Checklist：

#### Release notes：


wenhaocs (approved):

critical27:
Just review the last commit named `Switched to semifuture` is enough?

src/common/base/CollectNSucceeded-inl.h
```diff
@@ -11,59 +11,109 @@ namespace nebula {
...
       }
     });
+    VLOG(3) << "Set up future callback " << index;
   }
 
-  return ctx->promise.getFuture();
+  auto future = ctx->promise.getSemiFuture();
+  if (!executors.empty()) {
+    VLOG(3) << "Wiring the nested executors";
+    future = std::move(future).defer(
```
critical27, line 103: Any special reason to defer work on consumer side? I notice that you changed `result` to `std::vector<folly::Optional<folly::Try<FutureReturnType<FutureIter>>>>`.

src/common/base/CollectNSucceeded-inl.h
```diff
@@ -11,59 +11,109 @@ namespace nebula {
...
+          return;
         }
-        if ((++ctx->numCompleted) == ctx->nTotal || ctx->nSucceeded == n) {
-          // Done
-          VLOG(2) << "Set Value [completed=" << ctx->numCompleted << ", total=" << ctx->nTotal
-                  << ", Result list size=" << ctx->results.size() << "]";
-          ctx->promise.setValue(std::move(ctx->results));
+
+        bool fulfilled = false;
+        if (!ctx->fulfilled.compare_exchange_strong(fulfilled, true)) {
```
critical27, line 77: One of a big issue of `collectNSucceeded` is that it is no thread-safe, which requires all threads need to run on same executor and hurts performance quite a lot. We need to reconsider about it later, for example, use `SharedPromise`.

src/common/base/CollectNSucceeded-inl.h
```diff
@@ -11,59 +11,109 @@ namespace nebula {
...
+  auto ctx = std::make_shared<Context>(total, std::forward<ResultEval>(eval));
   for (size_t index = 0; first != last; ++first, ++index) {
-    first->setCallback_([n, ctx, index](auto, folly::Try<FutureReturnType<FutureIter>>&& t) {
-      if (!ctx->promise.isFulfilled()) {
+    first->setCallback_([n, ctx, index](auto&&,
+                                        folly::Try<FutureReturnType<FutureIter>>&& t) {
+      VLOG(3) << "Received the callback from the future " << index;
+
+      // relaxed because this guards control but does not guard data
+      auto const c = 1 + ctx->nCompleted.fetch_add(1, std::memory_order_relaxed);
```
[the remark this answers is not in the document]
critical27: Maybe you can't use `memory_order_relaxed` here, consider `3 replica, total = 2, n = 1`
1. both thread get `c = 0` (the `1 + ...` is not counted)
2. thread 1 failed because of exception, and return in line 77
3. thread 2 failed, but the we can still pass the `if` statement in line 76, and return in line 77

At this point the promise of `ctx` will never be set.

sherman-the-tank:
This PR is too old. Close it

PR #3415, opened 2021-12-06
[Raft] fix some remaining problems after pre-vote merged

critical27:
I am reviewing current implementation,  some issues are found:
1. We didn't save the info of voting for myself when start formal election, which could lead to split brain: 
    * There are ABCDE 5 replicas, and BC vote for A.
    * If A does not voted itself, another host could be the leader, for example, E could be the leader with votes from ADE.
2. #3322 has missed one rollback, leader will check whether term/role changed after we get AppendLogResponse, if failed, we need to rollback, otherwise, we can't write any logs to wal since `lastLogId_ != wal_->lastLogId()`.
3. When we start up, if `lastLogId_ < committedLogId_`, only lastLogId_ is set to committedLogId_.

And some minor changes, set `leader_` whenever term changed, this is a prepare for later AppendLogs refactor.

### DIFF
--- a/src/kvstore/raftex/RaftPart.cpp
+++ b/src/kvstore/raftex/RaftPart.cpp
@@ -273,7 +273,7 @@ void RaftPart::start(std::vector<HostAddr>&& peers, bool asLearner) {
     LOG(INFO) << idStr_ << "Reset lastLogId " << lastLogId_ << " to be the committedLogId "
               << committedLogId_;
     lastLogId_ = committedLogId_;
-    lastLogTerm_ = term_;
+    lastLogTerm_ = logIdAndTerm.second;
     wal_->reset();
   }
   LOG(INFO) << idStr_ << "There are " << peers.size() << " peer hosts, and total "
@@ -293,6 +293,7 @@ void RaftPart::start(std::vector<HostAddr>&& peers, bool asLearner) {
   if (asLearner) {
     role_ = Role::LEARNER;
   }
+  leader_ = HostAddr("", 0);
   startTimeMs_ = time::WallClock::fastNowInMilliSec();
   // Set up a leader election task
   size_t delayMS = 100 + folly::Random::rand32(900);
@@ -852,6 +853,7 @@ void RaftPart::processAppendLogResponses(const AppendLogResponses& resps,
       std::lock_guard<std::mutex> g(raftLock_);
       res = canAppendLogs(currTerm);
       if (res != AppendLogResult::SUCCEEDED) {
+        wal_->rollbackToLog(lastLogId_);
         break;
       }
       lastLogId_ = lastLogId;
@@ -996,6 +998,9 @@ bool RaftPart::prepareElectionRequest(cpp2::AskForVoteRequest& req,
     req.set_term(term_ + 1);
   } else {
     req.set_term(++term_);
+    // vote for myself
+    votedAddr_ = addr_;
+    votedTerm_ = term_;
   }
   req.set_last_log_id(lastLogId_);
   req.set_last_log_term(lastLogTerm_);
@@ -1073,6 +1078,7 @@ bool RaftPart::processElectionResponses(const RaftPart::ElectionResponses& resul
       LOG(INFO) << idStr_ << "Partition is elected as the new leader for term " << proposedTerm;
       term_ = proposedTerm;
       role_ = Role::LEADER;
+      leader_ = addr_;
       isBlindFollower_ = false;
     }
     return true;
@@ -1106,6 +1112,7 @@ folly::Future<bool> RaftPart::leaderElection(bool isPreVote) {
     // So we need to go back to the follower state to avoid the case.
     std::lock_guard<std::mutex> g(raftLock_);
     role_ = Role::FOLLOWER;
+    leader_ = HostAddr("", 0);
     inElection_ = false;
     return false;
   }
@@ -1303,10 +1310,13 @@ void RaftPart::processAskForVoteRequest(const cpp2::AskForVoteRequest& req,
     return;
   }
 
+  auto oldRole = role_;
   auto oldTerm = term_;
   // req.get_term() >= term_, we won't update term in prevote
   if (!req.get_is_pre_vote()) {
     term_ = req.get_term();
+    role_ = Role::FOLLOWER;
+    leader_ = HostAddr("", 0);
   }
 
   // Check the last term to receive a log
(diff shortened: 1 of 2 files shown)


src/kvstore/raftex/RaftPart.cpp
```diff
@@ -1345,29 +1346,33 @@ void RaftPart::processAskForVoteRequest(const cpp2::AskForVoteRequest& req,
     resp.set_error_code(cpp2::ErrorCode::E_TERM_OUT_OF_DATE);
     return;
```
kikimo, line 1347: We cannot just return here. When `req.get_term() > term_`，we should also update term_ and step to follower.

critical27: My bad, hold on

critical27: My bad, there is a missing local commit. Hold on one second.

src/kvstore/raftex/RaftPart.cpp
```diff
@@ -1303,10 +1310,13 @@ void RaftPart::processAskForVoteRequest(const cpp2::AskForVoteRequest& req,
...
   }
 
+  auto oldRole = role_;
   auto oldTerm = term_;
   // req.get_term() >= term_, we won't update term in prevote
   if (!req.get_is_pre_vote()) {
     term_ = req.get_term();
+    role_ = Role::FOLLOWER;
```
kikimo, line 1318: This is also problermatic. Suppose `term_ == req.get_term()` and they are two competing candidates, you cannot simply convert the voter into follower even though its log is older.

critical27: You mean only convert to follower when req.term is greater, but when terms are equal check normally?

kikimo: Right, you cannot convert a candidate into follower when seeing a RequestVote rpc of same term(but you should conver to 
 follower when seeing AppendEntries rpc of the same term).

critical27: I don't think there is a big difference between AskForVote and AppendLog when term are same. Think of this case:
ABCDE are on same term and all of them start election, let's say A start first and convert BCDE into follower, there could be two result:
1. A is elected, BC voted for A, but DE rejected because their logs are more up-to-date. But this is ok, because logs on   DE is only minority.
2. A is not elected, in this case, one of BCDE will start a new round election.

pengweisong: when term equal, could become leader only when this peer vote for the candidate.

critical27: Yeah, if leader of same term step down, that would be quite unstable. Fixed

kikimo: Suppose in a cluster of 5 noedes: A B C D E;
A, B both starts an election at the same term 5, B's log if newer than A,
C, D, E voted for A, but before A converting to leader, B send a RequestVote rpc to A, and if we convert A to a follower,
B will also failed the election at term 5 because C, D, E had voted for A at term 5,
we will have to start a new election. 

We must process differently on seeing same term for RequestVote and AppendEntries rpc request. The reason is that, when you see an AppendEntries rpc of the same term, that mean a leader has benn elected, so you must step follower if your are not; but seeing same term int RequestVote rpc request does not means that a leader has been elected for this term, so, don't do it.

kikimo: Another very very very important problem: **reset `votedAddr_` when you update `term_`**.  

It's disastrous to just update `term_` but keep the value of `votedAddr_` from older term, a cluster might takes very long time to vote a leader.

Think we must do a thoughly check for this in our Raft implementation.

critical27: > Another very very very important problem: reset votedAddr_ when you update term_

Why? We have discussed this issue in pre-vote, see the line 1351. 

`voteAddr_` and `votedTerm_` should check together, and they are only updated when voted for others.

kikimo: > 

Oh, I forgot `votedTerm_`

src/kvstore/raftex/RaftPart.cpp
```diff
@@ -1345,29 +1355,34 @@ void RaftPart::processAskForVoteRequest(const cpp2::AskForVoteRequest& req,
...
 
-  // Ok, no reason to refuse, we will vote for the candidate
-  LOG(INFO) << idStr_ << "The partition will vote for the candidate " << candidate
-            << ", isPreVote = " << req.get_is_pre_vote();
-
-  // Before change role from leader to follower, check the logs locally.
-  if (role_ == Role::LEADER && wal_->lastLogId() > lastLogId_) {
+  // not a pre-vote, need to rollback wal if necessary
+  // role_ and term_ has been set above
+  if (oldRole == Role::LEADER && wal_->lastLogId() > lastLogId_) {
```
kikimo, line 1371: Still not very clear about this part, let me try to clarify:
1. is `lastLogId_` == commitIndex?
2. wal->lastLogId() returns entries appended but might not have been commited?

if that's the case, then two questions:

1. `wal_->rollbackToLog(lastLogId_)` means to rolling back entries that haven't been commited yet?
2. is it possible that the entries not commited have been replicated among the majority of the raft peers?

critical27: It is hard to clarify, in short version:
    * As for leader, `lastLogId_` is `commitIndex` in paper
    * As for follower, it is not

So rollback means to rolling back entries that haven't been committed yet. And yes, the logs which may have been replicated to majority. What will happen is that, no matter we rollback or not, we will report error to client. So whether the log will be applied, depends on new leader.

panda-sheep (approved):
Good job!

src/kvstore/raftex/RaftPart.cpp
kikimo: I know why we should roll back here but still kind of warry about the potential impact of this part though I cann't tell what the problem is. I'll do more test.

kikimo (approved):
LGTM, I'll do more test together with the following raft pr.

wenhaocs (approved):

src/kvstore/raftex/RaftPart.cpp
```diff
@@ -852,6 +853,7 @@ void RaftPart::processAppendLogResponses(const AppendLogResponses& resps,
       std::lock_guard<std::mutex> g(raftLock_);
       res = canAppendLogs(currTerm);
       if (res != AppendLogResult::SUCCEEDED) {
+        wal_->rollbackToLog(lastLogId_);
```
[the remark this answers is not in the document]
critical27: Yeah, this will make leader fail to write any wal. Bacause new logs's first Id will be (lastLogId_ + 1), but (wal_->lastLogId +1 != lastLogId_ + 1)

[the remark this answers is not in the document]
panda-sheep: Get it，thx

PR #3429, opened 2021-12-08
fix crash when the expression exceed the depth

### LINKED ISSUE #3393: Limit the depth of the expression, report an error when it exceeds but not crash

heroicNeZha:
#### What type of PR is this?
- [x] bug
- [ ] feature
- [ ] enhancement

#### What does this PR do?
fix crash When the expression's depth exceeds the maximum limit
（The depth is related to the stack size. This pr limited 815 is calculated based on the stack size of the development machine）

fixed #3393 
  
#### Checklist：

#### Release notes：


src/graph/visitor/DeduceTypeVisitor.cpp
```diff
@@ -94,6 +94,16 @@ static const std::unordered_map<Value::Type, Value> kConstantValues = {
   }                                                                  \
   type_ = detectVal.type()
 
+#define CHECK_DEPTH(depth)                                 \
```
Shylock-Hg, line 97: Better to replace by method.

Shylock-Hg:
And maybe not all expression will be checked by `DeduceTypeVisitor`.

src/graph/visitor/DeduceTypeVisitor.cpp
heroicNeZha: Sorry I am new in c++,  I wanna return in the macro and don't know how to replace it with a method

Shylock-Hg: As bellow:
```c++
checkDepth();
if (status_.ok() == false) {
    return;
}
```

heroicNeZha: thanks！

heroicNeZha:
> And maybe not all expression will be checked by `DeduceTypeVisitor`.

The `DeduceTypeVisitor` extends pure virtual class `ExprVisitor` which seems contains all expressions

And I didn't check non-recursive expressions in `DeduceTypeVisitor`

src/graph/visitor/DeduceTypeVisitor.cpp
```diff
@@ -116,6 +116,10 @@ void DeduceTypeVisitor::visit(ConstantExpression *expr) {
 }
 
 void DeduceTypeVisitor::visit(UnaryExpression *expr) {
+  checkDepth();
+  if (!status_.ok()) {
+    return;
```
yixinglu, line 121: should always recover depth when exit current function:
```c++
checkDepth();
SCOPED_EXIT {
  recoverDepth();
}
```

However I also don't think we should do this thing in this visitor.

heroicNeZha: ok I will create a new visitor

critical27:
Generally LGTM

src/graph/visitor/CheckDepthVisitor.cpp
```diff
@@ -0,0 +1,251 @@
+/* Copyright (c) 2020 vesoft inc. All rights reserved.
```
Aiee, line 1: 2021

src/graph/visitor/CheckDepthVisitor.h
```diff
@@ -0,0 +1,99 @@
+/* Copyright (c) 2020 vesoft inc. All rights reserved.
```
Aiee, line 1: ditto

tests/tck/features/expression/Depth.feature
```diff
@@ -0,0 +1,160 @@
+# Copyright (c) 2021 vesoft inc. All rights reserved.
+#
+# This source code is licensed under Apache 2.0 License.
+@li
```
Aiee, line 4: delete testing mark

src/graph/visitor/CheckDepthVisitor.h
```diff
@@ -0,0 +1,99 @@
...
+ *
+ * This source code is licensed under Apache 2.0 License.
+ */
+
+#ifndef GRAPH_VISITOR_CHECKDEPTHVISITOR_H_
+#define GRAPH_VISITOR_CHECKDEPTHVISITOR_H_
```
Aiee, line 7: use `#pragma once`

src/graph/validator/FindPathValidator.cpp
```diff
@@ -46,6 +46,7 @@ Status FindPathValidator::validateWhere(WhereClause* where) {
   }
   where->setFilter(ExpressionUtils::rewriteLabelAttr2EdgeProp(expr));
   auto filter = where->filter();
+  NG_RETURN_IF_ERROR(checkExprDepth(filter));
 
   auto typeStatus = deduceExprType(filter);
```
yixinglu, line 51: It seems that you only need to do this `checkExprDepth()` inside `deduceExprType()`.  So you don't have to add this check everywhere.

heroicNeZha: done

src/graph/visitor/CheckDepthVisitor.cpp
```diff
@@ -0,0 +1,251 @@
...
+  const_cast<LabelExpression *>(expr->left())->accept(this);
+  if (!ok()) return;
+  const_cast<ConstantExpression *>(expr->right())->accept(this);
+}
+
+void CheckDepthVisitor::visit(FunctionCallExpression *expr) {
+  checkDepth();
+  SCOPE_EXIT { recoverDepth(); };
+  if (!ok()) return;
+  for (auto &arg : expr->args()->args()) {
```
jiayuehua, line 108: consider use any_of

src/graph/visitor/CheckDepthVisitor.cpp
```diff
@@ -0,0 +1,251 @@
...
+  if (!ok()) return;
+  expr->left()->accept(this);
+  if (!ok()) return;
+  expr->right()->accept(this);
+}
+
+void CheckDepthVisitor::visit(SubscriptExpression *expr) {
+  checkDepth();
+  SCOPE_EXIT { recoverDepth(); };
+  if (!ok()) return;
```
jiayuehua, line 69: consider refactor to two function that return bool and use: fa() and fb();

heroicNeZha: this pr is closed

Issue #3210, opened 2021-10-26
Variable not referred in from clause


czpmango:
```
(czp@nebula) [nba]> $var = GO FROM "Tim Duncan" OVER * yield like._dst as dst;go from "Tim Duncan" over like yield $var.dst

[ERROR (-1009)]: SemanticError: A variable must be referred in FROM before used in WHERE or YIELD

(czp@nebula) [nba]> $var=go from "Tim Duncan" over * yield like._dst as dst;$var1=go from $var.dst over like yield like._dst as dst1;go from $var1.dst1 over like yield $var.dst
+----------+
| $var.dst |
+----------+
| __NULL__ |
+----------+
| __NULL__ |
+----------+
| __NULL__ |
+----------+
| __NULL__ |
+----------+
| __NULL__ |
+----------+
| __NULL__ |
+----------+
| __NULL__ |
+----------+
Got 7 rows (time spent 23942/25187 us)
```

`$var.dst` never be referred in from clause of the third sequential sentence but used in yield clause.

czpmango:
Duplicated issue, fixed in [https://github.com/vesoft-inc/nebula/pull/3311](https://github.com/vesoft-inc/nebula/pull/3311)

PR #3430, opened 2021-12-08
fix inconsistent variable in yield and from clause

czpmango:
#### What type of PR is this?
- [x] bug
- [ ] feature
- [ ] enhancement

#### What does this PR do?
Fix inconsistent variable in yield and from clause.

[https://github.com/vesoft-inc/nebula/issues/3210](https://github.com/vesoft-inc/nebula/issues/3210)

#### Checklist：
- [x] Need to cherry-pick （If need to cherry-pick to some branches, please label the destination version(s).)

#### Release notes：
                       `

### DIFF
--- a/src/graph/validator/GoValidator.cpp
+++ b/src/graph/validator/GoValidator.cpp
@@ -36,9 +36,23 @@ Status GoValidator::validateImpl() {
     return Status::SemanticError("$- must be referred in FROM before used in WHERE or YIELD");
   }
 
-  if (!exprProps.varProps().empty() && goCtx_->from.fromType != kVariable) {
-    return Status::SemanticError(
-        "A variable must be referred in FROM before used in WHERE or YIELD");
+  if (!exprProps.varProps().empty()) {
+    if (goCtx_->from.fromType != kVariable) {
+      return Status::SemanticError(
+          "A variable must be referred in FROM before used in WHERE or YIELD");
+    }
+    auto varPropsMap = exprProps.varProps();
+    std::vector<std::string> keys;
+    for (const auto& [k, v] : varPropsMap) {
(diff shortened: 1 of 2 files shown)


Shylock-Hg (approved):

src/graph/validator/GoValidator.cpp
```diff
@@ -36,9 +36,23 @@ Status GoValidator::validateImpl() {
...
+    }
+    auto varPropsMap = exprProps.varProps();
+    std::vector<std::string> keys;
+    for (const auto& [k, v] : varPropsMap) {
+      keys.emplace_back(k);
+    }
+    if (keys.size() > 1) {
+      return Status::SemanticError("Multiple variable property is not supported in WHERE or YIELD");
+    }
+    if (keys.front() != goCtx_->from.userDefinedVarName) {
```
nevermore3, line 52: This condition may never be reached

czpmango: Why?

nevermore3: An error will be reported in the validatorFrom. when keys.front() != userdefinedVarName,

czpmango: I can debug this piece of code use this statement:
```
      $A = GO FROM 'Tim Duncan' OVER like YIELD like._dst AS dst;
       $B = GO FROM $A.dst OVER like YIELD like._dst AS dst;
        GO FROM $A.dst over like YIELD like._dst AS dst, $B.dst
```

nevermore3: my fault, deduceProp do not work on from clause

nevermore3 (approved):

PR #3435, opened 2021-12-08
[Raft] refactor processAppendLogRequest

critical27:
Change notes:

1. Main change is in  `processAppendLogRequest`

    A brief flow as follows (not exactly same in code):
    processAppendLogRequest

    To achieve this goal, some others need to be modified:
    * In AppendLogRequest, we need to carry each log's term (which is missing previously), so as a proper iterator.
    * For similarity, we need to return `committedLogId` and `committedLogTerm` in resp. But committedLogTerm is missing previously, so I add one.

2. We need to make sure `lastLogId_ == wal_->lastLogId()` in almost every case (except leader is replicating), we did this by calling `wal_->rollbackToLog`, since we handle `AppendLog` correctly, we don't need to rollback anymore, just set `lastLogId_ = wal_->lastLogId()` is enough.
3. All request and resp should carry the term, step down as follower when a higher term found, otherwise we may failed to elect a leader.
4. micor changes: make `RaftPart::cleanup` atomic

### DIFF
--- a/src/interface/raftex.thrift
+++ b/src/interface/raftex.thrift
@@ -46,6 +46,7 @@ enum ErrorCode {
     E_PERSIST_SNAPSHOT_FAILED = -13;
     E_RPC_EXCEPTION = -14;              // An thrift internal exception was thrown
     E_NO_WAL_FOUND = -15;
+    E_APPLY_FAIL = -16;
 }
 
 typedef i64 (cpp.type = "nebula::ClusterID") ClusterID
--- a/src/interface/raftex.thrift
+++ b/src/interface/raftex.thrift
@@ -47,6 +47,7 @@ enum ErrorCode {
     E_RPC_EXCEPTION = -14;              // An thrift internal exception was thrown
     E_NO_WAL_FOUND = -15;
     E_APPLY_FAIL = -16;
+    E_HOST_PAUSED = -17;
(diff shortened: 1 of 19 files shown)


critical27:
Ready to review

src/interface/raftex.thrift
```diff
@@ -46,6 +46,8 @@ enum ErrorCode {
     E_PERSIST_SNAPSHOT_FAILED = -13;
     E_RPC_EXCEPTION = -14;              // An thrift internal exception was thrown
     E_NO_WAL_FOUND = -15;
+    E_APPLY_FAIL = -16;
```
SuperYoko, line 49: Not used, maybe you miss it or will use later?

critical27: Ah, it is used in `RaftPart.cpp`

kikimo (dismissed):
LGTM

wenhaocs (dismissed):
What's the general rule of setting rpc timeout?

critical27:
> What's the general rule of setting rpc timeout?

Use the same raft rpc timeout is ok, previously there is a bug, so election use an another timeout.

SuperYoko (approved):

kikimo (approved):
LGTM

Issue #3434, opened 2021-12-08
Storage hang at exit


kikimo:
**Please check the FAQ documentation before raising an issue**

Storage may hang forever at exit, this happends regularly during our test when you kill a storage leader instance while inserting data. The probelm might be caused by a dead lock generated by the signal handler in nebula, **we think it might also exists in metad and graphd since they share similar signal handler processing mechanism**. When this happends, we can always see a similar stack:

```txt
(rr) bt
#0  0x0000000070000002 in syscall_traced ()
#1  0x0000250910a18725 in _raw_syscall () at /root/src/rr/src/preload/raw_syscall.S:120
#2  0x0000250910a144ff in traced_raw_syscall (call=call@entry=0x7f3c1c8fd970) at /root/src/rr/src/preload/syscallbuf.c:278
#3  0x0000250910a15f98 in sys_statfs (call=<optimized out>) at /root/src/rr/src/preload/syscallbuf.c:3148
#4  syscall_hook_internal (call=0x7f3c1c8fd970) at /root/src/rr/src/preload/syscallbuf.c:3415
#5  syscall_hook (call=0x7f3c1c8fd970) at /root/src/rr/src/preload/syscallbuf.c:3454
#6  0x0000250910a14340 in _syscall_hook_trampoline () at /root/src/rr/src/preload/syscall_hook.S:313
#7  0x0000250910a1439f in __morestack () at /root/src/rr/src/preload/syscall_hook.S:458
#8  0x0000250910a143bb in _syscall_hook_trampoline_48_3d_00_f0_ff_ff () at /root/src/rr/src/preload/syscall_hook.S:477
#9  0x00002509109f537c in futex_wait_cancelable (private=<optimized out>, expected=0, futex_word=0x250910833a2c) at ../sysdeps/nptl/futex-internal.h:183
#10 __pthread_cond_wait_common (abstime=0x0, clockid=0, mutex=0x250910833930, cond=0x250910833a00) at pthread_cond_wait.c:508
#11 __pthread_cond_wait (cond=0x250910833a00, mutex=0x250910833930) at pthread_cond_wait.c:638
#12 0x0000000004ed23e0 in std::condition_variable::wait(std::unique_lock<std::mutex>&) ()
#13 0x00000000046d867b in apache::thrift::concurrency::ThreadManager::Impl::removeWorkerImpl(std::unique_lock<std::mutex>&, unsigned long, bool) ()
[...]

PR #3437, opened 2021-12-08
simplify storage SIGINT signal handler

liuyu85cn:
#### What type of PR is this?
- [X] bug
- [ ] feature
- [ ] enhancement

#### What does this PR do?

In short: 

1. let's say we(storage server) have a thread running some task.

2. user send kill SIGINT signal, linux schedule the signal handler at the thread we talked about.

3. now, we interrupt the previous task, executing the signal handler, 
    it calls storage server->stop()
    trying to clear everything storage server has(also include this thread!!!!). 

4. Then we stuck.

[Storage hang at exit #3434](https://github.com/vesoft-inc/nebula/issues/3434)

### DIFF
--- a/src/daemons/StorageDaemon.cpp
+++ b/src/daemons/StorageDaemon.cpp
@@ -180,7 +180,7 @@ void signalHandler(int sig) {
     case SIGTERM:
       FLOG_INFO("Signal %d(%s) received, stopping this server", sig, ::strsignal(sig));
       if (gStorageServer) {
-        gStorageServer->stop();
+        gStorageServer->notifyStop();
       }
       break;
     default:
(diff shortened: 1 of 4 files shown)


src/daemons/StorageDaemon.cpp
```diff
@@ -180,7 +180,7 @@ void signalHandler(int sig) {
     case SIGTERM:
       FLOG_INFO("Signal %d(%s) received, stopping this server", sig, ::strsignal(sig));
       if (gStorageServer) {
-        gStorageServer->stop();
+        gStorageServer->notifyStop();
```
kikimo, line 183: This fix looks good. My concern, don't forget to fix metad too!

yixinglu (approved):
LGTM

kikimo (approved):
LGTM

critical27 (approved):
Good job

PR #3438, opened 2021-12-08
Fix the latency type in the executionResponse

Aiee:
#### What type of PR is this?
- [x] bug
- [ ] feature
- [ ] enhancement

#### What does this PR do?
Fix the latency type in the `executionResponse`. The inconsistent datatype of latency between the definition in `thrift` file and `Response.h` causes client transport `invalid data length`error.

This PR https://github.com/vesoft-inc/nebula/pull/2858 only modified the `latency` in the `thrift` file but left other definitions behind.
  
#### Checklist：

#### Release notes：

### DIFF
--- a/src/common/graph/ExecutionResponseOps-inl.h
+++ b/src/common/graph/ExecutionResponseOps-inl.h
@@ -34,10 +34,10 @@ struct TccStructTraits<::nebula::ExecutionResponse> {
     if (false) {
     } else if (_fname == "error_code") {
       fid = 1;
-      _ftype = apache::thrift::protocol::T_I64;
+      _ftype = apache::thrift::protocol::T_I32;
     } else if (_fname == "latency_in_us") {
       fid = 2;
-      _ftype = apache::thrift::protocol::T_I32;
+      _ftype = apache::thrift::protocol::T_I64;
     } else if (_fname == "data") {
       fid = 3;
       _ftype = apache::thrift::protocol::T_STRUCT;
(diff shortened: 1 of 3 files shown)


yixinglu (dismissed):

src/common/graph/ExecutionResponseOps-inl.h
```diff
@@ -34,7 +34,7 @@ struct TccStructTraits<::nebula::ExecutionResponse> {
     if (false) {
     } else if (_fname == "error_code") {
       fid = 1;
-      _ftype = apache::thrift::protocol::T_I32;
+      _ftype = apache::thrift::protocol::T_I64;
```
Shylock-Hg, line 37: error_code?

src/common/graph/ExecutionResponseOps-inl.h
```diff
@@ -34,7 +34,7 @@ struct TccStructTraits<::nebula::ExecutionResponse> {
...
     } else if (_fname == "latency_in_us") {
       fid = 2;
       _ftype = apache::thrift::protocol::T_I32;
```
HarrisChu, line 40: should change it? not error code.

src/common/graph/ExecutionResponseOps-inl.h
Aiee: Good catch

src/common/graph/ExecutionResponseOps-inl.h
Aiee: Fixed.

yixinglu (approved):

czpmango (approved):
Good job!

PR #3460, opened 2021-12-14
Allow users to choose linker like lld to make build faster.

jiayuehua:
#### What type of PR is this?
- [ ] bug
- [* ] feature
- [ ] enhancement

#### What does this PR do?
user can chose linker like lld to make build more faster.

none
  
If ci can install lld before build in the future, then can make lld default linker. Now still use bfd as default.

#### Additional context:
none

#### Checklist：
- [* ] Documentation affected （Please add the label if documentation needs to be modified.)

user in the future need to install lld first to link faster.
#### Release notes：
support lld linker and make build a litter faster


Sophie-Xie:
[* ] Documentation affected

--> Does this PR affect the document?

yixinglu (commented):
I think you shouldn't change the linker config, just by passing the `NEBULA_USE_LINKER ` into cmake in configuring stage.

please see the .github/workflows/pull_request.yml content

jiayuehua:
> I think you shouldn't change the linker config, just by passing the `NEBULA_USE_LINKER ` into cmake in configuring stage.
> 
> please see the .github/workflows/pull_request.yml content
No, the linker config is explicit telling what kind of linker is supported. If set it to CI, Many developer will not notice it, Also Then Can Set NEBULA_USE_LINKER to wrong value like 'notvalidld'. Set it in linkerconfig file developer can now what kind of linker supported.

yixinglu (approved):

yixinglu:
> > I think you shouldn't change the linker config, just by passing the `NEBULA_USE_LINKER ` into cmake in configuring stage.
> > please see the .github/workflows/pull_request.yml content
> 
> No, the linker config is explicit telling what kind of linker is supported. If set it to CI, Many developer will not notice it, Also Then Can Set NEBULA_USE_LINKER to wrong value like 'notvalidld'. Set it in linkerconfig file developer can know what kind of linker supported.

make sense. And only thing needing to do is just to verify the compiling in multi-platform.

Aiee (approved):

PR #3489, opened 2021-12-16
upgradeV2

panda-sheep:
#### What type of PR is this?
- [ ] bug
- [X] feature
- [ ] enhancement

#### What does this PR do?
dst_part_file is a json format file, the file name is as follows:

image

#### Checklist:

#### Release notes:

### DIFF
--- a/src/tools/db-upgrade/DbUpgraderTool.cpp
+++ b/src/tools/db-upgrade/DbUpgraderTool.cpp
@@ -392,7 +392,7 @@ int main(int argc, char* argv[]) {
   LOG(INFO) << "================== Prepare phase end ==================";
 
   // Upgrade data
-  LOG(INFO) << "================== Upgrade phase bengi ==================";
+  LOG(INFO) << "================== Upgrade phase bengin ==================";
 
   // The data path in storage conf is generally one, not too many.
   // So there is no need to control the number of threads here.
--- a/src/tools/db-upgrade/DbUpgraderTool.cpp
+++ b/src/tools/db-upgrade/DbUpgraderTool.cpp
@@ -3,8 +3,12 @@
  * This source code is licensed under Apache 2.0 License.
  */
 
+#include <folly/FileUtil.h>
+#include <folly/json.h>
+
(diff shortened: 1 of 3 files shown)


src/tools/db-upgrade/DbUpgraderTool.cpp
```diff
@@ -121,6 +233,32 @@ int main(int argc, char* argv[]) {
...
+  options.skipConfig_ = true;
+  auto metaClient =
+      std::make_unique<nebula::meta::MetaClient>(ioExecutor, std::move(addrs.value()), options);
+  CHECK_NOTNULL(metaClient);
+  if (!metaClient->waitForMetadReady(1)) {
+    LOG(ERROR) << "Meta is not ready: " << FLAGS_upgrade_meta_server;
+    return EXIT_FAILURE;
+  }
+
+  if (!FLAGS_auto_gen_part_schema) {
```
critical27, line 253: 👍

src/tools/db-upgrade/DbUpgraderTool.cpp
```diff
@@ -67,28 +74,133 @@ void printHelp() {
...
+    return false;
+  }
+
+  auto spaceVidLenRet = metaClient->getSpaceVidLen(spaceId);
+  if (!spaceVidLenRet.ok()) {
+    LOG(ERROR) << "Get space vid len failed, space id " << spaceId;
+    return false;
+  }
+  auto spaceVidLen = spaceVidLenRet.value();
+  std::unique_ptr<nebula::kvstore::RocksEngine> writeEngine(
```
critical27, line 122: I guess the `spaceVidLen` is not actually used? Maybe you could pass a fake one, so we don't need to pass the meta client

panda-sheep: Yes, it seems that it can only be used when the prefix is used.

panda-sheep:
Submitted at https://github.com/vesoft-inc/nebula-ent/pull/381, so close this

Issue #3475, opened 2021-12-15
storaged always sends heartbeat to metad follower if not registered


HarrisChu:
3 storaged and 3 metad.

1. start all the services and do not `add hosts`
2. storaged sends heartbeat to follower, and then return `MetaClient.cpp:70] Heartbeat failed, status:Machine not existed!`
3. storaged sends heartbeat to follower for ever.
4. `add host`

expected result:
1. storaged register succussfully.

actual result:
1. sends to follower for ever and cannot register successfully.

logs:

meta leader:
```
I1215 10:26:24.830297 30602 HBProcessor.cpp:33] Receive heartbeat from "192.168.8.63":9779, role = STORAGE
E1215 10:26:24.830343 30602 HBProcessor.cpp:38] Machine "192.168.8.63":9779 is not registed
I1215 10:26:25.358093 30602 HBProcessor.cpp:33] Receive heartbeat from "192.168.8.62":9779, role = STORAGE
E1215 10:26:25.358165 30602 HBProcessor.cpp:38] Machine "192.168.8.62":9779 is not registed
I1215 10:26:25.643523 30602 HBProcessor.cpp:33] Receive heartbeat from "192.168.8.63":9669, role = GRAPH
I1215 10:26:25.859187 30602 HBProcessor.cpp:33] Receive heartbeat from "192.168.8.61":9669, role = GRAPH
I1215 10:26:27.900756 30602 HBProcessor.cpp:33] Receive heartbeat from "192.168.8.62":9669, role = GRAPH
I1215 10:26:34.832269 30602 HBProcessor.cpp:33] Receive heartbeat from "192.168.8.63":9779, role = STORAGE
E1215 10:26:34.832338 30602 HBProcessor.cpp:38] Machine "192.168.8.63":9779 is not registed
I1215 10:26:35.359666 30602 HBProcessor.cpp:33] Receive heartbeat from "192.168.8.62":9779, role = STORAGE
E1215 10:26:35.359714 30602 HBProcessor.cpp:38] Machine "192.168.8.62":9779 is not registed
I1215 10:26:35.661341 30602 HBProcessor.cpp:33] Receive heartbeat from "192.168.8.63":9669, role = GRAPH
I1215 10:26:35.876652 30602 HBProcessor.cpp:33] Receive heartbeat from "192.168.8.61":9669, role = GRAPH
[...]

PR #3501, opened 2021-12-20
fix add hosts

HarrisChu:
#### What type of PR is this?
- [x] bug
- [ ] feature
- [ ] enhancement

#### What does this PR do?

if  the storage does not register to the cluster, it would return `E_MACHINE_NOT_FOUND`, and would always send heartbeat to the follower.
a little ugly

1. if run `add host` with the storage, it will be ok after updateLeader.
2. if not, it would always update leader, and send heartbeat to meta randomly.


HarrisChu:
fix https://github.com/vesoft-inc/nebula/issues/3475

critical27 (approved):

czpmango (approved):

yixinglu (approved):

PR #3504, opened 2021-12-20
Push aggregate in LookupIndex

Nivras:
#### What type of PR is this?
- [ ] bug
- [x] feature
- [ ] enhancement

#### What does this PR do?
LookupIndexRequest can receive the statProps info, and do aggregate in storage.

This PR add a new node(IndexAggregateNode) for LookupIndexPlan, it will check and append the columns which in statProp but not in returnColumns, and append it in returnColumns. 

IndexProjectNode before this PR will return the row in the order of returnColumns, after this PR, it will return returnColumns + statProps(which the statProp not in returnColumns). and IndexAggregateNode will truncate the Row, only real returnColumns append into resultDataSet, and calculate the aggregate expression and return the result.

#### Checklist:

#### Release notes:


src/storage/exec/IndexAggregateNode.cpp
```diff
@@ -0,0 +1,125 @@
...
+
+IndexAggregateNode::IndexAggregateNode(
+    RuntimeContext* context,
+    const std::vector<std::pair<std::string, cpp2::StatType>>& statInfos,
+    size_t returnColumnsCount)
+    : IndexNode(context, "IndexAggregateNode"),
+      statInfos_(statInfos),
+      returnColumnsCount_(returnColumnsCount) {}
+
+nebula::cpp2::ErrorCode IndexAggregateNode::init(InitContext& ctx) {
```
liwenhui-soul, line 25: will ctx change? input parameters should usually be values or const references, and input/output parameters should usually be non-const pointers.

src/storage/exec/IndexAggregateNode.h
```diff
@@ -0,0 +1,49 @@
...
+class IndexAggregateNode : public IndexNode {
+ public:
+  IndexAggregateNode(const IndexAggregateNode& node);
+  explicit IndexAggregateNode(RuntimeContext* context,
+                              const std::vector<std::pair<std::string, cpp2::StatType>>& statInfos,
+                              size_t returnColumnsCount);
+
+  nebula::cpp2::ErrorCode init(InitContext& ctx) override;
+  void initStatValue();
+  void addStatValue(const Value& value, ColumnStat& stat);
```
liwenhui-soul, line 33: same as init()

src/storage/BaseProcessor.h
```diff
@@ -101,6 +101,9 @@ class BaseProcessor {
 
   void handleAsync(GraphSpaceID spaceId, PartitionID partId, nebula::cpp2::ErrorCode code);
 
+  nebula::cpp2::ErrorCode checkStatType(const meta::SchemaProviderIf::Field* field,
```
liwenhui-soul, line 104: an input parameter should be const references.
please consult google style https://google.github.io/styleguide/cppguide.html#Inputs_and_Outputs

src/interface/storage.thrift
```diff
@@ -495,6 +495,7 @@ struct LookupIndexResp {
     // Each column represents one property. the column name is in the form of "tag_name.prop_alias"
     // or "edge_type_name.prop_alias" in the same order which specified in return_columns of request
     2: optional common.DataSet          data,
+    3: optional common.DataSet          stat_data,
```
critical27, line 498: Base on implementation, stat_data only has one row, each column in the row is one stat, right?

Maybe add some brief description in thrift file

src/storage/exec/IndexAggregateNode.cpp
critical27: I guess this because in other node the ctx may be modified, although not modified in this node.

critical27 (commented):
Goood job~~ Generally LGTM, clean code

src/storage/exec/IndexAggregateNode.cpp
Nivras: this node also modify the context, to give the columns in aggregate but not in returnColumns to IndexScanNode

cangfengzhs: I do not recommend writing this way. If the scalability of the program is considered, expression calculations may be implemented in the projection node later. For example, max(a)-min(a), the subtraction needs to be done in the projection node. If it is not placed in the returnColumn, it will bring additional complexity to the projection node.

src/storage/exec/IndexNode.h
```diff
@@ -69,6 +69,8 @@ struct InitContext {
   std::vector<std::string> returnColumns;
   // The index of name in `returncolumns`
   Map<std::string, size_t> retColMap;
+  // The columns in statColumns
+  Set<std::string> statColumns;
```
cangfengzhs, line 73: I think `statColumns` is redundant. It should be included in `returnColumns`. In my initial design, `returnColumns` contains more than just `property name`, it can be any expression. For example, `max(a)-min(a)`, we need an Aggregate node to calculate max(a) and min(a), and then a Projection is needed to calculate max(a)-min(a) (only subtraction is calculated) At this time, Aggregate should put max(a) and min(a) into `returnColumn` and `retColMap`, and then Projection obtains the values of max(a) and min(a) directly from them.

In this scheme `returnColumns` may require some design. But I think this is the most scalable and least coupled solution. Whether it is `projection {expr}`, `having {expr}`, or `order by {expr}` in the follow-up, the function independence between nodes can be maintained, and the intrusiveness is the lowest.

cangfengzhs (changes requested):

src/storage/exec/IndexNode.h
cangfengzhs: We can remember a TODO first and then merge it first

Nivras: Sounds good! I will add a TODO and fix it in next PR

src/interface/storage.thrift
```diff
@@ -495,6 +495,8 @@ struct LookupIndexResp {
     // Each column represents one property. the column name is in the form of "tag_name.prop_alias"
     // or "edge_type_name.prop_alias" in the same order which specified in return_columns of request
     2: optional common.DataSet          data,
```
pengweisong, line 497: give some brief description?

critical27 (approved):
Generally LGTM

cangfengzhs (approved):
LGTM

critical27 (approved):

cangfengzhs (approved):

PR #3527, opened 2021-12-21
Fix the issue that the query results will still contain the expired e…

luyade:
Fix an issue that the query results still contain the expired edges when use go reversely

#### What type of PR is this?
- [ ] bug

#### Checklist:

#### Release notes:


luyade:
close the PR because the commit use is not right

PR #3533, opened 2021-12-22
Disable breakpad in no-release version

Shylock-Hg:
#### What type of PR is this?
- [ ] bug
- [ ] feature
- [x] enhancement

#### Checklist:

#### Release notes:


jievince:
Why disable it?

Shylock-Hg:
> Why disable it?

Debug build don't need it.

CPWstatic:
Would it be better by providing an specific definition for breakpad? Such as `ENABLE_BREAKPAD`.

Shylock-Hg:
> Would it be better by providing an specific definition for breakpad? Such as `ENABLE_BREAKPAD`.

Ok, done.

package/package.sh
```diff
@@ -104,6 +104,12 @@ EOF
 
 function _build_graph {
     pushd ${build_dir}
+    enable_breakpad=OFF
+    if [ "x"$build_type = "xDebug" ] || [ "x"$build_type = "xRelWithDebInfo" ]; then
```
CPWstatic, line 108: The breakpad relys on the symbol table, so it only works in Debug/RelWithDebInfo mode. We should not modify this shell script, but the workflow of publish in github. Turn on the breakpad in the publishing workflow.

Shylock-Hg: ok.

CPWstatic (dismissed):

CPWstatic (approved):

package/package.sh
```diff
@@ -110,6 +115,12 @@ EOF
 
 function _build_graph {
     pushd ${build_dir}
+    if [ "x"$enable_breakpad == "xON" ]; then
+        if [ "x"$build_type != "xDebug" ] && [ "x"$build_type != "xRelWithDebInfo" ]; then
+            echo 'Breakpad need debug info.'
```
yixinglu, line 120: It's better to check this condition in `BreakpadConfig.cmake` to avoid user to misuse the `CMAKE_BUILD_TYPE` and `ENABLE_BREAKPAD`.

src/daemons/GraphDaemon.cpp
```diff
@@ -33,7 +33,7 @@ using nebula::network::NetworkUtils;
 static void signalHandler(nebula::graph::GraphServer *graphServer, int sig);
 static Status setupSignalHandler(nebula::graph::GraphServer *graphServer);
 static void printHelp(const char *prog);
-#if defined(__x86_64__)
+#if defined(__x86_64__) && defined(ENABLE_BREAKPAD)
```
yixinglu, line 36: Maybe we only need to check the `ENABLE_BREAKPAD` option, and the x86 arch could also be checked in `BreakpadConfig.cmake`

package/package.sh
Shylock-Hg: We can't turn on or turn off breakpad by `cmake_build_type`.

src/daemons/GraphDaemon.cpp
Shylock-Hg: It's not easy to get target arch.

src/daemons/SetupBreakpad.cpp
```diff
@@ -3,7 +3,7 @@
  * This source code is licensed under Apache 2.0 License.
  */
 
-#if defined(__x86_64__)
+#if defined(__x86_64__) && defined(ENABLE_BREAKPAD)
```
yixinglu, line 6: could u miss this `x86` macro check?

cmake/nebula/ThirdPartyConfig.cmake
```diff
@@ -106,7 +106,7 @@ find_package(FLEX REQUIRED)
 find_package(LibLZMA REQUIRED)
 find_package(Fizz REQUIRED)
 find_package(Sodium REQUIRED)
-if (${CMAKE_HOST_SYSTEM_PROCESSOR} MATCHES "x86_64")
+if (${CMAKE_HOST_SYSTEM_PROCESSOR} MATCHES "x86_64" AND ENABLE_BREAKPAD)
     find_package(Breakpad REQUIRED)
 endif()
```
yixinglu, line 111: Could we find this package in `BreakpadConfig.cmake` since we have checked the platform there?

CPWstatic (dismissed):

yixinglu (approved):

CPWstatic (approved):

PR #3542, opened 2021-12-23
simplify graph signal handler

### LINKED ISSUE #3441: graphd might block at exit

Take a look at this issue #3434 , after reviewing GraphDaemon's code, we suspect that it may suffer from the same problem since it call `gServer->stop()` in signal handler(`gServer` is a ThriftServer instance, if the handler is triggered and scheduled to execute in one of the ThriftServer's worker thread, it may block forever):

https://github.com/vesoft-inc/nebula/blob/cd2f4a66bb86b8478f0943a02f58dd16a99f7162/src/daemons/GraphDaemon.cpp#L199-L209

* OS: `uname -a`
* CPU: `lscpu`
* Commit id (e.g. `a3ffc7d8`)

1. Step 1
2. Step 2
3. Step 3

**Expected behavior**

**Additional context**

heroicNeZha:
#### What type of PR is this?
- [x] bug
- [ ] feature
- [ ] enhancement

#### What does this PR do?
closes #3441 
simplify graphd signal handler：Set stop condition instead of waiting all workers stop

In #3437 , it simplify storage SIGINT signal handler
  
#### Checklist:

#### Release notes:


src/daemons/GraphDaemon.cpp
```diff
@@ -166,33 +174,56 @@ int main(int argc, char *argv[]) {
...
+    gServer->setNumAcceptThreads(FLAGS_num_accept_threads);
+    gServer->setListenBacklog(FLAGS_listen_backlog);
+    if (FLAGS_enable_ssl || FLAGS_enable_graph_ssl) {
+      gServer->setSSLConfig(nebula::sslContextConfig());
+    }
+    setupThreadManager();
+    serverStatus_.store(STATUS_RUNNING);
+    FLOG_INFO("Starting nebula-graphd on %s:%d\n", localhost.host.c_str(), localhost.port);
+    try {
+      gServer->serve();  // Blocking wait until shut down via gServer->stop()
```
Shylock-Hg, line 198: I think you just need call `serve` in this thread.

heroicNeZha: When I call `serve()`, the deamon will be blocked here, I don’t know how to ensure that the `stop` call is not a worker thread

src/daemons/GraphDaemon.cpp
```diff
@@ -43,6 +46,11 @@ extern Status setupBreakpad();
 DECLARE_string(flagfile);
 DECLARE_bool(containerized);
 
+enum ServiceStatus { STATUS_UNINITIALIZED = 0, STATUS_RUNNING = 1, STATUS_STOPPED = 2 };
+std::atomic<ServiceStatus> serverStatus_{STATUS_UNINITIALIZED};
+std::mutex muStop_;
+std::condition_variable cvStop_;
```
yixinglu, line 52: Prefer to provide a class to handle this signal process

heroicNeZha: done

src/graph/service/GraphServer.cpp
```diff
@@ -24,7 +24,7 @@ bool GraphServer::start() {
   int numThreads =
       FLAGS_num_worker_threads > 0 ? FLAGS_num_worker_threads : gServer_->getNumIOWorkerThreads();
   std::shared_ptr<apache::thrift::concurrency::ThreadManager> threadManager(
-      PriorityThreadManager::newPriorityThreadManager(numThreads, false /*stats*/));
```
Shylock-Hg: Why change this?

heroicNeZha: After updating third party 3.0，newPriorityThreadManager remove second argument

src/graph/service/GraphServer.h
```diff
@@ -0,0 +1,45 @@
...
+
+ private:
+  HostAddr localHost_;
+
+  std::shared_ptr<folly::IOThreadPoolExecutor> ioThreadPool_;
+  std::shared_ptr<apache::thrift::concurrency::ThreadManager> workers_;
+  std::unique_ptr<apache::thrift::ThriftServer> gServer_;
+  std::unique_ptr<std::thread> graphThread_;
+
+  enum ServiceStatus { STATUS_UNINITIALIZED = 0, STATUS_RUNNING = 1, STATUS_STOPPED = 2 };
```
yixinglu, line 39: enum ServiceStatus : uint8_t {}

src/graph/service/GraphServer.h
```diff
@@ -0,0 +1,45 @@
...
+  void notifyStop();
+
+  void waitUntilStop();
+
+ private:
+  HostAddr localHost_;
+
+  std::shared_ptr<folly::IOThreadPoolExecutor> ioThreadPool_;
+  std::shared_ptr<apache::thrift::concurrency::ThreadManager> workers_;
+  std::unique_ptr<apache::thrift::ThriftServer> gServer_;
```
yixinglu, line 36: thriftServer_

src/graph/service/GraphServer.cpp
```diff
@@ -0,0 +1,106 @@
...
+#include "GraphServer.h"
+
+#include <memory>
+
+#include "graph/service/GraphFlags.h"
+#include "graph/service/GraphService.h"
+namespace nebula {
+namespace graph {
+
+GraphServer::GraphServer(HostAddr localHost) : localHost_(localHost) {}
```
yixinglu, line 14: std::move(localHost)

src/daemons/GraphDaemon.cpp
```diff
@@ -153,47 +152,23 @@ int main(int argc, char *argv[]) {
...
-  FLOG_INFO("Starting nebula-graphd on %s:%d\n", localhost.host.c_str(), localhost.port);
-  try {
-    gServer->serve();  // Blocking wait until shut down via gServer->stop()
-  } catch (const std::exception &e) {
-    FLOG_ERROR("Exception thrown while starting the RPC server: %s", e.what());
+  gServer = std::make_unique<nebula::graph::GraphServer>(localhost);
+
+  if (!gServer->start()) {
+    LOG(ERROR) << "The graph server start failed";
+    gServer->stop();
```
yixinglu, line 166: Need not to call stop explicitly

src/daemons/GraphDaemon.cpp
```diff
@@ -29,20 +29,19 @@ using nebula::fs::FileUtils;
...
 static void printHelp(const char *prog);
-static void setupThreadManager();
 #if defined(__x86_64__)
 extern Status setupBreakpad();
 #endif
 
 DECLARE_string(flagfile);
 DECLARE_bool(containerized);
 
+std::unique_ptr<nebula::graph::GraphServer> gServer;
```
yixinglu, line 43: don't forget to update the Standalone daemon introduced in #3310

src/graph/service/GraphServer.cpp
```diff
@@ -0,0 +1,106 @@
...
+    }
+  }
+
+  gServer_->stop();
+
+  graphThread_->join();
+}
+
+void GraphServer::notifyStop() {
+  std::unique_lock<std::mutex> lkStop(muStop_);
```
yixinglu, line 84: Replace this lock with atomic CAS operation.

yixinglu (commented):
A little complicated implementation, could use `init` and `start` function to do this. And need not to spawn a new thread for thrift server. you can talk with me offline.

src/graph/service/GraphServer.cpp
CPWstatic: This argument affects the performance a lot, we must know how it behaves now. And make sure if it has another way to set the value. A benchmark must done for this change.

heroicNeZha: https://github.com/facebook/fbthrift/commit/bfb527c7b254230cf5ea6ceb818c8ce57a3f9a84?diff=split
Through this commit，thrift removed stats, and the function is the same as setting it to false

src/graph/service/GraphServer.cpp
heroicNeZha: if remove this lock , it will be thread unsafe

CPWstatic (dismissed):

src/graph/service/GraphServer.cpp
```diff
@@ -0,0 +1,105 @@
...
+}
+
+void GraphServer::stop() {
+  if (serverStatus_.load() == ServiceStatus::STATUS_STOPPED) {
+    LOG(INFO) << "The graph server has been stopped";
+    return;
+  }
+
+  ServiceStatus serverExpected = ServiceStatus::STATUS_RUNNING;
+  serverStatus_.compare_exchange_strong(serverExpected, STATUS_STOPPED);
```
Shylock-Hg, line 97: Don't wait it to running?

yixinglu (approved):

src/graph/service/GraphServer.cpp
heroicNeZha: it wait in thriftServer.stop

Shylock-Hg (approved):

PR #3545, opened 2021-12-23
fix crash when the expression exceed the depth

heroicNeZha:
#### What type of PR is this?
- [x] bug
- [ ] feature
- [ ] enhancement

#### What does this PR do?
closing  #3393
```
# when the expression exceed the depth
[ERROR (-1004)]: SyntaxError: The above expression is not a valid expression, because its depth exceeds the maximum depth! near `1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+1+'
```
#### Checklist:

#### Release notes:


src/parser/parser.yy
```diff
@@ -555,6 +555,9 @@ expression
     }
     | compound_expression {
         $$ = $1;
+        if(!$$->checkDepth()){
+            throw nebula::GraphParser::syntax_error(@1, "The above expression is not a valid expression, because its depth exceeds the maximum depth!");
```
jievince, line 559: Although the expression exceeds the maximum depth, it's still is a valid expression in theory.  So it's better to adjust the error message.

src/parser/parser.yy
```diff
@@ -566,112 +569,220 @@ expression
             $$ = UnaryExpression::makeNegate(qctx->objPool(), $3);
         }
         scanner.setUnaryMinus(false);
+        if(!$$->checkDepth()){
+            throw nebula::GraphParser::syntax_error(@1, "The above expression is not a valid expression, because its depth exceeds the maximum depth!");
+        }
```
jievince, line 574: I am sure is it better to wrap the check as a macro function. Writing the same check and error messages too many times seems not well.

heroicNeZha: done

PR #3444, opened 2021-12-09
Support zone operations

darionyaphet:
#### What type of PR is this?
- [ ] bug
- [x] feature
- [ ] enhancement

#### Checklist：
- [ ] Documentation affected （Please add the label if documentation needs to be modified.)
- [ ] Incompatible （If it is incompatible, please describe it and add corresponding label.）
- [ ] Need to cherry-pick （If need to cherry-pick to some branches, please label the destination version(s).)
- [ ] Performance impacted: Consumes more CPU/Memory

#### Release notes：
Please confirm whether to reflect in release notes and how to describe:
>                                                                 `

PR #3548, opened 2021-12-23
Support auto registe when storage start up

darionyaphet:
#### What type of PR is this?
- [ ] bug
- [ ] feature
- [ ] enhancement

#### Checklist:

#### Release notes:


critical27 (commented):
Maybe reuse the old parameter name `hosts_whitelist_enabled`

src/meta/processors/admin/HBProcessor.cpp
```diff
@@ -34,7 +36,7 @@ void HBProcessor::process(const cpp2::HBReq& req) {
             << ", role = " << apache::thrift::util::enumNameSafe(role);
 
   if (role == cpp2::HostRole::STORAGE) {
-    if (!ActiveHostsMan::machineRegisted(kvstore_, host)) {
+    if (!FLAGS_auto_registe && !ActiveHostsMan::machineRegisted(kvstore_, host)) {
```
panda-sheep, line 39: I have a problem
For other roles such as storage_listener, do I need to add hosts?

panda-sheep: In my test, other roles are not cpp2::HostRole::STORAGE, add host succeeds, but show does not come out

darionyaphet: https://github.com/vesoft-inc/nebula/pull/3444  have fix it

panda-sheep: I think you should merge that bug in first, otherwise, the entire cluster will be useless.

jievince (approved):

Shylock-Hg (approved):

jievince (approved):

Issue #3373, opened 2021-11-29
All storage instances crashed after 7 hours of pressure test


kikimo:
**Please check the FAQ documentation before raising an issue**

**Describe the bug (__required__)**

A nebula cluster of 3storages + 1graph + 1meta, we keep inserting edge and trigger leader, after running for about 7 hours, all storage instances crash almost at the same time and the all have similar crash stack:

```txt
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib64/libthread_db.so.1".
--Type <RET> for more, q to quit, c to continue without paging--c
Core was generated by `/data/src/wwl/nebula/build/bin/nebula-storaged --flagfile /data/src/wwl/test/et'.
Program terminated with signal SIGSEGV, Segmentation fault.
#0  0x0000000002fb8260 in folly::ThreadLocalPtr<folly::SingletonThreadLocal<nebula::meta::MetaClient::ThreadLocalInfo, folly::detail::DefaultTag, folly::detail::DefaultMake<nebula::meta::MetaClient::ThreadLocalInfo>, void>::Wrapper, void, void>::get (this=0x7fbd204c6ee0) at /data/src/wwl/nebula/build/third-party/install/include/folly/ThreadLocal.h:153
153	    return static_cast<T*>(w.ptr);
[Current thread is 1 (Thread 0x7fbc115ff700 (LWP 3293767))]
(gdb) bt
#0  0x0000000002fb8260 in folly::ThreadLocalPtr<folly::SingletonThreadLocal<nebula::meta::MetaClient::ThreadLocalInfo, folly::detail::DefaultTag, folly::detail::DefaultMake<nebula::meta::MetaClient::ThreadLocalInfo>, void>::Wrapper, void, void>::get (this=0x7fbd204c6ee0)
    at /data/src/wwl/nebula/build/third-party/install/include/folly/ThreadLocal.h:153
#1  0x0000000002f2ffe4 in folly::ThreadLocal<folly::SingletonThreadLocal<nebula::meta::MetaClient::ThreadLocalInfo, folly::detail::DefaultTag, folly::detail::DefaultMake<nebula::meta::MetaClient::ThreadLocalInfo>, void>::Wrapper, void, void>::get (this=0x7fbd204c6ee0)
    at /data/src/wwl/nebula/build/third-party/install/include/folly/ThreadLocal.h:69
[...]

PR #3553, opened 2021-12-23
Issue3373 storage exit crash

cangfengzhs:
#### What type of PR is this?
- [x] bug
- [ ] feature
- [ ] enhancement

#### What does this PR do?
Use RCU replace ThreadLocal in MetaClient

https://github.com/vesoft-inc/nebula/issues/3373
https://github.com/vesoft-inc/nebula/issues/3497
  
At the beginning, we found that storage would crash after running for a long time (a large number of insert edge operations were performed at the same time). At the same time, Storage's memory usage will be very high. So we guess that there is a memory leak after the system OOM. However, it was later discovered that this is not the problem. Even if the Storage does not have OOM, it will crash when it is stopped. All coredump stacks destruct a static thread variable when the thread exits. This is a variable of type folly::SingletonThreadLocal introduced in MetaClient.

At the same time, in another scenario, if compaction is triggered when storage is started, it will crash directly, and the coredump stack and stop will be the same.

After a long time of investigation, we did not find the specific cause of this problem, but we found that this was a problem that only appeared after the introduction of folly::SingletonThreadLocal, so we chose to deprecate folly::SingletonThreadLocal and replace it with RCU it.

After using RCU, there is indeed no crash. I am not sure whether it was really fixed or just because the probability of crash has decreased and I did not find it.

In addition, the performance of using RCU should also be better than the performance of ThreadLocal, because no read-write lock means no blocking

#### Checklist:

#### Release notes:

>

### DIFF
--- a/src/clients/meta/MetaClient.cpp
+++ b/src/clients/meta/MetaClient.cpp
@@ -195,9 +195,11 @@ bool MetaClient::loadUsersAndRoles() {
   // Remove expired users from cache
   auto removeExpiredUser = [&](folly::ConcurrentHashMap<std::string, uint32>& userMap,
                                const std::unordered_set<std::string>& userList) {
-    for (auto& ele : userMap) {
-      if (userList.count(ele.first) == 0) {
-        userMap.erase(ele.first);
+    for (auto iter = userMap.begin(); iter != userMap.end();) {
+      if (!userList.count(iter->first)) {
+        iter = userMap.erase(iter);
+      } else {
+        ++iter;
       }
     }
   };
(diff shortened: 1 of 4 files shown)


critical27 (commented):
Good job.
1. I think `localCacheLock_` is useless now, maybe we could remove it.
2. move `killedPlans_` and `killedPlans_` to the same rcu

jackwener:
LGTM

critical27 (changes requested):
Just for test

critical27 (approved):
A long story... Good job~~ LGTM

CPWstatic (approved):

PR #3568, opened 2021-12-27
fix clang-format bug

jiayuehua:
#### What type of PR is this?
- [x] bug
- [ ] feature
- [ ] enhancement

#### What does this PR do?
clang format should not sort include files

#### Checklist:

#### Release notes:


yixinglu:
which bug ?

PR #3471, opened 2021-12-15
add tck ssl client

HarrisChu:
#### What type of PR is this?
- [ ] bug
- [ ] feature
- [x] enhancement

#### What does this PR do?
use ssl client in tck test 
Close #3433 

should not be merged utils nebula-python support `password`

PR #3570, opened 2021-12-27
Fix/fix

Shylock-Hg:
#### What type of PR is this?
- [ ] bug
- [ ] feature
- [ ] enhancement

#### What does this PR do?
Fit the library ssl configuration upgrading.
Need https://github.com/vesoft-inc/nebula/pull/3471 , seems don't support plaintext in ssl port again.

#### Checklist:

#### Release notes:

### DIFF
--- a/src/common/ssl/SSLConfig.cpp
+++ b/src/common/ssl/SSLConfig.cpp
@@ -18,7 +18,7 @@ namespace nebula {
 std::shared_ptr<wangle::SSLContextConfig> sslContextConfig() {
   auto sslCfg = std::make_shared<wangle::SSLContextConfig>();
   sslCfg->addCertificate(FLAGS_cert_path, FLAGS_key_path, FLAGS_password_path);
-  sslCfg->clientCAFile = FLAGS_ca_path;
+  sslCfg->clientVerification = folly::SSLContext::VerifyClientCertificate::DO_NOT_REQUEST;
   sslCfg->isDefault = true;
   return sslCfg;
 }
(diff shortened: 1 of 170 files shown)


src/common/ssl/SSLConfig.cpp
```diff
@@ -18,6 +18,7 @@ namespace nebula {
 std::shared_ptr<wangle::SSLContextConfig> sslContextConfig() {
   auto sslCfg = std::make_shared<wangle::SSLContextConfig>();
   sslCfg->addCertificate(FLAGS_cert_path, FLAGS_key_path, FLAGS_password_path);
+  sslCfg->clientVerification = folly::SSLContext::VerifyClientCertificate::DO_NOT_REQUEST;
```
Aiee, line 21: Does this mean that the certificate from the client could be omitted?

Aiee (commented):
Could we unify ca-signed and self-signeed as one?

HarrisChu (approved):

src/common/ssl/SSLConfig.cpp
HarrisChu: how about using "always" and failed if verification fails

Shylock-Hg: Not, it just check CA of server (which is new feature after library upgrading), so I disable it to keep same with before.

Aiee (approved):
I still believe we should not distinguish ca-signed and self-signed cert, maybe address this in another PR.

Aiee (approved):

HarrisChu (approved):

Shylock-Hg:
> I still believe we should not distinguish ca-signed and self-signed cert, maybe address this in another PR.

Could you detail this proposal?

PR #3573, opened 2021-12-27
Add constraints on invalid password attempts

Aiee:
#### What type of PR is this?
- [ ] bug
- [x] feature
- [ ] enhancement

#### What does this PR do?
1. Limit login attempts
2. Block the account after a few consecutive login attempts provide an incorrect password 

Close https://github.com/vesoft-inc/nebula/issues/2442
  
Introducing 2 new gflags: 
* `failed_login_attempts` 
how many consecutive incorrect passwords input to a **SINGLE** graph service node cause the account to become locked.

Please note that this flag is functioning on a **SINGLE** graph node, for example, if there are 3 graph nodes and each node has a  `failed_login_attempts` set to 3, this allows a user to input an invalid password for a total of 9 times (3 times per node).

* `password_lock_time` 
how long to lock the account after too many consecutive login attempts provide an incorrect password

By default, both glags have a value of 0, which means there is no limitation on login attempts.
For now, these configs function on all users rather than bound to a specific user.

#### Checklist:

#### Release notes:


src/clients/meta/MetaClient.cpp
```diff
@@ -37,6 +39,13 @@ DEFINE_int32(meta_client_retry_interval_secs, 1, "meta client sleep interval bet
 DEFINE_int32(meta_client_timeout_ms, 60 * 1000, "meta client timeout");
 DEFINE_string(cluster_id_path, "cluster.id", "file path saved clusterId");
 DEFINE_int32(check_plan_killed_frequency, 8, "check plan killed every 1<<n times");
+DEFINE_uint32(failed_login_attempts,
+              5,
+              "how many consecutive incorrect passwords cause the account to become locked");
+DEFINE_uint32(password_lock_time,
```
yixinglu, line 45: password_lock_time_in_secs

src/clients/meta/MetaClient.cpp
```diff
@@ -2334,16 +2380,80 @@ std::vector<cpp2::RoleItem> MetaClient::getRolesByUserFromCache(const std::strin
...
+            lockedSince + FLAGS_password_lock_time * 3600);
+      }
+      LOG(ERROR) << "Invalid password, remaining attempts: " << passwordAttemtRemain;
+      return Status::Error("Invalid password, remaining attempts: %d", passwordAttemtRemain);
+    }
+  }
+
+  // Authentication succeed, check if the account has been locked
+  auto curTimestamp = time::WallClock::fastNowInSec();
+  if (curTimestamp < lockedSince + FLAGS_password_lock_time * 3600) {
```
yixinglu, line 2445: Should check the lock time before password comparison

src/clients/meta/MetaClient.cpp
Aiee: MySQL uses the day as the unit of password_lock_time, and this is because MySQL supports unblocking a user manually. Since we don't have this ability right now, I use hours. But seems using seconds is sufficient for the block time window and will make testing easier. I'll change it in the next commit.

src/clients/meta/MetaClient.cpp
```diff
@@ -2353,16 +2412,73 @@ std::vector<cpp2::RoleItem> MetaClient::getRolesByUserFromCache(const std::strin
...
+  // Check user existence
   auto iter = threadLocalInfo.userPasswordMap_.find(account);
   if (iter == threadLocalInfo.userPasswordMap_.end()) {
-    return false;
+    return Status::Error("User not exist");
+  }
+
+  folly::RWSpinLock::WriteHolder holder(localCacheLock_);
+
+  auto& lockedSince = userLoginLockTime_[account];
```
jievince, line 2430: what if the account does not exist in the map?

Aiee: user existence will be checked in line 2422

```
> ./nebula-console -addr 192.168.8.6 -port 29562 -u ro -p 1
2021/12/29 22:48:57 Fail to create a new session from connection pool, fail to authenticate, error: User not exist
panic: Fail to create a new session from connection pool, fail to authenticate, error: User not exist
```

src/clients/meta/MetaClient.cpp
```diff
@@ -37,6 +39,27 @@ DEFINE_int32(meta_client_retry_interval_secs, 1, "meta client sleep interval bet
 DEFINE_int32(meta_client_timeout_ms, 60 * 1000, "meta client timeout");
 DEFINE_string(cluster_id_path, "cluster.id", "file path saved clusterId");
 DEFINE_int32(check_plan_killed_frequency, 8, "check plan killed every 1<<n times");
+DEFINE_uint32(failed_login_attempts,
+              0,
```
HarrisChu, line 43: what's the meaning for "0"

src/clients/meta/MetaClient.cpp
```diff
@@ -37,6 +39,27 @@ DEFINE_int32(meta_client_retry_interval_secs, 1, "meta client sleep interval bet
...
+              "how many consecutive incorrect passwords input to a SINGLE graph service node cause "
+              "the account to become locked.");
+DEFINE_uint32(
+    password_lock_time_in_secs,
+    0,
```
HarrisChu, line 48: ditto

src/clients/meta/MetaClient.cpp
jievince: See the desc of the PR.

Aiee: 0 means disable this feature. I had comments in src/graph/service/GraphFlags.h

tests/tck/features/admin/Authentication.feature
```diff
@@ -0,0 +1,44 @@
+# Copyright (c) 2021 vesoft inc. All rights reserved.
+#
+# This source code is licensed under Apache 2.0 License.
+Feature: Test Authentication
+
+  Background:
+    Given a nebulacluster with 1 graphd and 1 metad and 1 storaged
```
HarrisChu, line 7: prefer using params, so no need to add `self.graphd_param['failed_login_attempts'] = '5'` in nebula_service.py.

e.g.
```
    Given a nebulacluster with 1 graphd and 1 metad and 1 storaged:
      """
      graphd:failed_login_attempts=5
      """

File: tests/tck/features/admin/Authentication.feature [hunk 7]
```diff
@@ -0,0 +1,44 @@
...
+      Invalid password, remaining attempts: 2
+      """
+    When login "graphd[0]" with "user1" and "wrongPassword" should fail:
+      """
+      Invalid password, remaining attempts: 1
+      """
+    When login "graphd[0]" with "user1" and "wrongPassword" should fail:
+      """
+      5 times consecutive incorrect passwords has been input, user name: user1 has been blocked, try again in 10 seconds
+      """
```
HarrisChu, line 38: could you add a step - login with "user1" and "nebula1"

src/clients/meta/MetaClient.h
```diff
@@ -143,6 +144,10 @@ using IndexStatus = std::tuple<std::string, std::string, std::string>;
 using UserRolesMap = std::unordered_map<std::string, std::vector<cpp2::RoleItem>>;
 // get user password by account
 using UserPasswordMap = std::unordered_map<std::string, std::string>;
+// Mapping of user name and remaining wrong password attempts
+using UserPasswordAttemptsRemain = std::unordered_map<std::string, uint32>;
+// Mapping of user name and the timestamp when the account is locked
+using UserLoginLockTime = std::unordered_map<std::string, uint32>;
```
jievince, line 151: Could these 3 map be merged to one map?
eg.
std::unordered_map<std::string, PassInfo>.

Aiee: I've considered this, but `UserPasswordMap` is used in the `localThreadCache` as a read-only container, and the rest two objects require both read and write. I believe we could keep it like this for now.

src/clients/meta/MetaClient.cpp
```diff
@@ -2353,16 +2412,73 @@ std::vector<cpp2::RoleItem> MetaClient::getRolesByUserFromCache(const std::strin
...
+          "%d times consecutive incorrect passwords has been input, user name: %s has been "
+          "locked, try again in %ld seconds",
+          FLAGS_failed_login_attempts,
+          account.c_str(),
+          remainingLockTime);
+    }
+    // Clear lock state and reset attempts
+    lockedSince = 0;
+    passwordAttemtRemain = FLAGS_failed_login_attempts;
   }
```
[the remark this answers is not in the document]
Aiee: It is processed in line 2454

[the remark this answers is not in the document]
Aiee: if (FLAGS_failed_login_attempts == 0 || FLAGS_password_lock_time_in_secs == 0) the `lockedSince` will always be 0 so this branch won't be hit

tests/tck/features/admin/Authentication.feature
Aiee: Good to know this.

jievince (approved):
LGTM

HarrisChu (dismissed):
LGTM

jievince (approved):

HarrisChu (approved):

PR #3587, opened 2021-12-29
chore: update_issue_templates

wey-gu:
#### What type of PR is this?
- [ ] bug
- [ ] feature
- [ ] enhancement

#### What does this PR do?
polish issue template

https://github.com/vesoft-inc/nebula-community/issues/81
It will look like this:
Screen Shot 2021-12-29 at 4 19 23 PM

  
n/a

#### Additional context/ Design document:
n/a

#### Checklist:

#### Release notes:

n/a


Shylock-Hg:
Emmm... I still prefer to markdown.

wey-gu:
> Emmm... I still prefer to markdown.

In my first view of this web table, I thought that's cool(to have required fields and optional fields and a clean layout).

While after you mentioned your preference of markdown, I am not sure now, as I consider myself as a markdown fun :-), and sometimes when I was raising an issue, in case I was sure "one sentence" would explain needed information, I did remove the templates(and I think that's OK 😛), and this form actually prevent that possiblities.

Let's put it here for discussion with others.

What do you think, please, for those mates on the reviewer list?

Thanks.

Sophie-Xie:
I thought that's cool too, but it depends on developers' opinions mainly.

Shylock-Hg:
> I thought that's cool too, but it depends on developers' opinions mainly.

Yes, in my mind, template in Markdown format is the standard in now, change to new one is not friendly to developer and community.

wey-gu:
> 

Sure, just for one thing that with this PR's change, all fields are still in ascii and markdown syntax(apart from the log field in the new bug report template), the only differences are:

- Fields are separated into different parts(and will eventually be rendered as one single markdown after submitted).
- Certain fields are mandatory now.

Shylock-Hg:
> > Yes, in my mind, template in Markdown format is the standard in now, change to new one is not friendly to developer and community.
> 
> Sure, just for one thing that with this PR's change, all fields are still in ascii and markdown syntax(apart from the log field in the new bug report template), the only differences are:
> 
> * Fields are separated into different parts(and will eventually be rendered as one single markdown after submitted).
> * Certain fields are mandatory now.

Yes, Mardown wrapped by YAML is not Markdown.:)

wey-gu:
> > > Yes, in my mind, template in Markdown format is the standard in now, change to new one is not friendly to developer and community.
> > 
> > 
> > Sure, just for one thing that with this PR's change, all fields are still in ascii and markdown syntax(apart from the log field in the new bug report template), the only differences are:
> > 
> > * Fields are separated into different parts(and will eventually be rendered as one single markdown after submitted).
> > * Certain fields are mandatory now.
> 
> Yes, Mardown wrapped by YAML is not Markdown.:)

Ha, I got you, regarding `markdown` you meant the template itself rather than the issue/pr that it's templating for.

What I mentioned above is about how the template will be rendered(for issue-raisers) rather than how the template was encoded(maintaining the templates), let me check whether the equivalent format could be in .md and come back to you ;-).

The reason I have made it into YAML rather than Markdown was the similar template I referred are YAML

wey-gu:
update:

There is no way to use .md templates to render issue into separated forms(each form are still in markdown format and the whole issue is persistent in markdown, too) rather than YAML: [syntax-for-githubs-form-schema](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/syntax-for-githubs-form-schema).

If **template files in .md** should be kept then, we cannot change it to forms(separated markdown fields).

Although I consider `separated fields` is more than just cool but bringing issue-raiser better experiences yet may help enforce more information in the issue ;-).

Thanks~
BR, Wey

wey-gu:
Will close it if it's not the expected way.

PR #3600, opened 2021-12-29
disable an unstable test case

liuyu85cn:
#### What type of PR is this?
- [ X] bug

#### What does this PR do?

It is reported some CI failed of chain_delete_test.Test5 

can't repro on 211,  just disable it for a while. 

(It is weird that test5 and test6 is almost same... why test 6 not failed???)

#### Checklist:

#### Release notes:


jievince (approved):

darionyaphet (dismissed):

src/storage/test/ChainDeleteEdgesTest.cpp
```diff
@@ -219,64 +219,60 @@ TEST(ChainDeleteEdgesTest, Test4) {
   std::this_thread::sleep_for(std::chrono::milliseconds(300));
 }
 
-// add some edges, then delete one of them, not execute local commit
-TEST(ChainDeleteEdgesTest, Test5) {
```
yixinglu, line 223: use DISABLE_Test5 to disable this test case

yixinglu (approved):

src/storage/test/ChainDeleteEdgesTest.cpp
```diff
@@ -271,9 +236,9 @@ TEST(ChainDeleteEdgesTest, Test5) {
   FakeInternalStorageClient::hookInternalStorageClient(env, iClient);
   ChainResumeProcessor resumeProc(env);
   resumeProc.process();
+  // std::this_thread::sleep_for(std::chrono::milliseconds());
```
darionyaphet, line 239: why don't remove directly?

liuyu85cn: This PR is just for not cause other CI failed. Will dig more until it stable.
Thus it will have some debug code. Not sure if this is the unstable point.

Shylock-Hg: You could convert it to a draft PR.

critical27 (approved):

Issue #3476, opened 2021-12-15
storaged cannot be exited gracefully


HarrisChu:
https://github.com/vesoft-inc/nebula/issues/3475

After this, cannot be exited gracefully, not sure if it depends on heartbeat thread.

pstrack log as below:
```
Thread 3 (Thread 0x7fc024fdd700 (LWP 32381)):
#0  0x00007fc034d760e3 in epoll_wait () from /lib64/libc.so.6
#1  0x00000000021fc345 in ?? ()
#2  0x00000000021f30e8 in event_base_loop ()
#3  0x000000000212a466 in folly::EventBase::loopBody(int, bool) ()
#4  0x000000000212a8ff in folly::EventBase::loop() ()
#5  0x000000000212ca04 in folly::EventBase::loopForever() ()
#6  0x00000000020be8e9 in folly::IOThreadPoolExecutor::threadRun(std::shared_ptr<folly::ThreadPoolExecutor::Thread>) ()
#7  0x00000000020cae27 in void folly::detail::function::FunctionTraits<void ()>::callBig<std::_Bind<void (folly::ThreadPoolExecutor::*(folly::ThreadPoolExecutor*, std::shared_ptr<folly::ThreadPoolExecutor::Thread>))(std::shared_ptr<folly::ThreadPoolExecutor::Thread>)> >(folly::detail::function::Data&) ()
#8  0x00000000026c2570 in execute_native_thread_routine ()
#9  0x00007fc03504cea5 in start_thread () from /lib64/libpthread.so.0
#10 0x00007fc034d75b0d in clone () from /lib64/libc.so.6
Thread 2 (Thread 0x7fc0247dc700 (LWP 32390)):
#0  0x00007fc034d760e3 in epoll_wait () from /lib64/libc.so.6
#1  0x00000000021fc345 in ?? ()
#2  0x00000000021f30e8 in event_base_loop ()
#3  0x000000000212a466 in folly::EventBase::loopBody(int, bool) ()
#4  0x000000000212a8ff in folly::EventBase::loop() ()
#5  0x000000000212ca04 in folly::EventBase::loopForever() ()
#6  0x00000000020be8e9 in folly::IOThreadPoolExecutor::threadRun(std::shared_ptr<folly::ThreadPoolExecutor::Thread>) ()
[...]

PR #3621, opened 2022-01-04
add notifyStop() for metaClient

liuyu85cn:
#### What type of PR is this?
- [X ] bug
- [ ] feature
- [ ] enhancement

Issue(s) number: 

Description:

It is reported if not "add host" metaClient will stuck at "waitForMetaready()" and can't stop. 
This pr try to fix this. 

#### How do you solve it?

waitForMetaready() will detect a flag in each loop, before retry. Set this flag. 
  
#### Special notes for your reviewer, ex. impact of this fix, design document, etc:

Because no UT call metaClient.notifyStop() before, and it is hard to know which UT will affect by this, just run CI 

#### Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

#### Release notes:

> ex. Fixed the bug .....

### DIFF
--- a/src/storage/StorageServer.cpp
+++ b/src/storage/StorageServer.cpp
@@ -363,6 +363,9 @@ void StorageServer::notifyStop() {
     serverStatus_ = STATUS_STOPPED;
     cvStop_.notify_one();
   }
+  if (metaClient_) {
+    metaClient_->notifyStop();
+  }
 }
 
 void StorageServer::stop() {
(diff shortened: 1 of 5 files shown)


yixinglu (approved):

HarrisChu:
fix https://github.com/vesoft-inc/nebula/issues/3476

critical27 (approved):
👍 , quick fix

panda-sheep (approved):
Good job！

src/storage/StorageServer.cpp
```diff
@@ -363,6 +363,9 @@ void StorageServer::notifyStop() {
     serverStatus_ = STATUS_STOPPED;
     cvStop_.notify_one();
   }
+  if (metaClient_) {
+    metaClient_->notifyStop();
+  }
```
panda-sheep, line 368: 👍

PR #3637, opened 2022-01-05
Add ci for standalone compile and ut.

SuperYoko:
#### What type of PR is this?
- [x] bug
- [ ] feature
- [ ] enhancement

Issue(s) number: 

Description:
Add workflow to check standalone compile and ut. PYTEST and TCK comes later.

#### Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

#### Release notes:

> ex. Fixed the bug .....


.github/workflows/pull_request.yml
```diff
@@ -194,3 +194,72 @@ jobs:
...
+            esac
+            ;;
+          esac
+      - name: Make
+        run: |
+          ccache -z
+          ninja -j $(nproc)
+          ccache -s
+        working-directory: build/
+      - name: CTest
```
Shylock-Hg, line 254: Need to test tck too.

SuperYoko: I'll add it in another pr.

.github/workflows/pull_request.yml
```diff
@@ -199,3 +199,92 @@ jobs:
         with:
           name: ${{ matrix.os }}-${{ matrix.compiler }}-nebula-test-logs
           path:  ./build/server_*/logs/
+
+  standalone:
```
Shylock-Hg, line 203: I think you could build standalone nebula in origin ci job and test after origin test step.

SuperYoko: Another compile workflow is needed.And there is a metirc to combine workflows in origin ci job. I don't want this too heavy.

tests/tck/features/mutate/InsertWithTimeType.feature
```diff
@@ -57,6 +57,7 @@ Feature: Insert with time-dependent types
     Then a ExecutionError should be raised at runtime:Storage Error: The data type does not meet the requirements. Use the correct type of data.
     And drop the used space
 
+  @distonly
```
Shylock-Hg, line 60: It should support in standalone version too.

SuperYoko: Fixed.

src/clients/storage/StorageClientBase-inl.h
```diff
@@ -240,16 +240,18 @@ void StorageClientBase<ClientType, ClientManagerType>::getResponseImpl(
     DCHECK(!!ioThreadPool_);
     evb = ioThreadPool_->getEventBase();
   }
+  auto reqPtr = std::make_shared<std::pair<HostAddr, Request>>(std::move(request.first),
```
critical27, line 243: Would you explain more about why we need to use smart pointer here? `collectResponse` need to send request to multiple storage, but `getResponse` does not need to.

critical27 (commented):
Generally LGTM

src/clients/storage/StorageClientBase-inl.h
SuperYoko: RPC make a copy of requests, request can be moved after send and noone will use. While local function call need requests to stay where it is until function call is over.

SuperYoko: Maybe I should isolute this with MACRO.

critical27: Ah I get your your point. Because in stand alone mode, the `client` is the `GraphStorageLocalServer`, so the request need to out-live until the function call is over.

I don't quite remember the `request` parameter type in `remoteFunc`. If it only accepts const reference, maybe we we do need to extend the life cycle of `request`. If it can be rvalue, we can move it. This could be check later. The change here is ok to me for now.

critical27 (approved):
Good job~

critical27 (approved):

tests/common/nebula_service.py
```diff
@@ -424,6 +557,64 @@ def start(self):
...
+        # get session from the pool
+        client = client_pool.get_session('root', 'nebula')
+
+        hosts = ",".join(
+            [
+                "127.0.0.1:{}".format(str(storaged.storage_port))
+                for storaged in self.graphd_processes
+            ]
+        )
+        cmd = "ADD HOSTS {} INTO NEW ZONE \"default_zone\"".format(hosts)
```
darionyaphet, line 600: currently `ADD HOSTS {} INTO NEW ZONE`  not supported

SuperYoko: I keep it the same with basic script, this failure won't affect ci. I'll make change together with distribution version later.

pengweisong (approved):
Good Job

critical27 (approved):

PR #3643, opened 2022-01-05
add comments & accumulate bug fix for TOSS

liuyu85cn:
1. add comments & adjust some log level 
2. accumulate bug fix for TOSS
    2.1 one space may be affected by another space. when recover prime edges. 
    2.2 when call kvstore->get(), if leaderLeaseInvalid, it will report leader_change, change it to leader_lease_failed.
    2.3 print all execution trace in ChainProcessor::finish(), make much more grep friendly. 
    2.4 some function of memory lock is hard to understand (forceLock, forceUnlock) , change to "setAutoUnlock(bool)"

#### What type of PR is this?
- [ ] bug
- [ ] feature
- [x] enhancement

Issue(s) number: 

Description:

#### Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

#### Release notes:

> ex. Fixed the bug .....


src/storage/transaction/ChainAddEdgesLocalProcessor.cpp
```diff
@@ -363,9 +288,31 @@ cpp2::AddEdgesRequest ChainAddEdgesLocalProcessor::reverseRequest(
...
+            << ", rcRemote_=" << apache::thrift::util::enumNameSafe(rcRemote_)
+            << ", rcCommit_=" << apache::thrift::util::enumNameSafe(rcCommit_);
+  }
+  do {
+    if (rcPrepare_ != Code::SUCCEEDED) {
+      break;  // nothing written, no need to recover.
+    }
+
+    if (rcCommit_ != Code::SUCCEEDED) {
+      reportFailed(ResumeType::RESUME_CHAIN);
```
critical27, line 304: I see `addPrime` is called in `reportFailed`, why we need to do it again?

critical27 (commented):
Any special reason to use `rcPrepare_`, `rcRemote_` and `rcCommit_`, it actually make the parameter in `processRemote` and `processLocal` is useless, and the chain is broken. Is it mainly for we could set the result in `finish` method?

liuyu85cn:
> Any special reason to use `rcPrepare_`, `rcRemote_` and `rcCommit_`, it actually make the parameter in `processRemote` and `processLocal` is useless, and the chain is broken. Is it mainly for we could set the result in `finish` method?

use this style because previously, when pass a code in the chain, it's hard to tell where an error happens. 
for example if processLocal() get an error code "leader change", 
it can't determined if this error happened in prepareLocal() or processRemote().
actually, I use some other member to identify this error ("primeInserted" for example.)

It is the same reason why set result in finish(). Because I could get a clear view of what this processor has done.
The previous version, will report to transaction manager at the moment some error happens. 
This is really error prone because I need to carefully check not to report a same error twice.

src/storage/transaction/ChainAddEdgesLocalProcessor.cpp
liuyu85cn: In my opinion,  "addPrime()" is really hard to understand.
"reportFailed" is just a self explain/comment function .

critical27 (approved):

critical27:
@kikimo you could do a final test later

src/storage/test/CMakeLists.txt
```diff
@@ -754,20 +754,20 @@ nebula_add_executable(
...
-    OBJECTS
-        ${storage_test_deps}
-    LIBRARIES
-        ${ROCKSDB_LIBRARIES}
-        ${THRIFT_LIBRARIES}
-        ${PROXYGEN_LIBRARIES}
-        wangle
-        gtest
-)
+# nebula_add_executable(
```
panda-sheep, line 757: Why comment out? 
If don't need it, can delete it or add a comment

critical27 (dismissed):

kikimo:
> @kikimo you could do a final test later

I'll do it this afternoon

src/storage/test/CMakeLists.txt
liuyu85cn: ok, delete it for short

panda-sheep (dismissed):
Good job!

kikimo:
> > @kikimo you could do a final test later
> 
> I'll do it this afternoon

Checked, no problem.

kikimo (dismissed):
LGTM

kikimo (approved):
LGTM

critical27 (approved):

PR #3655, opened 2022-01-07
Support multi thread ingest

darionyaphet:
#### What type of PR is this?
- [ ] bug
- [ ] feature
- [X] enhancement

Issue(s) number: 

Description:

#### Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

#### Release notes:

> ex. Fixed the bug .....

### DIFF
--- a/src/kvstore/NebulaStore.cpp
+++ b/src/kvstore/NebulaStore.cpp
@@ -781,32 +781,41 @@ nebula::cpp2::ErrorCode NebulaStore::ingest(GraphSpaceID spaceId) {
   if (!ok(spaceRet)) {
     return error(spaceRet);
   }
+
   auto space = nebula::value(spaceRet);
+  std::vector<std::thread> threads;
+  nebula::cpp2::ErrorCode code = nebula::cpp2::ErrorCode::SUCCEEDED;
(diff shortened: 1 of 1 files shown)


src/kvstore/NebulaStore.cpp
```diff
@@ -781,32 +781,41 @@ nebula::cpp2::ErrorCode NebulaStore::ingest(GraphSpaceID spaceId) {
...
-      if (!ok(ret)) {
-        return error(ret);
-      }
-
-      auto path = folly::stringPrintf("%s/download/%d", value(ret)->getDataRoot(), part);
-      if (!fs::FileUtils::exist(path)) {
-        LOG(INFO) << path << " not existed";
-        continue;
-      }
+    threads.emplace_back(std::thread([&engine, &code, this, spaceId] {
```
Shylock-Hg, line 790: I think you could run it in folly Executor to avoid create too many threads.

darionyaphet: each engine will create a thread so it seems fine

darionyaphet: using Executor is a good idea thanks

PR #3662, opened 2022-01-07
add comments and adjust log level for meta/processors/admin

pengweisong:
#### What type of PR is this?
- [ ] bug
- [ ] feature
- [x] enhancement

Issue(s) number: 

Description:
Adjust codes in src/meta/processors/admin.:
1. change all ERROR to INFO
2. add comments for public function and some classes.

#### Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

#### Release notes:

>  No need


Aiee:
what's the reason to change all error log into info level?

pengweisong:
@Aiee Because we think error means the fault could not be recovered. The message before may could bring some confused for the user.

critical27 (approved):

liuyu85cn (approved):

PR #3666, opened 2022-01-07
fix issue 3601

### LINKED ISSUE #3601: storage did not release memory

All of the `Expression` are allocated in `ObjectPool`. And, when creating a tag/edge who contain a prop with a `default value`, the `default value` is stored as an `Expressoin` in Schema.

In graphd, the `ObjectPool` is created in every query, and released after query. But, in `MetaClient`, the pool is created when load schema, and released when reload schema if there is any update about schema. So, if schema is never never updated, the pool will never be released.

When writing any data who has a `default value` and ngql not fill it, storaged will compute `default value`. But, the `Expression` about `default value`  will be clone and allocated in `MetaClient`'s pool. This causes the pool to take up more and more memory, and it will never be released unless the schema is changed.

cangfengzhs:
#### What type of PR is this?
- [x] bug
- [ ] feature
- [ ] enhancement

Issue(s) number: 
close #3601 
Description:

#### Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

#### Release notes:

> ex. Fixed the bug .....


critical27 (approved):
Good job~~

critical27:
Remember to write a brief retrospective post in confluence

CPWstatic (approved):

Issue #3654, opened 2022-01-06
balance remove error


HarrisChu:
just record it.

issue 1:
src and dst are in the same.

issue 2:
after metad leader changed, cannot stop the job.

```
(root@nebula) [(none)]>       CREATE SPACE test(vid_type=int, replica_factor=3, partition_num=10) on "z1","z2","z3","z4";
Execution succeeded (time spent 10524/20016 us)

Thu, 06 Jan 2022 17:47:06 CST

(root@nebula) [(none)]> use test
Execution succeeded (time spent 2591/13982 us)

Thu, 06 Jan 2022 17:47:10 CST

(root@nebula) [test]>
(root@nebula) [test]> show parts
+--------------+-------------------+-----------------------------------------------------+-------+
| Partition ID | Leader            | Peers                                               | Losts |
+--------------+-------------------+-----------------------------------------------------+-------+
| 1            | "127.0.0.1:12391" | "127.0.0.1:12391, 127.0.0.1:18320, 127.0.0.1:17318" | ""    |
| 2            | "127.0.0.1:18320" | "127.0.0.1:12391, 127.0.0.1:18320, 127.0.0.1:19179" | ""    |
| 3            | "127.0.0.1:17318" | "127.0.0.1:12391, 127.0.0.1:17318, 127.0.0.1:19179" | ""    |
| 4            | "127.0.0.1:19179" | "127.0.0.1:18320, 127.0.0.1:17318, 127.0.0.1:19179" | ""    |
| 5            | "127.0.0.1:17318" | "127.0.0.1:12391, 127.0.0.1:18320, 127.0.0.1:17318" | ""    |
| 6            | "127.0.0.1:18320" | "127.0.0.1:12391, 127.0.0.1:18320, 127.0.0.1:19179" | ""    |
| 7            | "127.0.0.1:12391" | "127.0.0.1:12391, 127.0.0.1:17318, 127.0.0.1:19179" | ""    |
| 8            | "127.0.0.1:18320" | "127.0.0.1:18320, 127.0.0.1:17318, 127.0.0.1:19179" | ""    |
| 9            | "127.0.0.1:17318" | "127.0.0.1:12391, 127.0.0.1:18320, 127.0.0.1:17318" | ""    |
| 10           | "127.0.0.1:18320" | "127.0.0.1:12391, 127.0.0.1:18320, 127.0.0.1:19179" | ""    |
+--------------+-------------------+-----------------------------------------------------+-------+
[...]

PR #3668, opened 2022-01-07
Balance fix

liwenhui-soul:
#### What type of PR is this?
- [ ] bug
- [ ] feature
- [ ] enhancement

Issue(s) number: 

Description:

#### Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

#### Release notes:

> ex. Fixed the bug .....

### DIFF
--- a/src/kvstore/raftex/RaftexService.cpp
+++ b/src/kvstore/raftex/RaftexService.cpp
@@ -155,10 +155,20 @@ void RaftexService::addPartition(std::shared_ptr<RaftPart> part) {
 }
 
 void RaftexService::removePartition(std::shared_ptr<RaftPart> part) {
-  folly::RWSpinLock::WriteHolder wh(partsLock_);
-  parts_.erase(std::make_pair(part->spaceId(), part->partitionId()));
+  using FType = decltype(folly::makeFuture());
+  using FTValype = typename FType::value_type;
   // Stop the partition
-  part->stop();
+  folly::makeFuture()
+      .thenValue([this, &part](FTValype) {
+        folly::RWSpinLock::WriteHolder wh(partsLock_);
+        parts_.erase(std::make_pair(part->spaceId(), part->partitionId()));
+      })
+      // the part->stop() would wait for requestOnGoing_ in Host, and the requestOnGoing_ will
+      // release in response in ioThreadPool,this may cause deadlock, so doing it in another
+      // threadpool to avoid this condition
+      .via(folly::getGlobalCPUExecutor())
+      .thenValue([part](FTValype) { part->stop(); })
+      .wait();
 }
 
 std::shared_ptr<RaftPart> RaftexService::findPart(GraphSpaceID spaceId, PartitionID partId) {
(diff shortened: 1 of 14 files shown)


Sophie-Xie:
Which issues does this PR solve?

HarrisChu:
fix https://github.com/vesoft-inc/nebula/issues/3654

src/meta/processors/job/BalanceJobExecutor.cpp
```diff
@@ -88,6 +88,36 @@ nebula::cpp2::ErrorCode BalanceJobExecutor::save(const std::string& k, const std
   return rc;
 }
 
+void BalanceJobExecutor::insertOneTask(
+    const BalanceTask& task, std::map<PartitionID, std::vector<BalanceTask>>* existTasks) {
+  std::vector<BalanceTask>& taskVec = existTasks->operator[](task.getPartId());
```
kikimo, line 93: Just curious, why not use reference instead?

liwenhui-soul: > Just curious, why not use reference instead?

just follow google style, an output param should be non-const pointer

src/meta/processors/job/JobManager.cpp
```diff
@@ -144,7 +144,7 @@ void JobManager::scheduleThread() {
 
 // @return: true if all task dispatched, else false
 bool JobManager::runJobInternal(const JobDescription& jobDesc, JbOp op) {
-  std::lock_guard<std::mutex> lk(muJobFinished_);
+  std::lock_guard<std::recursive_mutex> lk(muJobFinished_);
```
critical27, line 147: Why we need to use recursive_mutex here?

critical27 (commented):
Generally LGTM, maybe add some comments

src/meta/processors/job/JobManager.cpp
liwenhui-soul: if all balance tasks success and the job is stopped before job status updated, the balance tasks will get muJobFinished_ through call back in the same thread

liwenhui-soul: have added comment at its declare place

critical27:
Close #3672

critical27 (dismissed):
Generally LGTM

src/kvstore/raftex/RaftexService.cpp
```diff
@@ -155,10 +155,17 @@ void RaftexService::addPartition(std::shared_ptr<RaftPart> part) {
...
+  using FType = decltype(folly::makeFuture());
+  using FTValype = typename FType::value_type;
   // Stop the partition
-  part->stop();
+  folly::makeFuture()
+      .thenValue([this, &part](FTValype) {
+        folly::RWSpinLock::WriteHolder wh(partsLock_);
+        parts_.erase(std::make_pair(part->spaceId(), part->partitionId()));
+      })
+      .via(getThreadManager().get())
```
critical27, line 166: Maybe add some comments here. Do you think we should use a global executor here?

liwenhui-soul: done

liuyu85cn (dismissed):

critical27 (approved):
Good job

PR #3681, opened 2022-01-11
[kvstore] adjust log level and add header comments

critical27:
#### What type of PR is this?
- [ ] bug
- [ ] feature
- [X] enhancement

Issue(s) number: 

Description:

1. Adjust log level and add header comments in `src/codec` and `src/kvstore` directory
2. Remove the HBase directory
3. Move the implementation of `AtomicLogBuffer` to source file

#### Special notes for your reviewer, ex. impact of this fix, design document, etc:

If any interface's comments is not clear, feel free to ask me to add some description

#### Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [X] N/A

Affects:

#### Release notes:

No


src/codec/RowReaderV1.cpp
```diff
@@ -219,7 +219,7 @@ Value RowReaderV1::getValueByIndex(const int64_t index) const noexcept {
     case PropertyType::STRING:
       return getString(index);
     default:
-      LOG(ERROR) << "Unknown type: " << apache::thrift::util::enumNameSafe(vType);
+      LOG(FATAL) << "Unknown type: " << apache::thrift::util::enumNameSafe(vType);
```
panda-sheep, line 222: why ？ This will coredump.

src/kvstore/KVEngine.h
```diff
@@ -15,18 +15,44 @@
...
+   * @brief Encode the operation of put key/value into write batch
+   *
+   * @param key
+   * @param value
+   * @return nebula::cpp2::ErrorCode
+   */
   virtual nebula::cpp2::ErrorCode put(folly::StringPiece key, folly::StringPiece value) = 0;
 
+  /**
+   * @brief Encode the operation of remove key/value into write batch
```
panda-sheep, line 35: `key/value`maybe  `key` ?

src/kvstore/KVEngine.h
```diff
@@ -15,18 +15,44 @@
...
+   *
+   * @param key
+   * @param value
+   * @return nebula::cpp2::ErrorCode
+   */
   virtual nebula::cpp2::ErrorCode remove(folly::StringPiece key) = 0;
 
-  // Remove all keys in the range [start, end)
+  /**
+   * @brief Encode the operation of range remove key/value into write batch
```
panda-sheep, line 44: ditto

src/kvstore/KVEngine.h
```diff
@@ -15,18 +15,44 @@
...
+   * @return nebula::cpp2::ErrorCode
+   */
   virtual nebula::cpp2::ErrorCode put(folly::StringPiece key, folly::StringPiece value) = 0;
 
+  /**
+   * @brief Encode the operation of remove key/value into write batch
+   *
+   * @param key
+   * @param value
+   * @return nebula::cpp2::ErrorCode
```
panda-sheep, line 39: The parameter is incorrect

panda-sheep (commented):
Great job

src/codec/RowReaderV1.cpp
panda-sheep: BTW, should `LOG(FATAL`) appear as little as possible in the code?  
Because LOG(FATAL) directly leads to service coredump

src/codec/RowReaderWrapper.cpp
```diff
@@ -107,7 +107,7 @@ bool RowReaderWrapper::reset(meta::SchemaProviderIf const* schema,
     currReader_ = &readerV2_;
     return true;
   } else {
-    LOG(ERROR) << "Unsupported row reader version " << readerVer;
+    LOG(WARNING) << "Unsupported row reader version " << readerVer;
```
darionyaphet, line 110: How to distinguish between warning and error？

src/codec/RowReaderV1.cpp
critical27: Actually for v1, there won't be other types, I think we could fail only in this case. I am ok if use `LOG(ERROR)` here.

src/kvstore/KVEngine.h
critical27: fixed.

src/codec/RowReaderV1.cpp
critical27: Fixed

src/kvstore/KVEngine.h
critical27: My bad, fixed

src/codec/RowReaderWrapper.cpp
critical27: It depends on whether administrator need to be involved. ERROR definitely need to be checked, but WARNING is not mandatory.

src/kvstore/KVStore.h
```diff
@@ -58,165 +58,376 @@ struct StoreCapability {
...
 
-  // Read a single key
+  /**
+   * @brief Read a single key
+   *
+   * @param spaceId
+   * @param partId
+   * @param key
+   * @param value
+   * @param canReadFromFollower Whether check if current kvstore is leader of given partition
```
wenhaocs, line 107: remove "whether".

This param is used in func `checkLeader`. If it is set to true, the current part can serve the read regardless of its role. So essentially, this param just indicates whether the current part can serve the read if the it's not leader, not the same as the current description. I think even just deleting the current description is good enough.

src/kvstore/KVStore.h
```diff
@@ -58,165 +58,376 @@ struct StoreCapability {
...
-  // If key[i] does not exist, the i-th value in return value would be
-  // Status::KeyNotFound
+  /**
+   * @brief Read a list of keys
+   *
+   * @param spaceId
+   * @param partId
+   * @param keys Keys to read
+   * @param values Pointers of value
+   * @param canReadFromFollower Whether check if current kvstore is leader of given partition
```
wenhaocs, line 123: ditto

src/kvstore/KVStore.h
```diff
@@ -58,165 +58,376 @@ struct StoreCapability {
...
-  // Get all results in range [start, end)
+  /**
+   * @brief Get all results in range [start, end)
+   *
+   * @param spaceId
+   * @param partId
+   * @param start Start key, inclusive
+   * @param end End key, exclusive
+   * @param iter Iterator in range [start, end), returns by kv engine
+   * @param canReadFromFollower Whether check if current kvstore is leader of given partition
```
wenhaocs, line 143: ditto

src/kvstore/Listener.cpp
```diff
@@ -209,7 +209,8 @@ void Listener::doApply() {
           break;
         }
         default: {
-          LOG(WARNING) << idStr_ << "Unknown operation: " << static_cast<int32_t>(log[0]);
+          VLOG(2) << idStr_
```
wenhaocs, line 212: since it is some kind of exception, what's the reason of changing from log warning to vlog?

wenhaocs (commented):
Awesome work!

panda-sheep (approved):
Impressive work!

src/common/utils/NebulaKeyUtils.cpp
```diff
@@ -72,20 +72,31 @@ std::string NebulaKeyUtils::edgeKey(size_t vIdLen,
...
   std::string key;
   key.reserve(kTagLen + vIdLen);
   key.append(reinterpret_cast<const char*>(&item), sizeof(int32_t))
       .append(vId.data(), vId.size())
       .append(vIdLen - vId.size(), pad);
   return key;
 }
+
+// static
+std::string NebulaKeyUtils::vertexPrefix(PartitionID partId) {
```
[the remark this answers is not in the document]
critical27: Yeah, it is not used for now. To scan all vertex in one part if necessary

src/kvstore/raftex/test/TestShard.cpp
```diff
@@ -194,7 +194,7 @@ std::tuple<nebula::cpp2::ErrorCode, LogID, TermID> TestShard::commitLogs(
           folly::RWSpinLock::WriteHolder wh(&lock_);
           currLogId_ = iter->logId();
           data_.emplace_back(currLogId_, log.toString());
-          VLOG(1) << idStr_ << "Write: " << log << ", LogId: " << currLogId_
+          VLOG(2) << idStr_ << "Write: " << log << ", LogId: " << currLogId_
```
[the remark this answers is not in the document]
critical27: Indeed, the main reason is that the log level in `RaftPart` is lower or equal to `VLOG(1)`, it would be inappropriate if UT's log level is higher.

src/kvstore/plugins/elasticsearch/ESListener.cpp
```diff
@@ -212,7 +212,7 @@ bool ESListener::writeData(const std::vector<nebula::plugin::DocItem>& items) co
   if (isNeedWriteOneByOne) {
     return writeDatum(items);
   }
-  LOG(ERROR) << "A fatal error . Full-text engine is not working.";
+  LOG(INFO) << idStr_ << "Failed to bulk into es.";
```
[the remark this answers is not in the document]
critical27: 👌

src/kvstore/KVStore.h
critical27: 👌

src/kvstore/Listener.cpp
critical27: Two reasons here:
The max log level is limited to VLOG(1)
If the exception happens, log here will be printed every time for each log entry, which is unexpectd.

wenhaocs (approved):

wenhaocs (approved):

src/kvstore/raftex/RaftLogIterator.h
```diff
@@ -12,20 +12,49 @@
 namespace nebula {
 namespace raftex {
 
+/**
+ * @brief The wal log iterator used when follower received logs from leader by rpc
+ */
 class RaftLogIterator final : public LogIterator {
  public:
+  /**
+   * @brief Construct a new raf log iterator
```
liuyu85cn, line 21: miss a 't'

src/kvstore/wal/FileBasedWal.h
```diff
@@ -50,80 +50,131 @@ class FileBasedWal final : public Wal, public std::enable_shared_from_this<FileB
...
+   * @param msg Log messgage to append
+   * @return Wheter append succeed
+   */
   bool appendLog(LogID id, TermID term, ClusterID cluster, std::string msg) override;
 
-  // Append a list of log messages to the WAL
-  // This method **IS NOT** thread-safe
-  // we **DO NOT** expect multiple threads will append logs
-  // simultaneously
+  //
```
liuyu85cn, line 113: can be deleted

liuyu85cn (approved):

PR #3696, opened 2022-01-11
add session tool to manage and analysis sessions

a516072575:
#### What type of PR is this?
- [ ] bug
- [ ] feature
- [x] enhancement

Issue(s) number: 

Description:
Metad OOM because some users have frequently created sessions and did not release those sessions. 
Finally by more than 8000000 sessions were created. When I tried to restart cluster, metad OOM occured again.
I found that graphd/storaged called waitForMetadReady function of MetaClient to wait for Metad become be available, 
Metad loaded all sessions to memory in waitForMetadReady function. Too many sessions caused metad OOM.

#### How do you solve it?
I develop a session tool to delete extra sessions in the way offline. It directly  delete sessions from rocksdb data of all metad after cluster shutdown. It can delete all sessions and delete the sessions for the specified user. like this:

**delete all sessions:**
session_tool -path data/meta/nebula/0/data -operator delete (need execute it on every metad node)
image

**delete all sessions of root user:**
session_tool -path data/meta/nebula/0/data -operator delete -user root (need execute it on every metad node)
image

In addition it also counts the number of sessions in rocksdb of metad in real time, like this:
**The total number of sessions in cluster：**
session_tool -path data/meta/nebula/0/data -operator count
image

**The number of sessions per user:**
session_tool -path data/meta/nebula/0/data -operator groupCount
image

**It can also show sessions info, like this:**
Show N sessions：
session_tool -path data/meta/nebula/0/data -operator list -limit N
image

**Show N sessions of user root:**
session_tool -path data/meta/nebula/0/data -operator list -limit N -user root 
image

#### Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [x] N/A

Affects:

#### Release notes:

> add a session tool to manage and analysis sessions


Hera-han:
@CPWstatic this issue has been open for a long time, please take a look and give feedback. thanks~

src/tools/session-tool/SessionTool.cpp
```diff
@@ -0,0 +1,225 @@
...
+            break;
+          }
+          iter->Next();
+        }
+      }
+    }
+
+    {
+      if ("groupCount" == op) {
+        auto map = std::unordered_map<std::string, int>();
```
darionyaphet, line 98: why don't use `std::unordered_map<std::string, int> map` directly?

CPWstatic:
The session could be manipulated via admin command. We support read sessions for now, i suggest you implement a `delete` command rather than a tool.

a516072575:
> The session could be manipulated via admin command. We support read sessions for now, i suggest you implement a `delete` command rather than a tool.

Too many sessions causing Metad OOM. Cluster can't start normally，admin command is not used before cluster is available. I can only use offline repair mode to delete sessions to ensure that the cluster is started normally. Yoy can get the reason for the cluster cannot be started normally from issue description

a516072575:
> The session could be manipulated via admin command. We support read sessions for now, i suggest you implement a `delete` command rather than a tool.

@CPWstatic 
Too many sessions causing Metad OOM. Cluster can't start normally，admin command is not used before cluster is available. I can only use offline repair mode to delete sessions to ensure that the cluster is started normally. Yoy can get the reason for the cluster cannot be started normally from issue description

xtcyclist:
This PR has been opened for too long. It is addressing a valid problem, although I'm not sure whether this problem is still there after almost a year. I've rebased the PR to check the CI results.

xtcyclist:
Given that the CI here run into errors and this issue has been opened for too long, I'm closing this PR fow now. It is welcomed for anyone to pick this up later, fix related issues, and reopen it.

PR #3707, opened 2022-01-13
Forbid the loopback in production conf

### LINKED ISSUE #3142: If the clients config the same ip/port, an overwritten problem will occur when show hosts.

If the clients config the same ip/port, an overwritten problem will occur when show hosts.

1. Given a same ip/port when start up graph or storage
2. Show host

**Expected behavior**

We should not allow the host join the cluster when it was configured a same ip/port with an existed host.

jackwener:
#### What type of PR is this?
- [x] bug
- [ ] feature
- [ ] enhancement

Issue(s) number: close #3142

Description:

In case that duplicated ip:port

#### How do you solve it?

modify the production confg
  
#### Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

#### Release notes:

> ex. Fixed the bug .....


jackwener:
The reason why we just delete config:
1. It is very difficult to check at the code level
2. The number and specific address of the loopback address are uncertain, because it can be configured by yourself
3. Trying to get the **real** ip port (Data Link Layer) at runtime(we is in application) is hard

Issue #3710, opened 2022-01-13
Raft snapshot data inconsistency


critical27:
When sending raft snapshot, all data need to use the same rocksdb snapshot, but we use multiple prefix to read data/index... which may lead to data inconsistency. See `NebulaSnapshotManager::accessAllRowsInSnapshot`.

critical27:
Fix it in 3.0 if possible

PR #3717, opened 2022-01-13
Add snapshot for prefix iterate.

SuperYoko:
#### What type of PR is this?
- [x] bug
- [ ] feature
- [ ] enhancement

Issue(s) number: 

Description:
Use prefix to iter data on the same snapshot of rocksdb.

#### Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

#### Release notes:

> ex. Fixed the bug .....


critical27:
Close https://github.com/vesoft-inc/nebula/issues/3710

critical27:
Very quick fix 👍

src/kvstore/RocksEngine.h
```diff
@@ -166,16 +174,19 @@ class RocksEngine : public KVEngine {
                                 std::unique_ptr<KVIterator>* iter) override;
 
   nebula::cpp2::ErrorCode prefix(const std::string& prefix,
-                                 std::unique_ptr<KVIterator>* iter) override;
+                                 std::unique_ptr<KVIterator>* iter,
```
darionyaphet, line 177: why don't use KVIterator* directly?

src/kvstore/RocksEngine.h
```diff
@@ -146,6 +146,14 @@ class RocksEngine : public KVEngine {
     return walPath_.c_str();
   }
 
+  const void* GetSnapshot() override {
+    return db_->GetSnapshot();
+  }
+
+  void ReleaseSnapshot(const void* snapshot) override {
+    db_->ReleaseSnapshot(reinterpret_cast<const rocksdb::Snapshot*>(snapshot));
```
darionyaphet, line 154: `ReleaseSnapshot`  would remove from file system?

SuperYoko: No, It's only a memory snapshot.

src/kvstore/RocksEngine.h
SuperYoko: Yes, it is bad to pass pointer of smart pointer. I don't know why it is coded this way. Maybe I should try to refactor it. But it do affect too many files.  Don't want do this in this pr.

liwenhui-soul:
what does void* mean? it doesn't have a type?

SuperYoko:
> what does void* mean? it doesn't have a type?

I wan't to  create a wrapper like iterator first, but it didn't  help much understanding. 
Anyway, Rocksdb Snapshot pointer should pass to prefix function, and there should be a container to save pointer. 

Maybe a warpper of vector <boost::any> is okey, but it still pass a pointer. I don't have another clear idear. WDYT?

critical27 (approved):

src/kvstore/RocksEngine.h
darionyaphet: AFAIK snapshot will create some hard link about the previous SST files why don't remove them?

SuperYoko: > AFAIK snapshot will create some hard link about the previous SST files why don't remove them?

https://github.com/facebook/rocksdb/wiki/Snapshot

> A snapshot captures a point-in-time view of the DB at the time it's created. Snapshots do not persist across DB restarts.

darionyaphet (approved):

HarrisChu:
https://github.com/vesoft-inc/nebula/issues/3782

seems db_upgrader crashed by this

panda-sheep:
Good job

PR #3735, opened 2022-01-17
Fix race of create tag and edge

### LINKED ISSUE #3376: Concurrent create tag and edge with same name may be both succeed.

Please check the [FAQ](https://github.com/vesoft-inc/nebula/blob/master/docs/manual-EN/1.overview/2.quick-start/2.FAQ.md) documentation and old issues before raising an issue in case someone has asked the same question that you are asking.

```cpp
 // CreateTagProcessor.cpp
  {
    // if there is an edge of the same name
    // TODO: there exists race condition, we should address it in the future
    folly::SharedMutex::ReadHolder rHolder(LockUtils::edgeLock());
    auto conflictRet = getEdgeType(spaceId, tagName);
    ......
  }  // <------------------------Unlock!!!!!!!!!
  auto columns = req.get_schema().get_columns();
  ......
```
Because `edgeLock` unlock before finish create tag, if there are two threads create tag/edge concurrency, may be both succeed.

* OS: `uname -a`
* CPU: `lscpu`
* Commit id (e.g. `a3ffc7d8`)

1. Step 1
2. Step 2
3. Step 3

**Expected behavior**

**Additional context**

Provide logs and configs, or any other context to trace the problem.

jackwener:
#### What type of PR is this?
- [x] bug
- [ ] feature
- [ ] enhancement

Issue(s) number:  close #3376

Description:

Merge two lock

#### Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

#### Release notes:

> ex. Fixed the bug .....

### DIFF
--- a/src/meta/processors/Common.h
+++ b/src/meta/processors/Common.h
@@ -25,8 +25,7 @@ class LockUtils {
   GENERATE_LOCK(id);
   GENERATE_LOCK(workerId);
   GENERATE_LOCK(localId);
-  GENERATE_LOCK(tag);
-  GENERATE_LOCK(edge);
+  GENERATE_LOCK(tagAndEdge);
   GENERATE_LOCK(tagIndex);
   GENERATE_LOCK(edgeIndex);
   GENERATE_LOCK(service);
--- a/src/meta/processors/schema/CreateEdgeProcessor.cpp
+++ b/src/meta/processors/schema/CreateEdgeProcessor.cpp
@@ -14,8 +14,7 @@ void CreateEdgeProcessor::process(const cpp2::CreateEdgeReq& req) {
   GraphSpaceID spaceId = req.get_space_id();
   CHECK_SPACE_ID_AND_RETURN(spaceId);
   const auto& edgeName = req.get_edge_name();
-  folly::SharedMutex::ReadHolder rHolder(LockUtils::tagAndEdgeLock());
-
+  folly::SharedMutex::WriteHolder holder(LockUtils::tagAndEdgeLock());
   // Check if the tag with same name exists
   auto conflictRet = getTagId(spaceId, edgeName);
   if (nebula::ok(conflictRet)) {
--- a/src/meta/processors/schema/CreateEdgeProcessor.cpp
+++ b/src/meta/processors/schema/CreateEdgeProcessor.cpp
@@ -15,24 +15,24 @@ void CreateEdgeProcessor::process(const cpp2::CreateEdgeReq& req) {
   CHECK_SPACE_ID_AND_RETURN(spaceId);
   const auto& edgeName = req.get_edge_name();
   folly::SharedMutex::ReadHolder rHolder(LockUtils::tagAndEdgeLock());
-  {
-    auto conflictRet = getTagId(spaceId, edgeName);
-    if (nebula::ok(conflictRet)) {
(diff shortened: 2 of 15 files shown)


jackwener:
why not increase the scope of lock.
Because now
```
(Edge RL) --> (Tag WL)
```
if 
```
(Edge RL --> (Tag WL)  ）
```
This will be dead lock

src/meta/processors/schema/CreateEdgeProcessor.cpp
```diff
@@ -13,28 +13,26 @@ namespace meta {
...
-    // TODO: there exists race condition, we should address it in the future
-    folly::SharedMutex::ReadHolder rHolder(LockUtils::tagLock());
-    auto conflictRet = getTagId(spaceId, edgeName);
-    if (nebula::ok(conflictRet)) {
-      LOG(ERROR) << "Failed to create edge `" << edgeName
-                 << "': some tag with the same name already exists.";
-      resp_.id_ref() = to(nebula::value(conflictRet), EntryType::EDGE);
-      handleErrorCode(nebula::cpp2::ErrorCode::E_CONFLICT);
+  const auto& edgeName = req.get_edge_name();
+  folly::SharedMutex::ReadHolder rHolder(LockUtils::tagAndEdgeLock());
```
Shylock-Hg, line 17: WriteHolder.

src/meta/processors/schema/CreateTagProcessor.cpp
```diff
@@ -13,28 +13,26 @@ namespace meta {
...
-    // TODO: there exists race condition, we should address it in the future
-    folly::SharedMutex::ReadHolder rHolder(LockUtils::edgeLock());
-    auto conflictRet = getEdgeType(spaceId, tagName);
-    if (nebula::ok(conflictRet)) {
-      LOG(ERROR) << "Failed to create tag `" << tagName
-                 << "': some edge with the same name already exists.";
-      resp_.id_ref() = to(nebula::value(conflictRet), EntryType::TAG);
-      handleErrorCode(nebula::cpp2::ErrorCode::E_CONFLICT);
+  const auto& tagName = req.get_tag_name();
+  folly::SharedMutex::ReadHolder rHolder(LockUtils::tagAndEdgeLock());
```
Shylock-Hg, line 17: ditto.

jackwener: fixed

Shylock-Hg (approved):

critical27 (approved):
LGTM, good job. 

BTW, we could unify all locks in meta with only one lock later, several reasons:
1. Dead lock may happen
2. Many interfaces does not use right lock

Contact me if you're interested

src/meta/processors/Common.h
```diff
@@ -25,8 +25,7 @@ class LockUtils {
   GENERATE_LOCK(id);
   GENERATE_LOCK(workerId);
   GENERATE_LOCK(localId);
-  GENERATE_LOCK(tag);
-  GENERATE_LOCK(edge);
+  GENERATE_LOCK(tagAndEdge);
   GENERATE_LOCK(tagIndex);
   GENERATE_LOCK(edgeIndex);
```
critical27, line 30: You could unify tagIndex and edgeIndex as one. Although tag/edge index is hard to misused, but they do will have same name IIRC.

Issue #3704, opened 2022-01-12
Optimizer rules compatibility


czpmango:
**Describe the bug (__required__)**
This pr (https://github.com/vesoft-inc/nebula/pull/3255) should fix all related optimizer rules, because the expression related processing logic has been changed(VarPropertyExpression/InputPropertyExpression -> LabelTagPropertyExpression). 

**Your Environments (__required__)**
Versions after this pr(https://github.com/vesoft-inc/nebula/pull/3255)

**How To Reproduce(__required__)**
Just check all expression related OptRule code logic.

**Additional context**
1. Some logic of this pr maybe also need been retested
```
(root@nebula) [nba]> with {a:1, b:{c:3, d:{e:5}}} as x return x.b.d.e
[ERROR (-1009)]: SemanticError: The type of `x' should be tag
```

2. Expression design should be reconsidered
    - reconsider data dependency design
    - avoid redundant design of expression types, eg. VarProperty/InputProperty/LabelTagProperty/LabelProperty, etc
    
3. Define the promises to optimizer to avoid errors like this pr
    - what types of plannode/expression should be taken into account (eg. `LabelTagProperty` or `VarProperty` in this pr)
    - data dependency and execution dependency conflict
    - column names should be set before optimization phase (this currently may cause some potential optimization rule bugs)
    - in what scenario dose reuse of executor dataset should be concerned(eg. Filter outputVar)
    - execution plans should be guaranteed to be stateless (eg. `validator` or `planner` setting some context states on which execution plans depend should be disabled)
    - expression conversion to `Value`, which is really different from expression rewriting, needs to be disabled or concerned. (eg. ColumnHints design)
    - clause level and planner's cutting of statements needed to be redesigned(eg. WHERE is not a clause in its own right — rather, it’s part of MATCH, OPTIONAL MATCH, START and WITH), this may improve some code logic related to `DataCollect`
    - ...

Shylock-Hg:
To avoid these happen again, we could add more test case which compare the optimized plan.

czpmango:
> To avoid these happen again, we could add more test cases which compare the optimized plan.

Each optimization rule has a corresponding test case to verify the optimized execution plan. 
This problem is not due to the lack of test cases, but rather the new features break some of the optimizer's guarantees.

PR #3736, opened 2022-01-17
fix with & optimizer rule error

nevermore3:
#### What type of PR is this?
- [x] bug
- [ ] feature
- [ ] enhancement

Issue(s) number: 

close https://github.com/vesoft-inc/nebula/issues/3704

Description:

CollapseProjectRule:

Test the following scenarios:
Project($-.v as v)<-Project(v.player.age)  optimized to Project( **$-.v**.player.age)
v.player.age is LabelTagPropertyExpression, where member variable **v** is a variablePropertyExpression
we should rewrite variablePropertyExpression($v)  to inputPropertyExpression($-.v) when this rule applies

#### Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

#### Release notes:

> ex. Fixed the bug .....


src/graph/validator/MatchValidator.cpp
```diff
@@ -819,7 +822,7 @@ Status MatchValidator::checkAlias(
       auto name = static_cast<const VariablePropertyExpression *>(labelExpr)->prop();
       auto res = getAliasType(aliasesAvailable, name);
       NG_RETURN_IF_ERROR(res);
-      if (res.value() != AliasType::kNode) {
+      if (res.value() == AliasType::kEdge || res.value() == AliasType::kPath) {
```
Aiee, line 825: What if `AliasType` is default?

nevermore3: it's the map type, it will not enter this branch

CPWstatic (approved):

jievince (approved):

PR #3738, opened 2022-01-17
Support more compiler launchers: distcc sccache

jiayuehua:
#### What type of PR is this?
- [ ] bug
- [ ] feature
- [x] enhancement

Issue(s) number: 

Description:
support more compiler launcher like distcc and sccache.

#### Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

#### Release notes:

> ex. Fixed the bug .....


cmake/nebula/CcacheConfig.cmake
```diff
@@ -1,16 +0,0 @@
-# To detect if ccache is available
-find_program(ccache_program_found "ccache")
```
darionyaphet, line 2: no longer using the `ccache` ?

yixinglu (approved):
LGTM

CPWstatic (approved):

PR #3741, opened 2022-01-17
Update client version to v3.0.0

Aiee:
#### What type of PR is this?
- [ ] bug
- [ ] feature
- [ ] enhancement

Issue(s) number: 

Description:
1. Update client version white list
2. Change the generated package name of python and golang

Depends on nebula-python update: https://github.com/vesoft-inc/nebula-python/pull/195

Please notice if this PR is merged then only clients with a version of v3.0.0 can connect to the service.

#### Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [X] N/A

Affects:

#### Release notes:

> This PR has been tested in the local environment with a personal nebula-python branch


src/graph/service/GraphFlags.cpp
```diff
@@ -56,7 +56,7 @@ DEFINE_bool(enable_optimizer, false, "Whether to enable optimizer");
 DEFINE_uint32(ft_request_retry_times, 3, "Retry times if fulltext request failed");
 DEFINE_bool(enable_client_white_list, true, "Turn on/off the client white list.");
 DEFINE_string(client_white_list,
-              nebula::getOriginVersion() + ":2.5.0:2.5.1:2.6.0",
+              nebula::getOriginVersion() + ":3.0.0",
```
HarrisChu, line 59: getOriginVersion would be 3.0.0 after we release, should we add this explicit version?

Aiee: I discussed this offline with @CPWstatic and we think explicitly adding the version is clearer.

Shylock-Hg (approved):

nevermore3 (approved):

PR #3775, opened 2022-01-21
Balance fix

liwenhui-soul:
#### What type of PR is this?
- [ ] bug
- [ ] feature
- [ ] enhancement

Issue(s) number: 

Description:

#### Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

#### Release notes:

> ex. Fixed the bug .....

### DIFF
--- a/src/meta/processors/admin/AdminClient.cpp
+++ b/src/meta/processors/admin/AdminClient.cpp
@@ -132,31 +132,20 @@ folly::Future<Status> AdminClient::addLearner(GraphSpaceID spaceId,
   return f;
 }
 
-folly::Future<Status> AdminClient::waitingForCatchUpData(GraphSpaceID spaceId,
-                                                         PartitionID partId,
-                                                         const HostAddr& target) {
+folly::Future<StatusOr<nebula::storage::cpp2::CatchUpResp>> AdminClient::waitingForCatchUpData(
+    GraphSpaceID spaceId, PartitionID partId, const HostAddr& target) {
   storage::cpp2::CatchUpDataReq req;
   req.space_id_ref() = spaceId;
   req.part_id_ref() = partId;
-  req.target_ref() = target;
-  auto partHosts = getPeers(spaceId, partId);
-  if (!nebula::ok(partHosts)) {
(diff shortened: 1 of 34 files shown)


src/meta/processors/admin/AdminClient.cpp
```diff
@@ -346,7 +368,7 @@ folly::Future<Status> AdminClient::getResponse(const HostAddr& host,
        respGen = std::move(respGen),
        this]() mutable {
         auto client = clientsMan_->client(host, evb);
-        remoteFunc(client, std::move(req))
+        remoteFunc(host, client, std::move(req))
```
critical27, line 371: If this changes is only for logging, why not just print them in related interfaces?

src/meta/processors/job/ZoneBalanceJobExecutor.cpp
```diff
@@ -305,16 +307,20 @@ Status ZoneBalanceJobExecutor::buildBalancePlan() {
...
+  }
 
   bool emty = std::find_if(existTasks.begin(),
                            existTasks.end(),
                            [](std::pair<const PartitionID, std::vector<BalanceTask>>& p) {
                              return !p.second.empty();
                            }) == existTasks.end();
-  if (emty || rc != nebula::cpp2::ErrorCode::SUCCEEDED) {
+  if (emty) {
+    executorOnFinished_(meta::cpp2::JobStatus::FINISHED);
```
critical27, line 320: I can't recall `executorOnFinished_`, what is this for? Why only some of code branch need to call `executorOnFinished_`?

And do we need to handle `LeaderBalanceJobExecutor`?

liwenhui-soul: > LeaderBalanceJobExecutor

finish or stop need to call executorOnFinished_, failed don't need to.

src/meta/processors/admin/AdminClient.cpp
critical27: 👌

liwenhui-soul:
test branch, close now

Issue #3782, opened 2022-01-21
db_upgrader crash when upgrade storage data 


HarrisChu:
1. use copy v2.6.1 cluster
2. replace bin folder
3. start metad
4. run command 
```
nebula1/bin/db_upgrader --src_db_path=/data/nebula/nebula_multiple_harris/nebula1/data/storage \
                --dst_db_path=/data/nebula/nebula_multiple_tmp/nebula1/data/storage/ \
                --upgrade_meta_server=192.168.15.13:1552,192.168.15.13:2552,192.168.15.13:3552 \
                --upgrade_version=2:3
```
db_upgrader crashed

```
[vesoft@vesoft-192-168-15-13 nebula_multiple_harris]$ nebula1/bin/nebula-graphd -version
nebula-graphd version Git: a06b059a, Build Time: Jan 21 2022 22:42:26
This source code is licensed under Apache 2.0 License.
```

```
(gdb) bt
#0  0x0000000001d24fc0 in rocksdb::DBImpl::NewIterator(rocksdb::ReadOptions const&, rocksdb::ColumnFamilyHandle*) ()
#1  0x00000000013f9d7d in nebula::kvstore::RocksEngine::prefixWithoutExtractor(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, void const*, std::unique_ptr<nebula::kvstore::KVIterator, std::default_delete<nebula::kvstore::KVIterator> >*) ()
#2  0x0000000001d0a7d7 in nebula::storage::UpgraderSpace::runPartV3() ()
#3  0x00000000024429f8 in folly::ThreadPoolExecutor::runTask(std::shared_ptr<folly::ThreadPoolExecutor::Thread> const&, folly::ThreadPoolExecutor::Task&&) ()
#4  0x000000000243520f in ?? ()
#5  0x00000000024ada81 in bool folly::AtomicNotificationQueue<folly::Function<void ()> >::drive<folly::EventBase::FuncRunner&>(folly::EventBase::FuncRunner&) ()
#6  0x00000000024aeaad in non-virtual thunk to folly::EventBaseAtomicNotificationQueue<folly::Function<void ()>, folly::EventBase::FuncRunner>::handlerReady(unsigned short) ()
#7  0x0000000002569993 in ?? ()
#8  0x000000000256a067 in event_base_loop ()
#9  0x00000000024a802e in folly::EventBase::loopBody(int, bool) ()
[...]

PR #3785, opened 2022-01-23
Fix iterator with snapshot(parameter position)

SuperYoko:
#### What type of PR is this?
- [x] bug
- [ ] feature
- [ ] enhancement

Issue(s) number: 
#3782
Description:
see related issue.

#### Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

#### Release notes:

> ex. Fixed the bug .....


HarrisChu:
fix https://github.com/vesoft-inc/nebula/issues/3782#issuecomment-1019487975

critical27:
If you think `void* ` is dangerous, maybe just remove the snapshot related interface from `KVStore` and `KVEngine`, just only implement them in `NebulaStore` and `RocksEngine`. This could be considered in the new version of GQL. For now, it is ok.

critical27:
Maybe we could also move `KVIterator` in the `NebulaStore::prefix` from parameter to return value later as well

critical27 (approved):

panda-sheep (approved):

Issue #3725, opened 2022-01-14
delete `MetaHttpReplaceHostHandler` after br release ?


HarrisChu:
as we add zone and `add hosts`, `MetaHttpReplaceHostHandler` has no effort at all.
shall we remove it?

critical27:
Talked with @pengweisong offline, there is already a interfaces which handle hosts correctly, will reuse it later

PR #3794, opened 2022-01-24
fix meta http replace host handler

pengweisong:
#### What type of PR is this?
- [x] bug
- [ ] feature
- [ ] enhancement

Issue(s) number: 
#3725

Description:

After adding zone, we should replace host info in zone table when replace host.

#### Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

#### Release notes:

> ex. Fixed the bug .....


src/meta/http/MetaHttpReplaceHostHandler.cpp
```diff
@@ -156,12 +171,54 @@ bool MetaHttpReplaceHostHandler::replaceHost(std::string ipv4From, std::string i
...
+
+  std::vector<nebula::kvstore::KV> data;
+  while (iter->valid()) {
+    bool needUpdate = false;
+    auto zoneName = MetaKeyUtils::parseZoneName(iter->key());
+    auto hosts = MetaKeyUtils::parseZoneHosts(iter->val());
+    for (auto& host : hosts) {
+      if (host.host == ipv4From) {
+        host.host = ipv4To;
+        needUpdate = true;
```
SuperYoko, line 200: break out?

SuperYoko: Better check that, I do not know the constraints on hosts.

pengweisong: After checking, I'll we could not add break. Because multi-services could lay in the same host, in which they share the same ip.

SuperYoko (approved):

critical27 (approved):
Good job

panda-sheep (approved):

Issue #2986, opened 2021-09-30
When creating user, the username can‘t be a string in double quotes, that's different with password


Sophie-Xie:
now
```
create user test2 with password "12"
```

suggest
```
create user "test1" with password "12"
```

https://discuss.nebula-graph.com.cn/t/topic/3732/2

Ishvina:
Hi! I am in an intro to swe class that is requiring me to work on my first open-source project. I was wondering if this issue is still open and if so could is this beginner-friendly?

wenhaocs:
> Hi! I am in an intro to swe class that is requiring me to work on my first open-source project. I was wondering if this issue is still open and if so could is this beginner-friendly?

Hi, thanks for looking into this. This issue is definitely a good beginner issue. I think it's still open. Please confirm it. @Sophie-Xie

Sophie-Xie:
> Hi! I am in an intro to swe class that is requiring me to work on my first open-source project. I was wondering if this issue is still open and if so could is this beginner-friendly?

@Ishvina Thanks. This issue is still open.  If you are interested, you can try it.
And I think we should support three formats(now, "", '') for compatibility.

wey-gu:
@Ishvina welcome and thank you for choosing to join the nebula community as your first open-source project, this means a lot to us!

BR//Wey

lisareichelson:
Hi! Could I have some guidance on figuring out which files to look in to work on this bug? I am also new to open source and a bit overwhelmed haha

lisareichelson:
Hi! Also new to open source. Could I have some guidance on finding where this issue is occurring? I believe I've narrowed it down to src/graph/validator/ACLValidator.cpp. I can see that ValidateImpl() is checking the length of the username/password. I'm having trouble finding where the formats are checked for compatibility. Any help would be appreciated :)

wey-gu:
Dear @lisareichelson,

It should be related to:

https://github.com/vesoft-inc/nebula/blob/3094bea4611042c1acb4c040b91c1fc489cf0872/src/parser/parser.yy#L421-L424

Where `name_label` was typed as a label rather a `STRING`:

https://github.com/vesoft-inc/nebula/blob/3094bea4611042c1acb4c040b91c1fc489cf0872/src/parser/parser.yy#L3630-L3637

There are some posts in the nebula blog in tag source-code: https://nebula-graph.io/tags/source-code/ for your references, too.

Have fun!

Thanks!

[...]

PR #3799, opened 2022-01-24
changed label to string

lisareichelson:
#### What type of PR is this?
- [ ] bug
- [ ] feature
- [X] enhancement

Issue(s) number: #2986

Description:
"When creating user, the username can‘t be a string in double quotes, that's different with password"

#### How do you solve it?
Changed username "name_label" to be stored as a string rather then a label, which should allow for single and double quotes to be a part of the username.

  
#### Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [X] N/A

Affects:

#### Release notes:

> ex. Fixed the bug .....

### DIFF
--- a/src/parser/parser.yy
+++ b/src/parser/parser.yy
@@ -419,7 +419,7 @@ static constexpr size_t kCommentLengthLimit = 256;
 %%
 
 name_label
-    : LABEL { $$ = $1; }
+    : STRING { $$ = new std::string("$1"); }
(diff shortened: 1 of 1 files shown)


CPWstatic:
This is not the right way to fix #2986, you should fix the `create_user_sentence`.

xtcyclist:
This PR has been idle for too long. Close it for now. Please refer to #2986 for the conclusion on how the user name shall be treated.

PR #3808, opened 2022-01-25
Add http port for show

### LINKED ISSUE #3149: support use nGQL to show service's http_port and disable http2_port
when I want to use graphd's httpPort to check status or do other things, I only can read flagfile to find, can we get it through ngql query?

jackwener:
#### What type of PR is this?
- [ ] bug
- [ ] feature
- [x] enhancement

Issue(s) number:  close #3149

Description:

#### Checklist:
Tests:
- [x] Function test
- [ ] Performance test
- [ ] N/A

Affects:

#### Release notes:

> ex. Fixed the bug .....


nevermore3 (dismissed):

darionyaphet (dismissed):

src/graph/executor/admin/ShowHostsExecutor.cpp
```diff
@@ -10,6 +10,7 @@
 #include "common/time/ScopedTimer.h"
 #include "graph/context/QueryContext.h"
 #include "graph/planner/plan/Admin.h"
+#include "webservice/WebService.h"
```
yixinglu, line 13: DECLARE this flag rather than including the header directly.

yixinglu (approved):

jievince (approved):

PR #3816, opened 2022-01-26
fix v3.0.0 ci

zhaohaifei:
#### What type of PR is this?
- [x] bug
- [ ] feature
- [ ] enhancement

Issue(s) number: 

Description:

#### Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

#### Release notes:

> ex. Fixed the bug .....

### DIFF
--- a/tests/Makefile
+++ b/tests/Makefile
@@ -34,7 +34,7 @@ install-deps:
 install-nebula-py: install-deps
 	git clone --branch master https://github.com/vesoft-inc/nebula-python $(CURR_DIR)/nebula-python
 	cd $(CURR_DIR)/nebula-python \
-		&& pip3 install --user --upgrade . -i $(PYPI_MIRROR)
+		&& pip3 install --user --force-reinstall . -i $(PYPI_MIRROR)
 	rm -rf $(CURR_DIR)/nebula-python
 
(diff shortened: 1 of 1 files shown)


Sophie-Xie (approved):

tests/Makefile
```diff
@@ -34,7 +34,7 @@ install-deps:
 install-nebula-py: install-deps
 	git clone --branch master https://github.com/vesoft-inc/nebula-python $(CURR_DIR)/nebula-python
 	cd $(CURR_DIR)/nebula-python \
-		&& pip3 install --user . -i $(PYPI_MIRROR) --upgrade
+		&& pip3 install --user --force-reinstall . -i $(PYPI_MIRROR)
```
darionyaphet, line 37: so we need reinstall nebula python each time?

zhaohaifei: I think so. Some runners will keep the previous nebula-python, resulting in a version mismatch.

HarrisChu: github runner would cache $HOME, and we install python package with `--user` option...

zhaohaifei: yeah.

PR #3818, opened 2022-01-26
abandon upgrader for old version

cangfengzhs:
#### What type of PR is this?
- [ ] bug
- [x] feature
- [ ] enhancement

Issue(s) number: 

Description:
The 3.0 upgrader can only be upgraded from 2.0GA, and it is forbidden to upgrade from an older version. Prevent user misoperation

#### Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

#### Release notes:

> ex. Fixed the bug .....


src/tools/db-upgrade/DbUpgraderTool.cpp
```diff
@@ -10,39 +10,37 @@
...
 
        --upgrade_meta_server=<ip:port,...>
          A list of meta severs' ip:port separated by comma.
          Default: 127.0.0.1:45500
 
-       --upgrade_version=<1:2|2RC:2|2:3>
-         This tool can only upgrade 1.x data, 2.0 RC, or 2.0 GA data.
-         1:2        upgrade the data from 1.x to 2.0GA
-         2RC:2      upgrade the data from 2.0RC to 2.0GA
+       --upgrade_version=<2:3>
```
critical27, line 42: I don't know if we should delete this parameter, maybe keep it?

cangfengzhs: I think we should keep it. This is a reminder to users.

critical27 (approved):

liuyu85cn (approved):

PR #3819, opened 2022-01-26
raft fix: fix always loop when no log need to send

liwenhui-soul:
#### What type of PR is this?
- [x] bug
- [ ] feature
- [ ] enhancement

Issue(s) number: https://github.com/vesoft-inc/nebula/issues/3671

Description:
if one host succeed but others failed, raft part would try again for all hosts, including the successful host, which would lead the error E_RAFT_NO_WAL_FOUND and fall into infinite loop

#### Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

#### Release notes:

> ex. Fixed the bug .....

### DIFF
--- a/src/kvstore/raftex/Host.cpp
+++ b/src/kvstore/raftex/Host.cpp
@@ -270,9 +270,27 @@ Host::prepareAppendLogRequest() {
   VLOG(2) << idStr_ << "Prepare AppendLogs request from Log " << lastLogIdSent_ + 1 << " to "
           << logIdToSend_;
 
+  auto makeReq = [this]() -> std::shared_ptr<cpp2::AppendLogRequest> {
+    auto req = std::make_shared<cpp2::AppendLogRequest>();
+    req->space_ref() = part_->spaceId();
+    req->part_ref() = part_->partitionId();
+    req->current_term_ref() = logTermToSend_;
+    req->committed_log_id_ref() = committedLogId_;
+    req->leader_addr_ref() = part_->address().host;
+    req->leader_port_ref() = part_->address().port;
+    req->last_log_term_sent_ref() = lastLogTermSent_;
+    req->last_log_id_sent_ref() = lastLogIdSent_;
+    return req;
+  };
+
   // We need to use lastLogIdSent_ + 1 to check whether need to send snapshot
   if (UNLIKELY(lastLogIdSent_ + 1 < part_->wal()->firstLogId())) {
     return startSendSnapshot();
+  }
+
+  if (lastLogIdSent_ == logIdToSend_) {
+    auto req = makeReq();
+    return req;
   }
 
   if (lastLogIdSent_ + 1 > part_->wal()->lastLogId()) {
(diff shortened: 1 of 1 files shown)


src/kvstore/raftex/Host.cpp
```diff
@@ -280,7 +280,18 @@ Host::prepareAppendLogRequest() {
...
+    req->part_ref() = part_->partitionId();
+    req->current_term_ref() = logTermToSend_;
+    req->committed_log_id_ref() = committedLogId_;
+    req->leader_addr_ref() = part_->address().host;
+    req->leader_port_ref() = part_->address().port;
+    req->last_log_term_sent_ref() = lastLogTermSent_;
+    // the purpose of setting last_log_id to -1 is to make remote host return committedLogId which
+    // could be used as lastLogIdSent in next loop
+    req->last_log_id_sent_ref() = -1;
+    return req;
```
critical27, line 294: Em... I don't understand why the change in the PR could solve `lastLogIdSent_ + 1 > part_->wal()->lastLogId()`?

Even if the peer return `committedLogId`, but leader still doesn't have the wal, won't it still loop forever?

```
send a request with -1 
                                          -> return committedLogId
send a request with -1 
                                          -> return committedLogId
...
```

src/kvstore/raftex/RaftPart.cpp
```diff
@@ -1001,6 +1001,12 @@ void RaftPart::processAppendLogResponses(const AppendLogResponses& resps,
     LOG_EVERY_N(WARNING, 100) << idStr_ << "Only " << numSucceeded
                               << " hosts succeeded, Need to try again";
     usleep(1000);
+    {
+      std::lock_guard<std::mutex> g(raftLock_);
+      // wal could be rollback between two cycles, if we still use the old lastLogId, we may always
+      // get E_RAFT_NO_WAL_FOUND
+      lastLogId = std::min(wal_->lastLogId(), lastLogId);
```
critical27, line 1008: There are two question here:
1. If you only modify lastLogId, but the `iter` is not modified, I really don't know whether we can succeed.
2. We will check the if the term has changed in `replicateLogs` (if rollback really happen, the term must have been updated), so I really doubt whether the problem should be fixed in this way. Or you could describe it in more details.

liwenhui-soul: we got E_RAFT_NO_WAL_FOUND  because there are some logs we are going to send don't exist in wal, the reason is that we alway use the old lastLogId, the lastLogId may be greater than part_->wal()->lastLogId().so update the lastLogId every circle, and if the wal is updated in this circle, continue in next circle with new lastLogId, and maybe there is a time that part_->wal() would not be updated and we send the logs.

src/kvstore/raftex/Host.cpp
liwenhui-soul: yes, only when when committedLogId == part_->wal()->lastLogId() could lead this condition.
just set req->last_log_id_sent_ref() = lastLogIdSent_ could solve the problem.

src/kvstore/raftex/RaftPart.cpp
liwenhui-soul: actually,  host use a new iterator from the wal, the iter is not used any more.

src/kvstore/raftex/Host.cpp
```diff
@@ -270,11 +270,29 @@ Host::prepareAppendLogRequest() {
...
+  };
+
   // We need to use lastLogIdSent_ + 1 to check whether need to send snapshot
   if (UNLIKELY(lastLogIdSent_ + 1 < part_->wal()->firstLogId())) {
     return startSendSnapshot();
   }
 
+  if (lastLogIdSent_ == logIdToSend_) {
+    auto req = makeReq();
+    return req;
```
panda-sheep, line 293: I am very curious, why do you need to use `AppendLogRequest` to send empty data, is it a waste of rpc?
Do not send AppendLogRequest, or return directly in RaftPart::replicateLogs?

liwenhui-soul: > I am very curious, why do you need to use `AppendLogRequest` to send empty data, is it a waste of rpc? Do not send AppendLogRequest, or return directly in RaftPart::replicateLogs?
if change leader and change back, the remote's logs may rollback, we must confirm the remote host really don't need logs.

critical27 (approved):
LGTM, a good start~

kikimo:
Excellent, and it would be perfect if you could elaborate the problem in detail.
I'm testing now.

critical27:
Don't merge plz, some test is ongoing

kikimo:
HOLD THIS PR: we found that raft cluster failed to elect a leader after applying this pr, still investigating

kikimo (approved):
Tested, LGTM

PR #3834, opened 2022-01-27
time-trace option for clang

jiayuehua:
## What type of PR is this?
- [ ] bug
- [*] feature
- [ ] enhancement

#### Description:
add -time-trace option for clang when ENABLE_BUILD_WITH_TIME_TRACE is on.

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....


cmake/nebula/TimeTrace.cmake
```diff
@@ -0,0 +1,6 @@
+if(CMAKE_CXX_COMPILER_ID MATCHES ".*Clang")
+  option(ENABLE_BUILD_WITH_TIME_TRACE "Enable -ftime-trace to generate time tracing .json files on clang" OFF)
```
darionyaphet, line 2: It is off by default？

jiayuehua: yes, it's for developer to trace compile time, not intend to be used for user.

darionyaphet: reasonable

czpmango (dismissed):

Aiee (dismissed):

darionyaphet (dismissed):

yixinglu (approved):

Aiee (approved):

PR #3836, opened 2022-01-27
Erase all symbols of erased group node.

### LINKED ISSUE #3826: Optimizer now only release symbols of top layer nodes in all matched nodes.

* OS: `uname -a`
* CPU: `lscpu`
* Commit id (e.g. `a3ffc7d8`)

1. Step 1
2. Step 2
3. Step 3

**Expected behavior**

**Additional context**

Shylock-Hg:
## What type of PR is this?
- [x] bug
- [ ] feature
- [ ] enhancement

#### Issue(s) number: 
Close #3826 

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....


src/graph/optimizer/OptGroup.cpp
```diff
@@ -81,8 +81,12 @@ Status OptGroup::explore(const OptRule *rule) {
     NG_RETURN_IF_ERROR(resStatus);
     auto result = std::move(resStatus).value();
     if (result.eraseAll) {
-      for (auto gnode : groupNodes_) {
-        gnode->node()->releaseSymbols();
+      matched.eraseAll();  // erase current mached sub-plan
```
yixinglu, line 84: Different OptNodes could share the same OptGroup as their dependencies. If one OptNode should be removed, we could not delete the OptGroup depended by it directly.

Shylock-Hg: As I known, the replaced sub-plan which matched is invisible again in the whole plan. Delete the information of them in symbol table is unsafe?

yixinglu: Not the sub-plan, only the OptNode in the same OptGroup will be removed. The dependent OptGroup of the OptNode maybe be shared with other OptNodes in different OptGroups.

Shylock-Hg: But current implementation contains error. Optimizer will release the symbols(readBy, writtenBy) of replaced nodes, but now just release the top-layer nodes instead of all matched nodes. So some deprecated symbols(readBy, writtenBy) will keep alive in symbol table, which will prevent some valid optimization.

Shylock-Hg:
Maybe could clean useless symbols when generate the final plan.

Shylock-Hg:
Addition, we can't clean symbols after optimization, because the unused symbols prevent some valid optimization.

Issue #3774, opened 2022-01-20
reduce reinventing the wheel. ie, reduce use of code in src/common, util someday delete this folder totally.


jiayuehua:
**Introduction**

src/common contains many code that implemented not very well, that is some bad wheel be reinvented. Should reduce use of code in src/common, use standard lib or thirdparty lib instead.
**Contents**

For example, apply those changes :
concurrent -> boost thread
charset -> boost text or [ztd.text ](https://ztdtext.readthedocs.io/en/latest/)
thread -> boost thread
base/status statusor -> abseil::status abseil::statusor 
network -> boost asio
cpp/helpers.h -> boost noncopyable
time -> chrono 
encryption -> openssl
fs -> std filesystem
...
etc

**Related work**

Shylock-Hg:
The utilities of time are aimed to separate some environment like TZ from system. The standard library don't expose it, so I write new one.

jiayuehua:
https://github.com/HowardHinnant/date merged into c++20, in c++17 can use this lib instead.

jiayuehua:
https://github.com/vesoft-inc/nebula/pull/3867

jiayuehua:
https://github.com/vesoft-inc/nebula/pull/3848

PR #3848, opened 2022-01-29
remove concurrent   base/stringvalue, use boost::barrier instead

jiayuehua:
## What type of PR is this?
- [x] bug
- [ ] feature
- [x] enhancement

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....


third-party/install-gcc.sh
```diff
@@ -1,115 +0,0 @@
-#! /usr/bin/env bash
```
Shylock-Hg, line 1: Why delete this?

jiayuehua: now use system gcc or user install itselt

Shylock-Hg: User of course could try gcc him want. And this script is used by our infra now, plz check it first instead of delete directly.

darionyaphet:
it seems replace the concurrent with boost thread library? if so please update the title. thanks ~

zhaohaifei (approved):
look good to me

third-party/install-gcc.sh
yixinglu: It's better not to delete the script until we've finished the discussion about compiler usage.

src/common/cpp/helpers.h
```diff
@@ -8,24 +8,18 @@
 namespace nebula {
 namespace cpp {
 
-class NonCopyable {
- protected:
-  NonCopyable() {}
-  NonCopyable(const NonCopyable&) = delete;
-  NonCopyable& operator=(const NonCopyable&) = delete;
-};
+class NonMovable {
```
yixinglu, line 11: Is there no `nonmovable` implementation in boost ?

src/graph/optimizer/test/CMakeLists.txt
```diff
@@ -5,7 +5,7 @@
 
 set(OPTIMIZER_TEST_LIB
     $<TARGET_OBJECTS:base_obj>
-    $<TARGET_OBJECTS:concurrent_obj>
+
```
yixinglu, line 8: cleanup

src/graph/planner/test/CMakeLists.txt
```diff
@@ -42,7 +42,7 @@ nebula_add_test(
         $<TARGET_OBJECTS:time_obj>
         $<TARGET_OBJECTS:fs_obj>
         $<TARGET_OBJECTS:base_obj>
-        $<TARGET_OBJECTS:concurrent_obj>
+
```
yixinglu, line 45: ditto

src/kvstore/LogEncoder.h
```diff
@@ -141,10 +142,7 @@ HostAddr decodeHost(LogType type, const folly::StringPiece& encoded);
  */
 int64_t getTimestamp(const folly::StringPiece& log);
 
-/**
- * @brief A wrapper class of batchs of log, support put/remove/removeRange
- */
-class BatchHolder : public nebula::cpp::NonCopyable, public nebula::cpp::NonMovable {
+class BatchHolder : public boost::noncopyable, public nebula::cpp::NonMovable {
```
yixinglu, line 145: It seems that the comment of this class should not be deleted.

third-party/install-gcc.sh
jiayuehua: ok.

src/common/cpp/helpers.h
jiayuehua: yes I search boost and folly codebase ,find nonthing.

src/kvstore/LogEncoder.h
jiayuehua: has reverted

zhaohaifei (approved):

Aiee (approved):

yixinglu (approved):

jiayuehua:
https://github.com/vesoft-inc/nebula/issues/3774

PR #3855, opened 2022-02-07
skip heartbeat when role in meta client is not set

### LINKED ISSUE #3854: cannot backup nebula after upgrader from v2
1. db_upgrader would create a metaclient, and send heart beat to meta.
2. option doesn't set host and role.
3. and it would make backup failed.

image

image
image

critical27:
## What type of PR is this?
- [X] bug
- [ ] feature
- [ ] enhancement

#### Issue(s) number: 
Close #3854

#### Description:
The role in meta client is `UNKNOWN` by default, and many tools use the default `MetaClientOptions`, which makes the host info is `Host("", 0)`, and the info will be write to meta if we send heartbeat. 

## How do you solve it?
Just skip heartbeat when role is unknown

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....

### DIFF
--- a/src/clients/meta/MetaClient.cpp
+++ b/src/clients/meta/MetaClient.cpp
@@ -94,15 +94,19 @@ MetaClient::~MetaClient() {
 }
 
 bool MetaClient::isMetadReady() {
-  auto ret = heartbeat().get();
-  if (!ret.ok()) {
-    LOG(ERROR) << "Heartbeat failed, status:" << ret.status();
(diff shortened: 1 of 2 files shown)


wenhaocs (approved):

darionyaphet (approved):
LGTM

PR #3857, opened 2022-02-08
Try to improve the performance of appendVertices

jievince:
## What type of PR is this?
- [ ] bug
- [ ] feature
- [x] enhancement

#### Description:
The case `0-hop`, 3.0's performance is much lower than 2.6
```
MATCH (v: Person) return v
```
While the case `1-hop` and `2-hop`, 3.0's performance is flat with 2.0
The execution plans are here:
[2.6.txt](https://github.com/vesoft-inc/nebula/files/8020543/2.6.txt)
[3.0.txt](https://github.com/vesoft-inc/nebula/files/8020544/3.0.txt)

### DIFF
--- a/tests/tck/features/match/Base.feature
+++ b/tests/tck/features/match/Base.feature
@@ -704,7 +704,7 @@ Feature: Basic match
     When executing query:
       """
       USE nba;
-      MATCH (n:player) RETURN n LIMIT 1;
+      MATCH (n:player) WHERE id(n) == "Boris Diaw" RETURN n;
       """
     Then the result should be, in any order:
       | n                                                   |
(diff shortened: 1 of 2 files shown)


tests/tck/features/match/Base.feature
```diff
@@ -704,7 +704,7 @@ Feature: Basic match
     When executing query:
       """
       USE nba;
-      MATCH (n:player) RETURN n LIMIT 1;
```
Shylock-Hg, line 707: Why change this?

jievince: `LIMIT` is not stable.

yixinglu (approved):

czpmango (approved):
LGTM

Aiee (approved):

PR #3868, opened 2022-02-09
Forbid grant root user

### LINKED ISSUE #3235: Root user should not be reassigned other roles in space
At now root user can be assigned to other roles in a space, it's confused for user which the real access he owns in the space? 

Accoding to my test, if I'm the root user of one nebula db which includes a space `test`
```
$ GRANT ROLE User ON test TO root
```
SHOW ROLES IN test; It's seems that I'm the `user` role of the space `test`, but I can modified the schema becaused of the db root role.

 User Role in space
image

`Create TAG` is allowed
image

**Additional context**
According to the confusing behavior of the role managing, it may cause some problems in applications about nebula.🤝

jackwener:
## What type of PR is this?
- [ ] bug
- [ ] feature
- [x] enhancement

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....


src/graph/validator/test/ACLValidatorTest.cpp
```diff
@@ -142,24 +142,6 @@ TEST_F(ACLValidatorTest, Simple) {
     ASSERT_EQ(*changePassword->newPassword(), newPassword);
   }
 
-  // grant role
```
Shylock-Hg, line 145: Why delete this?

src/graph/validator/ACLValidator.cpp
```diff
@@ -104,10 +104,18 @@ Status ChangePasswordValidator::toPlan() {
 // grant role
 Status GrantRoleValidator::validateImpl() {
   const auto *sentence = static_cast<const GrantSentence *>(sentence_);
-  if (sentence->getAccount()->size() > kUsernameMaxLength) {
+  const std::string *account = sentence->getAccount();
+  if (account->size() > kUsernameMaxLength) {
     return Status::SemanticError("Username exceed maximum length %ld characters.",
                                  kUsernameMaxLength);
   }
+  auto metaClient = qctx_->getMetaClient();
```
Shylock-Hg, line 112: You could get roles from session directly.

Shylock-Hg:
Need add test case in tck.

src/graph/validator/ACLValidator.cpp
Shylock-Hg: Sorry, it's my mistake, the granted user is `Account` from sentence and is different from current user of session.

src/graph/validator/test/ACLValidatorTest.cpp
jackwener: Because there is metaclient in this function.

Shylock-Hg: This case should keep.

src/graph/validator/test/ACLValidatorTest.cpp
```diff
@@ -142,24 +142,6 @@ TEST_F(ACLValidatorTest, Simple) {
...
-  {
```
Shylock-Hg, line 146: This case should keep.

Shylock-Hg (approved):

yixinglu (approved):

Shylock-Hg (approved):

yixinglu (approved):

PR #3880, opened 2022-02-11
if leader changed during raft::appendLog() RPC, use a different error code

liuyu85cn:
## What type of PR is this?
- [X] bug
- [ ] feature
- [ ] enhancement

#### Description:

It is nature to return E_RAFT_TERM_OUT_OF_DATE or LEADER_CHANGE
when raft check leader, but this will make caller believe the append failed.

however, this log is replicated to followers(written in follower's WAL)
and the follower may become leader (use that WAL)
which means this log may eventually commit succeeded.

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

n/a


panda-sheep (approved):
Good！

critical27 (approved):
LGTM

PR #3890, opened 2022-02-14
Expose rocksdb feature of disabling OS page cache

wenhaocs:
## What type of PR is this?
- [ ] bug
- [x] feature
- [ ] enhancement

#### Description:
Expose rocksdb option of disabling os page cache.
Since OS page cache may be unstable, this feature is very useful in testing and in some production environment which require all memory controllable.

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> Users can now disable the OS page cache and only use block cache and nebula storage cache in shared environment, to avoid interference of the memory footprints between applications.


yixinglu (approved):

panda-sheep (approved):
Good job！

PR #3892, opened 2022-02-15
add nebula-console package action

czpmango:
## What type of PR is this?
- [ ] bug
- [x] feature
- [ ] enhancement

#### Description:
Package nebula-console into an RPM /deb package through action.

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:
 - The RPM/DEB package will contain the nebula-console binary


.github/workflows/nightly.yml
```diff
@@ -33,6 +33,9 @@ jobs:
         with:
           run: sh -c "find . -mindepth 1 -delete"
       - uses: actions/checkout@v2
+      - uses: actions/setup-go@v2
+      - name: compile nebula-console
```
yixinglu, line 37:
```yaml
- uses: actions/checkout
  with:
     repository: vesoft-inc/nebula-console
     ref: ${{ github.ref }}
     directory: nebula-console
- run: |
      make && mv nebula-console ../build/bin/nebula
   working-directory: nebula-console
```

yixinglu: maybe it's better to do it by cmake

.github/workflows/nightly.yml
```diff
@@ -33,6 +33,9 @@ jobs:
         with:
           run: sh -c "find . -mindepth 1 -delete"
       - uses: actions/checkout@v2
+      - uses: actions/setup-go@v2
```
yixinglu, line 36: specify golang version

.github/workflows/nightly.yml
czpmango: ACK.

.github/workflows/nightly.yml
czpmango: ACK.

czpmango:
Closed because A better implementation is here https://github.com/vesoft-inc/nebula/pull/3905

HarrisChu:
we should prevent thrift version are different between nebula graph and nebula console.

PR #3904, opened 2022-02-16
Merge locks into one

darionyaphet:
## What type of PR is this?
- [ ] bug
- [ ] feature
- [X] enhancement

#### Description:

Using a larger lock to control concurrent operation.

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....

### DIFF
--- a/src/meta/processors/Common.h
+++ b/src/meta/processors/Common.h
@@ -14,32 +14,11 @@ namespace meta {
 class LockUtils {
  public:
   LockUtils() = delete;
-#define GENERATE_LOCK(Entry)                 \
-  static folly::SharedMutex& Entry##Lock() { \
-    static folly::SharedMutex l;             \
-    return l;                                \
+
+  static folly::SharedMutex& lock() {
+    static folly::SharedMutex lock;
+    return lock;
   }
-
-  GENERATE_LOCK(lastUpdateTime);
-  GENERATE_LOCK(space);
-  GENERATE_LOCK(id);
-  GENERATE_LOCK(workerId);
-  GENERATE_LOCK(localId);
-  GENERATE_LOCK(tagAndEdge);
-  GENERATE_LOCK(tagIndex);
-  GENERATE_LOCK(edgeIndex);
-  GENERATE_LOCK(service);
-  GENERATE_LOCK(fulltextIndex);
-  GENERATE_LOCK(user);
-  GENERATE_LOCK(config);
-  GENERATE_LOCK(snapshot);
-  GENERATE_LOCK(group);
-  GENERATE_LOCK(zone);
-  GENERATE_LOCK(listener);
-  GENERATE_LOCK(session);
-  GENERATE_LOCK(machine);
-
-#undef GENERATE_LOCK
 };
 
 }  // namespace meta
--- a/src/meta/processors/Common.h
+++ b/src/meta/processors/Common.h
@@ -19,6 +19,16 @@ class LockUtils {
     static folly::SharedMutex lock;
     return lock;
   }
+
+  static folly::SharedMutex& snapshotLock() {
+    static folly::SharedMutex snapshotLock;
+    return snapshotLock;
+  }
+
+  static folly::SharedMutex& sessionLock() {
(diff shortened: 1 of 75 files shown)


src/meta/processors/admin/HBProcessor.cpp
```diff
@@ -32,7 +32,7 @@ void HBProcessor::process(const cpp2::HBReq& req) {
   auto role = req.get_role();
   LOG(INFO) << "Receive heartbeat from " << host
             << ", role = " << apache::thrift::util::enumNameSafe(role);
-
+  folly::SharedMutex::WriteHolder holder(LockUtils::lock());
```
critical27, line 35: Since heartbeat's num is way more than other meta processor, do we really have to add write lock here? What we need to write here is only related to one host.

critical27 (commented):
You could do some test about the impact of latency. Especially about the latency.

src/meta/processors/admin/CreateBackupProcessor.cpp
```diff
@@ -239,11 +237,7 @@ void CreateBackupProcessor::process(const cpp2::CreateBackupReq& req) {
   backup.backup_name_ref() = std::move(backupName);
   backup.full_ref() = true;
   bool allSpaces = backupSpaces == nullptr || backupSpaces->empty();
-  if (allSpaces) {
-    backup.all_spaces_ref() = true;
-  } else {
-    backup.all_spaces_ref() = false;
-  }
+  backup.all_spaces_ref() = allSpaces;
```
SuperYoko, line 240: Much better~

src/meta/processors/Common.h
```diff
@@ -14,32 +14,11 @@ namespace meta {
...
-  GENERATE_LOCK(config);
-  GENERATE_LOCK(snapshot);
-  GENERATE_LOCK(group);
-  GENERATE_LOCK(zone);
-  GENERATE_LOCK(listener);
-  GENERATE_LOCK(session);
-  GENERATE_LOCK(machine);
 
-#undef GENERATE_LOCK
+  static folly::SharedMutex& lock() {
```
SuperYoko, line 18: static method may have different version.

src/meta/processors/admin/HBProcessor.cpp
darionyaphet: Mainly to prevent inconsistencies when reading data, this writing method is more strict.

darionyaphet:
> You could do some test about the impact of latency. Especially about the latency.

Because it was a segmented lock structure before, and now it is modified to a large lock, the increase in delay is mainly in the position of competing locks. So what actually needs to be evaluated is the impact of multiple requests competing for locks.

liuyu85cn (approved):
good job, LGTM

liuyu85cn (dismissed):

src/meta/processors/admin/CreateSnapshotProcessor.cpp
```diff
@@ -31,7 +31,7 @@ void CreateSnapshotProcessor::process(const cpp2::CreateSnapshotReq&) {
   }
 
   auto snapshot = folly::sformat("SNAPSHOT_{}", MetaKeyUtils::genTimestampStr());
-  folly::SharedMutex::WriteHolder wHolder(LockUtils::snapshotLock());
+  folly::SharedMutex::WriteHolder holder(LockUtils::lock());
```
panda-sheep, line 34: Ask a question:
create snapshot and create backup is a time-consuming operation. This lock will be held here for a long time.
During this period, other heartbeats may fail?

darionyaphet: Good point. I don't know the time-consuming in the case of a large amount of data.

panda-sheep (approved):
Good job

liuyu85cn (dismissed):

critical27 (approved):
Good job

panda-sheep (approved):

Issue #3667, opened 2022-01-07
Speed up the compilation


Sophie-Xie:
1. Coding Principles (.h, statement......)
2.history cleaning
3. Built-in tool check

czpmango:
Fwiw, ninja is not as fast as expected and even becomes slower than make.

jiayuehua:
https://github.com/vesoft-inc/nebula/pull/3738 add distcc support

PR #3906, opened 2022-02-16
clean up include files

jiayuehua:
## What type of PR is this?
- [*] bug
- [ ] feature
- [*] enhancement

#### Issue(s) number: 
https://github.com/vesoft-inc/nebula/issues/3667
## How do you solve it?
use iwyu first, then by hand.

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....


dutor:
There are too many redundant `include` directives, although of which the content will not be parsed multiple times or in vain. This situation is possibly inroduced by the pattern below:
```cpp
// file b
#include <a>
// file c.cpp
#include <a>  // already included within file b
#include <b>
```

This PR may help to speed up the build. But I don't think it's a good thing that _include_ directives make up nearly 10% of LOC in our codebase. This is exactly what @sherman-the-tank concerned about.

PR #3910, opened 2022-02-17
add optimizer comments

czpmango:
## What type of PR is this?
- [ ] bug
- [ ] feature
- [x] enhancement

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:


src/graph/optimizer/rule/CollapseProjectRule.h
```diff
@@ -13,16 +13,36 @@ DECLARE_bool(enable_optimizer_collapse_project_rule);
...
-/**
- * Combines two [[Project]] nodes into one
- * Required conditions:
- *  1. Match the pattern
- *  2. Expressions between nodes cannot be referenced more than once
- * Benefits:
- *  1. reduce the copy of memory between nodes
- *  2. reduces expression overhead in some cases(similar to column pruning)
- */
+/*
```
Aiee, line 16: I believe we should use `//`

yixinglu (dismissed):

src/graph/optimizer/rule/CollapseProjectRule.h
czpmango: Any reference?
Google style?

CPWstatic: Google style doesn't declare that the comments should use `//`, but it suggests that the comments style should be unified. I think it makes sense. And the most open source projects i had ever seen use the `//`. For now i have seen much styles in our project, it's too ugly.
```
// 

/*

*/

/**
 *
 */
```

czpmango: ACK.

src/graph/optimizer/OptContext.h
```diff
@@ -48,8 +48,10 @@ class OptContext final : private boost::noncopyable, private cpp::NonMovable {
   const OptGroupNode *findOptGroupNodeByPlanNodeId(int64_t planNodeId) const;
 
  private:
+  // A global flag to record whether this iteration caused a change to the plan
   bool changed_{true};
   graph::QueryContext *qctx_{nullptr};
+  // Memo memory management in the Optimizer phase
```
jievince, line 54: What is memo?

jievince (approved):

src/graph/optimizer/OptContext.h
Aiee, line 54: typo?

src/graph/optimizer/rule/PushLimitDownScanAppendVerticesRule.h
```diff
@@ -11,6 +11,45 @@
...
+//  +--------+--------+
+//           |
+// +---------+---------+
+// |   AppendVertices  |
+// +---------+---------+
+//           |
+// +---------+---------+
+// |    ScanVertices   |
+// |     (limit=3)     |
+// +---------+---------+
```
Aiee, line 51: Will there still be a limit node after optimization?

src/graph/optimizer/OptContext.h
Shylock-Hg, line 54: What's this?

yixinglu (approved):

PR #3911, opened 2022-02-17
Merge active host manager write operations into one

darionyaphet:
## What type of PR is this?
- [ ] bug
- [ ] feature
- [ ] enhancement

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....


critical27 (approved):

src/meta/processors/job/BalanceJobExecutor.h
```diff
@@ -147,6 +147,28 @@ class BalanceJobExecutor : public MetaJobExecutor {
     return Status::OK();
   }
 
+  /**
+   * @brief Update last time in meta service
+   *
+   * @return
+   */
+  nebula::cpp2::ErrorCode updateLastTime() {
+    std::vector<kvstore::KV> data;
```
critical27, line 156: Add lock here~

src/meta/processors/job/ZoneBalanceJobExecutor.cpp
```diff
@@ -45,11 +45,6 @@ folly::Future<Status> ZoneBalanceJobExecutor::executeInternal() {
     }
   }
   plan_->setFinishCallBack([this](meta::cpp2::JobStatus status) {
-    folly::SharedMutex::WriteHolder holder(LockUtils::lock());
```
critical27, line 48: Why delete this

src/meta/ActiveHostsMan.cpp
```diff
@@ -258,36 +249,5 @@ ErrorOr<nebula::cpp2::ErrorCode, HostInfo> ActiveHostsMan::getHostInfo(kvstore::
   return HostInfo::decode(hostValue);
 }
 
-nebula::cpp2::ErrorCode LastUpdateTimeMan::update(kvstore::KVStore* kv,
```
critical27, line 261: We'd better save this interface, just modify the parameter type. It would be implicit that we write a `lastUpdateTimeKey`, the `LastUpdateTimeMan::update` is more explicit.

src/meta/processors/BaseProcessor-inl.h
```diff
@@ -13,17 +13,21 @@ namespace nebula {
...
+  auto ret = nebula::cpp2::ErrorCode::SUCCEEDED;
   kvstore_->asyncMultiPut(kDefaultSpaceId,
                           kDefaultPartId,
                           std::move(data),
-                          [this, &baton](nebula::cpp2::ErrorCode code) {
-                            this->handleErrorCode(code);
+                          [&ret, &baton](nebula::cpp2::ErrorCode code) {
+                            if (nebula::cpp2::ErrorCode::SUCCEEDED != code) {
+                              ret = code;
+                              LOG(INFO) << "Put data error on meta server";
```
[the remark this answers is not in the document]
darionyaphet: Other methods, such as `doGet` `doMultiGet` and `doPrefix` the log level is `VLOG(2)` so we should consistent with them

src/meta/processors/job/BalanceJobExecutor.h
darionyaphet: the lock have been support before it be called

src/meta/ActiveHostsMan.cpp
darionyaphet: I intend to eliminate the problem of writing many times.If a function only retains the function of assembling key values, should it really exist?

critical27 (approved):

critical27:
LGTM, good job

liuyu85cn (approved):

PR #3917, opened 2022-02-18
use rcu in DiskManager

liwenhui-soul:
## What type of PR is this?
- [ ] bug
- [ ] feature
- [ ] enhancement

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....

### DIFF
--- a/src/kvstore/DiskManager.h
+++ b/src/kvstore/DiskManager.h
@@ -15,6 +15,7 @@
 #include "common/base/StatusOr.h"
 #include "common/thread/GenericWorker.h"
 #include "common/thrift/ThriftTypes.h"
+#include "folly/synchronization/Rcu.h"
 #include "interface/gen-cpp2/meta_types.h"
 
 namespace nebula {
@@ -41,6 +42,8 @@ class DiskManager {
   DiskManager(const std::vector<std::string>& dataPaths,
               std::shared_ptr<thread::GenericWorker> bgThread = nullptr);
 
+  ~DiskManager();
+
   /**
    * @brief return canonical data path of given space
    *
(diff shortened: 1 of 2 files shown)


src/kvstore/DiskManager.h
```diff
@@ -90,37 +93,41 @@ class DiskManager {
...
-  // given a space and data path, return all parts in the path
-  std::unordered_map<GraphSpaceID, PartDiskMap> partPath_;
-
-  // the index in dataPaths_ for a given space + part
-  std::unordered_map<GraphSpaceID, std::unordered_map<PartitionID, size_t>> partIndex_;
-
   // lock used to protect partPath_ and partIndex_
   std::mutex lock_;
+
+  mutable folly::rcu_domain<> rcuDomain_;
```
critical27, line 130: Maybe you don't need to declare a `rcu_domain` explicitly, the default one is enough.

src/kvstore/DiskManager.h
```diff
@@ -90,37 +93,41 @@ class DiskManager {
...
   std::vector<std::atomic_uint64_t> freeBytes_;
 
-  // given a space and data path, return all parts in the path
-  std::unordered_map<GraphSpaceID, PartDiskMap> partPath_;
-
-  // the index in dataPaths_ for a given space + part
-  std::unordered_map<GraphSpaceID, std::unordered_map<PartitionID, size_t>> partIndex_;
-
   // lock used to protect partPath_ and partIndex_
   std::mutex lock_;
```
critical27, line 128: I think we don't need the mutex now?

src/kvstore/DiskManager.h
liwenhui-soul: > Maybe you don't need to declare a `rcu_domain` explicitly, the default one is enough.

metaclient also use default, they share the same one?

src/kvstore/DiskManager.h
liwenhui-soul: > I think we don't need the mutex now?

write & write need it

critical27: 👌 , got it

critical27 (approved):
Good job~

cangfengzhs (approved):
LGTM

PR #3926, opened 2022-02-22
Memory lock in raft

liuyu85cn:
## What type of PR is this?
- [ ] bug
- [ ] feature
- [X] enhancement

#### Description:

Let add tag / edge use atomic Op again. 

1. before and include Nebula 2.0, we use atomicOp to deal with some atomic operation,
    e.g. change a tag/edge and its index in a batch. 

    It works, but as we implement this by send raft log in a sync way. 
    (all atomic op should be sent seperately, even if they are disjoint.) this is really slow.

2. In 2.6.x we use memory lock to deal with concurrent control. 
    We check early(in processor) if a request can run or not. 
    If it can, then we do the get/put as a normal log, which we can treat in batch.
    if it can't, we return error. 

    How ever, some user complain that they meet so many "Conflict error". 
    That they need to retry, they believe it will slow down the bulk insert.
    We explained that those conflict has to be retry either in Nebula it self or client, 
    but  it looks like they didn't agree with us. 

3. So now we implement a hybrid mode for this. 
    We had a memory lock in raft. just like Solution2. we check every logs to see if it can be combine with previous logs.
    If it can, then we send them in batch. 
    if it can't, then we treat it like the atomicOp way (Solution1).

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:
- [X] Performance impacted: Consumes more CPU/Memory

## Release notes:

As describe in the "Description", conflict concurrent insert tag/edge will not report "Data conflict".
But execute in a queue.

### DIFF
--- a/src/kvstore/raftex/RaftPart.cpp
+++ b/src/kvstore/raftex/RaftPart.cpp
@@ -771,7 +771,7 @@ folly::Future<nebula::cpp2::ErrorCode> RaftPart::appendLogAsync(ClusterID source
   if (bufferOverFlow_) {
     VLOG_EVERY_N(2, 1000)
         << idStr_ << "The appendLog buffer is full. Please slow down the log appending rate."
-        << "replicatingLogs_ :" << std::boolaplha << replicatingLogs_;
+        << "replicatingLogs_ :" << std::boolalpha << replicatingLogs_;
     return nebula::cpp2::ErrorCode::E_RAFT_BUFFER_OVERFLOW;
   }
   {
@@ -782,7 +782,7 @@ folly::Future<nebula::cpp2::ErrorCode> RaftPart::appendLogAsync(ClusterID source
     if (logs_.size() >= FLAGS_max_batch_size) {
       // Buffer is full
       VLOG(2) << idStr_ << "The appendLog buffer is full. Please slow down the log appending rate."
-              << "replicatingLogs_ :" << std::boolaplha << replicatingLogs_;
+              << "replicatingLogs_ :" << std::boolalpha << replicatingLogs_;
       bufferOverFlow_ = true;
       return nebula::cpp2::ErrorCode::E_RAFT_BUFFER_OVERFLOW;
     }
--- a/src/kvstore/raftex/RaftPart.cpp
+++ b/src/kvstore/raftex/RaftPart.cpp
@@ -57,7 +57,7 @@ using OpProcessor = folly::Function<std::optional<std::string>(AtomicOp op)>;
  *  MERGE_PREV: can merge with previous, can't merge any more.  (has to be tail)
  *  MERGE_BOTH: can merge with any other
  *
- *  Normal / heartbeat will alwayse be MERGE_BOTH
+ *  Normal / heartbeat will always be MERGE_BOTH
  *  Command will alwayse be MERGE_PREV
  *  ATOMIC_OP can be either MERGE_NEXT or MERGE_BOTH
  *                          depends on if it read a key in write set.
--- a/src/kvstore/raftex/RaftPart.cpp
+++ b/src/kvstore/raftex/RaftPart.cpp
@@ -50,92 +50,81 @@ using nebula::wal::FileBasedWalPolicy;
 
 using OpProcessor = folly::Function<std::optional<std::string>(AtomicOp op)>;
 
+/**
+ * @brief code to describle if a log can be merged with others
+ *  NO_MERGE: can't merge with any other
+ *  MERGE_NEXT: can't previous logs, can merge with next. (has to be head)
+ *  MERGE_PREV: can merge with previous, can't merge any more.  (has to be tail)
+ *  MERGE_BOTH: can merge with any other
+ *
+ *  Normal / heartbeat will alwayse be MERGE_BOTH
+ *  Command will alwayse be MERGE_PREV
+ *  ATOMIC_OP can be either MERGE_NEXT or MERGE_BOTH
+ *                          depends on if it read a key in write set.
+ *  no log type will judge as NO_MERGE
+ */
+
+enum class MergeAbleCode {
+  NO_MERGE = 0,
+  MERGE_NEXT = 1,
+  MERGE_PREV = 2,
+  MERGE_BOTH = 3,
+};
+
+/**
+ * @brief this is an Iterator deal with memory lock.
+ */
 class AppendLogsIterator final : public LogIterator {
  public:
-  AppendLogsIterator(LogID firstLogId, TermID termId, RaftPart::LogCache logs, OpProcessor opCB)
-      : firstLogId_(firstLogId),
-        termId_(termId),
-        logId_(firstLogId),
-        logs_(std::move(logs)),
-        opCB_(std::move(opCB)) {
-    leadByAtomicOp_ = processAtomicOp();
-    valid_ = idx_ < logs_.size();
-    hasNonAtomicOpLogs_ = !leadByAtomicOp_ && valid_;
-    if (valid_) {
-      currLogType_ = lastLogType_ = logType();
-    }
-  }
-
+  AppendLogsIterator(LogID firstLogId, TermID termId, RaftPart::LogCache logs)
+      : firstLogId_(firstLogId), termId_(termId), logId_(firstLogId), logs_(std::move(logs)) {}
   AppendLogsIterator(const AppendLogsIterator&) = delete;
   AppendLogsIterator(AppendLogsIterator&&) = default;
 
   AppendLogsIterator& operator=(const AppendLogsIterator&) = delete;
   AppendLogsIterator& operator=(AppendLogsIterator&&) = default;
 
-  bool leadByAtomicOp() const {
-    return leadByAtomicOp_;
-  }
-
-  bool hasNonAtomicOpLogs() const {
-    return hasNonAtomicOpLogs_;
-  }
-
-  LogID firstLogId() const {
-    return firstLogId_;
-  }
-
-  LogID lastLogId() const {
-    return firstLogId_ + logs_.size() - 1;
-  }
-
-  // Return true if the current log is a AtomicOp, otherwise return false
-  bool processAtomicOp() {
-    while (idx_ < logs_.size()) {
-      auto& tup = logs_.at(idx_);
-      auto logType = std::get<1>(tup);
-      if (logType != LogType::ATOMIC_OP) {
-        // Not a AtomicOp
-        return false;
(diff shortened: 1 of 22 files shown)


src/kvstore/raftex/RaftPart.cpp
```diff
@@ -186,17 +135,172 @@ class AppendLogsIterator final : public LogIterator {
...
+        std::get<2>(logWrapper) = std::move(result);
+        /**
+         * @brief We accept same read/write key in a one log,
+         *        but reject if same in different logs.
+         */
+        for (auto& key : read) {
+          auto cit = memLock_.find(key);
+          // read after write is not acceptable.
+          if (cit != memLock_.end()) {
+            return MergeAbleCode::MERGE_NEXT;
```
critical27, line 211: Where do we clear `memLock_`?

src/kvstore/raftex/RaftPart.cpp
```diff
@@ -947,56 +1024,15 @@ void RaftPart::processAppendLogResponses(const AppendLogResponses& resps,
...
-        // 2. old iter is empty && logs_.size() > 0, but all logs in new iter is
-        // atomic op,
-        //    and all of them failed, which would make iter is empty again
-        if (iter.empty()) {
-          replicatingLogs_ = false;
-          VLOG(4) << idStr_ << "No more log to be replicated";
-          return;
-        }
-      }
+      iter.commit();
```
critical27, line 1032: Maybe we still need to build another iterator, otherwise we need to wait heartbeat to trigger replication.

src/kvstore/raftex/RaftPart.cpp
```diff
@@ -50,92 +50,65 @@ using nebula::wal::FileBasedWalPolicy;
...
-        // AtomicOp Succeeded
-        return true;
-      } else {
-        // AtomicOp failed, move to the next log, but do not increment the
-        // logId_
-        ++idx_;
+  void commit(nebula::cpp2::ErrorCode code = nebula::cpp2::ErrorCode::SUCCEEDED) {
+    for (auto it = logs_.begin(); it != logs_.end(); ++it) {
+      auto& promiseRef = std::get<4>(*it);
+      if (!promiseRef.isFulfilled()) {
```
critical27, line 90: Add a DCHECK(!isFulfilled) here?

src/storage/mutate/AddVerticesProcessor.cpp
```diff
@@ -180,134 +170,138 @@ void AddVerticesProcessor::doProcessWithIndex(const cpp2::AddVerticesRequest& re
...
-                                       UNUSED(l);
-                                       UNUSED(icw);
-                                       handleAsync(spaceId_, partId, retCode);
-                                     });
+    // step 3, Insert new vertex data
+    ret.writeSet.emplace_back(key);
+    batchHolder->put(std::move(key), std::move(value));
   }
-}  // namespace storage
+  ret.batch = encodeBatchValue(batchHolder->getBatch());
```
critical27, line 301: I'm considering whether we could move all log-encoding into raft later, because we need to first encode here, and decode in raft again, it also introduce some extra string copy.

critical27 (commented):
Good job, generally LGTM. We could get rid of the ugly iterator finally...

src/kvstore/raftex/RaftPart.cpp
```diff
@@ -186,17 +135,171 @@ class AppendLogsIterator final : public LogIterator {
...
+      case LogType::ATOMIC_OP: {
+        auto& atomOp = std::get<3>(logWrapper);
+        auto [code, result, read, write] = atomOp();
+        if (code != nebula::cpp2::ErrorCode::SUCCEEDED) {
+          DLOG(INFO) << "===> OOPs, atomOp failed!!!";
+          auto& promiseRef = std::get<4>(logWrapper);
+          if (!promiseRef.isFulfilled()) {
+            promiseRef.setValue(code);
+          }
+          return MergeAbleCode::MERGE_BOTH;
```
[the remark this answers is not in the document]
liuyu85cn: ok

src/kvstore/raftex/RaftPart.cpp
liuyu85cn: Change AppendIteratorFactory::make() to static. MemLock will release after exit that function.

src/kvstore/raftex/RaftPart.cpp
liuyu85cn: ok

src/kvstore/raftex/RaftPart.cpp
```diff
@@ -1961,10 +1999,10 @@ bool RaftPart::checkAppendLogResult(nebula::cpp2::ErrorCode res) {
     {
       std::lock_guard<std::mutex> lck(logsLock_);
       logs_.clear();
-      cachingPromise_.setValue(res);
-      cachingPromise_.reset();
+      // cachingPromise_.setValue(res);
```
critical27, line 2002: Do we need to set the promise in `logs_` and `sendingLogs_`?

critical27 (commented):
Some insert and push_back could be replaced

critical27 (dismissed):
Good job, we get rid of it finally...

critical27:
@kikimo Do I merge it first or wait until test? This is a important change, maybe risky since many code in raft has been modified.

critical27:
@liuyu85cn could write something about this PR in release note.

kikimo:
> @kikimo Do I merge it first or wait until test? This is a important change, maybe risky since many code in raft has been modified.



> @kikimo Do I merge it first or wait until test? This is a important change, maybe risky since many code in raft has been modified.

Hold, don't merge before I do the test.

kikimo:
> @kikimo Do I merge it first or wait until test? This is a important change, maybe risky since many code in raft has been modified.



> @kikimo Do I merge it first or wait until test? This is a important change, maybe risky since many code in raft has been modified.

Hold, don't merge before I do the test.

src/kvstore/raftex/RaftPart.cpp
```diff
@@ -186,17 +136,177 @@ class AppendLogsIterator final : public LogIterator {
...
   TermID termId_;
   LogID logId_;
   RaftPart::LogCache logs_;
-  OpProcessor opCB_;
+};
+
+class AppendLogsIteratorFactory {
+ public:
+  AppendLogsIteratorFactory() = default;
+  static void make(RaftPart::LogCache& cacheLogs, RaftPart::LogCache& sendLogs) {
```
panda-sheep, line 148: nubility

src/storage/test/RebuildIndexTest.cpp
```diff
@@ -511,6 +526,10 @@ TEST_F(RebuildIndexTest, RebuildEdgeIndexWithAppend) {
   RebuildIndexTest::env_->rebuildIndexGuard_->clear();
   writer->stop();
   sleep(1);
+  for (int i = 1; i <= 5; ++i) {
+    LOG(INFO) << "sleep for " << i << "s";
+    sleep(1);
```
panda-sheep, line 531: Why not just increase the sleep time

liuyu85cn: because when run this case manually(watching the execution), it may be confused when sleep more than 1 seconds.

src/kvstore/raftex/test/RaftexTestBase.cpp
```diff
@@ -309,6 +315,13 @@ bool checkLog(std::shared_ptr<test::TestShard>& copy,
   for (size_t i = start; i <= end; i++) {
     folly::StringPiece msg;
     if (!copy->getLogMsg(i, msg) || msgs[i] != msg.toString()) {
+      // LOG(INFO) << "checkLog return false;";
```
critical27, line 318: Clean them~

src/kvstore/raftex/RaftPart.cpp
```diff
@@ -50,92 +50,81 @@ using nebula::wal::FileBasedWalPolicy;
...
 using OpProcessor = folly::Function<std::optional<std::string>(AtomicOp op)>;
 
+/**
+ * @brief code to describle if a log can be merged with others
+ *  NO_MERGE: can't merge with any other
+ *  MERGE_NEXT: can't previous logs, can merge with next. (has to be head)
+ *  MERGE_PREV: can merge with previous, can't merge any more.  (has to be tail)
+ *  MERGE_BOTH: can merge with any other
+ *
+ *  Normal / heartbeat will alwayse be MERGE_BOTH
```
critical27, line 60: Type always

src/storage/mutate/AddEdgesProcessor.cpp
```diff
@@ -191,257 +173,174 @@ void AddEdgesProcessor::doProcessWithIndex(const cpp2::AddEdgesRequest& req) {
...
-                batchHolder->remove(std::move(oi));
+      }
+      for (const auto& index : indexes_) {
+        if (edgeType == index->get_schema_id().get_edge_type()) {
+          // step 1, Delete old version index if exists.
+          if (oldReader != nullptr) {
+            auto oldIndexKeys = indexKeys(partId, oldReader.get(), key, index, nullptr);
+            if (!oldIndexKeys.empty()) {
+              ret.writeSet.insert(ret.writeSet.end(), oldIndexKeys.begin(), oldIndexKeys.end());
+              // ret.writeSet.push_back(oldIndexKeys);
```
critical27, line 270: Ditto~

src/storage/mutate/AddEdgesProcessor.cpp
```diff
@@ -191,257 +173,174 @@ void AddEdgesProcessor::doProcessWithIndex(const cpp2::AddEdgesRequest& req) {
...
+              // not.
+              auto indexState = env_->getIndexState(spaceId_, partId);
+              if (env_->checkRebuilding(indexState)) {
+                auto delOpKey = OperationKeyUtils::deleteOperationKey(partId);
+                for (auto& idxKey : oldIndexKeys) {
+                  ret.writeSet.push_back(idxKey);
+                  batchHolder->put(std::string(delOpKey), std::move(idxKey));
+                }
+              } else if (env_->checkIndexLocked(indexState)) {
+                // return folly::Optional<std::string>();
```
critical27, line 281: Ditto

critical27 (approved):
Good job!

liwenhui-soul:
I think it's a good job, do you agree?

critical27 (approved):

panda-sheep (approved):
GGood job!

Issue #3935, opened 2022-02-23
Connect using a pre-v2.6.0 client will cause a service crash


Aiee:
**Please check the FAQ documentation before raising an issue**

**Describe the bug (__required__)**
Connecting with a pre-v2.5.0 client to v3.0.0 Nebula will cause a service crash:

![image](https://user-images.githubusercontent.com/18348405/155281424-92ee0803-90fc-480c-92df-6e66a39a991e.png)

**Your Environments (__required__)**

* OS: `uname -a`
* Compiler: `g++ --version` or `clang++ --version`
* CPU: `lscpu`
* Commit id (e.g. `a3ffc7d8`)

**How To Reproduce(__required__)**

1. Step 1
2. Step 2
3. Step 3

**Expected behavior**

**Additional context**

PR #3942, opened 2022-02-24
Fix service crash caused by connecting using a pre-v2.6.0 client

Aiee:
## What type of PR is this?
- [x] bug
- [ ] feature
- [ ] enhancement

#### Issue(s) number: 
Fix https://github.com/vesoft-inc/nebula/issues/3935
#### Description:
Fix service crash caused by connecting using a pre-v2.6.0 client.

Error message when connecting the v3.0.0 Nebula service with v2.5.0 console:
```
> ./nebula-console -addr 192.168.8.6 -port 29562 -u root -p nebula
2022/02/24 22:01:21 [INFO] connection pool is initialized successfully
2022/02/24 22:01:21 Fail to create a new session from connection pool, fail to authenticate, error: The version of the client sending request from "::ffff:192.168.8.6":49722 is lower than v2.6.0, please update the client.
panic: Fail to create a new session from connection pool, fail to authenticate, error: The version of the client sending request from "::ffff:192.168.8.6":49722 is lower than v2.6.0, please update the client.

goroutine 1 [running]:
log.Panicf(0x655651, 0x35, 0xc00010de78, 0x1, 0x1)
        /usr/local/go/src/log/log.go:361 +0xc5
```

## How do you solve it?
Add a flag in `mataClient` to record whether the client triggered `verifyClientVersion()`. If the client called `verifyClientVersion`, the client has a version >= `v2.6.0`. Otherwise the client version is lower than `v2.5.0`.

When doing `auth()`, check the IP address of the client to see if the client is later than `v2.5.0`. Return an error if the client version is lower than `v2.5.0`.

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....

### DIFF
--- a/src/graph/service/PasswordAuthenticator.cpp
+++ b/src/graph/service/PasswordAuthenticator.cpp
@@ -13,7 +13,16 @@ PasswordAuthenticator::PasswordAuthenticator(meta::MetaClient* client) {
 }
 
 Status PasswordAuthenticator::auth(const std::string& user, const std::string& password) {
-  return metaClient_->authCheckFromCache(user, password);
+  UNUSED(user);
+  UNUSED(password);
+  return Status::Error(
+      "Should not be called, this interface is implemented to override the parent class");
+}
+
+Status PasswordAuthenticator::auth(const std::string& user,
+                                   const std::string& password,
+                                   const HostAddr& clientIp) {
+  return metaClient_->authCheckFromCache(user, password, clientIp);
 }
 
 }  // namespace graph
--- a/src/graph/service/PasswordAuthenticator.cpp
+++ b/src/graph/service/PasswordAuthenticator.cpp
@@ -13,16 +13,7 @@ PasswordAuthenticator::PasswordAuthenticator(meta::MetaClient* client) {
 }
 
 Status PasswordAuthenticator::auth(const std::string& user, const std::string& password) {
-  UNUSED(user);
-  UNUSED(password);
-  return Status::Error(
-      "Should not be called, this interface is implemented to override the parent class");
-}
-
-Status PasswordAuthenticator::auth(const std::string& user,
-                                   const std::string& password,
-                                   const HostAddr& clientIp) {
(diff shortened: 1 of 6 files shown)


src/graph/service/PasswordAuthenticator.cpp
```diff
@@ -13,7 +13,16 @@ PasswordAuthenticator::PasswordAuthenticator(meta::MetaClient* client) {
 }
 
 Status PasswordAuthenticator::auth(const std::string& user, const std::string& password) {
```
Shylock-Hg, line 15: Don't modify `auth` function, do it outside.

CPWstatic (approved):
LGTM.

src/graph/service/PasswordAuthenticator.cpp
Aiee: I need to pass the client addr into `MetaClient::authCheckFromCache()`, and `auth` is the middleware

Shylock-Hg: You could add `clientIp` to meta client in `GraphService::auth`

Aiee: I have passed clientIp in [src/graph/service/GraphService.cpp line 122](https://github.com/vesoft-inc/nebula/pull/3942/files#diff-3a9c8c30a7bc9f5d28cd53b6f7f2f3b29a67827996e45f6e6bd2d2a749db19c5R212)
Besides, the  `GraphService::auth()` in the enterprise version also has `clientIp` as parameter. This should be fine

Shylock-Hg (approved):

CPWstatic (approved):

HarrisChu:
depends on client behavior, i.e. client should verify the version and then auth immediately.
 
prefer to combine these two interfaces in the future.
```
auth(user, password, version)
```

Issue #3591, opened 2021-12-29
enhance meta client heartbeat, always receive the meta leader


HarrisChu:
as title
1. storage can receive meta leader even if other error code.
2. then meta client could send heartbeat to meta leader.

PR #3950, opened 2022-02-28
Machine not registed should response leader address

darionyaphet:
## What type of PR is this?
- [ ] bug
- [ ] feature
- [x] enhancement

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....
Fix faq, see
https://docs.nebula-graph.com.cn/3.0.0/20.appendix/0.FAQ/#nebula_graph_nebula-storaged


src/clients/meta/MetaClient.cpp
```diff
@@ -738,11 +738,13 @@ void MetaClient::getResponse(Request req,
...
               auto&& resp = t.value();
-              if (resp.get_code() == nebula::cpp2::ErrorCode::SUCCEEDED) {
+              auto code = resp.get_code();
+              if (code == nebula::cpp2::ErrorCode::SUCCEEDED) {
                 // succeeded
                 pro.setValue(respGen(std::move(resp)));
                 return;
-              } else if (resp.get_code() == nebula::cpp2::ErrorCode::E_LEADER_CHANGED) {
+              } else if (code == nebula::cpp2::ErrorCode::E_LEADER_CHANGED ||
+                         code == nebula::cpp2::ErrorCode::E_MACHINE_NOT_FOUND) {
```
critical27, line 747: Em... if you modify in this way, how to distinguish from actual `MACHINE_NOT_FOUND` and `MACHINE_NOT_FOUND` with LEADER_CHANGE

critical27 (approved):

liuyu85cn (approved):
LGTM

src/meta/processors/admin/HBProcessor.cpp
```diff
@@ -114,5 +115,14 @@ void HBProcessor::process(const cpp2::HBReq& req) {
   onFinished();
 }
 
+void HBProcessor::setLeaderInfo() {
+  auto leaderRet = kvstore_->partLeader(kDefaultSpaceId, kDefaultPartId);
+  if (ok(leaderRet)) {
+    resp_.leader_ref() = toThriftHost(nebula::value(leaderRet));
+  } else {
+    resp_.code_ref() = nebula::error(leaderRet);
```
pengweisong, line 123: will this replace the "E_MACHINE_NOT_FOUND"? Which one we will want to return to the user?

src/meta/processors/admin/HBProcessor.cpp
```diff
@@ -39,6 +39,7 @@ void HBProcessor::process(const cpp2::HBReq& req) {
     if (!ActiveHostsMan::machineRegisted(kvstore_, host)) {
       LOG(INFO) << "Machine " << host << " is not registed";
       handleErrorCode(nebula::cpp2::ErrorCode::E_MACHINE_NOT_FOUND);
+      setLeaderInfo();
```
pengweisong, line 42: Why not use handleLeaderChanged directly?

darionyaphet: because it is not leader change, to avoid confuse

src/meta/processors/admin/HBProcessor.cpp
darionyaphet: In most cases, this will not happen. when the leader is not found, the priority is higher than machine not found.

HarrisChu:
fix https://github.com/vesoft-inc/nebula/issues/3591

Issue #3688, opened 2022-01-11
Redesign scan vertices


Shylock-Hg:
**Please check the FAQ documentation before raising an issue**

**Describe the bug (__required__)**

Before #3328 we scan vertices by scan tags, but now it's mistake. So we need to redesign scan vertices.

**Your Environments (__required__)**

* OS: `uname -a`
* Compiler: `g++ --version` or `clang++ --version`
* CPU: `lscpu`
* Commit id (e.g. `a3ffc7d8`)

**How To Reproduce(__required__)**

1. Step 1
2. Step 2
3. Step 3

**Expected behavior**

**Additional context**

cangfengzhs:
According to the current implementation, tags will be specified in Scan vertex, and only vertexes with these tags will be accessed. If it is traversed according to vertexKey, all points will be traversed. I'm not quite sure if this logic is appropriate.

Shylock-Hg:
> According to the current implementation, tags will be specified in Scan vertex, and only vertexes with these tags will be accessed. If it is traversed according to vertexKey, all points will be traversed. I'm not quite sure if this logic is appropriate.

Yes, current implementation will ignore vertex without any tags.

cangfengzhs:
> Yes, current implementation will ignore vertex without any tags.
So do we have to change it to traverse all the vertices regardless of whether there is a corresponding tag? I'm not sure how much impact this change will have

Shylock-Hg:
> 

Yes, ScanVertices should return one row for each vertex regardless of whether there are any tags on it.

xtcyclist:
I think we shouldn't do any redesign for 3.4. Sounds too big. @Shylock-Hg what do you think?

Shylock-Hg:
> I think we shouldn't do any redesign for 3.4. Sounds too big. @Shylock-Hg what do you think?

It's bug of vertex key, bug now it's removed. I close it. @cangfengzhs

PR #3965, opened 2022-03-02
refactor ScanVertex

cangfengzhs:
1. Returns a row of data as long as the vertex exists, regardless of whether the tag exists.
2. Adjust unittest.

## What type of PR is this?
- [ ] bug
- [ ] feature
- [x] enhancement

#### Issue(s) number: 
#3688 
## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:
In version 3.0, we introduced the vertex key to record the existence of a point. However, ScanVertex does not strictly follow this rule, but only judges whether the vertex exists according to the Tag. This PR modifies this part of the logic. Now ScanVertex results will include every point, whether or not it has the requested tag.


src/storage/exec/ScanNode.h
```diff
@@ -42,55 +42,44 @@ class ScanVertexPropNode : public QueryNode<Cursor> {
...
-      auto tagIdIndex = tagNodesIndex_.find(tagId);
-      if (tagIdIndex == tagNodesIndex_.end()) {
-        continue;
-      }
-      auto vertexId = NebulaKeyUtils::getVertexId(vIdLen, key);
-      if (vertexId != currentVertexId && !currentVertexId.empty()) {
-        collectOneRow(isIntId, vIdLen, currentVertexId);
-      }  // collect vertex row
-      currentVertexId = vertexId;
+      vertexId = key.subpiece(vertexPrefix.size(), vIdLen).toString();
```
Shylock-Hg, line 73: `std::string vertexId = ...`

src/storage/test/DeleteTagsTest.cpp
```diff
@@ -164,7 +164,8 @@ TEST(DeleteTagsTest, SimpleTest) {
     auto resp = std::move(fut).get();
 
     ASSERT_EQ(0, (*resp.result_ref()).failed_parts.size());
-    ASSERT_EQ(0, resp.props_ref()->size());
```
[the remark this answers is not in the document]
cangfengzhs: Because I need modify mock data and add vertex key into it. Before that, GetProp can't get any data if tag has been deleted.But now it can get a vertex without any tag, and resultDataSet is not empty.

Shylock-Hg (approved):

src/storage/exec/ScanNode.h
```diff
@@ -42,55 +42,43 @@ class ScanVertexPropNode : public QueryNode<Cursor> {
...
   nebula::cpp2::ErrorCode doExecute(PartitionID partId, const Cursor& cursor) override {
+    partId_ = partId;
     auto ret = RelNode::doExecute(partId);
     if (ret != nebula::cpp2::ErrorCode::SUCCEEDED) {
       return ret;
     }
 
     std::string start;
-    std::string prefix = NebulaKeyUtils::tagPrefix(partId);
+    std::string vertexPrefix = NebulaKeyUtils::vertexPrefix(partId);
```
critical27, line 52: I doubt whether we should scan in this way, because it converts from one sequential scan to one sequential scan + lots of point look-up. Is there any other reason we need to scan in this way?

cangfengzhs: May be we can san with two iterator.

cangfengzhs:
Because the graph calculation may still use the ScanVertex interface, it is not changed for the time being to ensure compatibility.

Issue #3649, opened 2022-01-06
After inserting a vertex without a tag, use`show stats`to count the numbers of vertices or edges that cannot be counted


foesa-yang:
**Please check the FAQ documentation before raising an issue**

**Describe the bug (__required__)**

```bash
(root@nebula) [basketballplayer]> Insert vertex VALUES "1":(),"2":()
Execution succeeded (time spent 1111/1503 us)

Thu, 06 Jan 2022 03:46:54 UTC

(root@nebula) [basketballplayer]>  show stats
+---------+------------+-------+
| Type    | Name       | Count |
+---------+------------+-------+
| "Tag"   | "person"   | 0     |
| "Tag"   | "player"   | 51    |
| "Tag"   | "t2"       | 1     |
| "Tag"   | "team"     | 30    |
| "Edge"  | "follow"   | 81    |
| "Edge"  | "like"     | 0     |
| "Edge"  | "serve"    | 152   |
| "Space" | "vertices" | 82    |
| "Space" | "edges"    | 233   |
+---------+------------+-------+
Got 9 rows (time spent 797/1346 us)

Thu, 06 Jan 2022 03:46:58 UTC

```
**Your Environments (__required__)**

* OS: `uname -a`
* Compiler: `g++ --version` or `clang++ --version`
* CPU: `lscpu`
* Commit id (e.g. `a3ffc7d8`)

**How To Reproduce(__required__)**

1. Insert vertex without Tag.
2. `show stats` to count the numbers of vertices or edges.

**Expected behavior**

**Additional context**

critical27:
What's wrong with the result?
51 + 1 + 30 = 82
81 + 152 = 233

foesa-yang:
I insert a vertex without a tag, why can't I find it with `show stats`?

panda-sheep:
This has something to do with recent changes

When supporting the insertion of vertex without tag, the stats job needs to be compatible

czpmango:
Detailed test!

critical27:
`show stats` only scan tag/edge (`"Space" | "vertices"` is tag sum), we could handle the actual `vertex` in stats job later

PR #3967, opened 2022-03-02
fix Issue3649

cangfengzhs:
## What type of PR is this?
- [x] bug
- [ ] feature
- [ ] enhancement

#### Issue(s) number: 
#3649 
## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:
vertex count in `show stats` will counted by vertex key instead of tag key.

### DIFF
--- a/src/storage/admin/StatsTask.cpp
+++ b/src/storage/admin/StatsTask.cpp
@@ -105,6 +105,8 @@ nebula::cpp2::ErrorCode StatsTask::genSubTask(GraphSpaceID spaceId,
   std::unique_ptr<kvstore::KVIterator> tagIter;
   auto edgePrefix = NebulaKeyUtils::edgePrefix(part);
   std::unique_ptr<kvstore::KVIterator> edgeIter;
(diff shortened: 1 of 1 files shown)


src/storage/admin/StatsTask.cpp
```diff
@@ -219,7 +216,10 @@ nebula::cpp2::ErrorCode StatsTask::genSubTask(GraphSpaceID spaceId,
     }
     edgeIter->next();
   }
-
+  while (vertexIter && vertexIter->valid()) {
+    spaceVertices++;
+    vertexIter->next();
+  }
```
panda-sheep, line 222: This is not handled correctly

panda-sheep: Please ignore the comments above, I see that `vertexkey` is written by default in AddVerticesProcessor.

yixinglu (approved):

panda-sheep (approved):

PR #3974, opened 2022-03-03
optimize go yield

nevermore3:
## What type of PR is this?
- [ ] bug
- [ ] feature
- [ ] enhancement

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....

### DIFF
--- a/src/graph/validator/GoValidator.cpp
+++ b/src/graph/validator/GoValidator.cpp
@@ -139,6 +139,7 @@ Status GoValidator::validateYield(YieldClause* yield) {
       return Status::SemanticError("`%s' is not support in go sentence.", col->toString().c_str());
     }
 
+    col->setExpr(rewriteVertexEdge2EdgeProp(col->expr()));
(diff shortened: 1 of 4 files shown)


src/graph/validator/GoValidator.cpp
```diff
@@ -183,6 +185,24 @@ void GoValidator::extractPropExprs(const Expression* expr,
   const_cast<Expression*>(expr)->accept(&visitor);
 }
 
+Expression* GoValidator::rewriteVertexEdge2EdgeProp(const Expression* expr) {
+  auto pool = qctx_->objPool();
+  const auto& name = expr->toString();
+  if (name == "id($^)" || name == "src(edge)") {
```
Shylock-Hg, line 191: Compare expression structure instead of string.

nevermore3: why

CPWstatic (approved):

yixinglu (approved):

Issue #3743, opened 2022-01-18
Call thrift RunAdminJob to create a balance job, if space not existed, the job is still created


kqzh:
nba33 is not existed, but still created job
![image](https://user-images.githubusercontent.com/35095889/149864534-0cd233e5-a111-4fc6-8799-ced92f43ce63.png)

![image](https://user-images.githubusercontent.com/35095889/149864548-e28a1538-126a-4418-9dc4-1a8868682016.png)

PR #3976, opened 2022-03-03
refactor job manager part1

panda-sheep:
## What type of PR is this?
- [ ] bug
- [ ] feature
- [X] enhancement

#### Description:
refactor job manager, this pr mainly did the following parts
1. separation by space
2. Remove job_concurrency
Because only compact and flush, stats syntax supports job_concurrency, and job passes job_concurrency to task, but the corresponding task does not use job_concurrency.
Even using job_concurrency does not make much sense, because in a space, compactTask and flushTask are concurrent on the storage according to the number of engines. statsTask is concurrent according to the number of parts. FLAGS_max_concurrent_subtasks is enough

3. Fix bugs, such as https://github.com/vesoft-inc/nebula/issues/3743

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....


src/meta/processors/job/AdminJobProcessor.cpp
```diff
@@ -27,45 +24,17 @@ void AdminJobProcessor::process(const cpp2::AdminJobReq& req) {
...
-        errorCode = nebula::error(jobId);
-        break;
-      }
-
-      JobDescription jobDesc(nebula::value(jobId), cmd, paras);
-      errorCode = jobMgr->addJob(jobDesc, adminClient_);
-      if (errorCode == nebula::cpp2::ErrorCode::SUCCEEDED) {
-        result.job_id_ref() = nebula::value(jobId);
-      }
+      errorCode = addJobProcess(req, result);
```
pengweisong, line 33: Will be 'addJobProcess' more clear and enough?

pengweisong (approved):
Good Job, LGTM

src/meta/processors/job/AdminJobProcessor.cpp
```diff
@@ -142,5 +111,49 @@ void AdminJobProcessor::process(const cpp2::AdminJobReq& req) {
...
 }
 
+nebula::cpp2::ErrorCode AdminJobProcessor::addJobProcess(const cpp2::AdminJobReq& req,
+                                                         cpp2::AdminJobResult& result) {
+  auto cmd = req.get_cmd();
+  auto paras = req.get_paras();
+
+  // All jobs here are the space level, so the last parameter is spaceName.
+  if (paras.empty()) {
+    LOG(INFO) << "Parameter should be not empty";
```
SuperYoko, line 121: Prefer log with specific operation, this is too generic. 
If following logs metion this operation, Ignore this is okay.

src/meta/processors/job/AdminJobProcessor.cpp
```diff
@@ -27,45 +24,17 @@ void AdminJobProcessor::process(const cpp2::AdminJobReq& req) {
   }
   LOG(INFO) << __func__ << "() " << oss.str();
 
-  JobManager* jobMgr = JobManager::getInstance();
+  cpp2::AdminJobResult result;
+  auto errorCode = nebula::cpp2::ErrorCode::SUCCEEDED;
```
darionyaphet, line 28: using `nebula::cpp2::ErrorCode` to replace `auto` would be better?

critical27 (approved):
Generally LGTM.

src/meta/processors/job/AdminJobProcessor.cpp
```diff
@@ -142,5 +111,49 @@ void AdminJobProcessor::process(const cpp2::AdminJobReq& req) {
...
+  }
+
+  folly::SharedMutex::WriteHolder holder(LockUtils::lock());
+  auto jobId = autoIncrementId();
+  if (!nebula::ok(jobId)) {
+    return nebula::error(jobId);
+  }
+
+  JobDescription jobDesc(nebula::value(jobId), cmd, paras);
+  auto errorCode = jobMgr_->addJob(jobDesc, adminClient_);
```
critical27, line 151: The job manager will have one queue for one space, right?

panda-sheep: Yes, the later pr will do the following work:
Each space has a priority queue.
Jobs in the same space are executed serially according to priority.
Jobs in different spaces are executed in parallel according to the space.

src/meta/processors/job/AdminJobProcessor.cpp
panda-sheep: My original intention is: for different job operations, separate out a separate interface.
So `addJobProcess` handles the operation of `add job`.
Of course, within the add job, there is its own check.

src/meta/processors/job/AdminJobProcessor.cpp
panda-sheep: Good point.
The following pr will move the last `spaceName` of parameter out.
Different `AdminJobOp` operations, check the validity of their own parameters.
Therefore, the following pr will uniformly change this error message to make the error message more clear.

liuyu85cn (approved):
good job

CPWstatic (approved):

PR #3977, opened 2022-03-03
Refactor RPC client getResponse implementation

yixinglu:
## What type of PR is this?
- [ ] bug
- [ ] feature
- [X] enhancement

#### Issue(s) number: 
none
#### Description:

Use folly::collect and deferValue to simplify the collectResponse implementation. The promise usage is not user-friendly and hard to debug.

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....
none


src/clients/storage/StorageClientBase-inl.h
```diff
@@ -120,180 +74,135 @@ StorageClientBase<ClientType, ClientManagerType>::collectResponse(
...
+    auto fut = getResponse(evb, req.first, req.second, std::move(remoteFunc))
+                   .ensure([totalLatencies, i, start]() {
+                     (*totalLatencies)[i] = time::WallClock::fastNowInMicroSec() - start;
+                   });
+
+    respFutures.emplace_back(std::move(fut));
+  }
+
+  return folly::collectAll(respFutures)
+      .deferValue([this, requests = std::move(requests), totalLatencies, hosts](
```
critical27, line 100: Do we need to add `via` after `collectAll` and before `deferValue`? I can't recall that which thread would trigger the callback in `defer`.

My point is the rpc of `respFutures` will run on the parameter `evb`, what if `collectAll` is running on same `evb`, can we make sure that `rpc1, rpc2, ..., collectAll, rpcn` won't happen? This would cause deadlock

yixinglu: If we call the via for the SemiFuture of collectAll, these tasks will be executed immediately on evb. that's not our expected result. We want to let user of this interface to specify which thread these tasks work on. It's same as the original logics.

critical27: 👌

critical27 (approved):
Good job~

jievince (approved):
Good job!

Issue #3869, opened 2022-02-09
StatsManager is not thread safe


cangfengzhs:
**Describe the bug (__required__)**
```
sm.stats_.emplace(
      counterName,
      std::make_pair(std::make_unique<std::mutex>(),
                     std::make_unique<StatsType>(
                         60,
                         std::initializer_list<StatsType::Duration>(
                             {seconds(5), seconds(60), seconds(600), seconds(3600)}))));
```
`stats_` is a `std::unordered_map`. And some time will read/write without lock

Shylock-Hg:
It's protected by nameMapLock_.

cangfengzhs:
> 

No nameMapLock_ protect another map. You can find some other places where stat_ is used without lock.

Shylock-Hg:
> > 
> 
> No, nameMapLock_ protect another map. You can find some other places where stat_ is used without lock.

It also protects stats_ too, add/delete counter will lock nameMapLock_ firstly.

cangfengzhs:
> 

But reading stat_ doesn't with Lock. If one thread read stat_'s element while another thread insert/erase some element, this may be lead to structure modify of unordered_map. Reading thread will access some released memory.

Shylock-Hg:
> > 
> 
> But reading stat_ doesn't with Lock. If one thread read stat_'s element while another thread insert/erase some element, this may be lead to structure modify of unordered_map. Reading thread will access some released memory.

Yes, read need protection too.

cangfengzhs:
> Yes, read need protection too.
And I think this part requires some complex design rather than simply using read-write locks. The performance of the lock is very poor, and the metric cannot affect the performance of the system.

Shylock-Hg:
> > Yes, read need protection too.
> 
> And I think this part requires some complex design rather than simply using read-write locks. The performance of the lock is very poor, and the metric cannot affect the performance of the system.

Agree, in my mind, static metric is enough.

PR #3992, opened 2022-03-09
refactor stats manager for safe and performance

cangfengzhs:
## What type of PR is this?
- [ ] bug
- [ ] feature
- [x] enhancement

In this PR, I refactor the hole StatsManager.

Function:
1. The API directly distinguishes between counter type metrics and histogram type metrics, and adds gauge type metrics
2. RCU is used to replace the read-write lock. This will have a very large performance improvement.
3. Metrics is registered during initialization. However, it supports dynamic addition and deletion of labels. A good trade-off is made between performance and ease of use.
4. Rename statsmanager to metrics

Safe:
Remove all previous thread unsafe hidden dangers. However, the cost is that the read performance will be slightly reduced, but the write performance will be better than before.

In addition, because the stats related classes provided by folly are thread unsafe, we need additional locks when writing to multiple threads. This can become a performance bottleneck. After a simple test, the limit QPS of a single counter / histogram under a single thread is 2 million, and the limit under 64 threads is 1 million. In terms of throughput, it is enough. However, in the multithreading scenario, the sys occupation of the system CPU will become very, very high.

If you want to further optimize the performance, you can use ThreadLocal. This implementation is very complex. However, because the write QPS of metrics is much larger than the read QPS, the multi and scalability will become very good.

#### Issue(s) number: 
#3869 
## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....

### DIFF
--- a/src/common/stats/CMakeLists.txt
+++ b/src/common/stats/CMakeLists.txt
@@ -7,5 +7,12 @@ nebula_add_library(
     OBJECT
     StatsManager.cpp
 )
-
+nebula_add_library(
+    metrics_obj
+    OBJECT
+    Metrics.cpp
+    Histogram.cpp
+    Counter.cpp
+    Gauge.cpp
+)
 nebula_add_subdirectory(test)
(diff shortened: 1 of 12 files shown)


jievince (commented):
Good job!

Issue #3988, opened 2022-03-08
Get json result of a `profile` query causes crash


Aiee:
**Please check the FAQ documentation before raising an issue**

**Describe the bug (__required__)**

Get JSON result of a `profile` query causes crash
![Screen Shot 2022-03-08 at 16 10 58](https://user-images.githubusercontent.com/18348405/157194227-d111298a-b72f-40bb-a98f-382d2f941f35.png)

**Your Environments (__required__)**

* OS: `uname -a`
* Compiler: `g++ --version` or `clang++ --version`
* CPU: `lscpu`
* Commit id (e.g. `a3ffc7d8`)

**How To Reproduce(__required__)**

call `executeJson("profile yield 1")` in the client

**Expected behavior**

**Additional context**

PR #3998, opened 2022-03-10
Fix crash when use `executeJson()` with profile

Aiee:
## What type of PR is this?
- [x] bug
- [ ] feature
- [ ] enhancement

#### Issue(s) number: 
Fix https://github.com/vesoft-inc/nebula/issues/3988
#### Description:
Some members of the `PlanDescription` are initialized as `null`

After fixing:
Screen Shot 2022-03-10 at 14 03 08

## How do you solve it?
Check pointers before constructing the dynamic object.

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

Fix crash when use `executeJson()` with profile


Shylock-Hg:
Add test case.

Shylock-Hg (approved):

PR #4000, opened 2022-03-10
Fix lambda capture bug in GraphStorageLocalServer

yixinglu:
This will crash the daemon

## What type of PR is this?
- [X] bug
- [ ] feature
- [ ] enhancement

#### Description:

The request object in lambda function is captured by reference, this will crash the server when this async interface finishes to run and the task in threadManager haven't be executed.

## How do you solve it?

Captured by copy.

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....


jievince (approved):

SuperYoko (approved):
If request won't cost too much,capture by value is better.
I'm okay with this fix, thanks.

PR #4010, opened 2022-03-11
Binary pack script opt

### LINKED ISSUE #4003: Add service stop in binary/rpm/deb uninstall
This confused new users when they did this:

- remove everything on nebula and rpm -e
- install nebula
- The state of the cluster is the same as it was before, and it turned out the process are still the old ones.

We may add post script in remove phase to stop services.

### LINKED ISSUE #4009: Add order for script/service start all
To make storaged being started slightly slower than meta, to avoid storaged being waiting for a long time(really long) after the initial attempt of checking meta to be ready failed(due to meta is not yet up)

wey-gu:
## What type of PR is this?
- [ ] bug
- [ ] feature
- [x] enhancement

#### Issue(s) number: 
close #4003 
close #4009 

#### Description:

- stop service before uninstalling packages
- try best to avoid storage initial start attempt failed due to meta is not yet ready(and the retry next attempt will be quite long, later)

## How do you solve it?

- Add pre uninstall script to stop service before uninstall
- Softly enforce waiting for port listening during service start with a best effort/timer

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [x] Manual Test(to be done later)

Affects:

## Release notes:

> ex. Fixed the bug .....
- Binary package script optimized.


wey-gu:
will manually test when I got bandwidth to, plz don't review now.

package/prerm
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
+install_dir=/usr/local/nebula
+daemons=(storaged metad graphd)
+for daemon in ${daemons[@]}
+do
+    if [[ ! -f $install_dir/scripts/nebula.service ]] ; then
+        $install_dir/scripts/nebula.service stop ${daemon}
```
CPWstatic, line 8: This operation can be dangerous in a production environment. I suggest that we do not do this implicitly.

scripts/nebula.service
```diff
@@ -141,6 +141,8 @@ function start_daemon {
     local command="${executable} --flagfile ${config}"
     INFO "Starting ${daemon_name}..."
     eval ${command}
+    local port=$(get_port_from_config ${config})
+    wait_for_port_listening ${port} 20 || WARN "Timed out waiting for ${daemon_name} to finish starting, but it may still be running"
```
HarrisChu, line 145: prefer to implement this logic in nebula process.
and the behavior should be same with running in container.

HarrisChu: or just like readinessprobe in k8s, curl a http interface?

wey-gu:
this will be revisited later when I have bandwidth to think about it.
Thanks for the reviews already ;)

wey-gu:
closing it for now

PR #4013, opened 2022-03-11
update time when alter space

liwenhui-soul:
## What type of PR is this?
- [x] bug
- [ ] feature
- [ ] enhancement

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....


src/meta/processors/parts/AlterSpaceProcessor.cpp
```diff
@@ -25,7 +25,18 @@ void AlterSpaceProcessor::process(const cpp2::AlterSpaceReq& req) {
     default:
       break;
   }
-  handleErrorCode(nebula::cpp2::ErrorCode::SUCCEEDED);
+  std::vector<kvstore::KV> data;
```
critical27, line 28: Write all data in a single batch (both data in `addZones` and update time), you could see other processors.

darionyaphet (approved):
LGTM

critical27 (approved):

PR #4014, opened 2022-03-11
show peers in raft stat

liwenhui-soul:
## What type of PR is this?
- [x] bug
- [ ] feature
- [ ] enhancement

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....


critical27 (approved):

src/interface/raftex.thrift
```diff
@@ -133,6 +133,7 @@ struct GetStateResponse {
     6: LogID            last_log_id;
     7: TermID           last_log_term;
     8: Status           status;
+    9: list<binary>     peers;
```
darionyaphet, line 136: `peers` have been used?

liwenhui-soul: what used mean?

src/kvstore/raftex/RaftPart.cpp
```diff
@@ -1067,6 +1067,15 @@ void RaftPart::getState(cpp2::GetStateResponse& resp) {
   resp.last_log_id_ref() = lastLogId_;
   resp.last_log_term_ref() = lastLogTerm_;
   resp.status_ref() = status_;
+  std::vector<std::string> peers;
```
pengweisong, line 1070: Is the peers containing itself?

src/interface/raftex.thrift
darionyaphet: where is the use of it

liwenhui-soul: monitor tools

pengweisong: Maybe you could add some explanation in the future pr

kikimo: > 

It's very important for us to retrieve raft peers info when trouble shooting some issue.
BTW, think we should also including this info in storage monitring metrics.

kikimo (approved):
LGTM

PR #4015, opened 2022-03-11
cleanWal by commitLogId

liwenhui-soul:
## What type of PR is this?
- [x] bug
- [ ] feature
- [ ] enhancement

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....


src/kvstore/NebulaStore.cpp
```diff
@@ -1133,7 +1133,7 @@ void NebulaStore::cleanWAL() {
       auto& part = partEntry.second;
       if (part->needToCleanWal()) {
         // clean wal by expired time
-        part->wal()->cleanWAL();
+        part->cleanWal();
```
critical27, line 1136: Maybe unify all usage, either implement `cleanWAL()` in `Part` and `Listener`, or just call `part->wal()->cleanWAL(id)`.

critical27 (commented):
Generally LGTM

src/kvstore/NebulaStore.cpp
```diff
@@ -1133,16 +1133,16 @@ void NebulaStore::cleanWAL() {
...
     }
   }
   for (const auto& spaceEntry : spaceListeners_) {
     for (const auto& partEntry : spaceEntry.second->listeners_) {
       for (const auto& typeEntry : partEntry.second) {
         const auto& listener = typeEntry.second;
-        // clean wal by log id
-        listener->wal()->cleanWAL(listener->getApplyId());
+        // clean wal by commit log id
+        listener->cleanWal();
```
critical27, line 1145: `Listener` need to override it, because it use apply id, not commitId

critical27 (approved):
Good job

src/kvstore/wal/FileBasedWal.cpp
```diff
@@ -672,23 +672,30 @@ void FileBasedWal::cleanWAL() {
...
             << walFiles_.rbegin()->second->lastId();
     return;
   }
-
+  int walTTL = FLAGS_wal_ttl;
   // remove wal file that lastId is less than target
   auto iter = walFiles_.begin();
   while (iter != walFiles_.end()) {
-    if (iter->second->lastId() < id) {
+    if (iter->second->lastId() < id && index < size - 2 && (now - iter->second->mtime() > walTTL)) {
```
pengweisong, line 694: If we do not use walTTL to limit the cleanWAL, what will happen?

critical27: Then no wal would be deleted

pengweisong (approved):
Good Job

PR #4022, opened 2022-03-14
unify leader change and machine not found

darionyaphet:
## What type of PR is this?
- [ ] bug
- [ ] feature
- [ ] enhancement

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....

### DIFF
--- a/src/clients/meta/MetaClient.cpp
+++ b/src/clients/meta/MetaClient.cpp
@@ -743,7 +743,8 @@ void MetaClient::getResponse(Request req,
                 // succeeded
                 pro.setValue(respGen(std::move(resp)));
                 return;
-              } else if (code == nebula::cpp2::ErrorCode::E_LEADER_CHANGED) {
+              } else if (code == nebula::cpp2::ErrorCode::E_LEADER_CHANGED ||
+                         code == nebula::cpp2::ErrorCode::E_MACHINE_NOT_FOUND) {
                 updateLeader(resp.get_leader());
                 if (retry < retryLimit) {
                   evb->runAfterDelay(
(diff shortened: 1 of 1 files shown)


HarrisChu (approved):

Aiee (approved):

src/clients/meta/MetaClient.cpp
```diff
@@ -743,7 +743,8 @@ void MetaClient::getResponse(Request req,
                 // succeeded
                 pro.setValue(respGen(std::move(resp)));
                 return;
-              } else if (code == nebula::cpp2::ErrorCode::E_LEADER_CHANGED) {
+              } else if (code == nebula::cpp2::ErrorCode::E_LEADER_CHANGED ||
+                         code == nebula::cpp2::ErrorCode::E_MACHINE_NOT_FOUND) {
```
critical27, line 747: If so, why not just handle only return `E_MACHINE_NOT_FOUND` when the key not in `ActiveHostsMan::machineRegisted`...

HarrisChu:
BTW, prefer to handle error code in heart beat.

```
    if (!ActiveHostsMan::machineRegisted(kvstore_, host)) {
      # return leader change code if meta is follower


```

darionyaphet:
> BTW, prefer to handle error code in heart beat.
> 
> ```
>     if (!ActiveHostsMan::machineRegisted(kvstore_, host)) {
>       # return leader change code if meta is follower
> ```

`setLeaderInfo` have processed this problem

src/clients/meta/MetaClient.cpp
darionyaphet: when the active hosts manager not found the hosts return `E_MACHINE_NOT_FOUND` and set leader message

critical27: I mean only leader could return `E_MACHINE_NOT_FOUND`, follower will return LEADER_CHANGED, so we don't need to have to add code in `MetaClient`

critical27:
Merge first, the code could be improved later

Issue #3993, opened 2022-03-09
Something error when the version of tag/edge schema is greater than 256


sworduo:
If the version of schema is greater than 256, let's say 500, then 'desc tag xxx' would only return the schema with version 255 but not the largest one. All the alter operations including 'alter tag/edge xxx add/drop/change' would have the same problem in v2.0.0. https://github.com/vesoft-inc/nebula/issues/3889#issuecomment-1062588737

```
(root@nebula) [schema_space]> alter tag tag_1 add (abcde string)
Execution succeeded (time spent 3135/3574 us)

Wed, 09 Mar 2022 14:44:05 CST

(root@nebula) [schema_space]> alter tag tag_1 drop (abcde)
[ERROR (-8)]: Not existed!

Wed, 09 Mar 2022 14:44:13 CST
```

Furthermore, in v2.0.0, when the version of shema is 255, the latest alter operation will based on 255 instead of 256 or higher.  When I execute 

```
(root@nebula) [schema_space]> alter tag tag_1 add (abcde string)
Execution succeeded (time spent 2741/3092 us)

Wed, 09 Mar 2022 15:14:55 CST

(root@nebula) [schema_space]> alter tag tag_1 add (abcd1e string)
Execution succeeded (time spent 3675/4125 us)

Wed, 09 Mar 2022 15:16:03 CST

(root@nebula) [schema_space]> alter tag tag_1 add (abcd12e string)
Execution succeeded (time spent 3381/3771 us)

Wed, 09 Mar 2022 15:16:05 CST

(root@nebula) [schema_space]> alter tag tag_1 add (abcd123e string)
Execution succeeded (time spent 2668/2999 us)

Wed, 09 Mar 2022 15:16:07 CST

(root@nebula) [schema_space]> alter tag tag_1 add (abcd1234e string)
Execution succeeded (time spent 3078/3454 us)

Wed, 09 Mar 2022 15:16:10 CST
```

The log is
```

I0309 15:14:14.469921 25032 AlterTagProcessor.cpp:101] Alter Tag tag_1, tagId 531 , new version: 256
I0309 15:14:53.660425 25032 AlterTagProcessor.cpp:101] Alter Tag tag_1, tagId 531 , new version: 256
I0309 15:14:54.676265 25032 AlterTagProcessor.cpp:101] Alter Tag tag_1, tagId 531 , new version: 256
[...]

PR #4023, opened 2022-03-14
fix-bug-when-scheam-version-greater-than-256

sworduo:
## What type of PR is this?
- [☑️] bug
- [ ] feature
- [ ] enhancement

#### Issue(s) number: 
https://github.com/vesoft-inc/nebula/issues/3993
## How do you solve it?
If the schema version got by doprefix is less than 255, just return. Otherwise, it will iterate all the schema belonging to the queried tag to find the largest one.

## Checklist:
Tests:
- [☑️] Unit test(positive and negative cases)
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....


critical27:
Thx for contribution

src/common/utils/MetaKeyUtils.cpp
```diff
@@ -560,6 +560,24 @@ SchemaVer MetaKeyUtils::parseEdgeVersion(folly::StringPiece key) {
          *reinterpret_cast<const SchemaVer*>(key.begin() + offset);
 }
 
+SchemaVer MetaKeyUtils::getLatestEdgeVersion(kvstore::KVIterator* iter, folly::StringPiece& val) {
+  SchemaVer maxVer = MetaKeyUtils::parseEdgeVersion(iter->key());
+  val = iter->val();
+  if (maxVer >= borderScheVer) {
```
critical27, line 566: Actually I think maybe you need to iterate no matter what is in maxVer.
For example, if there are `0x7fffffffffffffff` schemas, `parseEdgeVersion` will return 0....

The root cause is that we encode numbers in little endian, which is not the same order of rocksdb in bytes order.

critical27, line 566: Actually I think maybe you need to iterate no matter what is in maxVer.
For example, if there are 0x7fffffffffffffff schemas (which is std::numeric_limits<SchemaVer>::max()), when we encode into rocksdb, the bytes are `00 00 00 00 00 00 00 00 `, and this will be the first schema since we use rocksdb in byte order, so parseEdgeVersion will return 0....

The root cause is that we encode numbers in little endian, which is not the same order of rocksdb in bytes order.

critical27: Actually it is more like a math problem 😢 , I'm really not good at it...

src/common/utils/MetaKeyUtils.cpp
```diff
@@ -560,6 +560,24 @@ SchemaVer MetaKeyUtils::parseEdgeVersion(folly::StringPiece key) {
          *reinterpret_cast<const SchemaVer*>(key.begin() + offset);
 }
 
+SchemaVer MetaKeyUtils::getLatestEdgeVersion(kvstore::KVIterator* iter, folly::StringPiece& val) {
```
panda-sheep, line 563: Maybe function name `getLatestEdgeVersion` needs to be rename.
Here, not only the maximum version is obtained, but the val corresponding to the maximum version is also obtained.

panda-sheep (commented):
Good cache.

The root cause is a sorting problem caused by little endian mode.
Take the int32_t value `255` or `256` as an example:

int 32_t 255（`0x000000FF`） 
int 32_t 256（`0x00000100`）

The corresponding value  in little endian mode
255:  0x FF  00 00 00 
256   0x 00  01 00  00   

The corresponding value  in big endian mode:
255:  0x 00  00 00 FF 
256   0x 00  00 01 00   

The little endian mode we use, when sorted in ascending order in rocksdb,
255 is after 256, (we would have expected 255 to be before 256)

src/common/utils/MetaKeyUtils.cpp
panda-sheep: Agree with what critical27 said.
Or distinguish the endianness used.
Or traverse all the keys of the tagId and take the largest version.
The second is easier

panda-sheep (approved):
Good job!

critical27 (approved):

Issue #4037, opened 2022-03-16
incorrect version in docker image


HarrisChu:
```
[vesoft@qa-tools nebula-docker-compose]$ docker run -it --rm vesoft/nebula-graphd:v3 --version
nebula-graphd version 2022.03.15-nightly, Git: 2e938c7, Build Time: Mar 15 2022 11:04:03
This source code is licensed under Apache 2.0 License.
```

root cause:
1. we build docker image in rc workflow without version, so the version is nightly in `package.sh`.
2. in release workflow, just copy the docker image.

changes:
1. build rc image with version.

HarrisChu:
BTW
v3 is the latest version for v3.x. i.e. it's the same with v3.0.1 now.
maybe we could just create a v3 branch

Sophie-Xie:
Is only docker wrong, other packages are right?

HarrisChu:
> Is only docker wrong, other packages are right?

yes

PR #4038, opened 2022-03-16
fix version in docker image

HarrisChu:
## What type of PR is this?
- [ ] bug
- [ ] feature
- [ ] enhancement

#### Description:
fix https://github.com/vesoft-inc/nebula/issues/4037

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,11 +1,14 @@
 FROM vesoft/nebula-dev:centos7 as builder
 
 ARG BRANCH=master
+ARG VERSION=
 
 COPY . /home/nebula/BUILD
 
 RUN cd /home/nebula/BUILD/package \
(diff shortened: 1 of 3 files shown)


Aiee (approved):

docker/Dockerfile
```diff
@@ -1,11 +1,14 @@
 FROM vesoft/nebula-dev:centos7 as builder
 
 ARG BRANCH=master
+ARG VERSION=
```
yixinglu, line 4: Why to specify the VERSION argument for docker images? If there's no version parameter for package.sh script, this script will use git to generate it according to the branch.

PR #4040, opened 2022-03-16
remove learner parts when removing space

liwenhui-soul:
## What type of PR is this?
- [ ] bug
- [ ] feature
- [ ] enhancement

#### Description:
till now storage remove space by meta info, which does not contain learner, so the learner parts still exist after remove the space,  this will cause core in next balance.

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....

### DIFF
--- a/src/kvstore/NebulaStore.cpp
+++ b/src/kvstore/NebulaStore.cpp
@@ -39,12 +39,12 @@ NebulaStore::~NebulaStore() {
   raftService_->stop();
   LOG(INFO) << "Waiting for the raft service stop...";
   raftService_->waitUntilStop();
-  spaces_.clear();
-  spaceListeners_.clear();
   bgWorkers_->stop();
   bgWorkers_->wait();
   storeWorker_->stop();
   storeWorker_->wait();
+  spaces_.clear();
+  spaceListeners_.clear();
   LOG(INFO) << "~NebulaStore()";
 }
 
@@ -437,6 +437,13 @@ void NebulaStore::removeSpace(GraphSpaceID spaceId, bool isListener) {
   if (!isListener) {
     auto spaceIt = this->spaces_.find(spaceId);
     if (spaceIt != this->spaces_.end()) {
+      for (auto& [partId, part] : spaceIt->second->parts_) {
+        // before calling removeSpace, meta client would call removePart to remove all parts in
+        // meta cache, which do not contain learners, so we remove them here
+        if (part->isLearner()) {
+          removePart(spaceId, partId, false);
+        }
+      }
       auto& engines = spaceIt->second->engines_;
       for (auto& engine : engines) {
         auto parts = engine->allParts();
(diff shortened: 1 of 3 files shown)


critical27 (approved):
LGTM

src/kvstore/NebulaStore.cpp
```diff
@@ -392,6 +392,9 @@ void NebulaStore::removeSpace(GraphSpaceID spaceId, bool isListener) {
   if (!isListener) {
     auto spaceIt = this->spaces_.find(spaceId);
     if (spaceIt != this->spaces_.end()) {
+      for (auto& [partId, part] : spaceIt->second->parts_) {
+        removePart(spaceId, partId, false);
```
pengweisong, line 396: If added removePart here,  the `engine->removePart(partId);` will not needed below?

liwenhui-soul: actually need it too

pengweisong: Why? Because there is also `e->removePart(partId);` in each `NebulaStore::removePart`, are they duplicated?

src/kvstore/NebulaStore.h
```diff
@@ -581,7 +581,7 @@ class NebulaStore : public KVStore, public Handler {
    * @param spaceId
    * @param partId
    */
-  void removePart(GraphSpaceID spaceId, PartitionID partId) override;
+  void removePart(GraphSpaceID spaceId, PartitionID partId, bool needLock = true) override;
```
liuyu85cn, line 584: how about add some commet to explain in what condition needLock, or not.

src/kvstore/NebulaStore.cpp
liwenhui-soul: engine remove parts by its own part's info, removing old codes in case of uncertainty is not a good idea.

liwenhui-soul: are you certain that this code could be removed? if you are certain I will remove it.

pengweisong: I think it is, but need some tests. You could also sort out the logic to point out it, after all, we could not at code without considering the context.

liwenhui-soul: then who could do the tests? I am certain that this code could stay.

src/kvstore/NebulaStore.cpp
```diff
@@ -490,8 +497,11 @@ nebula::cpp2::ErrorCode NebulaStore::clearSpace(GraphSpaceID spaceId) {
   return nebula::cpp2::ErrorCode::SUCCEEDED;
 }
 
-void NebulaStore::removePart(GraphSpaceID spaceId, PartitionID partId) {
-  folly::RWSpinLock::WriteHolder wh(&lock_);
+void NebulaStore::removePart(GraphSpaceID spaceId, PartitionID partId, bool needLock) {
+  folly::RWSpinLock::WriteHolder wh(nullptr);
```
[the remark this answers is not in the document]
liwenhui-soul: do nothing, you can see its code

pengweisong (approved):
LGTM

src/kvstore/NebulaStore.cpp
```diff
@@ -490,8 +497,11 @@ nebula::cpp2::ErrorCode NebulaStore::clearSpace(GraphSpaceID spaceId) {
...
+  if (needLock) {
+    wh.reset(&lock_);
+  }
```
darionyaphet, line 504: would you support more explanation? thanks

liwenhui-soul: if need lock, wh will hold lock_

src/kvstore/NebulaStore.cpp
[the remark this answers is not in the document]
darionyaphet: if so why don't remove this line?

critical27 (approved):

PR #4045, opened 2022-03-17
refactor job manager part3 

panda-sheep:
## What type of PR is this?
- [ ] bug
- [ ] feature
- [x] enhancement

#### Description:
rename `AdminCmd` to `JobType`, 
1） `adminclient` not only handles jobs, but also other handles commands. So the `AdminCmd` here is easily misunderstood. 
2） `AdminCmd` is just the type of job，It is more appropriate to use `JobType`

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....


liuyu85cn (approved):
good job

critical27 (approved):

CPWstatic (approved):

PR #4053, opened 2022-03-20
Stop scan edge opt when build path

wey-gu:
## What type of PR is this?
- [x] bug
- [ ] feature
- [ ] enhancement

## How do you solve it?

Stop scanV -> scanE optRule when buildPath
> update: after eliminate AV being merged, the PR is changed during rebase:

- Changed GetEdgeTransformRule:
  - Add a Project to be matched to exclude `match p=()-[e]->() return p limit 3`
  - Removed AV(AVorLimit --> Limit), as AV in this rule will end up losing limit-count to lead scanEdge(limit:-1) in `match ()-[e]->() return e,e limit 3`
- Added GetEdgesTransformAppendVerticesLimitRule: (SV, TR, AV, L, Prj) ->(SE, Prj, AV, L, Prj)
  - doing the same thing as the changed GetEdgeTransformRule for `match ()-[e]->() return e limit 3`

> previous version before eliminate AV being merged
>
> - delete: av->traverse->scanV
> - add: project-> limit -> av->traverse->sv
>   - exclude optRule for buildPath in project.colums()

```sql
(root@nebula) [basketballplayer]> show tag indexes
+------------------+----------+----------+
| Index Name       | By Tag   | Columns  |
+------------------+----------+----------+
| "player_index_0" | "player" | []       |
| "player_index_1" | "player" | ["name"] |
+------------------+----------+----------+
Got 2 rows (time spent 2891/3892 us)

## before this PR, below return path was pushed down, while it cannot be pushed down, we should prevent it.
(root@nebula) [basketballplayer]> match p=(:`team`)-->() return p limit 3
+----------+
| p        |
+----------+
| BAD_TYPE |
| BAD_TYPE |
| BAD_TYPE |
+----------+
Got 3 rows (time spent 6144/26093 us)

## afterwards

# with this PR, this will not be optimized to scanE, which led the wrong pushdown, instead it use scanV(and here for p=(:`player`)-->() return p, it will leverage tag-index)

(root@nebula) [basketballplayer]> match p=(:`team`)-->() return p limit 3
[ERROR (-1005)]: Scan vertices or edges need to specify a limit number, or limit number can not push down.

Sun, 20 Mar 2022 07:24:38 UTC

```

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A
- [x] Manual Test

Affects:

## Release notes:

> ex. Fixed the bug .....

Fixed the bug on the indexless case of `match p=(:team)--() return p limit 1` end up with badtype, instead, it should be rejected due to being unable to pushdown.


Shylock-Hg:
Format the code  by `make clang-format`

wey-gu:
> Format the code by `make clang-format`

will do so, thanks!

wey-gu:
Now I see a regression case, will look into that now.

```
E   AssertionError: Fail to exec: 

MATCH ()-[e:is_teacher]->() RETURN type(e) AS Type, e.start_year AS StartYear, e.end_year AS EndYear LIMIT 3

error: Scan vertices or edges need to specify a limit number, or limit number can not push down.
```

wey-gu:
After looking into it, it turned out when the returned part referred to `e` for more than once, i.e. `return e, e as e_`, it will:

- a. Create two ProjectBelow -> ProjectAbove in a row, the duplicated pathBuild will be created in ProjectBelow
- b. (thanks @Shylock-Hg , pointed me to this)There is already a `CollapseProjectRule` to make the two projects into one, while, it will not do the transform when there is one prop(`e` here, for example) being referred for more than once

For fact b. I don't quite understand it now, was asking @czpmango on this offline.

The solution may be either create another rule to fit the `project-project-appendV-traverse-scanV` pattern or add `project-project(buildPath) ==> project(without buildPath in case no path output)` rule.

> a 
```
(root@nebula) [basketballplayer]> explain match p=()-[e]->() return e, e.start_year as st
Execution succeeded (time spent 12523634/12521037 us)

Execution Plan (optimize time 12520901 us)

-----+----------------+--------------+----------------+---------------------------------------------------------------
| id | name           | dependencies | profiling data | operator info                                                |
-----+----------------+--------------+----------------+---------------------------------------------------------------
|  7 | Project        | 6            |                | outputVar: [                                                 |
|    |                |              |                |   {                                                          |
|    |                |              |                |     "colNames": [                                            |
|    |                |              |                |       "e",                                                   |
|    |                |              |                |       "st"                                                   |
|    |                |              |                |     ],                                                       |
|    |                |              |                |     "name": "__Project_7",                                   |
|    |                |              |                |     "type": "DATASET"                                        |
|    |                |              |                |   }                                                          |
|    |                |              |                | ]                                                            |
|    |                |              |                | inputVar: __Project_6                                        |
|    |                |              |                | columns: [                                                   |
|    |                |              |                |   "$e",                                                      |
|    |                |              |                |   "$e.start_year"                                            |
|    |                |              |                | ]                                                            |
-----+----------------+--------------+----------------+---------------------------------------------------------------
|  6 | Project        | 5            |                | outputVar: [                                                 |
|    |                |              |                |   {                                                          |
|    |                |              |                |     "colNames": [                                            |
|    |                |              |                |       "e",                                                   |
|    |                |              |                |       "p"                                                    |
|    |                |              |                |     ],                                                       |
|    |                |              |                |     "name": "__Project_6",                                   |
|    |                |              |                |     "type": "DATASET"                                        |
|    |                |              |                |   }                                                          |
|    |                |              |                | ]                                                            |
|    |                |              |                | inputVar: __AppendVertices_5                                 |
|    |                |              |                | columns: [                                                   |
|    |                |              |                |   "$-.e[0] AS e",                                            |
|    |                |              |                |   "PathBuild[$-.__VAR_0,$-.e,$-.__VAR_1] AS p"               |
|    |                |              |                | ]                                                            |
-----+----------------+--------------+----------------+---------------------------------------------------------------
```

> b
```c++
  // disable this case to avoid the expression in ProjBelow being eval multiple
  // times
  std::unordered_set<std::string> uniquePropRefNames;
  std::unordered_set<std::string> multiRefColNames;
  for (auto p : allPropRefNames) {
    if (!uniquePropRefNames.insert(p).second) {
      // Records PropRefNames that are referenced multiple times
      multiRefColNames.insert(p);
    }
  }

  // 2. find link according to propRefNames and colNames in ProjBelow
  std::unordered_map<std::string, Expression*> rewriteMap;
  auto colNames = projBelow->colNames();
  for (size_t i = 0; i < colNames.size(); ++i) {
    if (uniquePropRefNames.count(colNames[i])) {
      auto colExpr = colsBelow[i]->expr();
      // disable this case to avoid the expression in ProjBelow being eval multiple
      // times
      if (!graph::ExpressionUtils::isPropertyExpr(colExpr) && multiRefColNames.count(colNames[i])) {
        return TransformResult::noTransform();
      }
      rewriteMap[colNames[i]] = colExpr;
    }
  }
```

wey-gu:
I am stuck here as when the return part comes with more than one `e` with limit, it won't be matched to `prj->limit->av->traverse->scanVertices`, instead it matched the pattern of `prj-->av->traverse->scanVertices`, where the limit information cannot be fetched(for all node, it's -1).

The strange thing is the final opt plan with `explain MATCH ()-[e:serve]->() RETURN e,properties(e) limit 1` is `prj->limit->av->traverse->scanVertices`, if I disabled `prj-->av->traverse->scanVertices`, although it will be never matched to this patter.

I will talk to @Shylock-Hg closely tomorrow.

I may not be catching up with this release cycle but I will try my best to do it.

Sorry.

wey-gu:
After discussions with @Shylock-Hg and @yixinglu, something should be wrong somewhere else that had led to the pattern not being mapped, will look into it later.

Sophie-Xie:
Will you continue to fix it? It needs to be merged into version 3.2. Thanks.

wey-gu:
> Will you continue to fix it? It needs to be merged into version 3.2. Thanks.

I will look into it again before the end of July, in case collaboration is needed, will request it soon.

wey-gu:
Update:
Today I spent some time figuring out the issue before:

issue:
- ` MATCH ()-[e:serve]->() RETURN e LIMIT 3` works fine, it will be matched to new rule `GetEdgesTransformLimitRule`
- `MATCH ()-[e:serve]->() RETURN e,e LIMIT 3` will end up entering `GetEdgesTransformRule` now, due to `GetEdgesTransformRule` was changed to match with extra project, the `traverse->limit()` and all other nodes' limit will be -1, thus the scanEdge.limit will be `-1`, and due to the LIMIT isn't pushdown, the query cannot be made

Because the goal of `GetEdgesTransformRule` was only to address the case with a limit sample when the edge comes without indexes, thus I could remove `GetEdgesTransformRule`. Actually, the pattern can be matched to `GetEdgesTransformLimitRule` anyway when there is no `GetEdgesTransformRule` to short circuit it, and everything works as expected.

wey-gu:
The [UT Failure](https://github.com/vesoft-inc/nebula/runs/6641638067?check_suite_focus=true) seems irrelevant

```python
=================================== FAILURES ===================================
_______________________ test_lookuptest_yieldclausetest ________________________
[gw4] linux -- Python 3.8.10 /usr/bin/python3
/github/home/.local/lib/python3.8/site-packages/pytest_bdd/scenario.py:177: in scenario_wrapper
    _execute_scenario(feature, scenario, request, encoding)
/github/home/.local/lib/python3.8/site-packages/pytest_bdd/scenario.py:143: in _execute_scenario
    _execute_step_function(request, scenario, step, step_func)
/github/home/.local/lib/python3.8/site-packages/pytest_bdd/scenario.py:113: in _execute_step_function
    return_value = step_func(**kwargs)
tck/conftest.py:736: in execution_should_be_succ
    check_resp(rs, stmt)
common/utils.py:334: in check_resp
    assert resp.is_succeeded(), msg
E   AssertionError: Fail to exec: INSERT VERTEX student(number, age), teacher(number, age)  VALUES "220":(1, 20, 1, 30), "221":(2, 22, 2, 32), error: Storage Error: Tag not found.
___________________ test_indextest_rebuildtagindexstatusinfo ___________________
```

tests/tck/features/optimizer/GetEdgesTransformLimitRule.feature
```diff
@@ -0,0 +1,23 @@
+# Copyright (c) 2021 vesoft inc. All rights reserved.
```
Shylock-Hg, line 1: Put this file in `bugfix` folder.

wey-gu: OK, got it.

wey-gu:
@Shylock-Hg 
Regarding the rebase, I didn't manually test yet as now busy with other things, will do it later. I have below thoughts, could you help double-check if it is correct, first?

```
    Rebase on #4277 Remove useless AppendVertices

    As this rule is only applied for
       match ()-[e]->() return e limit 1
    when 1. there is no index on all edges and 2. limit clause
    exists, thus we could assert that AppendVertices
    will be removed by EliminateAppendVerticesRule

    //  Tranformation:
    //  Before:
    // +---------+---------+
    // |      Project      |
    // +---------+---------+
    //           |
    // +---------+---------+
    // |       Limit       |
    // +---------+---------+
    //           |
    // +---------+---------+
    // |      Traverse     |
    // +---------+---------+
    //           |
    // +---------+---------+
    // |    ScanVertices   |
    // +---------+---------+
    //
    //  After:
    // +---------+---------+
    // |      Project      |
    // +---------+---------+
    //           |
    // +---------+---------+
    // |       Limit       |
    // +---------+---------+
    //           |
    // +---------+---------+
    // |      Project      |
    // +---------+---------+
    //           |
    // +---------+---------+
    // |      ScanEdges    |
    // +---------+---------+
```

Shylock-Hg:
> @Shylock-Hg Regarding the rebase, I didn't manually build/test yet as now busy with other things, will do it later. I have below thoughts, could you help double-check if it is correct, first?
> 
> ```
>     Rebase on #4277 Remove useless AppendVertices
> 
>     As this rule is only applied for
>        match ()-[e]->() return e limit 1
>     when 1. there is no index on all edges and 2. limit clause
>     exists, thus we could assert that AppendVertices
>     will be removed by EliminateAppendVerticesRule
> 
>     //  Tranformation:
>     //  Before:
>     // +---------+---------+
>     // |      Project      |
>     // +---------+---------+
>     //           |
>     // +---------+---------+
>     // |       Limit       |
>     // +---------+---------+
>     //           |
>     // +---------+---------+
>     // |      Traverse     |
>     // +---------+---------+
>     //           |
>     // +---------+---------+
>     // |    ScanVertices   |
>     // +---------+---------+
>     //
>     //  After:
>     // +---------+---------+
>     // |      Project      |
>     // +---------+---------+
>     //           |
>     // +---------+---------+
>     // |       Limit       |
>     // +---------+---------+
>     //           |
>     // +---------+---------+
>     // |      Project      |
>     // +---------+---------+
>     //           |
>     // +---------+---------+
>     // |      ScanEdges    |
>     // +---------+---------+
> ```



> @Shylock-Hg Regarding the rebase, I didn't manually build/test yet as now busy with other things, will do it later. I have below thoughts, could you help double-check if it is correct, first?
> 
> ```
>     Rebase on #4277 Remove useless AppendVertices
> 
>     As this rule is only applied for
>        match ()-[e]->() return e limit 1
>     when 1. there is no index on all edges and 2. limit clause
>     exists, thus we could assert that AppendVertices
>     will be removed by EliminateAppendVerticesRule
> 
>     //  Tranformation:
>     //  Before:
>     // +---------+---------+
>     // |      Project      |
>     // +---------+---------+
>     //           |
>     // +---------+---------+
>     // |       Limit       |
>     // +---------+---------+
>     //           |
>     // +---------+---------+
>     // |      Traverse     |
>     // +---------+---------+
>     //           |
>     // +---------+---------+
>     // |    ScanVertices   |
>     // +---------+---------+
>     //
>     //  After:
>     // +---------+---------+
>     // |      Project      |
>     // +---------+---------+
>     //           |
>     // +---------+---------+
>     // |       Limit       |
>     // +---------+---------+
>     //           |
>     // +---------+---------+
>     // |      Project      |
>     // +---------+---------+
>     //           |
>     // +---------+---------+
>     // |      ScanEdges    |
>     // +---------+---------+
> ```

See the comment

src/graph/optimizer/rule/GetEdgesTransformLimitRule.cpp
```diff
@@ -24,59 +24,73 @@ using nebula::graph::Traverse;
...
+    std::unique_ptr<GetEdgesTransformLimitRule>(new GetEdgesTransformLimitRule());
 
-GetEdgesTransformRule::GetEdgesTransformRule() {
+GetEdgesTransformLimitRule::GetEdgesTransformLimitRule() {
   RuleSet::QueryRules().addRule(this);
 }
 
-const Pattern &GetEdgesTransformRule::pattern() const {
-  static Pattern pattern =
-      Pattern::create({PlanNode::Kind::kAppendVertices, PlanNode::Kind::kLimit},
```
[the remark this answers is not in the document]
Shylock-Hg: Why don't you keep the origin code ?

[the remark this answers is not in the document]
wey-gu: Previously I used to keep the original rule (with an extra build-path condition check) but found the original rule will anyway “short circuit” this new rule, no matter how the order in the CMake list of the two rules, while when the original rule being applied with transform for `MATCH ()-[e:serve]->() RETURN e,e LIMIT 3`(e being returned more than once, or `return e.PropA, properties(e)` ), there is no `LIMIT` info in that pattern, thus it will end up with limit(-1) to lead it cannot be pushed down. Then I found we actually only care about this pattern when 1. there is no edge index; 2. there is a `LIMIT` clause, thus I could actually remove the original short rule, and it worked fine for all regression cases.

Now, with your change on that rule being merged, the `orLimit` is in the pattern, maybe I should revisit to see if I could  leave the two, and add `buildPath` check in the original rule instead of removing them, I have two talks to be prepared for tonight and tomorrow night, I will come back to this later soon.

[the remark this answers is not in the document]
wey-gu: > Ok, and you should format the feature file by `make -C tests fmt` to avoid CI lint error.

I don't know this, too, thanks!!

wey-gu:
update: I managed to make it work in my env, just pushed to check UT(don't have time to run UT locally before going to bed), and will continue tomorrow.

src/graph/optimizer/rule/GetEdgesTransformRule.cpp
```diff
@@ -32,18 +32,20 @@ GetEdgesTransformRule::GetEdgesTransformRule() {
 }
 
 const Pattern &GetEdgesTransformRule::pattern() const {
-  static Pattern pattern =
-      Pattern::create({PlanNode::Kind::kAppendVertices, PlanNode::Kind::kLimit},
-                      {Pattern::create(PlanNode::Kind::kTraverse,
-                                       {Pattern::create(PlanNode::Kind::kScanVertices)})});
+  static Pattern pattern = Pattern::create(
```
wey-gu, line 35: Note:
AV was removed here because I found that when handling `match ()-[e]->() return e, e limit 3` in this RULE transform, the limit count is lost in all operators, thus it will end up cannot be pushdown with SE(limit=-1).
When AV is removed from here, the other rule GetEdgesTransformAppendVerticesLimitRule will take care of the cases with AV.

Also, `match ()-[e]->() return e limit 3` is quite different from `match ()-[e]->() return e,e limit 3`

src/graph/optimizer/rule/GetEdgesTransformAppendVerticesLimitRule.h
```diff
@@ -0,0 +1,95 @@
...
+
+  bool match(OptContext *ctx, const MatchedResult &matched) const override;
+  StatusOr<TransformResult> transform(OptContext *ctx, const MatchedResult &matched) const override;
+
+  std::string toString() const override;
+
+ private:
+  GetEdgesTransformAppendVerticesLimitRule();
+
+  static graph::ScanEdges *traverseToScanEdges(const graph::Traverse *traverse,
```
Shylock-Hg, line 84: I think you could reuse these methods.

wey-gu: Sure, will reuse it/avoid duplicating the same function

wey-gu: It's now reused.

Shylock-Hg: I think you could define a common father class to do it.

wey-gu: OK, will do so.

Shylock-Hg (approved):

CPWstatic (approved):

PR #4055, opened 2022-03-21
fix kv snapshot data inconsist

liwenhui-soul:
## What type of PR is this?
- [ ] bug
- [ ] feature
- [ ] enhancement

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....

### DIFF
--- a/src/kvstore/Part.cpp
+++ b/src/kvstore/Part.cpp
@@ -433,9 +433,19 @@ nebula::cpp2::ErrorCode Part::cleanup() {
   LOG(INFO) << idStr_ << "Clean rocksdb part data";
   auto batch = engine_->startBatchWrite();
   // Remove the vertex, edge, index, systemCommitKey, operation data under the part
+
+  const auto& kvPre = NebulaKeyUtils::kvPrefix(partId_);
+  auto ret =
+      batch->removeRange(NebulaKeyUtils::firstKey(kvPre, 128), NebulaKeyUtils::lastKey(kvPre, 128));
+  if (ret != nebula::cpp2::ErrorCode::SUCCEEDED) {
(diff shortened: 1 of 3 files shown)


src/kvstore/Part.cpp
```diff
@@ -433,9 +433,19 @@ nebula::cpp2::ErrorCode Part::cleanup() {
   LOG(INFO) << idStr_ << "Clean rocksdb part data";
   auto batch = engine_->startBatchWrite();
   // Remove the vertex, edge, index, systemCommitKey, operation data under the part
+
+  const auto& kvPre = NebulaKeyUtils::kvPrefix(partId_);
+  auto ret =
+      batch->removeRange(NebulaKeyUtils::firstKey(kvPre, 128), NebulaKeyUtils::lastKey(kvPre, 128));
```
darionyaphet, line 439: why `128`? Is it a magic number?

liwenhui-soul: relax....it's not certain yet...

critical27: I am not sure how to handle it …… For example, if we want to delete part 3's kv data. The prefix is `06 03 00 00` in hex, can we delete the range by `[06 03 00 00, 06 03 00 01]`?

critical27:
Could merge it for now, the `firstKey` and `lastKey` could check later.

critical27 (approved):

kikimo (approved):
LGTM

src/kvstore/Part.cpp
1150310621: why not just part_id plus 1 as the last key?

1150310621: sorry, forget that part_id is encoded in little endian.

PR #4061, opened 2022-03-21
fix bug stop job before it start

liwenhui-soul:
## What type of PR is this?
- [x] bug
- [ ] feature
- [ ] enhancement

#### Description:
this pr is to solve two points:
1.  if stop a job quickly when schedule is going to start the job, it may:
    the job does not exist in runningJobs_ , but exist in inFlightJobs_, so we can't create a new one even if we stopped it.
2. the start time of the job is 0 before starting, so it would always be expired through the calculation of duration, the job's info  would be erased and we can't see it in show list.

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....

### DIFF
--- a/src/meta/processors/job/JobManager.cpp
+++ b/src/meta/processors/job/JobManager.cpp
@@ -126,6 +126,7 @@ void JobManager::scheduleThread() {
     auto jobOp = std::get<0>(opJobId);
     auto jodId = std::get<1>(opJobId);
     auto spaceId = std::get<2>(opJobId);
+    std::lock_guard<std::recursive_mutex> lk(muJobFinished_[spaceId]);
     auto jobDescRet = JobDescription::loadJobDescription(spaceId, jodId, kvStore_);
     if (!nebula::ok(jobDescRet)) {
       LOG(INFO) << "[JobManager] load an invalid job from space " << spaceId << " jodId " << jodId;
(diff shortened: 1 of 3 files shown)


src/meta/processors/job/JobManager.cpp
```diff
@@ -232,8 +231,9 @@ nebula::cpp2::ErrorCode JobManager::jobFinished(JobID jobId, cpp2::JobStatus job
 
   auto it = runningJobs_.find(jobId);
   if (it == runningJobs_.end()) {
-    LOG(INFO) << folly::sformat("can't find jobExecutor, jobId={}", jobId);
-    return nebula::cpp2::ErrorCode::E_UNKNOWN;
+    // the job has not started yet
```
panda-sheep, line 234: If here `jobId`  is not found from `runningJobs_`, JobExecutor should be created.

For example, meta leader change, reportTaskFinish is sent to the new meta leader, and the runningJobs_ in the new meta leader is empty.

liwenhui-soul: that's a problem

panda-sheep: If you can't find it here, create a new JobExecutor

liwenhui-soul: yes, but only for storage jobs

src/meta/processors/job/JobManager.cpp
```diff
@@ -513,10 +514,10 @@ ErrorOr<nebula::cpp2::ErrorCode, std::vector<cpp2::JobDesc>> JobManager::showJob
 
 bool JobManager::isExpiredJob(JobDescription& jobDesc) {
   auto status = jobDesc.getStatus();
-  if (status == cpp2::JobStatus::QUEUE || status == cpp2::JobStatus::RUNNING) {
+  auto jobStart = jobDesc.getStartTime();
+  if (status == cpp2::JobStatus::QUEUE || status == cpp2::JobStatus::RUNNING || jobStart == 0) {
```
panda-sheep, line 518: The job of stop, starttime is 0, can never be deleted

panda-sheep (approved):
Good

critical27 (approved):

PR #4078, opened 2022-03-25
remove bzip2 fatal dependency

jiayuehua:
## What type of PR is this?
- [*] bug
- [ ] feature
- [ ] enhancement

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....

### DIFF
--- a/cmake/nebula/ThirdPartyConfig.cmake
+++ b/cmake/nebula/ThirdPartyConfig.cmake
@@ -78,9 +78,7 @@ endif()
 
 message("")
 
-find_package(Bzip2 REQUIRED)
 find_package(DoubleConversion REQUIRED)
-find_package(Fatal REQUIRED)
 find_package(Fbthrift REQUIRED)
 find_package(Folly REQUIRED)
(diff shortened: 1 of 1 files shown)


jiayuehua:
https://discuss.nebula-graph.com.cn/t/topic/8063

jiayuehua:
https://discuss.nebula-graph.com.cn/t/topic/7858

yixinglu:
Why remote these libraries? they have been pre-installed in our thirdparty libraries:

image

jiayuehua:
the project don't use them. And Users has report build error.

yixinglu:
> the project don't use them. And Users has report build error.

Some libraries are just not used by default in some configuration. it doesn't mean they are useless later. You should help the users to fix the build error rather than removing these libraries.

jiayuehua:
those two libs are totally not needed by the project. If needed in the future, than should added in the future.

PR #4093, opened 2022-03-29
Add standalone docker build

SuperYoko:
## What type of PR is this?
- [ ] bug
- [ ] feature
- [x] enhancement

#### Description:
Support standalone version docker for cloud use.

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -42,6 +42,7 @@ jobs:
           - metad
           - storaged
           - tools
+          - standalone
     steps:
       - uses: webiny/action-post-run@2.0.1
         with:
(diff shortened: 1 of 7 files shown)


.github/workflows/release.yml
```diff
@@ -42,6 +42,7 @@ jobs:
           - metad
           - storaged
           - tools
+          - standalone
```
HarrisChu, line 45: should also change workflow in rc.yml.

workflow in `release` is just copy rc to release.

SuperYoko:
close to do it in Ent

PR #4101, opened 2022-03-30
add code for part peers backward compatible

pengweisong:
## What type of PR is this?
- [x] bug
- [ ] feature
- [ ] enhancement

#### Description:
When persist peers info in the storage, we should remember to fill the data which upgrade from the old version.

## How do you solve it?

Move the balance peer info to another system table: balance table.
When start the storaged, we should:
1. load the balance parts.
2. load the local parts. 
    - If the part already exist in the balance parts, skip it.
    - Else if the part the part not in the part manager, skip it.
3. load the parts from part manager.
    - If the part already been created, skip it.
    - Will not erase any part in this phase.

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....


src/kvstore/NebulaStore.cpp
```diff
@@ -104,35 +107,72 @@ void NebulaStore::loadPartFromDataPath() {
...
-          bool isNormalPeer = true;
-
-          Peer raftPeer;
-          bool exist = raftPeers.get(raftAddr_, &raftPeer);
-          if (exist) {
-            if (raftPeer.status != Peer::Status::kNormalPeer) {
-              isNormalPeer = false;
+
+        // load balancing part info which persisted to local engine.
+        for (auto& [partId, raftPeers] : engine->balancePartPeers()) {
```
liwenhui-soul, line 114: sorry, I don't understand, what does balancePartPeers mean, shouldn't we load allPartPeers?

liwenhui-soul: I known, ignore it

liwenhui-soul:
good~

src/kvstore/Common.h
```diff
@@ -106,26 +108,37 @@ inline std::ostream& operator<<(std::ostream& os, const Peer& peer) {
 struct Peers {
  private:
   std::map<HostAddr, Peer> peers;
+  int createdTime;
 
  public:
-  Peers() {}
+  Peers() {
+    createdTime = static_cast<int>(time::WallClock::fastNowInSec());
```
panda-sheep, line 115: `static_cast<int> ` can be removed? 
Use `int64_t createdTime` to avoid conversion

src/kvstore/NebulaStore.cpp
pengweisong: That interface has been removed~

src/kvstore/Common.h
pengweisong: int64_t is a little difficult to encode and decode?

critical27 (commented):
Generally LGTM, you could modify the `int` to `int32_t` or `int64_t`

src/kvstore/NebulaStore.cpp
```diff
@@ -104,35 +107,78 @@ void NebulaStore::loadPartFromDataPath() {
...
+
+          if (raftPeers.isExpired()) {
+            LOG(INFO) << "Space: " << spaceId << ", part:" << partId
+                      << " balancing info expired, ignore it.";
+            continue;
+          }
+
+          auto spacePart = std::make_pair(spaceId, partId);
+          if (partSet.find(spacePart) == partSet.end()) {
+            partSet.insert(std::make_pair(spaceId, partId));
```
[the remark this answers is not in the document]
pengweisong: OK

src/kvstore/Common.h
pengweisong: done

critical27 (approved):
LGTM

liwenhui-soul (approved):

PR #4104, opened 2022-03-31
Limits for add and recover balance data or zone balance jobs

panda-sheep:
## What type of PR is this?
- [ ] bug
- [ ] feature
- [x] enhancement

#### Description:
```
Only affects balance data or zone balance job

1) When adding a job
If a space has failed balance data or zone balance jobs
You cannot add new balance data or zone balance jobs to this space, 
you must first recover the failed balance data or zone balance jobs.
If the job keeps failing, you can stop the job first, and then add a new job of this type

2) When recovering the job
For  balance data or zone balance job.
If you recover a failed job, you can recover.
If you recover a stopped job, compare it by starttime to see if there is an updated failed job and finished job of this type after the stopped job.
If there is, it cannot be recovered, otherwise it can be recovered

For the same type of job, when recovering multiple stopped jobs at a time. 
Only recover the latest stopped job. 
 
The same type of STOPPED job exists in the following form:
job1(stopped) job2(failed)
recover job job1 failed
recover job job2 success

job1(stopped) job2(fininshed) job3(stopped)
recover job job1 failed
recover job job3 success
recover job job1, job3 Only job3 can recover
```
## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....


src/meta/processors/job/AdminJobProcessor.cpp
```diff
@@ -125,12 +125,19 @@ nebula::cpp2::ErrorCode AdminJobProcessor::addJobProcess(const cpp2::AdminJobReq
...
+  auto runningJobExist = jobMgr_->checkOnRunningJobExist(spaceId_, type, paras, jId);
+  if (runningJobExist) {
     LOG(INFO) << "Job has already exists: " << jId;
     result.job_id_ref() = jId;
     return nebula::cpp2::ErrorCode::SUCCEEDED;
   }
+  if (type == cpp2::JobType::DATA_BALANCE || type == cpp2::JobType::ZONE_BALANCE) {
+    auto retCode = jobMgr_->checkNotFinishedJobExist(spaceId_, type);
+    if (retCode != nebula::cpp2::ErrorCode::SUCCEEDED) {
+      LOG(INFO) << "There are not finished data balance jobs or zone balance jobs.";
```
wenhaocs, line 137: ditto

src/meta/processors/job/JobDescription.h
```diff
@@ -159,8 +159,7 @@ class JobDescription {
   cpp2::JobDesc toJobDesc();
 
   bool operator==(const JobDescription& that) const {
-    return space_ == that.space_ && type_ == that.type_ && paras_ == that.paras_ &&
-           status_ == that.status_;
```
wenhaocs, line 163: want to double check this is intended

src/meta/processors/job/AdminJobProcessor.cpp
```diff
@@ -125,12 +125,19 @@ nebula::cpp2::ErrorCode AdminJobProcessor::addJobProcess(const cpp2::AdminJobReq
...
   JobID jId = 0;
-  auto jobExist = jobMgr_->checkJobExist(spaceId_, type, paras, jId);
-  if (jobExist) {
+  auto runningJobExist = jobMgr_->checkOnRunningJobExist(spaceId_, type, paras, jId);
+  if (runningJobExist) {
     LOG(INFO) << "Job has already exists: " << jId;
     result.job_id_ref() = jId;
     return nebula::cpp2::ErrorCode::SUCCEEDED;
   }
+  if (type == cpp2::JobType::DATA_BALANCE || type == cpp2::JobType::ZONE_BALANCE) {
```
wenhaocs, line 134: is this if statement redundant as it is checked in the function `checkNotFinishedxxx`

src/meta/processors/job/JobManager.cpp
```diff
@@ -587,6 +587,36 @@ bool JobManager::checkJobExist(GraphSpaceID spaceId,
...
+        continue;
+      }
+      auto tup = MetaKeyUtils::parseJobVal(iter->val());
+      auto type = std::get<0>(tup);
+      auto status = std::get<2>(tup);
+      if (type == cpp2::JobType::DATA_BALANCE || type == cpp2::JobType::ZONE_BALANCE) {
+        if (status != cpp2::JobStatus::FINISHED && status != cpp2::JobStatus::INVALID) {
+          LOG(INFO) << "There are some data balance or zone balance jobs that did not complete "
+                       "successfully";
+          return nebula::cpp2::ErrorCode::E_JOB_MAYBE_RECOVER;
```
wenhaocs, line 612: Do not understand "maybe recover". May change to "E_JOB_UNFINISHED"

src/clients/meta/MetaClient.cpp
```diff
@@ -915,6 +915,8 @@ Status MetaClient::handleResponse(const RESP& resp) {
       return Status::Error("No valid job!");
     case nebula::cpp2::ErrorCode::E_JOB_NOT_IN_SPACE:
       return Status::Error("Job not existed in chosen space!");
+    case nebula::cpp2::ErrorCode::E_JOB_MAYBE_RECOVER:
+      return Status::Error("There is not finished data balance or zone balance job!");
```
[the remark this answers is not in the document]
panda-sheep: Good, 👍

src/meta/processors/job/JobDescription.h
panda-sheep: Yes.
The jobs in `inFlightJobs_` are all jobs that need to be run and have three states:
`QUEUE`: when adding a job
`FAILED`, `STOPPED`: recover job

So we use `checkOnRunningJobExist` to check if the job in `inFlightJobs_` is running, should not check the status.
For example, there are recover jobs of `FAILED` and `STOPPED` in `inFlightJobs_`. At this time, add the same job again. It was successful in the past, but now we do not allow it because they are the same job.

src/meta/processors/job/JobManager.cpp
panda-sheep: change to E_JOB_NEED_RECOVER

src/interface/common.thrift
```diff
@@ -391,7 +391,7 @@ enum ErrorCode {
     E_JOB_NOT_FINISHED                = -2048,
     E_TASK_REPORT_OUT_DATE            = -2049,
     E_JOB_NOT_IN_SPACE                = -2050,
-    E_JOB_MAYBE_RECOVER               = -2051,
+    E_JOB_NEED_RECOVER               = -2051,
```
darionyaphet, line 394: maybe need alignment

critical27:
Ready?

critical27 (commented):
Generally LGTM, @liwenhui-soul please confirm if this PR satisfy.

src/interface/common.thrift
panda-sheep: ok

panda-sheep:
> Ready?

yeah

critical27 (approved):

src/meta/processors/job/JobManager.cpp
```diff
@@ -587,6 +587,34 @@ bool JobManager::checkJobExist(GraphSpaceID spaceId,
...
+
+    for (; iter->valid(); iter->next()) {
+      if (!MetaKeyUtils::isJobKey(iter->key())) {
+        continue;
+      }
+      auto tup = MetaKeyUtils::parseJobVal(iter->val());
+      auto type = std::get<0>(tup);
+      auto status = std::get<2>(tup);
+      if (type == cpp2::JobType::DATA_BALANCE || type == cpp2::JobType::ZONE_BALANCE) {
+        if (status == cpp2::JobStatus::FAILED || status == cpp2::JobStatus::STOPPED) {
```
liwenhui-soul, line 609: how about remove the condition of STOPPED, because if the job always fail, we can't create a new one, so if a job failed and we don't want to continue, we stop it and create a new one, this way could be equal to "reset balance plan" in the past.

panda-sheep: make sense

liwenhui-soul (approved):

src/meta/processors/job/JobManager.cpp
```diff
@@ -671,7 +701,121 @@ ErrorOr<nebula::cpp2::ErrorCode, uint32_t> JobManager::recoverJob(
...
+      totalJobKVs.emplace_back(std::make_pair(jobKeys[i], jobVals[i]));
+    }
+
+    // For DATA_BALANCE and ZONE_BALANCE job, jobs with STOPPED, FAILED, QUEUE status
+    // The following situations can be recovered:
+    // Queue: The job has not been executed, the machine restarted.
+    // FAILED: The failed jobid will be executed. FAILED and QUEUE jobs will not exist at the same
+    // time. STOPPED: If only one stopped jobId is specified, there will not be finished job or
+    // failed jobId after the job. If multiple jobIds are specified, only last jobId will be
+    // recovered. there will not be finished job or failed jobId after the job. The form in which
```
critical27, line 713: What does `there will not be finished job or failed jobId after the job.`  mean?
You mean once a job is STOPPED, we can't add any job?

BTW `there will not be finished job or failed jobId after the job.` is duplicated in the comment

critical27:
So complicated ... I'm confusing ... Several question, majorly related to RECOVER:
1. If we could recover multiple job, one failed, one recovered, this is weird.
2. Do we really need to support recover STOPPED job? @liuyu85cn PTAL

src/meta/processors/job/JobManager.cpp
panda-sheep: `You mean once a job is STOPPED, we can't add any job?`  can add this type job
but the specified type job failed， can't add this type job.

The following example, in the order of the starttime of the job
`job1(stopped) job2(failed)`
recover job job1     failed
recover job job2     success

`job1(stopped) job2(fininshed) job3(stopped)`
recover job job1     failed
recover job job3     success
recover job job1, job3   Only job3 can recover

panda-sheep:
> So complicated ... I'm confusing ... Several question, majorly related to RECOVER:
> 
> 1. If we could recover multiple job, one failed, one recovered, this is weird.
> 2. Do we really need to support recover STOPPED job? @liuyu85cn PTAL

The stop job can be recovered. Just need some conditions to limit whether it can be recoveredd.

src/meta/processors/job/JobManager.cpp
critical27: 👌 , the comment in code is not exactly same with the comment here, which confuses me ...

panda-sheep: yeah, my fault, I have updated the comment

critical27:
Could 

> > So complicated ... I'm confusing ... Several question, majorly related to RECOVER:
> > 
> > 1. If we could recover multiple job, one failed, one recovered, this is weird.
> > 2. Do we really need to support recover STOPPED job? @liuyu85cn PTAL
> 
> 1. Multiple jobs can be recovered at one time. For multiple jobs of the same type, it is only necessary to recover the job with the latest starttime as much as possible.
> 2. The stop job can be recovered. Just need some conditions to limit whether it can be recovered.

Yeah, but there is recovering one of the jobs actually failed, this does not make sense. I mean we could restrict only recover one job is allowed later if necessary, we could simplify the code, and user won't be confused.

BTW, if we are recovering one job, and recover failed, what message do we get from console? "Recovered job failed"?

critical27 (approved):

panda-sheep:
> 



> > > So complicated ... I'm confusing ... Several question, majorly related to RECOVER:
> > > 
> > > 1. If we could recover multiple job, one failed, one recovered, this is weird.
> > > 2. Do we really need to support recover STOPPED job? @liuyu85cn PTAL
> > 
> > 
> > 
> > 1. Multiple jobs can be recovered at one time. For multiple jobs of the same type, it is only necessary to recover the job with the latest starttime as much as possible.
> > 2. The stop job can be recovered. Just need some conditions to limit whether it can be recovered.
> 
> Yeah, some of the recovered jobs actually failed, this does not make sense. I mean we could restrict only recover one job is allowed later if necessary, we could simplify the code, and user won't be confused.
> 
> BTW, if we are recovering one job, and recover failed, what message do we get from console? "Recovered job failed"?

1) Yes, we can support it later. Only one jobId can be recovered at a time. This makes the error clearer.
2) At present, recover jobs only return how many job can be recovered. 
     When recover multiple jobids at a time, a certain jobId cannot be executed and no error is reported.

liwenhui-soul (approved):

PR #4105, opened 2022-03-31
Trigger release workflow manually

### LINKED ISSUE #3815: Optimize branch name and pull method
1. The branch name is the same as the tag name now, we should prefix the branch name with "release-". Doc and workflow, etc. also need to be modified.
2. The vX.Y version should be in a branch, and the bug fix version is identified by tag.

yixinglu:
## What type of PR is this?
- [ ] bug
- [ ] feature
- [X] enhancement

#### Issue(s) number: 

close #3815 

#### Description:

when releasing a nebula PATCH version(z version of x.y.z), create git tag on one same branch rather than a new branch, such as:

```text
release-3.2   -----+----------+------- branch
                    \          \
                    v3.2.0    v3.2.1   tag
```

## How do you solve it?

use github action manual trigger to package rc version.

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [X] N/A

Affects:

## Release notes:
None

> ex. Fixed the bug .....
None

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -5,7 +5,7 @@ on:
     types: [synchronize, reopened, labeled]
     branches:
       - master
-      - 'v[0-9]+.*'
+      - 'release-**'
 
 concurrency:
   group: ${{ github.head_ref }}
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -185,8 +185,7 @@ jobs:
       - name: coverage
         if: ${{ matrix.compiler == 'gcc-9.3' && matrix.os == 'ubuntu2004' }}
         run: |
-            pip3 install fastcov
-            fastcov -d build -l -o fastcov.info -p --exclude /usr/include --exclude=/opt/vesoft --exclude scanner.lex
+            ~/.local/bin/fastcov -d build -l -o fastcov.info -p --exclude /usr/include --exclude=/opt/vesoft --exclude scanner.lex
       - uses: codecov/codecov-action@v2
         if: ${{ matrix.compiler == 'gcc-9.3' && matrix.os == 'ubuntu2004' }}
         with:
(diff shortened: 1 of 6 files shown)


Sophie-Xie:
Thanks!

And I have some questions, if I enter the version number, is it v3.0.2 or 3.0.2?  And Is the name of github tag still v3.0.2?

yixinglu:
> Thanks!
> 
> And I have some questions, if I enter the version number, is it v3.0.2 or 3.0.2? And Is the name of github tag still v3.0.2?

Input version of the workflow is following format: `3.0.2`, there's an example in this parameter description. The tag name should be same as before, such as: `v3.0.2`.

.gitignore
```diff
@@ -38,6 +38,7 @@ modules/
 
 # tests
 !tests/Makefile
+tests/secrets
```
Aiee, line 41: Ent version require this directory

.github/workflows/rc.yml
```diff
@@ -1,9 +1,12 @@
...
 on:
-  push:
-    branches:
-      - 'v[0-9]+.*'
+  workflow_dispatch:
+    inputs:
+      version:
+        description: "version such as '3.1.0'"
+        required: true
+        default: '3.1.0'
```
jievince, line 9: hardcode 3.1.0?

Aiee (dismissed):

.github/workflows/rc.yml
HarrisChu: prefer to do not use default, if not provide version, it would be errors

yixinglu: make sense!

.github/workflows/pull_request.yml
```diff
@@ -5,7 +5,7 @@ on:
     types: [synchronize, reopened, labeled]
     branches:
       - master
-      - 'v[0-9]+.*'
+      - 'release-**'
```
HarrisChu, line 8: typo?

should match release-3.1

HarrisChu (approved):

PR #4106, opened 2022-03-31
Fix optimizer symbols.

### LINKED ISSUE #3826: Optimizer now only release symbols of top layer nodes in all matched nodes.

Shylock-Hg:
## What type of PR is this?
- [x] bug
- [ ] feature
- [ ] enhancement

#### Issue(s) number: 
Close #3826 

#### Description:
Add two constraints to optimizer:
1. One OptGroup will share same output variable
2. There is no overlap output variable between OptGroup

And add one implicit transform:
1. The transformed result of optimizer rule will rebuild input relationship by dependencies.

And some improvement:
1. PlanNode clone will generate new output variable and use it
2. PlanNode only need one output variable so change it from vector to scalar.

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....

### DIFF
--- a/src/graph/optimizer/OptGroup.cpp
+++ b/src/graph/optimizer/OptGroup.cpp
@@ -9,6 +9,7 @@
 
 #include "graph/context/QueryContext.h"
 #include "graph/optimizer/OptContext.h"
+#include "graph/optimizer/OptGroup.h"
 #include "graph/optimizer/OptRule.h"
 #include "graph/planner/plan/Logic.h"
 #include "graph/planner/plan/PlanNode.h"
@@ -44,11 +45,21 @@ OptGroup::OptGroup(OptContext *ctx) noexcept : ctx_(ctx) {
 void OptGroup::addGroupNode(OptGroupNode *groupNode) {
   DCHECK(groupNode != nullptr);
   DCHECK(groupNode->group() == this);
+  if (outputVar_.empty()) {
+    outputVar_ = groupNode->node()->outputVar();
+  } else {
+    DCHECK_EQ(outputVar_, groupNode->node()->outputVar());
+  }
   groupNodes_.emplace_back(groupNode);
   groupNode->node()->updateSymbols();
 }
 
 OptGroupNode *OptGroup::makeGroupNode(PlanNode *node) {
+  if (outputVar_.empty()) {
+    outputVar_ = node->outputVar();
+  } else {
+    DCHECK_EQ(outputVar_, node->outputVar());
+  }
   groupNodes_.emplace_back(OptGroupNode::create(ctx_, node, this));
   return groupNodes_.back();
 }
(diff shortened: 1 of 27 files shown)


src/graph/optimizer/OptGroup.cpp
```diff
@@ -210,5 +225,28 @@ const PlanNode *OptGroupNode::getPlan() const {
...
+  if (std::find(boundary.begin(), boundary.end(), group_) != boundary.end()) {
+    return;
+  }
+  // Rebuild input
+  if (node_->inputVars().size() == dependencies_.size()) {
+    // Can't build input for some nodes, maybe need redesign it and corresponding rule.
+    for (std::size_t i = 0; i < dependencies_.size(); i++) {
+      const OptGroup *dep = dependencies_[i];
+      node_->setInputVar(dep->outputVar(), i);
+      for (auto *groupNode : dep->groupNodes()) {
```
yixinglu, line 244: I think we need not to rebuild all `OptGroupNodes` in the depended `OptGroup` since there are many unchanged nodes which are not in the path of the OptRule. It will save much time to rebuild the nodes of transformed by the OptRule. WDYT?

src/graph/optimizer/OptRule.cpp
```diff
@@ -31,6 +31,13 @@ const PlanNode *MatchedResult::planNode(const std::vector<int32_t> &pos) const {
   return DCHECK_NOTNULL(result->node)->node();
 }
 
+void MatchedResult::collectBoundary(std::vector<OptGroup *> &boundary) const {
```
yixinglu, line 34: It's better to collect all `OptGroupNode` rather than `OptGroup` in the path of matched result.

src/graph/optimizer/OptRule.h
```diff
@@ -30,6 +30,8 @@ class OptGroup;
 struct MatchedResult {
   const OptGroupNode *node{nullptr};
   std::vector<MatchedResult> dependencies;
+  // Boundary of the whole matched result, aka the dependencies of leaf node in MatchedResult.
+  std::vector<OptGroup *> boundary_;
```
yixinglu, line 34: It looks like there is no need to save this variable.

yixinglu (commented):
Compared to help the OptRule author to rebuild the dependency relationships, I prefer to check them in optimizer and report a big warning for them if they don't the right thing. This will help author know where the error is.

src/graph/optimizer/OptGroup.cpp
Shylock-Hg: We could let users do it in rule, and only check it here. As you said before.

src/graph/optimizer/OptRule.cpp
Shylock-Hg: But the `dependencies_` is `OptGroup`.

src/graph/optimizer/OptRule.h
Shylock-Hg: Yes, we could collect it in `match` method directly.

Shylock-Hg:
Replaced by https://github.com/vesoft-inc/nebula/pull/4146

PR #4107, opened 2022-04-01
fix apply outdate membership change

critical27:
## What type of PR is this?
- [X] bug
- [ ] feature
- [ ] enhancement

#### Description:

Introduced in #3846, my bad. The key point is that, whenever we start-up a partition, we use the peers in meta, which is a snapshot of state machine by applying all previous membership changes. So after we start-up, we should not apply any membership change before we started.

Here is a example:
1. We have done lots of membership change, so there are plenty logs like (removePeer, addLearner, addPeer, etc)
2. We start a new part, with 5 peers (from meta) besides it self
3. If we apply the membership change which happens before we start-up (those in step 1), our peers maybe total wrong. For example, after applying them, we have only 2 peers...
4. At this point, anything weird could happen...

However, this is not a silver bullet, the membership change need to be re-considered totally after this version. 

BTW:
Fix crash when trying to transfer leader to a empty host ("":0)

## How do you solve it?
Do not apply any membership change before a part started.

## Checklist:
Tests:
- [X] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

Introduced in the same version, not related


liwenhui-soul (approved):

pengweisong (approved):
LGTM

liwenhui-soul (approved):

PR #4110, opened 2022-04-01
disable console pkg

czpmango:
## What type of PR is this?
- [ ] bug
- [ ] feature
- [ ] enhancement

#### Description:
Temporarily disable packaging of the nebula-console.
The normal packaging process must artificially ensure that both nebula and nebula-console have available and consistent branches before packaging, but due to the dependencies between nebula and nebula-console, some errors will reduce the efficiency of ci.

## How do you solve it?
Just disable the packaging switch related with nebula-console.

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -35,7 +35,6 @@ target/
 cluster.id
 pids/
 modules/
-.linters/
 
 # tests
 !tests/Makefile
--- a/.gitignore
+++ b/.gitignore
@@ -35,6 +35,7 @@ target/
 cluster.id
 pids/
 modules/
+.linters/
 
 # tests
 !tests/Makefile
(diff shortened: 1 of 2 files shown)


Aiee (approved):

.gitignore
```diff
@@ -35,6 +35,7 @@ target/
 cluster.id
 pids/
 modules/
+.linters/
```
yixinglu, line 38: Why ignore this linters?

czpmango: Fixed.

Sophie-Xie (approved):

jievince (approved):

Issue #4097, opened 2022-03-30
where statement causes storage process to exit


rongjianxing:
**where语句导致storage进程退出**

**wsl ubuntu 20.04.4 lts**

* OS: `Linux DESKTOP-A2N2CP3 4.4.0-19041-Microsoft #488-Microsoft Mon Sep 01 13:43:00 PST 2020 x86_64 x86_64 x86_64 GNU/Linux`

**
// linux console
```
wget https://oss-cdn.nebula-graph.com.cn/package/3.0.1/nebula-graph-3.0.1.ubuntu2004.amd64.deb
sudo dpkg -i nebula-graph-3.0.1.ubuntu2004.amd64.deb
sudo /usr/local/nebula/scripts/nebula.service start all
./nebula-console -addr localhost -port 9669 -u root -p 123456
```
// nubula-console
```sql
ADD HOSTS 127.0.0.1:9779;

CREATE SPACE test (partition_num=1, replica_factor=1, vid_type=fixed_string(30));
use test;

create tag t_date_eq( \
    sts_date date,    \
    cost double,     \
);

create tag index idx_t_date_eq_sts_date on t_date_eq(sts_date);
create tag index idx_t_date_eq__cost on t_date_eq(cost);

lookup on t_date_eq                     \
    where t_date_eq.sts_date == date('2022-03-25') \
        and t_date_eq.cost > 0  \
    yield id(vertex) as vertex_id, properties(vertex).sts_date as sts_date;
```
**

rongjianxing:
可以给个回复吗。
想知道是什么原因导致的，是wsl的原因？还是单个storage的原因？还是nebula本身bug导致该语句无论如何都会有问题？

wey-gu:
Thanks @rongjianxing !

I can confirm(just now with your reproduce procedure, thanks!) this query crashed stroaged.

It should be a bug, and we will look into it for the fix.

BR//Wey

Aiee:
Hi @rongjianxing, 
Thank you for your feedback. It is caused by a bug in the constant folding of `date('2022-03-25')`. You may replace `date('2022-03-25')` with a string representation as a workaround for now, as we are fixing it. 

I'll link the PR to this issue when it is available.

PR #4111, opened 2022-04-01
Fix service crash caused by using function call as a part of the filter in `LOOKUP`

Aiee:
## What type of PR is this?
- [x] bug
- [ ] feature
- [ ] enhancement

#### Issue(s) number: 
Close https://github.com/vesoft-inc/nebula/issues/4097
#### Description:
There is a bug in the constant folding of function call expressions like `date('2022-03-25')` when checking purity.

## How do you solve it?
Fix purity checking by bonding purity to the number of arity of the function

## Checklist:
Tests:
- [x] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....

### DIFF
--- a/src/common/function/FunctionManager.cpp
+++ b/src/common/function/FunctionManager.cpp
@@ -456,7 +456,7 @@ FunctionManager::FunctionManager() {
     auto &attr = functions_["abs"];
     attr.minArity_ = 1;
     attr.maxArity_ = 1;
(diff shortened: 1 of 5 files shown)


jievince (approved):

CPWstatic (approved):
Excellent!

yixinglu (approved):
LGTM

PR #4124, opened 2022-04-08
change default variables in release

HarrisChu:
## What type of PR is this?
- [ ] bug
- [ ] feature
- [x] enhancement

#### Description:
1. there're too many variables when building release package, and when we build nebula in test platform, we should follow these options.
2. just use `package.sh` directly when building release.


yixinglu (approved):
LGTM

czpmango (approved):

PR #4129, opened 2022-04-11
default add hosts for standalone version

SuperYoko:
## What type of PR is this?
- [ ] bug
- [ ] feature
- [x] enhancement

#### Description:
Standalone version will default auto add local storage hosts now. use "--add_local_host" to control this.

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....


src/storage/StorageServer.cpp
```diff
@@ -168,6 +170,28 @@ bool StorageServer::start() {
   options.dataPaths_ = dataPaths_;
 
   metaClient_ = std::make_unique<meta::MetaClient>(ioThreadPool_, metaAddrs_, options);
+
+#ifdef BUILD_STANDALONE
+  if (FLAGS_add_local_host) {
```
critical27, line 175: 1. Does meta always started before storage in standalone?
2. Do we need to move `addHosts` after `waitForMetadReady`?

tests/common/nebula_service.py
```diff
@@ -321,6 +321,7 @@ def _make_sa_params(self, **kwargs):
         self.graphd_param['password_lock_time_in_secs'] = '10'
         self.graphd_param['raft_heartbeat_interval_secs'] = '30'
         self.graphd_param['skip_wait_in_rate_limiter'] = 'true'
+        self.graphd_param['add_local_host'] = 'false'
```
critical27, line 324: UT has already call add hosts explicitly?

src/storage/StorageServer.cpp
SuperYoko: 1. yes, standalone daemon will wait meta service start then start storage.
2. we can't if we don't addhosts first,  wait will stuck there.

tests/common/nebula_service.py
SuperYoko: Ent version hardcoded "add hosts to new zone default_zone", and in nebula repo "add hosts",  which will affect tck.  So I let "add hosts" done by tck start cluster scripts use "--add_local_host".

src/storage/StorageServer.cpp
critical27: 👌

critical27 (approved):

src/storage/StorageServer.cpp
```diff
@@ -42,6 +42,8 @@ DEFINE_bool(local_config, false, "meta client will not retrieve latest configura
 DEFINE_int32(storage_port, 44501, "Storage daemon listening port");
 DEFINE_int32(storage_num_worker_threads, 32, "Number of workers");
 DECLARE_bool(local_config);
+DEFINE_bool(add_local_host, true, "Whether add localhost automatically");
```
pengweisong, line 45: should we tell the user "in standalone version"?

SuperYoko: Only standalone version supply this config, and only standalone config file contain it.

pengweisong (approved):
LGTM

Issue #4123, opened 2022-04-08
Disk fault recovery


Sophie-Xie:
**Describe the background of the requirement**
- Considering that a node has multiple data paths, there is no need to rebuild the entire node when the disk corresponding to a data path fails.

**Describe the detail of the requirement**
- After the disk corresponding to the data path is repaired, the original part data can be automatically restored to the data path.

**Describe the scenarios of the requirement**
- Disk failback, not node failback

**Describe the priority**
- Urgency: hi
- Importance: hi

critical27:
Try to fix this, for example, a storage start with 2 data paths, say p1, p2.
1. create a space, so both p1 and p2 has some parts
2. stop storage, remove the data on p1
3. restart storage, ideally the deleted parts would start on p1, but based on current implementation, I guess the part will only put to p2.

See `NebulaStore::loadPartFromPartManager()` for details, we won't rebuild the kvengine in p1, so the partitions removed in step2 will only be added to p2

![image](https://user-images.githubusercontent.com/13706157/162412533-43d80309-faeb-40af-88a0-de8053f12cdd.png)

wenhaocs:
How about disk balance feature?

Sophie-Xie:
> 

Maybe in 5.x.

Sophie-Xie:
> How about disk balance feature?


Maybe in 5.x.

PR #4131, opened 2022-04-11
fix disk fault recovery

panda-sheep:
## What type of PR is this?
- [ ] bug
- [x] feature
- [ ] enhancement

#### Description:
fix  https://github.com/vesoft-inc/nebula/issues/4123

When a storage configure contains multiple dataPaths, this pr handles adding dataPaths or removing some dataPaths.

For example: 
dataPath in storage configure contains 2 dataPaths, namely dataPath1 and dataPath2, and each dataPath contains 50 parts.
At this time, dataPath3 was newly added.
                     
case1: The disk of dataPath1 is broken, but there is no repair. 
            Remove dataPath1 from dataPath in storage configure.
case2: The disk of dataPath1 is broken, but repaired.

At this time, the storage is started, and the distribution of the final part is as follows:
image

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....


panda-sheep:
Has tested, ready to review

src/kvstore/NebulaStore.cpp
```diff
@@ -359,14 +359,31 @@ ErrorOr<nebula::cpp2::ErrorCode, HostAddr> NebulaStore::partLeader(GraphSpaceID
...
+    // Iterate over all engines to ensure that each dataPath has an engine
     if (this->spaces_.find(spaceId) != this->spaces_.end()) {
       LOG(INFO) << "Data space " << spaceId << " has existed!";
-      return;
-    }
-    LOG(INFO) << "Create data space " << spaceId;
-    this->spaces_[spaceId] = std::make_unique<SpacePartInfo>();
-    for (auto& path : options_.dataPaths_) {
-      this->spaces_[spaceId]->engines_.emplace_back(newEngine(spaceId, path, options_.walPath_));
+      for (auto& path : options_.dataPaths_) {
```
critical27, line 365: Why not build the engine in `loadPartFromPartManager`, `addSpace` is called for each partition in meta, which is duplicated.

critical27: Would you add a test case if possible.

panda-sheep: First, `addSpace` is processed once per spaceId
Secondly, the `addSpace` interface is to create an engine for each dataPath of the current host, check and create an engine according to the dataPath and put it in addSpace to feel more unified.
wdyt?

critical27: I got it, my bad.

src/kvstore/NebulaStore.cpp
```diff
@@ -359,14 +364,31 @@ ErrorOr<nebula::cpp2::ErrorCode, HostAddr> NebulaStore::partLeader(GraphSpaceID
 void NebulaStore::addSpace(GraphSpaceID spaceId, bool isListener) {
   folly::RWSpinLock::WriteHolder wh(&lock_);
   if (!isListener) {
+    // Iterate over all engines to ensure that each dataPath has an engine
```
wenhaocs, line 367: It's better to make it clear in the comments that disk paths may not have kv engines created after recovering from repair.

wenhaocs (approved):
LGTM. Per discussion, it's better to make it clear that disks are empty after repair.

critical27 (approved):
Good job~

Issue #3778, opened 2022-01-21
When parser encounters some special property names, it will report an error


cangfengzhs:
https://discuss.nebula-graph.com.cn/t/topic/7162
```
(root@nebula) [test]> create EDGE edge_type1(e string, e1 string,p string, p1 string);
Execution succeeded (time spent 3092/3706 us)
                                                  
Fri, 21 Jan 2022 15:52:27 CST

(root@nebula) [test]> go from 1 over edge_type1 where edge_type1.e=="1" yield edge as e;
+---+
| e |
+---+
+---+
Empty set (time spent 5529/6241 us)

Fri, 21 Jan 2022 15:53:01 CST

(root@nebula) [test]> go from 1 over edge_type1 where edge_type1.e1=="1" yield edge as e;
[ERROR (-1004)]: SyntaxError: Out of range: near `.e1'

Fri, 21 Jan 2022 15:53:05 CST

(root@nebula) [test]> go from 1 over edge_type1 where edge_type1.p=="1" yield edge as e;
+---+
| e |
+---+
+---+
Empty set (time spent 5335/5965 us)

Fri, 21 Jan 2022 15:53:12 CST

(root@nebula) [test]> go from 1 over edge_type1 where edge_type1.p1=="1" yield edge as e;
+---+
| e |
+---+
+---+
Empty set (time spent 5281/6015 us)

```

CPWstatic:
It should be a matter of scientific notation.

PR #4136, opened 2022-04-12
fix the wrong regex pattern of scientific notation

jievince:
## What type of PR is this?
- [x] bug
- [ ] feature
- [ ] enhancement

#### Issue(s) number: 
Fix https://github.com/vesoft-inc/nebula/issues/3778
#### Description:
`.e1` is not a valid format of scientific notation.

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....


src/parser/scanner.lex
```diff
@@ -404,7 +404,8 @@ LABEL_FULL_WIDTH            {CN_EN_FULL_WIDTH}{CN_EN_NUM_FULL_WIDTH}*
 
 {DEC}*\.{DEC}+              |
 {DEC}+\.{DEC}*              |
-{DEC}*\.{DEC}*{EXP}         |
+{DEC}*\.{DEC}+{EXP}         |
```
Shylock-Hg, line 407: I think you could add `{EXP}?` directly.

jievince: nice!

Shylock-Hg (approved):

yixinglu (approved):

PR #4146, opened 2022-04-13
Refactor the process of symbols in optimizer.

### LINKED ISSUE #3826: Optimizer now only release symbols of top layer nodes in all matched nodes.

Shylock-Hg:
## What type of PR is this?
- [x] bug
- [ ] feature
- [ ] enhancement

#### Issue(s) number: 
Close #3826

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....

### DIFF
--- a/src/graph/optimizer/OptGroup.cpp
+++ b/src/graph/optimizer/OptGroup.cpp
@@ -44,11 +44,21 @@ OptGroup::OptGroup(OptContext *ctx) noexcept : ctx_(ctx) {
 void OptGroup::addGroupNode(OptGroupNode *groupNode) {
   DCHECK(groupNode != nullptr);
   DCHECK(groupNode->group() == this);
+  if (outputVar_.empty()) {
+    outputVar_ = groupNode->node()->outputVar();
+  } else {
+    DCHECK_EQ(outputVar_, groupNode->node()->outputVar());
+  }
   groupNodes_.emplace_back(groupNode);
   groupNode->node()->updateSymbols();
 }
 
 OptGroupNode *OptGroup::makeGroupNode(PlanNode *node) {
+  if (outputVar_.empty()) {
+    outputVar_ = node->outputVar();
+  } else {
+    DCHECK_EQ(outputVar_, node->outputVar());
+  }
   groupNodes_.emplace_back(OptGroupNode::create(ctx_, node, this));
   return groupNodes_.back();
 }
@@ -70,7 +80,8 @@ Status OptGroup::explore(const OptRule *rule) {
     NG_RETURN_IF_ERROR(groupNode->explore(rule));
 
     // Find more equivalents
-    auto status = rule->match(ctx_, groupNode);
+    std::vector<OptGroup *> boundary;
+    auto status = rule->match(ctx_, groupNode, boundary);
     if (!status.ok()) {
       ++iter;
       continue;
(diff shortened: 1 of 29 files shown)


src/graph/optimizer/OptGroup.cpp
```diff
@@ -70,7 +80,8 @@ Status OptGroup::explore(const OptRule *rule) {
     NG_RETURN_IF_ERROR(groupNode->explore(rule));
 
     // Find more equivalents
-    auto status = rule->match(ctx_, groupNode);
+    std::vector<OptGroup *> boundary;
+    auto status = rule->match(ctx_, groupNode, boundary);
```
yixinglu, line 84: Need not to collect the boundary of this rule. you could use the last MatchedResult to get the wanted OptGroup.

src/graph/optimizer/OptRule.cpp
```diff
@@ -59,26 +62,68 @@ StatusOr<MatchedResult> Pattern::match(const OptGroupNode *groupNode) const {
...
+  }
+  // Check dataflow
+  const auto *node = groupNode->node();
+  if (node->inputVars().size() == deps.size()) {
+    // Don't check when count of dependencies is different from count of input variables
+    for (std::size_t i = 0; i < deps.size(); i++) {
+      const OptGroup *dep = deps[i];
+      if (node->inputVar(i) != dep->outputVar()) {
+        return false;
+      }
```
yixinglu, line 111: Maybe we could do more things than this input variable validation, such as check whether these new generated OptGroupNodes reference other nodes outside since we guarantee that all plan nodes of the new subplan couldn't access other plan nodes except the intput/output nodes.

yixinglu (commented):
Good Job! Could this PR solve the optimization issue about match after the rules bug fixed?
if so, we could consider to cherry pick it into v3.1.0.

src/graph/optimizer/OptGroup.cpp
Shylock-Hg: Fixed.

src/graph/optimizer/OptRule.cpp
Shylock-Hg: I this has been done in current implementation. All input variables of plan nodes in current transform result had been checed.

Shylock-Hg:
> Good Job! Could this PR solve the optimization issue about match after the rules bug fixed? if so, we could consider to cherry pick it into v3.1.0.

Could resolve the case mentioned in #3826

src/graph/optimizer/OptRule.cpp
Shylock-Hg: Fixed.

src/graph/optimizer/OptRule.cpp
```diff
@@ -76,6 +86,48 @@ StatusOr<MatchedResult> Pattern::match(const OptGroup *group) const {
...
+      // Only use by father plan node
+      if (node->inputVars()[i]->readBy.size() != 1) {
```
yixinglu, line 119: it's better to check whether the input variables are from outside plan. But it's ok to process like this now.

yixinglu (approved):
LGTM

czpmango (approved):

Issue #4154, opened 2022-04-14
Optional match in master(2022.04.13-nightly) became non-optional logcially


wey-gu:
**Please check the FAQ documentation before raising an issue**

**Describe the bug (__required__)**

Today in the master docker build, I found the query of optional match is wrong.

> `2022.04.13-nightly` ad8d661 ❌

```
(root@nebula) [basketballplayer]> match (n:player) WHERE id(n) == "player100" OPTIONAL MATCH (n)<-[:serve]-(m) RETURN n, m
+---+---+
| n | m |
+---+---+
+---+---+
(root@nebula) [basketballplayer]> match (n:player) WHERE id(n) == "player100" RETURN n
+----------------------------------------------------+
| n                                                  |
+----------------------------------------------------+
| ("player100" :player{age: 42, name: "Tim Duncan"}) |
+----------------------------------------------------+
Got 1 rows (time spent 2469/6476 us)
```
> `2022.03.30-nightly` 1e68162 ✅
```
(root@nebula) [basketballplayer]> match (n) WHERE id(n) == "player100" OPTIONAL MATCH (n)<-[:serve]-(m) RETURN n, m
+----------------------------------------------------+----------+
| n                                                  | m        |
+----------------------------------------------------+----------+
| ("player100" :player{age: 42, name: "Tim Duncan"}) | __NULL__ |
+----------------------------------------------------+----------+
Got 1 rows (time spent 6296/50222 us)
```

**Your Environments (__required__)**

* OS: `uname -a`
* Compiler: `g++ --version` or `clang++ --version`
* CPU: `lscpu`
* Commit id: ad8d661 aka 2022.04.13-nightly

**How To Reproduce(__required__)**

`match (n:player) WHERE id(n) == "player100" OPTIONAL MATCH (n)<-[:serve]-(m) RETURN n, m`

**Expected behavior**

optional match shouldn't be mandatory, thus n should be returned and m to be empty.

**Additional context**

```
(root@nebula) [basketballplayer]> explain match (n:player) WHERE id(n) == "player100" OPTIONAL MATCH (n)<-[:serve]-(m) RETURN n, m
Execution succeeded (time spent 1259/15026 us)

Execution Plan (optimize time 465 us)

-----+----------------+--------------+----------------+------------------------------------------
| id | name           | dependencies | profiling data | operator info                           |
-----+----------------+--------------+----------------+------------------------------------------
| 13 | Project        | 12           |                | outputVar: {                            |
[...]

PR #4159, opened 2022-04-14
Fix optional of multi-match

czpmango:
## What type of PR is this?
- [x] bug
- [ ] feature
- [ ] enhancement

#### Issue(s) number: 
https://github.com/vesoft-inc/nebula/issues/4154

## How do you solve it?
The calculation of `Filter` before and after `InnerJoin/Cartesian` are equivalent *if the Filter can be delay*.

The calculation of `Filter` before and after `LeftJoin` are not equivalent, bcz `LeftJoin` always keep the left table rows.
So the correct solution is to introduce all visible variables via `Argument` and computes the `Filter` before `LeftJoin`.

This pr currently only fixes the case where `Filter` not referencing variables outside the definitions in the current match pattern.

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....

### DIFF
--- a/src/graph/planner/match/MatchPlanner.cpp
+++ b/src/graph/planner/match/MatchPlanner.cpp
@@ -83,11 +83,19 @@ Status MatchPlanner::connectMatchPlan(SubPlan& queryPlan, MatchClauseContext* ma
   }
   if (!intersectedAliases.empty()) {
     if (matchCtx->isOptional) {
+      // connect LeftJoin match filter
+      if (matchCtx->where != nullptr) {
+        auto wherePlanStatus =
+            std::make_unique<WhereClausePlanner>()->transform(matchCtx->where.get());
+        NG_RETURN_IF_ERROR(wherePlanStatus);
+        auto wherePlan = std::move(wherePlanStatus).value();
+        matchPlan = SegmentsConnector::addInput(wherePlan, matchPlan, true);
+      }
       queryPlan =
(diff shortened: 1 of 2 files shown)


jievince (approved):

src/graph/planner/match/MatchPlanner.cpp
```diff
@@ -83,11 +83,19 @@ Status MatchPlanner::connectMatchPlan(SubPlan& queryPlan, MatchClauseContext* ma
   }
   if (!intersectedAliases.empty()) {
     if (matchCtx->isOptional) {
+      // connect LeftJoin match filter
+      if (matchCtx->where != nullptr) {
```
Shylock-Hg, line 87: What will happen when filter reference variable outside current match cluase if you don't delay the filter

czpmango:
```
MATCH (a)--(b)--(c) 
OPTIONAL MATCH (a)--(b)--(d) WHERE d.prop>c.prop
RETURN count(*) AS count
```
This statement is currently not fixed, `count` will return 0.

Shylock-Hg: If this hotfix introduces a new bug , it's not necessary to merge. WDYT? @CPWstatic

CPWstatic: Do you mean the previous hotfix or this one?

CPWstatic (approved):

PR #4182, opened 2022-04-19
fix wrong space key after dropping hosts

liwenhui-soul:
## What type of PR is this?
- [x] bug
- [ ] feature
- [ ] enhancement

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....

### DIFF
--- a/src/meta/processors/zone/DropHostsProcessor.cpp
+++ b/src/meta/processors/zone/DropHostsProcessor.cpp
@@ -90,21 +90,13 @@ void DropHostsProcessor::process(const cpp2::DropHostsReq& req) {
           return std::find(hosts.begin(), hosts.end(), h) != hosts.end();
         })) {
       LOG(INFO) << "Drop zone " << zoneName;
-      code = checkRelatedSpaceAndCollect(zoneName, holder.get());
+      code = checkRelatedSpaceAndCollect(zoneName, &spaceMap);
       if (code != nebula::cpp2::ErrorCode::SUCCEEDED) {
         LOG(INFO) << "Check related space failed";
         break;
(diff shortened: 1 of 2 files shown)


critical27:
Do we support drop several hosts at the same time?

Suppose we have a space `s1` with only one replica on [z1, z2, z3], each zone has one host, which is [h1, h2, h3]. If we execute `DROP HOST h2, h3`, base on current implementation, `s1`'s zone list would be either `[z1, z2] or `[z1, z3]`. The reason is that `checkRelatedSpaceAndCollect` is modify the zone list based on data in rocksdb....

That's why the old data has the `spaceMap`... Correct me if I am wrong.... 

And we'd better add a test case in TCK later

HarrisChu:
> Do we support drop several hosts at the same time?

yes, support

liwenhui-soul:
> Do we support drop several hosts at the same time?
> 
> Suppose we have a space `s1` with only one replica on [z1, z2, z3], each zone has one host, which is [h1, h2, h3]. If we execute `DROP HOST h2, h3`, base on current implementation, `s1`'s zone list would be either `[z1, z2]` or `[z1, z3]`, not `[z1]`. The reason is that `checkRelatedSpaceAndCollect` is modify the zone list based on data in rocksdb....
> 
> That's why we need `spaceMap`. Correct me if I am wrong....
> 
> And we'd better add a test case in TCK later

fixed

critical27 (approved):
LGTM

liuyu85cn (approved):

Issue #4184, opened 2022-04-20
match v limit 20000 may crash the storaged


xiajingchun:
**Please check the FAQ documentation before raising an issue**

Run below query on SF300 or SF1000 dataset will crash all the storage services in the cluster.
`match (v) return v limit 20000`

The coredump is as below:
![match_coredump](https://user-images.githubusercontent.com/6269380/164144253-2c767e6b-0813-42b7-8634-67ec26597cab.png)

**Your Environments (__required__)**

* OS: CentOS 8
* Compiler: `g++ --version` or `clang++ --version`
* CPU: `lscpu`
* Commit id 02bdc67 (2022.4.19-nightly)

**How To Reproduce(__required__)**

Just run the query and wait.

**Expected behavior**

**Additional context**

PR #4190, opened 2022-04-20
Expression is stateful to store the result of evaluation, so we can't…

Shylock-Hg:
… share it inter threads.

## What type of PR is this?
- [x] bug
- [ ] feature
- [ ] enhancement

#### Issue(s) number: 

fixed https://github.com/vesoft-inc/nebula/issues/4184

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....

### DIFF
--- a/src/clients/meta/MetaClient.cpp
+++ b/src/clients/meta/MetaClient.cpp
@@ -170,10 +170,14 @@ void MetaClient::heartBeatThreadFunc() {
     bgThread_->addDelayTask(
         FLAGS_heartbeat_interval_secs * 1000, &MetaClient::heartBeatThreadFunc, this);
   };
-  auto ret = heartbeat().get();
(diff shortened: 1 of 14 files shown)


nevermore3:
This bug seems to have nothing to do with the data set, why did CI not report this problem

Shylock-Hg:
> This bug seems to have nothing to do with the data set, why did CI not report this problem

Dataset won't share inter threads, each job has independent dataset to write.

critical27:
> This bug seems to have nothing to do with the data set, why did CI not report this problem

It only will happens when storage execute in multiple thread.

critical27 (approved):

nevermore3 (approved):

PR #3881, opened 2022-02-11
Remove max hop

jackwener:
## What type of PR is this?
- [ ] bug
- [ ] feature
- [x] enhancement

## Checklist:
Tests:
- [ ] Unit test(positive and negative cases)
- [x] Function test
- [ ] Performance test
- [ ] N/A

Affects:
- [x] Documentation affected (Please add the label if documentation needs to be modified.)
- [ ] Incompatibility (If it breaks the compatibility, please describe it and add the label.）
- [ ] If it's needed to cherry-pick (If cherry-pick to some branches is required, please label the destination version(s).)
- [ ] Performance impacted: Consumes more CPU/Memory

## Release notes:

Please confirm whether to be reflected in release notes and how to describe:
> ex. Fixed the bug .....

PR #4200, opened 2022-04-22
check maximum length of variable length relationship

nevermore3:
## What type of PR is this?
- [ ] bug
- [ ] feature
- [x] enhancement

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....


Shylock-Hg:
It's modified in https://github.com/vesoft-inc/nebula/pull/3881

nevermore3:
> #3881

this PR removes the limit of maximum length,  there is a risk of memory OOM

Shylock-Hg:
> > #3881
> 
> this PR removes the limit of maximum length, there is a risk of OOM

We should avoid OOM by memory tacking and kill expensive query.

PR #4210, opened 2022-04-26
Fix/plan compare

Shylock-Hg:
## What type of PR is this?
- [ ] bug
- [ ] feature
- [x] enhancement

#### Description:
Fix the plan comparing in tck. Support to compare the nested dictionary in plan.

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....


jievince (approved):
Well done! Thanks for fixing it.

Aiee (dismissed):

tests/common/plan_differ.py
```diff
@@ -2,8 +2,11 @@
 #
 # This source code is licensed under Apache 2.0 License.
 
+from codecs import ignore_errors
```
Aiee, line 5: Why do we need this?

tests/common/plan_differ.py
```diff
@@ -2,8 +2,11 @@
...
 import re
 import json
+import deepdiff
```
Aiee, line 8: Is this for debugging?

Shylock-Hg: Not, to compare the dict

nevermore3 (approved):

tests/common/plan_differ.py
Shylock-Hg: Already removed

Aiee (approved):

PR #4226, opened 2022-05-05
modify subgraph loop condition

nevermore3:
## What type of PR is this?
- [ ] bug
- [ ] feature
- [x] enhancement

#### Description:

The last step of the subgraph is to get all the edges between the visited vetices, if only the output vertices, do not need to take the last step

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....

### DIFF
--- a/src/graph/planner/ngql/SubgraphPlanner.cpp
+++ b/src/graph/planner/ngql/SubgraphPlanner.cpp
@@ -71,12 +71,16 @@ StatusOr<SubPlan> SubgraphPlanner::nSteps(SubPlan& startVidPlan, const std::stri
   auto resultVar = qctx->vctx()->anonVarGen()->getVar();
   auto loopSteps = qctx->vctx()->anonVarGen()->getVar();
   subgraphCtx_->loopSteps = loopSteps;
-  auto* subgraph = Subgraph::make(qctx, gn, resultVar, loopSteps, steps.steps() + 1);
+
+  auto* subgraph = Subgraph::make(qctx, gn, resultVar, loopSteps, steps.steps());
   subgraph->setOutputVar(input);
   subgraph->setBiDirectEdgeTypes(subgraphCtx_->biDirectEdgeTypes);
   subgraph->setColNames({nebula::kVid});
-
-  auto* condition = loopCondition(steps.steps() + 1, gn->outputVar());
+  uint32_t maxSteps = steps.steps();
+  if (subgraphCtx_->getEdgeProp || subgraphCtx_->withProp) {
+    ++maxSteps;
+  }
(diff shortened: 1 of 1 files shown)


src/graph/planner/ngql/SubgraphPlanner.cpp
```diff
@@ -71,7 +71,11 @@ StatusOr<SubPlan> SubgraphPlanner::nSteps(SubPlan& startVidPlan, const std::stri
   auto resultVar = qctx->vctx()->anonVarGen()->getVar();
   auto loopSteps = qctx->vctx()->anonVarGen()->getVar();
   subgraphCtx_->loopSteps = loopSteps;
-  auto* subgraph = Subgraph::make(qctx, gn, resultVar, loopSteps, steps.steps() + 1);
+  uint32_t maxSteps = steps.steps();
+  if (subgraphCtx_->getEdgeProp || subgraphCtx_->withProp) {
+    ++maxSteps;
+  }
```
czpmango, line 77: Why does taking properties have to do with the max step of loop.

src/graph/planner/ngql/SubgraphPlanner.cpp
```diff
@@ -75,8 +75,11 @@ StatusOr<SubPlan> SubgraphPlanner::nSteps(SubPlan& startVidPlan, const std::stri
   subgraph->setOutputVar(input);
   subgraph->setBiDirectEdgeTypes(subgraphCtx_->biDirectEdgeTypes);
   subgraph->setColNames({nebula::kVid});
-
-  auto* condition = loopCondition(steps.steps() + 1, gn->outputVar());
+  uint32_t maxSteps = steps.steps();
+  if (subgraphCtx_->getEdgeProp || subgraphCtx_->withProp) {
+    ++maxSteps;
+  }
+  auto* condition = loopCondition(maxSteps, gn->outputVar());
```
Shylock-Hg, line 82: Steps of subgraph executor don't keep same with this?

src/graph/planner/ngql/SubgraphPlanner.cpp
nevermore3: getEdgeProp just indicates whether to output the edge。will fix name, avoid ambiguity

src/graph/planner/ngql/SubgraphPlanner.cpp
nevermore3: no effect

Shylock-Hg (approved):

czpmango (approved):

Issue #4230, opened 2022-05-05
lookup would crash storaged where there's an index on a nullable property


xiajingchun:
**Please check the FAQ documentation before raising an issue**

**Describe the bug (__required__)**

When a tag has a nullable bool property with an index on it, run a lookup query without where condition would crash the storaged. See the reproduce steps for more details.

**Your Environments (__required__)**

* OS: Centos 7
* Compiler: `g++ --version` or `clang++ --version`
* CPU: `lscpu`
* Commit id (e.g. `a3ffc7d8`) v3.1.0 - 33fd35e 

**How To Reproduce(__required__)**

Use below schema to create a space.
```
drop space test;
create space test(partition_num=10,replica_factor=1,vid_type=fixed_string(32));
:sleep 20
use test;
create tag v(flag bool NULL);
:sleep 20
create tag index v_index_0 on v(flag);
:sleep 20
insert vertex v(flag) values "001":(true);
insert vertex v() values "002":();
```
The run a lookup query like below:
```
lookup on v yield id(vertex) as vid;
```
Storaged will then crash.

**Expected behavior**

**Additional context**

Core dump is like:
```
(gdb) bt
#0  0x00007ff424a9b387 in raise () from /lib64/libc.so.6
#1  0x00007ff424a9ca78 in abort () from /lib64/libc.so.6
#2  0x00000000025b4dca in google::LogMessage::Fail() ()
#3  0x00000000025b7ec4 in google::LogMessage::SendToLog() ()
#4  0x00000000025b4899 in google::LogMessage::Flush() ()
#5  0x00000000025b8579 in google::LogMessageFatal::~LogMessageFatal() ()
#6  0x000000000150580e in apache::thrift::detail::validate_bool(unsigned char) [clone .part.0] ()
#7  0x000000000152f399 in apache::thrift::CompactProtocolWriter::writeBool(bool) ()
#8  0x000000000156a769 in unsigned int apache::thrift::Cpp2Ops<nebula::Value, void>::write<apache::thrift::CompactProtocolWriter>(apache::thrift::CompactProtocolWriter*, nebula::Value const*) ()
#9  0x000000000156ab6f in unsigned int apache::thrift::Cpp2Ops<nebula::List, void>::write<apache::thrift::CompactProtocolWriter>(apache::thrift::CompactProtocolWriter*, nebula::List const*) ()
[...]

PR #4234, opened 2022-05-06
fix lookup crash when has nullable cols

critical27:
## What type of PR is this?
- [X] bug
- [ ] feature
- [ ] enhancement

#### Issue(s) number: 

#4230 

#### Description:

The main problem has two parts:
1. The `indexNullable_` in `IndexScanNode` is not initialized correctly.
2. In `IndexScanNode`, the offset of nullable bits is not calculated correctly.

Check if there is an nullable property when build plan.

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

Fix storage may crash when lookup index with nullable property.

### DIFF
--- a/src/storage/exec/IndexScanNode.h
+++ b/src/storage/exec/IndexScanNode.h
@@ -60,8 +60,13 @@ class IndexScanNode : public IndexNode {
                 const std::string& name,
                 IndexID indexId,
                 const std::vector<cpp2::IndexColumnHint>& columnHints,
-                ::nebula::kvstore::KVStore* kvstore)
-      : IndexNode(context, name), indexId_(indexId), columnHints_(columnHints), kvstore_(kvstore) {}
+                ::nebula::kvstore::KVStore* kvstore,
+                bool hasNullableCol)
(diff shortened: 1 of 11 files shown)


src/storage/exec/IndexScanNode.h
```diff
@@ -153,6 +154,10 @@ class IndexScanNode : public IndexNode {
    */
   std::unique_ptr<kvstore::KVIterator> iter_;
   nebula::kvstore::KVStore* kvstore_;
+  /**
+   * @brief if index contain nullable field or not
```
SuperYoko, line 158: Why change this line's order, keep align with new added initilize?

SuperYoko (approved):

pengweisong (approved):

src/storage/exec/IndexScanNode.h
critical27: Yeah, just want to keep align the parameter order in constructor and member initialize list.

PR #4242, opened 2022-05-09
Optimize input/var property expression to get property from tuple by

### LINKED ISSUE #4174: Optimize retrieving Input/Variable property by index instead of property name lookup
**Introduction**

We could assume input dataset is immutable, so we could get column of input dataset by index directly.
Subjob of #4122  

ps. If we flatten dataset of GetNeighbors, we could optimize Tag/Edge property by this way too.
**Contents**

**Related work**

Shylock-Hg:
index to avoid hash table lookup.

## What type of PR is this?
- [ ] bug
- [ ] feature
- [x] enhancement

#### Issue(s) number: 
Close #4174

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....


Shylock-Hg:
Origin benchmark:
```
/nebula/src/graph/context/test/InputVarPropBenchmark.cpprelative  time/iter  iters/s
inputVarPropEval(test)                                      32.79ns   30.50M
```

src/common/expression/PropertyExpression.cpp
```diff
@@ -68,15 +68,29 @@ void TagPropertyExpression::accept(ExprVisitor* visitor) {
...
-  return ctx.getInputProp(prop_);
+  if (!propIndex_.has_value()) {
+    auto indexResult = ctx.getInputPropIndex(prop_);
+    if (!indexResult.ok()) {
+      return Value::kEmpty;
+    }
+    propIndex_ = indexResult.value();
+  }
+  return ctx.getColumn(propIndex_.value());
 }
```
jievince, line 79: I remember we have a `ColumnIndexExpr`?  It looks like that rewriteing `InputPropExpr` to `ColumnIndexExpr` is more easy

Shylock-Hg: Good job, we can add an optimize rule to do it. But current optimization is also useful now and no need to rewrite expression.

jievince (approved):

nevermore3 (approved):

PR #4245, opened 2022-05-10
fix listener can't receive snapshot anymore

critical27:
## What type of PR is this?
- [X] bug
- [ ] feature
- [ ] enhancement

#### Description:

During replication test, I found two problems here:
1. Once listener return FAILED when receiving snapshot, the sender won't retry (we only retry for timeout before). This is a minor issue, because usually, listener won't explicitly return error when receiving snapshot (at least listener should not @panda-sheep you need to check your `SyncListener` implementation)
2. What's worse, listener won't be able to receive snapshot. Because the listener is always in `WAITING_SNAPSHOT` state. So we just reuse the same logic in normal `RaftPart`.

## Checklist:
Tests:
- [X] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

Not related

### DIFF
--- a/src/kvstore/Listener.cpp
+++ b/src/kvstore/Listener.cpp
@@ -147,6 +147,9 @@ void Listener::doApply() {
   if (isStopped()) {
     return;
   }
+  if (needToCleanupSnapshot()) {
+    cleanupSnapshot();
+  }
   // todo(doodle): only put is handled, all remove is ignored for now
   folly::via(executor_.get(), [this] {
     SCOPE_EXIT {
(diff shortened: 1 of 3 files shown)


critical27:
@Sophie-Xie perhaps you need to cherry-pick it to the enterprise version.

liwenhui-soul (dismissed):

panda-sheep (approved):
Good

panda-sheep (approved):

liwenhui-soul (approved):

Issue #4241, opened 2022-05-09
More expressions in prop filtering for GO


wey-gu:
**Is your feature request related to a problem? Please describe.**

Expression to do things like starts_with or others in GO.

**Describe the solution you'd like**

- starts with
- wild card(if feasible)

**Describe alternatives you've considered**

**Additional context**

https://discuss.nebula-graph.com.cn/t/topic/8642?u=wey

sworduo:
Like 
`go from "Marco Belinelli", "Luka Doncic" over like where prefix($$.team.name "S") and wildcard($^.player.name, "x?")`
Just add a new expression and rewrite its `eval` is okay.

PR #4248, opened 2022-05-11
Start end contain

nevermore3:
## What type of PR is this?
- [ ] bug
- [ ] feature
- [ ] enhancement

#### Issue(s) number: 
close https://github.com/vesoft-inc/nebula/issues/4241
## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....

### DIFF
--- a/src/common/expression/RelationalExpression.cpp
+++ b/src/common/expression/RelationalExpression.cpp
@@ -113,82 +113,70 @@ const Value& RelationalExpression::eval(ExpressionContext& ctx) {
       break;
     }
     case Kind::kContains: {
-      if (lhs.isBadNull() || rhs.isBadNull()) {
-        result_ = Value::kNullBadType;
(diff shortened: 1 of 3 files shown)


src/common/expression/RelationalExpression.cpp
```diff
@@ -113,82 +113,70 @@ const Value& RelationalExpression::eval(ExpressionContext& ctx) {
       break;
     }
     case Kind::kContains: {
-      if (lhs.isBadNull() || rhs.isBadNull()) {
-        result_ = Value::kNullBadType;
-      } else if ((!lhs.isNull() && !lhs.isStr()) || (!rhs.isNull() && !rhs.isStr())) {
```
czpmango, line 118: Why change this?
```
return "czp " contains 3;
```

Issue #4122, opened 2022-04-08
Improve performance in the LDBC scenarios.


CPWstatic:
**Introduction**
Improve performance in the LDBC scenarios.

**Contents**
There are several measures to improve performance, we will have a doc.

**Related work**
subtask: https://github.com/vesoft-inc/nebula/issues/4137

xtcyclist:
Shall we close this issue?

PR #4249, opened 2022-05-11
Eliminate vid predication Filter.

Shylock-Hg:
## What type of PR is this?
- [ ] bug
- [ ] feature
- [x] enhancement

#### Issue(s) number: 
Sub job of #4122 

#### Description:
GetVertices will retrieve Vertex by Id set directly, so we don't filter them again in Filter operator, so we eliminate this Filter operator.

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....


nevermore3:
if the vertexIDSeek statement is mixed with other property filtering statements, can we delete the vertexID condition after selecting the vertexID index?

Shylock-Hg:
> if the vertexIDSeek statement is mixed with other property filtering statements, can we delete the vertexID condition after selecting the vertexID index?

We can, but need to refactor VidExtractVisitor and maybe lead to incompatible, so don't refactor it here.

nevermore3 (approved):

src/graph/util/ExpressionUtils.cpp
```diff
@@ -1205,7 +1205,40 @@ bool ExpressionUtils::checkExprDepth(const Expression *expr) {
...
+  if (expr->kind() == Expression::Kind::kRelIn) {
+    // id(V) IN [List]
+    if (relExpr->right()->kind() != Expression::Kind::kList) {
+      DLOG(ERROR) << "DEBUG POINT:3";
+      return false;
+    }
+  } else if (expr->kind() == Expression::Kind::kRelEQ) {
+    // id(V) = Value
+    if (relExpr->right()->kind() != Expression::Kind::kConstant) {
+      DLOG(ERROR) << "DEBUG POINT:4";
```
CPWstatic, line 1236: We don't need these debug log if we had done debug.

CPWstatic (commented):
LGTM. Please remove the debug logs.

src/graph/util/ExpressionUtils.cpp
Shylock-Hg: Fixed.

nevermore3 (approved):

CPWstatic (approved):

Issue #4273, opened 2022-05-20
Eliminate `AppendVertices` which's result is not used.


Shylock-Hg:
**Introduction**

e.g. `MATCH (v)-[:like]->() return v`

ps. Need care about `MATCH p = (v)-[:like]->() return p`

Sub job of #4122 

**Contents**

**Related work**

PR #4277, opened 2022-05-24
Remove useless AppendVertices.

Shylock-Hg:
## What type of PR is this?
- [ ] bug
- [ ] feature
- [x] enhancement

#### Issue(s) number: 
Close https://github.com/vesoft-inc/nebula/issues/4273

#### Description:
Remove the useless AppendVertices executor.

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....


czpmango (dismissed):

src/graph/optimizer/rule/EliminateAppendVerticesRule.cpp
```diff
@@ -0,0 +1,84 @@
...
+    if (expr != nullptr) {
+      return false;
+    }
+  }
+  const auto* appendVertices = static_cast<const AppendVertices*>(matched.planNode({0, 0}));
+  DCHECK_EQ(appendVertices->kind(), graph::PlanNode::Kind::kAppendVertices);
+  if (appendVertices->vFilter() != nullptr || appendVertices->filter() != nullptr) {
+    return false;
+  }
+  if (appendVertices->colNames().back()[0] != '_') {  // Anonymous node
```
CPWstatic, line 53: We should have a function to do this, like `AnonVarGenerator::isAnno(string varName)`

src/graph/optimizer/rule/EliminateAppendVerticesRule.cpp
```diff
@@ -0,0 +1,84 @@
...
+using nebula::graph::AppendVertices;
+using nebula::graph::PlanNode;
+using nebula::graph::Project;
+using nebula::graph::QueryContext;
+
+namespace nebula {
+namespace opt {
+
+// TODO If we prune properties in Optimizer Rules, we could check properties of AppendVertices,
+// and drop AppendVertices if it has no properties directly.
```
CPWstatic, line 22: That's right, but more conditions should be considered, such as if the var is used by other operators.
```
Match (a)-[]-(b) Return id(b);
```

CPWstatic (commented):
Excellent!

src/graph/optimizer/rule/EliminateAppendVerticesRule.cpp
Shylock-Hg: In this case, there is properties `_vid`, it's not empty too.

src/graph/optimizer/rule/EliminateAppendVerticesRule.cpp
Shylock-Hg: Done.

CPWstatic (approved):

czpmango (approved):

PR #4290, opened 2022-05-30
Implement inner parallel procesing by using Scatter/Gather model. 

CPWstatic:
## What type of PR is this?
- [ ] bug
- [ ] feature
- [ ] enhancement

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....

### DIFF
--- a/tests/tck/features/optimizer/PrunePropertiesRule.feature
+++ b/tests/tck/features/optimizer/PrunePropertiesRule.feature
@@ -14,7 +14,8 @@ Feature: Prune Properties rule
       MATCH p = (v:player{name: "Tony Parker"})-[e:like]->(v2)
       RETURN v2.player.age
       """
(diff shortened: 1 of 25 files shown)


tests/tck/features/optimizer/PrunePropertiesRule.feature
```diff
@@ -14,7 +14,8 @@ Feature: Prune Properties rule
       MATCH p = (v:player{name: "Tony Parker"})-[e:like]->(v2)
       RETURN v2.player.age
       """
-    Then the result should be, in order:
+    # Since in distributed system and scatter gather model, the order is not guarrented.
```
Shylock-Hg, line 17: Does OpenCypher require order of result?

PR #4305, opened 2022-06-07
fix storage crash

nevermore3:
## What type of PR is this?
- [x] bug
- [ ] feature
- [ ] enhancement

#### Description:
image

When the edge has an index and insert value exceeding length, insert edge will cause crash.

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....

### DIFF
--- a/src/storage/mutate/AddEdgesProcessor.cpp
+++ b/src/storage/mutate/AddEdgesProcessor.cpp
@@ -156,28 +156,19 @@ void AddEdgesProcessor::doProcessWithIndex(const cpp2::AddEdgesRequest& req) {
     auto partId = part.first;
     const auto& edges = part.second;
     // cache edgeKey
-    std::unordered_set<std::string> visited;
-    visited.reserve(edges.size());
     std::vector<kvstore::KV> kvs;
     kvs.reserve(edges.size());
-    auto code = nebula::cpp2::ErrorCode::SUCCEEDED;
-    deleteDupEdge(const_cast<std::vector<cpp2::NewEdge>&>(edges));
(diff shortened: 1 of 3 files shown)


src/storage/mutate/AddEdgesProcessor.cpp
```diff
@@ -386,13 +372,24 @@ std::vector<std::string> AddEdgesProcessor::indexKeys(
...
     while (iter != edges.end()) {
       auto edgeKeyRef = iter->key_ref();
+      if (!NebulaKeyUtils::isValidVidLen(
+              spaceVidLen_, edgeKeyRef->src_ref()->getStr(), edgeKeyRef->dst_ref()->getStr())) {
+        LOG(ERROR) << "Space " << spaceId_ << " vertex length invalid, "
+                   << "space vid len: " << spaceVidLen_
+                   << ", edge srcVid: " << *edgeKeyRef->src_ref()
+                   << ", dstVid: " << *edgeKeyRef->dst_ref() << ", ifNotExists_: " << std::boolalpha
+                   << ifNotExists_;
+        code = nebula::cpp2::ErrorCode::E_INVALID_VID;
```
pengweisong, line 390: May be not need a variable code in this function, just return `nebula::cpp2::ErrorCode::E_INVALID_VID;` will be ok

nevermore3: good idea

panda-sheep (approved):

PR #4307, opened 2022-06-08
fix  drop checkpoint failure

panda-sheep:
## What type of PR is this?
- [x] bug
- [ ] feature
- [ ] enhancement

#### Description:
fix drop checkpoint failure

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....

### DIFF
--- a/src/storage/admin/DropCheckpointProcessor.cpp
+++ b/src/storage/admin/DropCheckpointProcessor.cpp
@@ -13,7 +13,7 @@ void DropCheckpointProcessor::process(const cpp2::DropCPRequest& req) {
   auto spaceIdList = req.get_space_ids();
   auto& name = req.get_name();
   for (auto spaceId : spaceIdList) {
-    auto code = env_->kvstore_->dropCheckpoint(spaceId, std::move(name));
+    auto code = env_->kvstore_->dropCheckpoint(spaceId, name);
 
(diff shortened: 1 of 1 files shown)


critical27 (approved):

src/storage/admin/DropCheckpointProcessor.cpp
```diff
@@ -13,7 +13,7 @@ void DropCheckpointProcessor::process(const cpp2::DropCPRequest& req) {
   auto spaceIdList = req.get_space_ids();
   auto& name = req.get_name();
   for (auto spaceId : spaceIdList) {
-    auto code = env_->kvstore_->dropCheckpoint(spaceId, std::move(name));
```
critical27, line 16: Em, IIRC, in the old code, the second parameter in dropCheckpoint is copied not moved from the `name`. Are you sure this PR fix some bugs?

critical27:
Don't merge~

src/storage/admin/DropCheckpointProcessor.cpp
panda-sheep: I will test it

panda-sheep: My fault

First `name` is a const reference
Secondly, the second parameter of `dropCheckpoint` is also a const reference.
move is meaningless here.

PR #4311, opened 2022-06-09
fix when restoring, the machine key is lost

panda-sheep:
## What type of PR is this?
- [x] bug
- [ ] feature
- [ ] enhancement

#### Description:
fix when restoring, the machine key is lost.
Examples are as follows:

During restore, the mapping relationship of hosts is as follows
```
211 -> 213
212 -> 212
213 -> 211

```
When the restore is over, there are only `211, 212,` two machine keys left, and the `213` machine key is lost.

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....


pengweisong (approved):
Good Job

critical27 (approved):

PR #4328, opened 2022-06-15
Fix get edges transform rule.

Shylock-Hg:
1. Input/Ouput variables.
2. Keep column names of Limit same with input plan node.

## What type of PR is this?
- [x] bug
- [ ] feature
- [ ] enhancement

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....

### DIFF
--- a/src/graph/optimizer/rule/CollapseProjectRule.cpp
+++ b/src/graph/optimizer/rule/CollapseProjectRule.cpp
@@ -75,6 +75,7 @@ StatusOr<OptRule::TransformResult> CollapseProjectRule::transform(
   // 2. find link according to propRefNames and colNames in ProjBelow
   std::unordered_map<std::string, Expression*> rewriteMap;
   auto colNames = projBelow->colNames();
+  DCHECK_EQ(colNames.size(), colsBelow.size());
   for (size_t i = 0; i < colNames.size(); ++i) {
     if (uniquePropRefNames.count(colNames[i])) {
       auto colExpr = colsBelow[i]->expr();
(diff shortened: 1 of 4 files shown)


czpmango:
Nice fix. But since it is a bug fix, is there a corresponding issue to track? For the current pr, it is also necessary to add test cases to constrain subsequent development.

wey-gu:
> Nice fix. But since it is a bug fix, is there a corresponding issue to track? For the current pr, it is also necessary to add test cases to constrain subsequent development.

Thanks, I would like to learn how some of the fixes (addressing my previous faults I think) cover wrong behaviors, too!
Silly questions

- i.e. I don't understand when should we care about input var, output var, col, and when we don't have to, I need to look into its corresponding nodes' executor code or?
- and, I don't understand why for instance `PushTopN index scan` and `Push vFilter ScanV` are doing erase policy differently(one for current and other for all)

Shylock-Hg:
> Nice fix. But since it is a bug fix, is there any corresponding issue to track? For the current pr, it is also necessary to add test cases to constrain subsequent development.

In fact, the CI job in Debug building will report error, but now it always compile to OOM

Shylock-Hg:
> > Nice fix. But since it is a bug fix, is there a corresponding issue to track? For the current pr, it is also necessary to add test cases to constrain subsequent development.
> 
> Thanks, I would like to learn how some of the fixes (addressing my previous faults I think) cover wrong behaviors, too!
> 
> Silly questions from my side
> 
> * i.e. I don't understand when should we care about input var, output var, col, and when we don't have to, I need to look into its corresponding nodes' executor code or?
> * and, I don't understand why for instance `PushTopN index scan` and `Push vFilter ScanV` are doing erase policy differently(one for current and other for all)

1. You can check the constraints I mentioned in https://github.com/vesoft-inc/nebula/pull/4146#issue-1202863174
2. Now there are no differences, two methods to erase will lead to that subplan produced by rule will replace origin.

PR #4331, opened 2022-06-17
force cache the docker layer

HarrisChu:
## What type of PR is this?
- [ ] bug
- [ ] feature
- [ ] enhancement

#### Description:
force cache the docker layer.
the layer is about 20GB, and need another 20min to export the layer.
after then each job would need 5min with the cache layer.

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -82,6 +82,7 @@ jobs:
           tags: |
             vesoft/nebula-graphd:nightly
           target: graphd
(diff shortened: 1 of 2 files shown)


Aiee (approved):

HarrisChu:
fix https://github.com/vesoft-inc/nebula/runs/6924097368?check_suite_focus=true

Sophie-Xie (approved):

PR #4352, opened 2022-06-24
fix_shortest_path_crash

nevermore3:
## What type of PR is this?
- [x] bug
- [ ] feature
- [ ] enhancement

#### Description:

Graph crash when running tck manually 
image
Can't be reproduced consistently.
Tck is run with 16 concurrent threads.

## How do you solve it?

 shared variables are locked before use in multi-threading environment

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....

### DIFF
--- a/src/graph/executor/algo/ShortestPathBase.h
+++ b/src/graph/executor/algo/ShortestPathBase.h
@@ -43,15 +43,9 @@ class ShortestPathBase {
 
   Status handleErrorCode(nebula::cpp2::ErrorCode code, PartitionID partId) const;
 
-  void addStats(RpcResponse& resp,
-                std::unordered_map<std::string, std::string>* stats,
-                size_t stepNum,
-                int64_t timeInUSec,
-                bool reverse) const;
+  void addStats(RpcResponse& resp, size_t stepNum, int64_t timeInUSec, bool reverse) const;
 
-  void addStats(PropRpcResponse& resp,
-                std::unordered_map<std::string, std::string>* stats,
-                int64_t timeInUSec) const;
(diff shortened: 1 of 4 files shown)


src/graph/executor/algo/ShortestPathBase.h
```diff
@@ -86,6 +80,7 @@ class ShortestPathBase {
   std::unordered_map<std::string, std::string>* stats_{nullptr};
   size_t maxStep_;
   bool singleShortest_{true};
+  folly::SpinLock statsLock_;
```
Shylock-Hg, line 83: How about other memebers?

nevermore3: other members will not be written by multiple threads at the same time

src/graph/executor/algo/SingleShortestPath.cpp
```diff
@@ -264,6 +264,9 @@ folly::Future<bool> SingleShortestPath::buildEvenPath(size_t rowNum,
       return false;
     }
     for (auto& meetVertex : vertices) {
+      if (!meetVertex.isVertex()) {
```
Shylock-Hg, line 267: Why this isn't vertex?

Aiee (approved):

src/graph/executor/algo/SingleShortestPath.cpp
nevermore3: vertices may contain EMPTY values ​​in exceptional cases

Shylock-Hg (approved):

PR #4357, opened 2022-06-27
subgraph filter

nevermore3:
## What type of PR is this?
- [ ] bug
- [x] feature
- [ ] enhancement

#### Description:
support vertex filter & edge filter in subgraph sentence

1、subgraph operator consists of the getNeighbors operator, which loops the user-specified number of times inside the subgraph operator.
2、where clause does not support logicalOR expressions, bcz `$^.player.age > 30 OR like.likeness > 80` can't filter data only by $^.player.age
3、for the VID that does not satisfy the tagfilter condition, the filterEdge function will filter the edge where dst is VID in the previous step

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....


nevermore3:
waiting https://github.com/vesoft-inc/nebula/issues/4649 fix

Shylock-Hg (approved):

PR #4361, opened 2022-06-28
Implement inner parallel procesing by using Scatter/Gather model.

CPWstatic:
Some debug infos.

Multi job AppendVertices handle resp.

Multi job Project.

Enhance AppendVertices.

Multi job filter.

Fix row cnt and delete debug info.

Fix AppendV.

Add debug info.

Rm debug info.

Fix copy in traverse.

Set max_job_size to default 20.

Add runMultiJobs template.

Impl Project/Filter/AppendV.

Impl Join.

Fix double free for join.

Fix filter on GetNeighborsIter.

Fix prop iter.

Fix filter on prop iter.

Fix left join by clearing hash tabls manually.

Report error in filter.

Fix tests: the order is not guarentted in scatter gather model.

Iterates to begin out of job.

Modify the default config.

Fix appendv.

Support run multi job mode test.

Test multi job.

Rebase and compatible with single job.

## What type of PR is this?
- [ ] bug
- [ ] feature
- [x] enhancement

#### Description:

This pr implements inner parallel processing by using Scatter/Gather model and improves a lot for analytics processing scenario. Not all operators are implemented to parallel-processing for now. Those left operators will be implemented in futher future.

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

This pr implements inner parallel processing by using Scatter/Gather model and improves a lot for analytics processing scenario. Not all operators are implemented to parallel-processing for now. Those left operators will be implemented in futher future.
```
max_job_size
min_batch_size
```

### DIFF
--- a/src/graph/executor/Executor.h
+++ b/src/graph/executor/Executor.h
@@ -106,6 +106,17 @@ class Executor : private boost::noncopyable, private cpp::NonMovable {
   // Store the default result which not used for later executor
   Status finish(Value &&value);
 
+  size_t getBatchSize(size_t totalSize) const;
+
+  // ScatterFunc: A callback function that handle partial records of a dataset.
+  // GatherFunc: A callback function that gather all results of ScatterFunc, and do post works.
+  // Iterator: An iterator of a dataset.
+  template <
+      class ScatterFunc,
+      class ScatterResult = typename std::result_of<ScatterFunc(size_t, size_t, Iterator *)>::type,
+      class GatherFunc>
+  auto runMultiJobs(ScatterFunc &&scatter, GatherFunc &&gather, Iterator *iter);
+
   int64_t id_;
 
   // Executor name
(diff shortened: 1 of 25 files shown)


src/graph/executor/Executor.h
```diff
@@ -106,6 +106,17 @@ class Executor : private boost::noncopyable, private cpp::NonMovable {
...
+  size_t getBatchSize(size_t totalSize) const;
+
+  // ScatterFunc: A callback function that handle partial records of a dataset.
+  // GatherFunc: A callback function that gather all results of ScatterFunc, and do post works.
+  // Iterator: An iterator of a dataset.
+  template <
+      class ScatterFunc,
+      class ScatterResult = typename std::result_of<ScatterFunc(size_t, size_t, Iterator *)>::type,
+      class GatherFunc>
+  auto runMultiJobs(ScatterFunc &&scatter, GatherFunc &&gather, Iterator *iter);
```
nevermore3, line 118: Status runMultiJobs(...)
the return value type is explicit

src/graph/executor/query/TraverseExecutor.cpp
```diff
@@ -267,7 +291,133 @@ Status TraverseExecutor::buildInterimPath(GetNeighborsIter* iter) {
...
+  }
+  paths_.emplace_back();
+
+  auto scatter = [this, prev](
+                     size_t begin, size_t end, Iterator* tmpIter) mutable -> StatusOr<JobResult> {
+    return handleJob(begin, end, tmpIter, *prev);
+  };
+
+  auto gather = [this,
+                 pathCnt = pathCnt](std::vector<StatusOr<JobResult>> results) mutable -> Status {
```
nevermore3, line 315: [this, pathCnt](...)

src/graph/executor/query/TraverseExecutor.cpp
```diff
@@ -267,7 +291,133 @@ Status TraverseExecutor::buildInterimPath(GetNeighborsIter* iter) {
...
+    auto pathToSrcFound = prev.find(srcV.getVertex().vid);
+    if (pathToSrcFound == prev.end()) {
+      return Status::Error("Can't find prev paths.");
+    }
+    const auto& paths = pathToSrcFound->second;
+    for (auto& prevPath : paths) {
+      if (hasSameEdge(prevPath, e.getEdge())) {
+        continue;
+      }
+      if (uniqueDst.emplace(dst).second) {
```
nevermore3, line 393: this can guarantee the uniqueness within a thread, but the uniqueness between different threads cannot be guaranteed
Is it possible to deduplicate  in gather function one more time

nevermore3 (approved):

src/graph/executor/query/InnerJoinExecutor.h
```diff
@@ -31,6 +32,16 @@ class InnerJoinExecutor : public JoinExecutor {
                          Iterator* probeIter,
                          const std::unordered_map<Value, std::vector<const Row*>>& hashTable) const;
 
+  // joinMultiJobs/probe/singleKeyProbe implemented for multi jobs.
+  // For now, the InnerJoin implementation only implement the parallel processing on probe side.
+  folly::Future<Status> joinMultiJobs(const std::vector<Expression*>& hashKeys,
+                                      const std::vector<Expression*>& probeKeys,
+                                      const std::vector<std::string>& colNames);
+
+  folly::Future<Status> probe(const std::vector<Expression*>& probeKeys, Iterator* probeIter);
```
Shylock-Hg, line 41: Mark `this` to `const` to avoid modify hash table in multiple threads.

Shylock-Hg (approved):
LGTM. But we'd better to split data and executor in scheduler to avoid modify implementation of executor. And keep executer execute in single thread and view 'the whole data'.

PR #4362, opened 2022-06-28
Async gc.

CPWstatic:
## What type of PR is this?
- [ ] bug
- [ ] feature
- [x] enhancement

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

Clean the unused memory on background threads, this is helpful for big queries since the memory release of interim results may cost too much time. We sugguest enable async gc in analytics processing scenario.
```
enable_async_gc
gc_worker_size
```


nevermore3 (dismissed):

nevermore3 (approved):

src/graph/gc/GC.cpp
```diff
@@ -0,0 +1,34 @@
...
+  workers_.addRepeatTaskForAll(50, &GC::periodicTask, this);
+}
+
+void GC::clear(std::vector<Result>&& garbage) {
+  queue_.enqueue(std::move(garbage));
+}
+
+void GC::periodicTask() {
+  // TODO: maybe could release by batch
+  queue_.try_dequeue();
```
Shylock-Hg, line 31: There are multiple threads contention. I think you could move closure to executor directly.

PR #4392, opened 2022-07-08
Update pull_request.yml

kezhenxu94:
We have a dedicate GitHub actions for header check so this should be updated, and we don't recommend using `@main` so there is no unexpected behaviors when we have changes in our codebase.

See https://github.com/apache/skywalking-eyes/pull/123


kezhenxu94:
Duplicate of https://github.com/vesoft-inc/nebula/pull/4390

PR #4395, opened 2022-07-11
Replace obsolete RocksDB API

Qiaolin-Yu:
## What type of PR is this?
- [ ] bug
- [ ] feature
- [x] enhancement

#### Description:
- Replace obsolete RocksDB API "backupable db" to be compatible with RocksDB in the future

## How do you solve it?
- Replace with updated RocksDB API.

- RocksDB has supported the updated API in very early versions, so it will not cause existing compatibility problems.

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....

### DIFF
--- a/src/kvstore/RocksEngine.cpp
+++ b/src/kvstore/RocksEngine.cpp
@@ -538,7 +538,7 @@ void RocksEngine::openBackupEngine(GraphSpaceID spaceId) {
       }
     }
     rocksdb::BackupEngine* backupDb;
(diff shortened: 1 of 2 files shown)


panda-sheep (approved):
Thanks for your contribution.  Good.

In backupable_db.h, Use the following aliases for backward compatibility.
`using BackupableDBOptions = BackupEngineOptions;`

critical27 (approved):

Issue #4397, opened 2022-07-11
Storage Crash when deleting vertex does not exist and vid length more than we define in space.


Milittle:
Storage Crash when deleting vertex does not exist and vid length more than we define in space.

**How To Reproduce(__required__)**

Step 1 
Create one space: define vid length is 120.

create tag and create one index.(this bug need indexes in space.)

Step 2 
delete vertex 'xxxx' 

xxx length need more than vid length defined.

the storage will be dead.

I think we should add one verfication in the follow one line just like the second line:

one line: https://github.com/vesoft-inc/nebula/blob/7a7ab68a0497da8f13e1af7c5ca6b080352202a3/src/storage/mutate/DeleteVerticesProcessor.cpp#L117

second line: https://github.com/vesoft-inc/nebula/blob/7a7ab68a0497da8f13e1af7c5ca6b080352202a3/src/storage/mutate/DeleteVerticesProcessor.cpp#L59

Milittle:
> Storage Crash when deleting vertex does not exist and vid length more than we define in space.
> 
> **How To Reproduce(**required**)**
> 
> Steps to reproduce the behavior:
> 
> Step 1 Create one space: define vid length is 120.
> 
> create tag and create one index.(this bug need indexes in space.)
> 
> Step 2 delete vertex 'xxxx'
> 
> xxx length need more than vid length defined.
> 
> the storage will be dead.
> 
> I think we should add one verfication in the follow one line just like the second line:
> 
> one line:
> 
> https://github.com/vesoft-inc/nebula/blob/7a7ab68a0497da8f13e1af7c5ca6b080352202a3/src/storage/mutate/DeleteVerticesProcessor.cpp#L117
> 
> second line:
> 
> https://github.com/vesoft-inc/nebula/blob/7a7ab68a0497da8f13e1af7c5ca6b080352202a3/src/storage/mutate/DeleteVerticesProcessor.cpp#L59

If this worked, can i pull new request for this bug?

wey-gu:
@SuperYoko could you plz take a look at this?

SuperYoko:
Maybe a verify in graph is needed too. Since @Sophie-Xie  has add this to v3.2.0, I think it will be fixed soon. 

(And we have acctually met some crash with vid len check. : D, thanks for your help~)

PR #4406, opened 2022-07-12
Fix delete vertex or edge, the storage crashed.(issue: #4397)

Milittle:
## What type of PR is this?
- [ ] bug

#### Description:

* When our space has tag index or edge index.
* When delete vertex or edge, we specify one vid more than our defined
  in space.
* The sotrage will be crashed.

## How do you solve it?

Solution:
1. Add vid valid verfication.

## Checklist:
Tests:
- [ ] N/A

Affects:
- [ ] N/A

## Release notes:

> ex. Fixed storaged crashed, when we delete vertex and edge specify vid more than our defined before in space.

### DIFF
--- a/src/storage/mutate/DeleteEdgesProcessor.cpp
+++ b/src/storage/mutate/DeleteEdgesProcessor.cpp
@@ -103,6 +103,14 @@ void DeleteEdgesProcessor::process(const cpp2::DeleteEdgesRequest& req) {
 
       nebula::cpp2::ErrorCode err = nebula::cpp2::ErrorCode::SUCCEEDED;
       for (const auto& edgeKey : part.second) {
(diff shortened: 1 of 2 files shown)


SuperYoko (approved):
LGTM. It's now enough to stop crash due to this check. 

1. This check means that somebody must guarantee  it, and the earlier the better, maybe later there's some fix.
2. TCK needed, Maybe later this week I'll add it.

critical27 (approved):

wey-gu:
> LGTM. It's now enough to stop crash due to this check.
> 
> 1. This check means that somebody must guarantee  it, and the earlier the better, maybe later there's some fix.
> 2. TCK needed, Maybe later this week I'll add it.

Thanks @SuperYoko :)

Issue #4412, opened 2022-07-13
ColumnExpression encode and decode are not matched


Linary:
**Please check the FAQ documentation before raising an issue**

**Describe the bug (__required__)**

Can't decode ColumnExpression, it will report failed like:

```
F20220713 17:35:04.868080 830015 Value.cpp:749] Check failed: type_ == Type::INT (__EMPTY__ vs. INT) 
```

The reason is that the implementation of encode and decode methods do not match.

```
void ColumnExpression::writeTo(Encoder &encoder) const {
  encoder << kind_;
  // call Expression::Encoder::operator<<(size_t size)
  encoder << index_;
}

void ColumnExpression::resetFrom(Decoder &decoder) {
  // Expression::Decoder::readValue()
  index_ = decoder.readValue().getInt();
}
```

**Your Environments (__required__)**

* OS: Linux hw-ecs 4.18.0-348.7.1.el8_5.x86_64 #1 SMP
* Compiler: g++ (GCC) 8.5.0 20210514 (Red Hat 8.5.0-4)
* CPU: x86_64
* Commit id (e.g. `a3ffc7d8`)

**How To Reproduce(__required__)**

1. make a ColumnExpression object
2. encode it to string
3. decode string

**Expected behavior**

**Additional context**

PR #4413, opened 2022-07-13
Fix ColumnExpression encode and decode are not matched

Linary:
## What type of PR is this?
- [x] bug
- [ ] feature
- [ ] enhancement

#### Issue(s) number: 

[#4412](https://github.com/vesoft-inc/nebula/issues/4412)

## How do you solve it?

Let ColumnExpression writeTo encoder via `Value` instead of `size_t`

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....

### DIFF
--- a/src/common/expression/ColumnExpression.cpp
+++ b/src/common/expression/ColumnExpression.cpp
@@ -38,7 +38,7 @@ void ColumnExpression::writeTo(Encoder &encoder) const {
 }
 
 void ColumnExpression::resetFrom(Decoder &decoder) {
-  index_ = decoder.readValue().getInt();
+  index_ = decoder.readSize();
 }
 
 }  // namespace nebula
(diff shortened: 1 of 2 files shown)


src/common/expression/ColumnExpression.cpp
```diff
@@ -34,7 +34,7 @@ void ColumnExpression::accept(ExprVisitor *visitor) {
 
 void ColumnExpression::writeTo(Encoder &encoder) const {
   encoder << kind_;
-  encoder << index_;
+  encoder << Value(index_);
```
Shylock-Hg, line 37: ReadSize in decoder is better.

Linary: done.

Linary:
please review again?

Shylock-Hg (approved):

PR #4416, opened 2022-07-14
fix deadlock

critical27:
## What type of PR is this?
- [X] bug
- [ ] feature
- [ ] enhancement

#### Description:

There are two read lock in LeaderBalanceJobExecutor.cpp in same thread (grab the first one then another, first not released). Once a write lock happens between two read lock, deadlock would happen.

> btw, std::shared_lock won't happen, will dig more later

## Checklist:
Tests:
- [X] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....


kikimo (approved):
LGTM

liwenhui-soul (approved):

Issue #4424, opened 2022-07-14
test_single_shortestpaths3 fails with not found results


xtcyclist:
**Please check the FAQ documentation before raising an issue**

**Describe the bug (__required__)**

`E     NotFoundRow: 4: |<("Tim Duncan":bachelor{name:Tim Duncan,speciality:psychology}:player{age:42,name:Tim Duncan})-[:serve@0{end_year:2016,start_year:1997}]->("Spurs":team{name:Spurs})<-[:serve@0{end_year:2019,start_year:2017}]-("Rudy Gay":player{age:32,name:Rudy Gay})-[:serve@0{end_year:2017,start_year:2013}]->("Kings":team{name:Kings})>|`

**Your Environments (__required__)**

* OS: ubuntu 20.04
* Compiler: clang-10
* CPU: `lscpu`
* Commit id: 7981108f509f1e7e5c9134932f6b91fa2c177b9f

**How To Reproduce(__required__)**

1. Step 1
make RM_DIR=false DEBUG=false J=6 tck
or
pytest -k 'test_single_shortestpaths3' .
3. Step 2
4. Step 3

**Expected behavior**

**Additional context**

Not sure whether it's really a bug, or it's just the test results are not deterministic.

xtcyclist:
this turns out to be an expected result. 'match shortestPath()' returns only 1 matching result out of many, the contents of which depend on the execution env. 

thus, the match query in this test case may have multiple valid but different results.

considering changing the test case.

czpmango:
This is cypher's doc [https://neo4j.com/docs/cypher-manual/current/execution-plans/shortestpath-planning/](https://neo4j.com/docs/cypher-manual/current/execution-plans/shortestpath-planning/)

Please test neo4j behavior and align with @nevermore3  the original intention of the implementation.

czpmango:
Fwiw, you can refer to this post to understand the difference between `shortestpath pattern` and `allshortestpath pattern`. 
[Difference between `shortestpath` and `allshortestpaths` in cypher](https://community.neo4j.com/t5/neo4j-graph-platform/what-s-the-difference-between-shortestpath-and-allshortestpaths/m-p/36869)

PR #4431, opened 2022-07-15
modify the test case of shortestspaths3 to make its results determini…

xtcyclist:
…stic, given that match with shortestPath() may return one shortest path, not all of them.

## What type of PR is this?
- [x] bug
- [ ] feature
- [ ] enhancement

#### Issue(s) number: 
#4424 

#### Description:
The original test case 
```
      MATCH (a:player{name:"Tim Duncan"}), (b:team)
      MATCH p = shortestPath( (a)-[e:serve*..3]-(b) )
      RETURN p
```
may return only one shortestPath() between the src and the end, not all of them. The exact path returned depends on the state of the database. It's not deterministic. So it happens that this test case fails with an expected result.

The behavior of shortestPath() is to be discussed. Maybe it should return all shortest paths with the same shortest length. Or, maybe an allshortestpath() should be added?

For the time being, I recommend modifying this query to pass this test case.

## How do you solve it?

Modified the return values in the query to have deterministic results.
```
      MATCH (a:player{name:"Tim Duncan"}), (b:team)
      MATCH p = shortestPath( (a)-[e:serve*..3]-(b) )
      RETURN b, length(p)
```

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....


tests/tck/features/match/SingleShorestPath.feature
```diff
@@ -205,30 +205,51 @@ Feature: single shortestPath
...
+    #  | <("Tim Duncan" :bachelor{name: "Tim Duncan", speciality: "psychology"} :player{age: 42, name: "Tim Duncan"})-[:serve@0 {end_year: 2016, start_year: 1997}]->("Spurs" :team{name: "Spurs"})<-[:serve@0 {end_year: 2019, start_year: 2016}]-("Paul Gasol" :player{age: 38, name: "Paul Gasol"})-[:serve@0 {end_year: 2008, start_year: 2001}]->("Grizzlies" :team{name: "Grizzlies"})>                       |
+    #  | <("Tim Duncan" :bachelor{name: "Tim Duncan", speciality: "psychology"} :player{age: 42, name: "Tim Duncan"})-[:serve@0 {end_year: 2016, start_year: 1997}]->("Spurs" :team{name: "Spurs"})<-[:serve@0 {end_year: 2019, start_year: 2016}]-("Paul Gasol" :player{age: 38, name: "Paul Gasol"})-[:serve@0 {end_year: 2014, start_year: 2008}]->("Lakers" :team{name: "Lakers"})>                             |
```
nevermore3, line 234: u can delete these result

xtcyclist: deleted.

nevermore3 (dismissed):

nevermore3 (approved):

codesigner (approved):

Issue #4382, opened 2022-07-04
download should fail when hdfs client not being installed/called


wey-gu:
It succeeds even not being installed.

Ref: https://discuss.nebula-graph.com.cn/t/topic/9286/23?u=wey

critical27:
Bug confirmed, will try to fix it.

wey-gu:
thanks @critical27 , closed as fixed by #4442

Issue #4263, opened 2022-05-18
Job status is failed, but the error code of tasks is succeed.


slimshadylol0303:
![image](https://user-images.githubusercontent.com/95005615/168972175-85738f60-c1b8-46ad-9236-05fd25bbeab9.png)

PR #4442, opened 2022-07-19
fix download always return succeeded

### LINKED ISSUE #4321: Download sst return success but actually failed
As title

* Commit id (e.g. `a3ffc7d8`)
image

image

image

image

**Expected behavior**

**Additional context**

critical27:
## What type of PR is this?
- [X] bug
- [ ] feature
- [ ] enhancement

#### Issue(s) number: 
Close #4321, #4382

#### Description:
`ProcessUtils::runCommand` will always return ok. So we can't tell whether hdfs command succeeded or not.

## How do you solve it?
Use SubProcess to execute shell command.

After this PR, if the hdfs address is illegal, the `show job` is weird (job failed but task succeed), probably same as https://github.com/vesoft-inc/nebula/issues/4263, will try to fix it later.

image

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....


liwenhui-soul (approved):

src/common/hdfs/HdfsCommandHelper.cpp
```diff
@@ -5,40 +5,54 @@
...
   auto command = folly::stringPrintf(
       "hdfs dfs -ls hdfs://%s:%d%s", hdfsHost.c_str(), hdfsPort, hdfsPath.c_str());
   LOG(INFO) << "Running HDFS Command: " << command;
-  auto result = ProcessUtils::runCommand(command.c_str());
-  if (result.ok()) {
-    return result.value();
-  } else {
-    return Status::Error(folly::stringPrintf("Failed to run %s", command.c_str()));
+  try {
+    folly::Subprocess proc(std::vector<std::string>({command}));
```
panda-sheep, line 23: 👍

panda-sheep (approved):
Good

src/common/hdfs/HdfsCommandHelper.cpp
SuperYoko, line 23: Here actually may have some trouble with system environment variable.

PR #4443, opened 2022-07-19
Avoid folly::init overwirte log configuration.

Shylock-Hg:
## What type of PR is this?
- [x] bug
- [ ] feature
- [ ] enhancement

#### Description:
Avoid folly::init overwirte the glog configuration.

## How do you solve it?
Patch of #4409 

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....

### DIFF
--- a/src/daemons/MetaDaemon.cpp
+++ b/src/daemons/MetaDaemon.cpp
@@ -66,12 +66,7 @@ int main(int argc, char* argv[]) {
   // the 2nd will make the 1st failed to output log anymore
   gflags::ParseCommandLineFlags(&argc, &argv, false);
 
-  // Setup logging
-  auto status = setupLogging(argv[0]);
-  if (!status.ok()) {
-    LOG(ERROR) << status;
-    return EXIT_FAILURE;
-  }
+  Status status;
 
 #if defined(ENABLE_BREAKPAD)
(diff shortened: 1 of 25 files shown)


src/daemons/MetaDaemon.cpp
```diff
@@ -66,12 +66,7 @@ int main(int argc, char* argv[]) {
   // the 2nd will make the 1st failed to output log anymore
```
critical27, line 66: Please verify if the behavior in the comment.

Shylock-Hg: In current `folly::init` and `SetUpLogging` is called after pid checking too.

Shylock-Hg: But Standalone daemon don't require this constraint even before this PR, need to modify it?

nevermore3 (approved):

src/daemons/MetaDaemon.cpp
critical27: It is ok for now. I can't clearly remember why we move `setupLogging` very close to start position of the main.

critical27 (approved):

Issue #4247, opened 2022-05-11
Failed job cannot be stopped


NicolaCage:
<img width="1191" alt="image" src="https://user-images.githubusercontent.com/37568394/167779411-af493e05-3c96-4a1e-9ce4-319a3d8c9e0c.png">
<img width="206" alt="image" src="https://user-images.githubusercontent.com/37568394/167779460-6013d5c1-4e59-4de5-988f-b71f3ecc3ad9.png">

NicolaCage:
<img width="1173" alt="image" src="https://user-images.githubusercontent.com/37568394/167780710-1f6e01c1-2be0-48a9-869a-cb03aedb534c.png">
The running task cannot be stopped

NicolaCage:
nebula version： Community v3.1.0

critical27:
The error message has say it failed to persistent job info. ZONE_BALANCE support stopped (the stop job operation failed), see the sheet I send you guys before. It is quite a different problem with the title of the issue.

I'll close it later if there is no more update

critical27:
BTW, we can only stop the latest balance job. FAILED of outdate job should not stop. This is by design

critical27:
Some problem has been fixed in #4460, again, failed job cannot and should not be stopped

Issue #4232, opened 2022-05-06
show job the Status is FAILED but the Error Code is SUCCEEDED


dbacyj:
**Describe the bug (__required__)**
* balance leader and show job xx :

   <img width="776" alt="image" src="https://user-images.githubusercontent.com/72034305/171137418-b03c4de1-5c22-4934-8ba6-8ef3fa53bc5c.png">

* the meta log 
![image](https://user-images.githubusercontent.com/72034305/167078756-c0e1e763-d0b5-41ef-bba3-55cf916981c9.png)

* version info
![image](https://user-images.githubusercontent.com/72034305/167079244-257a7394-7fec-4b8c-a491-80dc3805c9f3.png)

critical27:
Marked

PR #4446, opened 2022-07-20
Job manager related fix

critical27:
## What type of PR is this?
- [X] bug
- [ ] feature
- [X] enhancement

#### Issue(s) number: 
https://github.com/vesoft-inc/nebula/issues/4247
https://github.com/vesoft-inc/nebula/issues/4232 (Part of the issue is addressed in #4444)
https://github.com/vesoft-inc/nebula-ent/issues/1052

#### Description:
1. When stop job a failed/finished job, we will return "E_SAVE_JOB_FAILURE" , which is obscure, return another error code.
2. Balance leader/data's error code is not passed out.  And balance leader has no tasks, which is quite same as #4442

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....

### DIFF
--- a/src/meta/processors/job/LeaderBalanceJobExecutor.cpp
+++ b/src/meta/processors/job/LeaderBalanceJobExecutor.cpp
@@ -227,17 +227,14 @@ nebula::cpp2::ErrorCode LeaderBalanceJobExecutor::finish(bool) {
   return nebula::cpp2::ErrorCode::SUCCEEDED;
 }
 
-folly::Future<Status> LeaderBalanceJobExecutor::executeInternal() {
-  folly::Promise<Status> promise;
+folly::Future<nebula::cpp2::ErrorCode> LeaderBalanceJobExecutor::executeInternal() {
+  folly::Promise<nebula::cpp2::ErrorCode> promise;
   auto future = promise.getFuture();
   // Space ID, Replica Factor and Dependent On Group
   std::vector<std::tuple<GraphSpaceID, int32_t, bool>> spaces;
   auto ret = getAllSpaces(spaces);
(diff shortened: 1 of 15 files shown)


wenhaocs (approved):
LGTM

src/meta/processors/job/LeaderBalanceJobExecutor.cpp
```diff
@@ -286,13 +283,13 @@ folly::Future<Status> LeaderBalanceJobExecutor::executeInternal() {
 
     inLeaderBalance_ = false;
     if (failed != 0) {
-      return Status::Error("partiton failed to transfer leader");
+      LOG(INFO) << folly::stringPrintf("%d partitons failed to transfer leader", failed);
```
panda-sheep, line 286: Why does it still return success here?

panda-sheep (commented):
Good job

src/meta/processors/job/LeaderBalanceJobExecutor.cpp
critical27: Because we can't expect transfer leader always succeed (it is not a definitely successful operations). Not quite related to this PR.

panda-sheep: ok

panda-sheep (approved):

Sophie-Xie:
@critical27 pls fix the tck.

panda-sheep (approved):

liwenhui-soul (approved):

wenhaocs (approved):

PR #4478, opened 2022-08-02
Parser2

peter-rich:
## What type of PR is this?
- [ ] bug
- [x] feature
- [ ] enhancement

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....


src/parser/TraverseSentences.h
```diff
@@ -9,9 +9,23 @@
 #include "parser/EdgeKey.h"
 #include "parser/MutateSentences.h"
 #include "parser/Sentence.h"
-
 namespace nebula {
 
+class IsomorSentence final : public Sentence {
+  public:
+    explicit IsomorSentence(NameLabelList* tags) {
+      tags_.reset(tags);
```
wenhaocs, line 17: Why do you take tags as input for isomor sentence? Shouldn't it be space name？

src/parser/parser.yy
```diff
@@ -1390,11 +1392,20 @@ truncate_clause
...
-        $$ = go;
+        auto go = new GoSentence($2, $3, $4, $5, $7);  
+        go->setYieldClause($6);  
+        $$ = go;  
     }
     ;
+ 
+ 
+ isomor_sentence
+    : KW_ISOMOR name_label_list {
```
wenhaocs, line 1403: Why do you only take one argument? Shouldn't it be 2 arguments?

wenhaocs (commented):
Please discuss it with me.

PR #4496, opened 2022-08-08
Build the planner and validator for subgraph Isomorphism

peter-rich:
## What type of PR is this? 
- [ ] bug
- [x] feature
- [ ] enhancement

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....

### DIFF
--- a/src/graph/planner/ngql/IsomorPlanner.h
+++ b/src/graph/planner/ngql/IsomorPlanner.h
@@ -1,0 +1,42 @@
+/* Copyright (c) 2021 vesoft inc. All rights reserved.
+ *
+ * This source code is licensed under Apache 2.0 License.
+ */
+
+#ifndef GRAPH_PLANNER_NGQL_FETCH_VERTICES_PLANNER_H_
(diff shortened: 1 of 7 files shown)


wenhaocs (approved):
LGTM. Will review later.

wenhaocs (commented):
You need to register plans.
Need to define the interface (input and output) in your plan rather than use others's input/output directly.

src/graph/planner/ngql/IsomorPlanner.h
```diff
@@ -0,0 +1,42 @@
+/* Copyright (c) 2021 vesoft inc. All rights reserved.
+ *
+ * This source code is licensed under Apache 2.0 License.
+ */
+
+#ifndef GRAPH_PLANNER_NGQL_FETCH_VERTICES_PLANNER_H_
```
wenhaocs, line 6: The macro should be updated. Otherwise, it will not included.

PR #4498, opened 2022-08-09
Feature/pattern expression ref local variable

Shylock-Hg:
## What type of PR is this?
- [ ] bug
- [x] feature
- [ ] enhancement

#### Description:
Support pattern expression reference local defined variable, implement this by filter path in list(the result of pattern expressioin)

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....


src/graph/util/ParserUtil.cpp
```diff
@@ -14,21 +14,21 @@ namespace graph {
 void ParserUtil::rewriteLC(QueryContext *qctx,
                            ListComprehensionExpression *lc,
                            const std::string &oldVarName) {
-  const auto &newVarName = qctx->vctx()->anonVarGen()->getVar();
-  qctx->ectx()->setValue(newVarName, Value());
+  qctx->vctx()->anonVarGen()->createVar(oldVarName);
+  qctx->ectx()->setValue(oldVarName, Value());
```
jievince, line 18: Can you explain this?

Shylock-Hg: In List Comprehesion `[i in [...] | ...]`, the local defined variable `i` will be used in sub-expr, so I keep its origin name.

jievince: What if the local var name same with the global var?

Shylock-Hg: Shadow global variable.

jievince (approved):

nevermore3 (approved):

PR #4515, opened 2022-08-12
refactor recover job

liwenhui-soul:
## What type of PR is this?
- [ ] bug
- [ ] feature
- [x] enhancement

## How do you solve it?
1. a meta job that used to be RUNNING would be set to FAILED when process restart

2.  only jobs that are failed or stopped could be recovered

3. all recovered jobs would be set to QUEUE

4. for a balance job: 
(1) if there's a newer finished balance job, the stopped or failed balance job can't be recovered.
(2) only the lasted stopped or failed balance job could be recovered

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....


src/meta/processors/job/JobManager.cpp
```diff
@@ -646,7 +652,7 @@ nebula::cpp2::ErrorCode JobManager::checkNeedRecoverJobExist(GraphSpaceID spaceI
       auto status = std::get<2>(tup);
       if (type == cpp2::JobType::DATA_BALANCE || type == cpp2::JobType::ZONE_BALANCE) {
         // QUEUE: The job has not been executed, the machine restarted
-        if (status == cpp2::JobStatus::FAILED || status == cpp2::JobStatus::QUEUE) {
+        if (status == cpp2::JobStatus::FAILED) {
```
SuperYoko, line 655: Modify this comment.

src/meta/processors/job/JobManager.cpp
```diff
@@ -709,187 +715,130 @@ nebula::cpp2::ErrorCode JobManager::stopJob(GraphSpaceID spaceId, JobID jobId) {
...
+    return retCode;
+  }
+  for (; iter->valid(); iter->next()) {
+    if (!MetaKeyUtils::isJobKey(iter->key())) {
+      continue;
     }
-    for (; iter->valid(); iter->next()) {
-      if (!MetaKeyUtils::isJobKey(iter->key())) {
+    auto optJobRet = JobDescription::makeJobDescription(iter->key(), iter->val());
+    if (!nebula::ok(optJobRet)) {
```
SuperYoko, line 738: Handle this.

SuperYoko (approved):

src/meta/processors/job/JobManager.cpp
```diff
@@ -83,14 +83,26 @@ nebula::cpp2::ErrorCode JobManager::handleRemainingJobs() {
       auto optJob = nebula::value(optJobRet);
       std::unique_ptr<JobExecutor> je =
           JobExecutorFactory::createJobExecutor(optJob, kvStore_, adminClient_);
-      // Only balance has been recovered
+      // Only balance would change
       if (optJob.getStatus() == cpp2::JobStatus::RUNNING && je->isMetaJob()) {
-        jds.emplace_back(optJob);
+        jds.emplace_back(std::move(optJob));
```
critical27, line 88: Won't this would break the rule below?
```
for a balance job:
(1) if there's a newer finished balance job, the stopped or failed balance job can't be recovered.
(2) only the lasted stopped or failed balance job could be recovered
```

src/meta/processors/job/JobManager.cpp
```diff
@@ -83,14 +83,26 @@ nebula::cpp2::ErrorCode JobManager::handleRemainingJobs() {
...
+        auto spaceId = optJob.getSpace();
+        auto jobId = optJob.getJobId();
+        enqueue(spaceId, jobId, JbOp::ADD, optJob.getJobType());
+        inFlightJobs_.emplace(std::move(jobId), std::move(optJob));
       }
     }
   }
   for (auto& jd : jds) {
-    jd.setStatus(cpp2::JobStatus::QUEUE, true);
+    jd.setStatus(cpp2::JobStatus::FAILED, true);
```
critical27, line 105: Why you prefer FAILED instead of FAILED?

liwenhui-soul: ？

src/meta/processors/job/JobManager.cpp
liwenhui-soul: the guarantee is at jobmanager.cpp:832

src/meta/processors/job/JobManager.cpp
liwenhui-soul: this is the rule 1

jinyingsunny:
checked ent-3.3.0

PR #4522, opened 2022-08-15
Fix some pointer partial specializations are missing in hash and equal_to in Value and Row

critical27:
## What type of PR is this?
- [X] bug
- [ ] feature
- [X] enhancement

#### Description:
Some pointer type partial specializations are missing in Value and Row, the most weird thing I encountered is that:
```
std::unordered_set<Value*> this_can_dedup;
// however
std::unordered_set<const Value*> this_can_not_dedup;
```
If you use a type as template argument different from the template parameter in partial specialization, you will use the primary template, aka, the one defined in STL:
```
  template<typename _Tp>
    struct equal_to : public binary_function<_Tp, _Tp, bool>
    {
      _GLIBCXX14_CONSTEXPR
      bool
      operator()(const _Tp& __x, const _Tp& __y) const
      { return __x == __y; }
    };
```

And, the dedup will failed surprisingly...

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....


jievince (approved):

Shylock-Hg (approved):

panda-sheep (commented):
Good job

Issue #4385, opened 2022-07-07
Enhance property pruning rule


czpmango:
At present, our  property pruning optimization only considers the symbols between operators without analyzing specific expressions, so some statements that can be optimized may be ignored. The following statement can prune all properties.
```
MATCH (v1)-[e:like*1..5]->(v2)
WHERE id(v1) == "Tim Duncan"
RETURN count(v2)
```
This optimization can not only reduce rpc overhead but also reduce memory pressure during graphd computation.

czpmango:
Some simple aggregations(such count(*)) can be considered pushed down to the storage layer(Scan or Traverse operator).
This may greatly improve the performance of statements like the above.

PR #4523, opened 2022-08-15
Prune properties

nevermore3:
## What type of PR is this?
- [ ] bug
- [ ] feature
- [x] enhancement

#### Issue(s) number: 
close https://github.com/vesoft-inc/nebula/issues/4385

#### Description:
1、attributes are collected from the root node, and the attributes required by each node are the union of itself and the parent node
2、when encountering UNION and BicaresionProduct operators, the left and right branches collect attributes independently Except for the join operator
3、when the vertexProp of appendVertices is null, it can be returned directly without going through RPC
4、delete _src from edge properties， instead of VID

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....


src/graph/context/Iterator.cpp
```diff
@@ -535,7 +537,7 @@ Value GetNeighborsIter::getEdge() const {
   DCHECK_EQ(edgeNamePropList.size(), propList.size());
   for (size_t i = 0; i < propList.size(); ++i) {
     auto propName = edgeNamePropList[i];
-    if (propName == kSrc || propName == kDst || propName == kRank || propName == kType) {
```
Shylock-Hg, line 538: Why change this?

src/graph/executor/query/AppendVerticesExecutor.cpp
```diff
@@ -26,10 +26,12 @@ StatusOr<DataSet> AppendVerticesExecutor::buildRequestDataSet(const AppendVertic
 
 folly::Future<Status> AppendVerticesExecutor::appendVertices() {
   SCOPED_TIMER(&execTime_);
-
   auto *av = asNode<AppendVertices>(node());
-  StorageClient *storageClient = qctx()->getStorageClient();
+  if (FLAGS_optimize_appendvertices && av != nullptr && av->props() == nullptr) {
```
Shylock-Hg, line 30: This will lead error if dangle edges presents.

tests/tck/features/optimizer/PrunePropertiesRule.feature
```diff
@@ -21,12 +21,12 @@ Feature: Prune Properties rule
...
-      | 4  | AppendVertices | 3            | {  "props": "[{\"props\":[\"age\"],\"tagId\":9}]" }                                                        |
-      | 3  | Traverse       | 7            | {  "vertexProps": "", "edgeProps": "[{\"props\":[\"_dst\", \"_rank\", \"_type\", \"_src\"],\"type\":3}]" } |
-      | 7  | IndexScan      | 2            |                                                                                                            |
-      | 2  | Start          |              |                                                                                                            |
+      | id | name           | dependencies | operator info                                                                                    |
+      | 8  | Project        | 4            |                                                                                                  |
+      | 4  | AppendVertices | 3            | {  "props": "[{\"props\":[\"age\"],\"tagId\":9}]" }                                              |
+      | 3  | Traverse       | 7            | {  "vertexProps": "", "edgeProps": "[{\"props\":[\"_dst\", \"_rank\", \"_type\"],\"type\":3}]" } |
```
Shylock-Hg, line 27: I think we need `_dst` of edge only in this case.

src/graph/executor/query/AppendVerticesExecutor.cpp
nevermore3: will be turned on when testing

tests/tck/features/optimizer/PrunePropertiesRule.feature
nevermore3: In order to distinguish the same edge in the path, _type and _rank also need to be

Shylock-Hg: ok

src/graph/planner/ngql/PathPlanner.cpp
```diff
@@ -355,23 +356,24 @@ PlanNode* PathPlanner::buildEdgePlan(PlanNode* dep, const std::string& input) {
   // col 0 of the project->output() is [edge...]
   auto* unwindExpr = ColumnExpression::make(pool, 0);
   auto* unwind = Unwind::make(qctx, project, unwindExpr);
+  unwind->setFromPipe(true);
   unwind->setColNames({"edges"});
 
   // extract src from edge
   auto srcArgs = ArgumentList::make(pool);
-  srcArgs->addArgument(ColumnExpression::make(pool, 1));
```
Shylock-Hg, line 362: Maybe we could use constant variable instead of number literal.

Shylock-Hg: Or property name directly. And optimize hash map lookup in inner implementation.

nevermore3: there is a detailed note above, 0 for column 0

Shylock-Hg: And user could turn off `properties pruning`, so the column index number will lead to mismatched property.

nevermore3: this is nothing to do with properties pruning, it is just an optimization of find path, no need to output additional data

src/graph/executor/query/AppendVerticesExecutor.cpp
Shylock-Hg: If so, I think you could eliminate AppendVertices operator in optimizer.

czpmango: We don't have test cases for dangling edges now. If this operator returns directly, the behavior is different from before.

czpmango: I also agree with @Shylock-Hg  that if property Pruning is a deterministic optimization, isn't it better practice to handle it in the optimizer?

src/graph/visitor/PropertyTrackerVisitor.cpp
```diff
@@ -193,6 +220,15 @@ void PropertyTrackerVisitor::visit(FunctionCallExpression *expr) {
...
 
+void PropertyTrackerVisitor::visit(AggregateExpression *expr) {
+  auto funName = expr->name();
+  std::transform(funName.begin(), funName.end(), funName.begin(), ::tolower);
+  if (funName == "count") {
+    return;
+  }
+  expr->arg()->accept(this);
+}
+
```
czpmango, line 231: Why return directly?
`count(*)` is different from `count(v.player.age)`.

nevermore3: good catch

src/graph/executor/query/AppendVerticesExecutor.cpp
nevermore3: there will be a problem of dangle edges. Currently it is manually controlled, and it is usually turned off. It is only turned on when the LDBC test is performed

czpmango:
https://github.com/vesoft-inc/nebula/blob/63179178d335abe73e7987d3c2b78fa2ae422f96/src/graph/optimizer/rule/EliminateAppendVerticesRule.cpp#L53-L55

I found that there is still room for optimization in these lines of code. If you are optimizing similar statements `MATCH (v:player)-[]-(n) RETUNR v.player.name`, it may help.

Shylock-Hg:
> https://github.com/vesoft-inc/nebula/blob/63179178d335abe73e7987d3c2b78fa2ae422f96/src/graph/optimizer/rule/EliminateAppendVerticesRule.cpp#L53-L55
> 
> I found that there is still room for optimization in these lines of code. If you are optimizing similar statements `MATCH (v:player)-[]-(n) RETUNR v.player.name`, it may help.

Yes, could prune properties before optimizer.

src/graph/executor/query/AppendVerticesExecutor.cpp
```diff
@@ -69,6 +71,39 @@ folly::Future<Status> AppendVerticesExecutor::appendVertices() {
       });
 }
 
+Status AppendVerticesExecutor::handleNullProp(const AppendVertices *av) {
```
Shylock-Hg, line 74: You could eliminate this operator in optimizer based on prune properties before optimizer rules.

jievince (approved):
Good job

Shylock-Hg (approved):

src/graph/visitor/PropertyTrackerVisitor.cpp
```diff
@@ -147,41 +189,26 @@ void PropertyTrackerVisitor::visit(AttributeExpression *expr) {
...
       break;
   }
 }
 
 void PropertyTrackerVisitor::visit(FunctionCallExpression *expr) {
-  static const std::unordered_set<std::string> kVertexIgnoreFuncs = {"id"};
-  static const std::unordered_set<std::string> kEdgeIgnoreFuncs = {
-      "src", "dst", "type", "typeid", "rank"};
+  static const std::unordered_set<std::string> ignoreFuncs = {
+      "src", "dst", "type", "typeid", "id", "rank", "length"};
```
czpmango, line 207: That is a bad implementation and bad fix. 
We should not ignore the inner expression behavior. It's really ambiguous.

Issue #4525, opened 2022-08-16
 Optimize GO statement join performance


dbacyj:
**Introduction**

nGQL:
```

LOOKUP ON COVID19 YIELD id(vertex) as id | \
GO FROM $-.id OVER visit \
WHERE (visit.leaveTime - visit.arriveTime) >= 600000 \
YIELD visit._src AS src_id, visit._dst AS dst_id, visit.arriveTime AS arriveTime, visit.leaveTime AS leaveTime, visit.totalVisitTime AS totalVisitTime | \
GO FROM $-.dst_id OVER visit REVERSELY \
WHERE ((visit.arriveTime < $-.arriveTime AND visit.leaveTime - $-.arriveTime >= 600000) \
OR (visit.arriveTime >= $-.arriveTime AND visit.leaveTime <= $-.leaveTime AND visit.leaveTime - visit.arriveTime >= 600000) \
OR (visit.arriveTime >= $-.arriveTime AND visit.leaveTime > $-.leaveTime AND $-.leaveTime - visit.arriveTime >= 600000)) \
AND ($-.totalVisitTime >=108000000 OR visit.totalVisitTime >=108000000) \YIELD $-.src_id as src, visit._dst as dst;


```

profile:
![image](https://user-images.githubusercontent.com/72034305/184788025-f5061bd5-2e0e-4c4b-bd50-172ab911121a.png)

stats:
![image](https://user-images.githubusercontent.com/72034305/184788127-6700b717-1f7e-4fb9-b5d3-e23e0dbf21ec.png)

PR #4531, opened 2022-08-17
GetDstBySrc

critical27:
## What type of PR is this?
- [ ] bug
- [X] feature
- [X] enhancement

#### Issue(s) number: 
Subtask of #4525

#### Description:
Add a new rpc interface called GetDstBySrc, the difference between it and GetNeighbors is that it **only** return dst of the edge in each row (even the srcId is not returned), see the interface in storage.thrift for details. And the dst has been deduped before returned to graphd to minimize memory cost.

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....

### DIFF
--- a/src/storage/exec/GetDstBySrcNode.h
+++ b/src/storage/exec/GetDstBySrcNode.h
@@ -18,7 +18,7 @@ class GetDstBySrcNode : public QueryNode<VertexID> {
   GetDstBySrcNode(RuntimeContext* context,
                   const std::vector<SingleEdgeNode*>& edgeNodes,
                   EdgeContext* edgeContext,
-                  nebula::List* result)
+                  nebula::DataSet* result)
       : context_(context), edgeNodes_(edgeNodes), edgeContext_(edgeContext), result_(result) {
     name_ = "GetDstBySrcNode";
   }
@@ -55,12 +55,14 @@ class GetDstBySrcNode : public QueryNode<VertexID> {
       auto props = context_->props_;
       DCHECK_EQ(props->size(), 1);
 
+      nebula::List list;
       // collect props need to return
       if (!QueryUtils::collectEdgeProps(
-               key, context_->vIdLen(), context_->isIntId(), reader, props, *result_)
+               key, context_->vIdLen(), context_->isIntId(), reader, props, list)
                .ok()) {
         return nebula::cpp2::ErrorCode::E_EDGE_PROP_NOT_FOUND;
       }
+      result_->rows.emplace_back(std::move(list));
     }
     return nebula::cpp2::ErrorCode::SUCCEEDED;
   }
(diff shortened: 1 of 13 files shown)


jievince (approved):
LGTM

src/storage/exec/GetDstBySrcNode.h
```diff
@@ -0,0 +1,95 @@
...
+      if (type != context_->edgeType_) {
+        // update info when edgeType changes while iterating over different edge types
+        setCurrentEdgeInfo(type);
+      }
+      auto key = iter_->key();
+      auto reader = iter_->reader();
+      auto props = context_->props_;
+      DCHECK_EQ(props->size(), 1);
+
+      nebula::List list;
```
Shylock-Hg, line 58: nebula::Row is more clear.

src/storage/exec/MultiTagNode.h
```diff
@@ -48,9 +48,6 @@ class MultiTagNode : public IterateNode<VertexID> {
 
     // add result of each tag node to tagResult
     for (auto* tagNode : tagNodes_) {
```
Shylock-Hg, line 50: Why remove this?

pengweisong (approved):
Good job

src/storage/exec/MultiTagNode.h
wenhaocs: The same Q

wenhaocs:
Generally, can you reuse `getNeighbors` but make `kVid` optional, ie. set `kVid` only when it's specified in req?

jievince:
> ee the interfac

They are different in memory occupation. See https://confluence.nebula-graph.io/pages/viewpage.action?pageId=47392379

wenhaocs:
> > ee the interfac
> 
> They are different in memory occupation. See https://confluence.nebula-graph.io/pages/viewpage.action?pageId=47392379

The response seems like the same:
image

The only difference is in request. But you can mark only returning kDst in the request.

jievince:
image
GetNeighbors takes up more memory.

src/storage/exec/MultiTagNode.h
critical27: They have been check on a higher level node, this is redundant. All the check only exists in output node, just to keep unify with others

critical27:
> Generally, can you reuse `getNeighbors` but make `kVid` optional, ie. set `kVid` only when it's specified in req?

We could indeed, there are several slight difference between them:
1. return format as @jievince mentioned
2. the StoragePlan is simplified for GetDstBySrc
3. the output has been deduped in GetDstBySrc, which we don't do it in getNeighbors
4. most of all, this interface is used for the interim result for the first (N - 1) steps in `go N steps`, if we reuse GetNeighbors, the impact will be not be trival

Shylock-Hg (approved):

PR #4550, opened 2022-08-17
Fix stats log print to console when we start process.

Milittle:
we before setupLogging, init our stats, when we adjust high level verbose
log. it will be output logs to console.

## What type of PR is this?
- [x] bug
- [ ] feature
- [ ] enhancement

#### Description:

we before setupLogging, init our stats, when we adjust high level verbose
log. it will be output logs to console.

## How do you solve it?

I adjust the init stats order from main process in Metad and Storaged

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....


panda-sheep (approved):
Good catch, thanks for your contribution

jievince (approved):

critical27 (approved):

Issue #4551, opened 2022-08-18
lookup statement cause graph crash


czpmango:
**Describe the bug (__required__)**
lookup statement cause graph crash when there is an and expression with an in expression in it.

**Your Environments (__required__)**
nebula 3.2 

**How To Reproduce(__required__)**
```
lookup on like where like.name in ["like1"] and like.score>2332 yield properties(edge)
```
**Expected behavior**
Not crash.

PR #4552, opened 2022-08-18
fix lookup crash

czpmango:
## What type of PR is this?
- [x] bug
- [ ] feature
- [ ] enhancement

#### Issue(s) number: 
#4551 
## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....


tests/tck/features/lookup/LookUp.feature
```diff
@@ -184,6 +184,13 @@ Feature: LookUpTest_Vid_String
     Then the result should be, in any order:
       | id    |
       | "220" |
+    When executing query:
+      """
+      LOOKUP ON lookup_tag_2 WHERE lookup_tag_2.col3 in [100.5] AND lookup_tag_2.col2 == 100 YIELD id(vertex) as id
+      """
+    Then the result should be, in any order:
+      | id    |
+      | "220" |
```
Aiee, line 193: What's the difference between this and 
```
    When profiling query:
      """
      LOOKUP ON player WHERE player.age IN [40] AND player.name == "Kobe Bryant" YIELD id(vertex) as id, player.age
      """
      | id            | player.age |
      | "Kobe Bryant" | 40         |
    And the execution plan should be:
      | id | name               | dependencies | operator info |
      | 3  | Project            | 4            |               |
      | 4  | TagIndexPrefixScan | 0            |               |
      | 0  | Start              |              |               |
```
Check TagIndexFullScan.feature

src/graph/optimizer/rule/OptimizeEdgeIndexScanByFilterRule.cpp
```diff
@@ -118,7 +118,7 @@ StatusOr<TransformResult> OptimizeEdgeIndexScanByFilterRule::transform(
         if (static_cast<ListExpression*>(inExpr->right())->size() > 1) {
           return TransformResult::noTransform();
         } else {
-          transformedExpr = graph::ExpressionUtils::rewriteInExpr(condition);
+          transformedExpr = graph::ExpressionUtils::rewriteInExpr(inExpr);
```
Aiee, line 121: This branch won't be reached for the case you added below because there is only 1 element in the list of the IN expr

tests/tck/features/lookup/LookUp.feature
czpmango: I wrote the wrong test case, I will change it.

jievince (approved):

nevermore3 (approved):

Issue #4565, opened 2022-08-22
Storaged crashes due to accessing non-existent property


czpmango:
**Describe the bug (__required__)**
Storaged crashes due to accessing non-existent property.

**Environments (__required__)**
nebula 3.2

**How To Reproduce(__required__)**
```
(root@nebula) [test]> create tag V(id int, name string)
Execution succeeded (time spent 1493/1783 us)

(root@nebula) [test]> desc tag V
+--------+----------+-------+---------+---------+
| Field  | Type     | Null  | Default | Comment |
+--------+----------+-------+---------+---------+
| "id"   | "int64"  | "YES" |         |         |
| "name" | "string" | "YES" |         |         |
+--------+----------+-------+---------+---------+
Got 2 rows (time spent 1298/1659 us)

(root@nebula) [test]> CREATE tag index v_id_index on V(id)
Execution succeeded (time spent 1493/1783 us)

(root@nebula) [test]> match (v:V{idx:1})-[]->() return v
+---+
| v |
+---+
+---+
Empty set (time spent 4459/4757 us)

(root@nebula) [test]> rebuild tag index v_id_index
+------------+
| New Job Id |
+------------+
| 16         |
+------------+
Got 1 rows (time spent 1529/1848 us)

(root@nebula) [test]> match (v:V{idx:1})-[]->() return v
[ERROR (-1005)]: Storage Error: part: 11, error: E_RPC_FAILURE(-3).

```

**Expected behavior**
Not crash.

critical27:
Which version do you use? 3.2 community version do not reproduce. I am sure that ent version has this problem because there is an extra filter push down rule called `PushFilterDownAppendVerticesRule`.

czpmango:
I reproduced that on commit `a2de29784`.
I only considered the `first step filter` of the `traverse` node and the `filter` of the `AppendVertices`. Will there be other scenarios to break the test?

critical27:
I'll try it, sync later

critical27:
I can't reproduce with the commit you mentioned. But something must go wrong here, there is a filter in the explained plan, but when storage receive the GetNeighbors request, there is no filter at all.
```
(root@nebula) [nba]> explain match (v:player{invalid_property:1})-[]->() return v
Execution succeeded (time spent 2046/2843 us)

Execution Plan (optimize time 611 us)

-----+----------------+--------------+----------------+----------------------------------------------
| id | name           | dependencies | profiling data | operator info                               |
[...]

PR #4567, opened 2022-08-22
Fix storage crash due to accessing non-existent property

czpmango:
## What type of PR is this?
- [x] bug
- [ ] feature
- [ ] enhancement

#### Issue(s) number: 
https://github.com/vesoft-inc/nebula/issues/4565

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....


critical27:
Thx for help, use https://github.com/vesoft-inc/nebula/pull/4568 will be more appropriate. Actually, there are more than one case here.

czpmango:
> Thx for help, use #4568 will be more appropriate. Actually, there are more than one case here.

ACK.

PR #4568, opened 2022-08-22
fix invalid filter in GetProp make storage crashed

critical27:
## What type of PR is this?
- [X] bug
- [ ] feature
- [ ] enhancement

#### Issue(s) number: 

https://github.com/vesoft-inc/nebula-ent/issues/1165
#4565 

#### Description:

Invalid filter will make storage crash, this only happens in enterprise version with additional filter push down rule.

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....

### DIFF
--- a/src/storage/exec/QueryUtils.h
+++ b/src/storage/exec/QueryUtils.h
@@ -73,6 +73,9 @@ class QueryUtils final {
 
       if (nullType == NullType::UNKNOWN_PROP) {
         VLOG(1) << "Fail to read prop " << propName;
+        if (!field) {
+          return value;
+        }
         if (field->hasDefault()) {
           DefaultValueContext expCtx;
           ObjectPool pool;
--- a/src/storage/test/GetPropTest.cpp
+++ b/src/storage/test/GetPropTest.cpp
@@ -802,6 +802,57 @@ TEST(GetPropTest, FilterTest) {
     // }
     ASSERT_EQ(expected, *resp.props_ref());
   }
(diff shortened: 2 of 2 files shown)


wenhaocs (approved):
LGTM

panda-sheep (approved):

SuperYoko (approved):

src/storage/exec/QueryUtils.h
```diff
@@ -73,6 +73,9 @@ class QueryUtils final {
 
       if (nullType == NullType::UNKNOWN_PROP) {
         VLOG(1) << "Fail to read prop " << propName;
+        if (!field) {
```
czpmango, line 76: Why not just return null outside the if block？

critical27: It is null here~ NullType::UNKNOWN_PROP

czpmango: IMO, NullType::UNKNOWN_PROP is not necessary, which can lead to more problems due to graphd's unreasonable handling of runtime errors.
For example, filter operator execution will throw a runtime error if UNKNOWN_PROP is fed to it. 
https://github.com/vesoft-inc/nebula/blob/7f185c969f33ac7bc85a57de0ec3d7620253c24b/src/graph/executor/query/FilterExecutor.cpp#L67-L69

PR #4570, opened 2022-08-23
unify pushed down filter behavior

critical27:
## What type of PR is this?
- [X] bug
- [ ] feature
- [ ] enhancement

#### Issue(s) number: 
https://github.com/vesoft-inc/nebula/issues/4565

#### Description:

Follow-up of #4568, the actual problem arises from #4270, if a filter contains an invalid property, it bypass property check because openCypher regard property not existed as null. However, when storage try to read it in `QueryUtils::readValue`, it does not check whether the field exists. (Before #4270, all field must exists)

## How do you solve it?
#4568 has check the filed existence in `QueryUtils::readValue`, just remove the property existence check in `checkExp`, and the redundant `readValue` is removed to unify usage.

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....

### DIFF
--- a/src/storage/query/QueryBaseProcessor-inl.h
+++ b/src/storage/query/QueryBaseProcessor-inl.h
@@ -152,7 +152,7 @@ nebula::cpp2::ErrorCode QueryBaseProcessor<REQ, RESP>::buildFilter(
     if (filter_ == nullptr) {
       return nebula::cpp2::ErrorCode::E_INVALID_FILTER;
     }
-    return checkExp(filter_, false, true, false, true);
+    return checkExp(filter_, false, true, false);
   }
   return nebula::cpp2::ErrorCode::SUCCEEDED;
 }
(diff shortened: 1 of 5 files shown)


src/storage/query/QueryBaseProcessor-inl.h
```diff
@@ -260,8 +260,10 @@ nebula::cpp2::ErrorCode QueryBaseProcessor<REQ, RESP>::getSpaceEdgeSchema() {
 }
 
 template <typename REQ, typename RESP>
-nebula::cpp2::ErrorCode QueryBaseProcessor<REQ, RESP>::checkExp(
-    const Expression* exp, bool returned, bool filtered, bool updated, bool allowNoexistentProp) {
```
Shylock-Hg, line 264: Update will report error when property is not existed.

SuperYoko (approved):

pengweisong (approved):

src/storage/query/QueryBaseProcessor-inl.h
critical27: Em, indeed, good catch... Has graphd verified it or not? If not, maybe we can't delete it simply

critical27: What if I handle separately in UpdateProcessor, WDYT?

Shylock-Hg: As I known, it won't.

Shylock-Hg: What's the mistake of origin implementation?

critical27: Ok, I will check it and keep the checkExp as simple as possible

critical27: Before https://github.com/vesoft-inc/nebula/pull/4270, all field must exists in expression. This has make storage hard to handle property not existent... #4565 and https://github.com/vesoft-inc/nebula-ent/issues/1165

PR #4574, opened 2022-08-23
Finish the bug in validator with good Executor Framework

peter-rich:
## What type of PR is this?
- [ ] bug
- [x] feature
- [x] enhancement

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....

### DIFF
--- a/src/graph/executor/Executor.cpp
+++ b/src/graph/executor/Executor.cpp
@@ -51,6 +51,7 @@
 #include "graph/executor/algo/ProduceAllPathsExecutor.h"
 #include "graph/executor/algo/ShortestPathExecutor.h"
 #include "graph/executor/algo/SubgraphExecutor.h"
+#include "graph/executor/algo/IsomorExecutor.h"
 #include "graph/executor/logic/ArgumentExecutor.h"
 #include "graph/executor/logic/LoopExecutor.h"
 #include "graph/executor/logic/PassThroughExecutor.h"
@@ -460,6 +461,10 @@ Executor *Executor::makeExecutor(QueryContext *qctx, const PlanNode *node) {
     case PlanNode::Kind::kSubgraph: {
       return pool->makeAndAdd<SubgraphExecutor>(node, qctx);
     }
+    case PlanNode::Kind::kIsomor: {
+      //  return pool->makeAndAdd<SubgraphExecutor>(node, qctx);
+      return pool->makeAndAdd<IsomorExecutor>(node, qctx);
+    }
     case PlanNode::Kind::kAddHosts: {
       return pool->makeAndAdd<AddHostsExecutor>(node, qctx);
     }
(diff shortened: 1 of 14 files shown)


src/graph/executor/Executor.cpp
```diff
@@ -550,9 +555,9 @@ Executor *Executor::makeExecutor(QueryContext *qctx, const PlanNode *node) {
     case PlanNode::Kind::kShortestPath: {
       return pool->makeAndAdd<ShortestPathExecutor>(node, qctx);
     }
-    // case PlanNode::Kind::kIsomor: {
-    //   return pool->makeAndAdd<SomeIsomorExecutor>(node, qctx);
-    // }
+    //  case PlanNode::Kind::kIsomor: {
```
wenhaocs, line 558: I think you can delete these lines.

src/graph/executor/Executor.h
```diff
@@ -12,6 +12,7 @@
 #include "common/time/ScopedTimer.h"
 #include "graph/context/ExecutionContext.h"
 #include "graph/context/QueryContext.h"
+
```
wenhaocs, line 15: extra line

src/graph/service/PermissionCheck.cpp
```diff
@@ -77,6 +77,7 @@ namespace graph {
     case Sentence::Kind::kCreateEdge:
     case Sentence::Kind::kAlterEdge:
     case Sentence::Kind::kDropTag:
+    case Sentence::Kind::kIsomor:
```
wenhaocs, line 80: The handling function for these cases are `canWriteSchema`.  You don't need to write schema. How about trying to add the case of Isomor around line 130 so `canReadSchemaOrData` is the handling function?

src/graph/validator/IsomorValidator.cpp
```diff
@@ -12,16 +12,19 @@ namespace graph {
...
+  NG_RETURN_IF_ERROR(validateTag(fSentence->graphs()));
   return Status::OK();
 }
 // Check validity of tags specified in sentence
 Status IsomorValidator::validateTag(const NameLabelList *nameLabels) {
-  if (nameLabels == nullptr) {
-// Wthether Found Tag in the storage?  --> need the coorperation of the storage section.
-  } else {
-  }
+  // Wthether Found Tag in the storage?  --> need the coorperation of the storage section.
```
wenhaocs, line 20: Is this comment outdated?

src/graph/validator/IsomorValidator.cpp
```diff
@@ -12,16 +12,19 @@ namespace graph {
...
+  auto graphs = nameLabels->labels();
+
+  // The first graph is data graph and the second graph is the query graph
```
wenhaocs, line 23: Is the sequence here wrong? Please double check.

src/graph/validator/IsomorValidator.cpp
```diff
@@ -12,16 +12,19 @@ namespace graph {
...
+  fetchCtx_->querySpace = qctx_->schemaMng()->toGraphSpaceID(*graphs[0]);
+  fetchCtx_->dataSpace = qctx_->schemaMng()->toGraphSpaceID(*graphs[1]);
   return Status::OK();
 }
+
```
wenhaocs, line 28: extra line

src/graph/executor/algo/IsomorExecutor.cpp
```diff
@@ -0,0 +1,88 @@
...
+
+#include "graph/planner/plan/Algo.h"
+
+namespace nebula {
+namespace graph {
+
+folly::Future<Status> IsomorExecutor::execute() {
+   // TODO: Replace the following codes with subgraph matching. Return type.
+   SCOPED_TIMER(&execTime_);
+   auto* subgraph = asNode<Subgraph>(node());
```
wenhaocs, line 15: Just to make sure you are going to replace `Subgraph` to `Isomor` plan node right?

peter-rich: This is the old version changed from the SubgraphExecutor which I need to change later against with adding the subgraph.cpp (subgraph matching algorithm).

PR #4578, opened 2022-08-23
fix scan vertex/edge do not handle ttl

critical27:
## What type of PR is this?
- [X] bug
- [ ] feature
- [ ] enhancement

#### Issue(s) number: 
https://github.com/vesoft-inc/nebula-ent/issues/1162

#### Description:
As title, scan vertex/edge does not handle ttl, and some of code in the community/ent version are not unified.

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....


src/storage/exec/ScanNode.h
```diff
@@ -262,9 +274,16 @@ class ScanEdgePropNode : public QueryNode<Cursor> {
     return nebula::cpp2::ErrorCode::SUCCEEDED;
   }
 
-  void collectOneRow(bool isIntId, std::size_t vIdLen) {
+  nebula::cpp2::ErrorCode collectOneRow(bool isIntId, std::size_t vIdLen) {
     List row;
     nebula::cpp2::ErrorCode ret = nebula::cpp2::ErrorCode::SUCCEEDED;
+    // Usually there is only one edge node, when all of the egdeNodes are invalid (e.g. ttl
+    // expired), just skip the row. If we don't skip it, there will be a whole line of empty value.
```
Shylock-Hg, line 281: In fact, we should return one only if vertex key exists.

critical27: Where did you handle this in scan before? BTW, the vertex key will be removed later

Shylock-Hg: When introducing this, there is no vertex key.

Shylock-Hg: It's a bug delayed https://github.com/vesoft-inc/nebula/issues/3688

pengweisong (approved):

src/storage/query/ScanVertexProcessor.cpp
```diff
@@ -75,6 +75,7 @@ nebula::cpp2::ErrorCode ScanVertexProcessor::checkAndBuildContexts(
       return nullptr;
     }
   });
+  buildTagTTLInfo();
```
panda-sheep, line 78: 👍

panda-sheep (approved):
Good job

PR #4601, opened 2022-08-31
Better error msg for validity check of geography

jievince:
## What type of PR is this?
- [ ] bug
- [ ] feature
- [ ] enhancement

#### Description:
When geography dataum is found to be illegal, print the specific cause of the error to facilitate troubleshooting.
eg.
```
E20220902 03:24:24.550307 350836 FunctionManager.cpp:2440] ST_GeogFromText error: The geography LINESTRING(0 1, 2 3) built from wkt LINESTRING(0 1, 2 3) is invalid, error: FindValidationError, S2 Errorcode: 1, message: Vertex 0 is not unit length
```

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....

### DIFF
--- a/src/common/function/FunctionManager.cpp
+++ b/src/common/function/FunctionManager.cpp
@@ -2511,7 +2511,8 @@ FunctionManager::FunctionManager() {
       if (!args[0].get().isGeography()) {
         return Value::kNullBadType;
       }
-      return args[0].get().getGeography().isValid();
+      auto status = args[0].get().getGeography().isValid();
+      return status.ok() ? true : false;
     };
   }
   // geo predicates
--- a/src/common/function/FunctionManager.cpp
+++ b/src/common/function/FunctionManager.cpp
@@ -2512,7 +2512,10 @@ FunctionManager::FunctionManager() {
         return Value::kNullBadType;
       }
       auto status = args[0].get().getGeography().isValid();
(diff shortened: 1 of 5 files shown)


Aiee (dismissed):

src/common/function/FunctionManager.cpp
```diff
@@ -2511,7 +2511,8 @@ FunctionManager::FunctionManager() {
       if (!args[0].get().isGeography()) {
         return Value::kNullBadType;
       }
-      return args[0].get().getGeography().isValid();
+      auto status = args[0].get().getGeography().isValid();
+      return status.ok() ? true : false;
```
Shylock-Hg, line 2515: It's seems not affect.

czpmango (approved):
LGTM

src/common/function/FunctionManager.cpp
czpmango, line 2515: return status.ok()

jievince: Missed it here, thanks

jievince: Fixed

Aiee (approved):

czpmango (approved):

PR #4605, opened 2022-09-01
fix filter executor

czpmango:
## What type of PR is this?
- [ ] bug
- [ ] feature
- [ ] enhancement

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....


Shylock-Hg:
Please add a comment.

czpmango:
> Please add a comment.

I explained in the description of pr why I need to fix it this way, as for the logic of fixing the code is relatively simple.

Shylock-Hg:
I think you could simply remove `modify` related operations from `Iterator` of DataSet, and keep `Iterator` is immutable. For     `Filter` you could copy valid rows to new DataSet, and same for `Limit` or other operators need modify input DataSet.

czpmango:
> I think you could simply remove `modify` related operations from `Iterator` of DataSet, and keep `Iterator` is immutable. For `Filter` you could copy valid rows to new DataSet, and same for `Limit` or other operators need modify input DataSet.

Some of the previous revisions considered performance issues more comprehensively, and it is feasible to revert this refactoring if the previous considerations can be covered.
I found some previous pr for your reference:
https://github.com/vesoft-inc/nebula-graph/pull/789
https://github.com/vesoft-inc/nebula-graph/pull/757
https://github.com/vesoft-inc/nebula-graph/pull/791

The current pr fix is feasible, and the code changes and performance impact are manageable.

jievince (dismissed):

src/graph/context/Result.h
```diff
@@ -63,6 +63,15 @@ class Result final {
     }
   }
 
+  std::vector<std::string> getColNames() const {
```
Aiee, line 66: `const std::vector<std::string>& getColNames() const`

src/graph/planner/plan/PlanNode.cpp
```diff
@@ -330,6 +330,17 @@ void PlanNode::readVariable(Variable* varPtr) {
...
 
+bool PlanNode::isConcurrencySafeInputVar() const {
+  if (inputVars_.size() != 1) {
+    return true;
+  }
+  auto inputVar = inputVars_.front();
+  if (inputVar->readBy.size() < 2) {
+    return true;
+  }
+  return false;
```
Aiee, line 341: `return inputVar->readBy.size() < 2`

src/graph/context/Result.h
czpmango: NCK.
Cannot return reference if data type is not dataset.

src/graph/planner/plan/PlanNode.cpp
czpmango: ACK.

czpmango: Already delete this interface and replace by Executor::movable.

Shylock-Hg:
`Intersection` will call `unstableErase` too, I think you should check all operators call `erase` and `eraseRange`

czpmango:
> eraseRange

The `IntersectExecutor` will only consume data through erase function, and will not cause concurrent read-write conflicts.

Shylock-Hg:
> > Intersection will call unstableErase too, I think you should check all operators call erase and eraseRange
> 
> The `IntersectExecutor` will only consume data through erase function, and will not cause concurrent read-write conflicts.

If other operator reference input of Intersection, it's of course read-write conflicts.

czpmango:
> > Intersection will call unstableErase too, I think you should check all operators call erase and eraseRange
> 
> The `IntersectExecutor` will only consume data through erase function, and will not cause concurrent read-write conflicts.

This issue will only be reproduced by the arguement operator of multiple match statements.
It will only crash in the following code block:
https://github.com/vesoft-inc/nebula/blob/fc5735a133a21d1c1093c4735e0b3b5e9609856a/src/graph/executor/logic/ArgumentExecutor.cpp#L18-L36

Shylock-Hg (dismissed):

src/graph/executor/query/FilterExecutor.cpp
```diff
@@ -84,6 +84,10 @@ Status FilterExecutor::handleSingleJobFilter() {
   bool canMoveData = movable(inputVar);
   Result result = ectx_->getResult(inputVar);
   auto *iter = result.iterRef();
+  // Always reuse getNeighbors's dataset to avoid some go statement execution plan related issues
+  if (iter->isGetNeighborsIter()) {
+    canMoveData = true;
```
Shylock-Hg, line 89: Why add this?

czpmango: Reuse getNeighbors's dataset. Go statements do not cause write-read conflicts.

jievince (approved):

Aiee (approved):

PR #4607, opened 2022-09-01
Fix 1212

xtcyclist:
…due to found syantax errors.

## What type of PR is this?
- [X] bug
- [ ] feature
- [ ] enhancement

#### Issue(s) number: 
https://github.com/vesoft-inc/nebula-ent/issues/1212

#### Description:
When some expression has a wrong syntax, it is merged with others and fed into further evaluation, hitting a LOG(FATAL).

## How do you solve it?
Stop further folding expressions and return in FoldConstantExprVisitor::fold(), if status_ is already set to an error state.

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A
- [X] tck 

Affects:

## Release notes:

Fixing a crash bug caused by evaluating an expression with wrong a syntax.


Aiee (approved):
LGTM

jievince (approved):

PR #4616, opened 2022-09-05
fix list tag/edge/tagindex/edgeindex

panda-sheep:
## What type of PR is this?
- [x] bug
- [ ] feature
- [ ] enhancement

#### Description:
fix `list tag/edge/tagindex/edgeindex` logic to display data correctly. 
Although it was discovered by the enterprise version, the community version also has the same problem.

`list tag/edge/tagindex/edgeindex` logic has problems, resulting in display wrong data

The test case is as follows(used for enterprise version):
1) The master cluster is in the synchronous state at the beginning, do the following:
```
show sync status;
create tag player1(name string, age int, p3 int, p4 int);
stop sync
```

2) Suppose the slave cluster did the following before synchronizing:
```
create tag index fix on player1(p3);
desc tag index fix
show tag indexes
```
image

3） The master cluster does the following:
```
create tag index fix on player1(age);
desc tag index fix
show tag indexes
```
image

4）Then the master cluster starts synchronization
` restart sync`
At this time, the following results are displayed from the slave cluster
`desc tag index fix`
image
The result is displayed correctly

`show tag indexes`
image

Dirty data is displayed, and the expected result is consistent with the master cluster

close #1236

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....

### DIFF
--- a/src/common/utils/MetaKeyUtils.cpp
+++ b/src/common/utils/MetaKeyUtils.cpp
@@ -768,6 +768,16 @@ std::string MetaKeyUtils::indexTagKey(GraphSpaceID spaceId, const std::string& n
   return key;
 }
 
+std::string MetaKeyUtils::indexTagPrefix(GraphSpaceID spaceId) {
+  EntryType type = EntryType::TAG;
+  std::string key;
+  key.reserve(128);
+  key.append(kIndexTable.data(), kIndexTable.size())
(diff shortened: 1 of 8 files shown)


liwenhui-soul (approved):

wenhaocs (approved):
LGTM

PR #4623, opened 2022-09-06
fix startup service failed when log directory not exists and enable_breakpad

panda-sheep:
…reakPad

## What type of PR is this?
- [x] bug
- [ ] feature
- [ ] enhancement

#### Description:
In the released installation package, when `ENABLE_BREAKPAD` is enabled, if the `log` directory does not exist, it will fail to start `meta/stroage (storagelistener)`.

Many versions have this problem.

close https://github.com/vesoft-inc/nebula-ent/issues/1257

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....

### DIFF
--- a/src/daemons/MetaDaemon.cpp
+++ b/src/daemons/MetaDaemon.cpp
@@ -68,14 +68,6 @@ int main(int argc, char* argv[]) {
 
   Status status;
 
-#if defined(ENABLE_BREAKPAD)
-  status = setupBreakpad();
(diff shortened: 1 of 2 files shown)


wenhaocs (approved):
Well done

Aiee (approved):

Issue #4274, opened 2022-05-23
the metrics of "commit_log_latency_us" is abnormal. Too large


xigongdaEricyang:
![image](https://user-images.githubusercontent.com/9858560/169739887-857530fb-9366-42e5-b9f1-ff98df4407af.png)

liwenhui-soul:
what's this mean?

xigongdaEricyang:
in stats metrics， the value of commit_log_latency_us is 7841184878, and it means commit log takes almost 7841s. The value is too large, so we think the value is wrong

shanlai:
the same problem exists with the metrics of "append_wal_latency_us", please fix it together

jinyingsunny:
![image](https://user-images.githubusercontent.com/106510322/196430192-2fb3a8e9-6727-425a-bdd5-01cdb71eb35e.png)
checked ent-3.3 when write with speed of 1/s

PR #4627, opened 2022-09-07
fix issue 4274

cangfengzhs:
## What type of PR is this?
- [ ] bug
- [ ] feature
- [ ] enhancement

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....

### DIFF
--- a/src/kvstore/Part.cpp
+++ b/src/kvstore/Part.cpp
@@ -12,6 +12,7 @@
 #include "common/utils/Utils.h"
 #include "kvstore/LogEncoder.h"
 #include "kvstore/RocksEngineConfig.h"
+#include "kvstore/stats/KVStats.h"
 
 DEFINE_int32(cluster_id, 0, "A unique id for each cluster");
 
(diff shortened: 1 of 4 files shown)


jievince (approved):

SuperYoko (approved):

critical27:
Close #4274

src/kvstore/raftex/RaftPart.cpp
```diff
@@ -1063,7 +1063,6 @@ void RaftPart::processAppendLogResponses(const AppendLogResponses& resps,
       */
       auto [code, lastCommitId, lastCommitTerm] = commitLogs(std::move(walIt), true, true);
       if (code == nebula::cpp2::ErrorCode::SUCCEEDED) {
-        stats::StatsManager::addValue(kCommitLogLatencyUs, execTime_);
```
critical27, line 1066: Could add the `kCommitLogLatencyUs` in the scope from line 1054 to 1082?

src/kvstore/Part.cpp
```diff
@@ -214,7 +215,9 @@ std::tuple<nebula::cpp2::ErrorCode, LogID, TermID> Part::commitLogs(
     std::unique_ptr<LogIterator> iter, bool wait, bool needLock) {
   // We should apply any membership change which happens before start time. Because when we start
   // up, the peers comes from meta, has already contains all previous changes.
-  SCOPED_TIMER(&execTime_);
+  SCOPED_TIMER([](uint64_t elapsedTime) {
+    stats::StatsManager::addValue(kCommitLogLatencyUs, elapsedTime);
+  });
```
panda-sheep, line 220: Good

panda-sheep (approved):

src/kvstore/raftex/RaftPart.cpp
cangfengzhs: There are two places that will call commitlogs, so if you only add it to the place you said, then the commit latency will record less data. If you add it to both places, then the meaning of this metric will be unclear

critical27: How about add it in `commitLogs` itself, that could be unified.

cangfengzhs: In this PR, it is modified in this way

critical27: 👌 my bad...

critical27 (approved):

PR #4629, opened 2022-09-08
Rollback vertex key

critical27:
## What type of PR is this?
- [ ] bug
- [ ] feature
- [X] enhancement

#### Description:
Rollback the vertex key, upgrade tool is removed directly.

There are still two tck cases I think it is wrong:
1. GoYieldVertexEdge.feature: multi edges over all
2. Go.feature: multi edges over all

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....


nevermore3:
```
When executing query:
      """
      GO FROM "Paul Gasol" OVER *
      WHERE $$.player.name IS NOT NULL
      YIELD edge as e
      """
      | e                                                     |
      | [:like "Paul Gasol"->"Kobe Bryant" @0 {likeness: 90}] |
      | [:like "Paul Gasol"->"Marc Gasol" @0 {likeness: 99}]  |
```
Since getProp previously returned EMPTY rows, now it return nothing,  should use is NULL  as filter conditions

nevermore3 (dismissed):

nevermore3 (approved):

cangfengzhs (approved):

foesa-yang:
https://github.com/vesoft-inc/nebula-docs-cn/pull/2192

PR #4652, opened 2022-09-16
fix tagless

nevermore3:
## What type of PR is this?
- [ ] bug
- [x] feature
- [ ] enhancement

#### Description:

getProp interface's behave is 

before tagless
1、  if <vid + tag> exists in kvstore return value  else 2
2、 if vid exisits in tagless vids return value else return nothing

after tagless
1、if <vid+tag> exists in kvstore return value else return nothing

When attribute clipping, if there is no specified tag at the  vertex, all tags in the schema need to be added

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....


jievince (approved):

shanlai:
the following cases still fail:

Issue #4168, opened 2022-04-18
optional match filter bug


czpmango:
Incorrect data filtering when the where clause of optional match refers to variables declared in other match.

**Environment**
version 3.0

**How To Reproduce**

```
(root@nebula) [nba]> MATCH (m)-[]-(n) WHERE id(m)=="Tim Duncan" OPTIONAL MATCH (n)-->(v) WHERE v.player.age < m.player.age RETURN count(*) AS count
+-------+
| count |
+-------+
| 19    |
+-------+
Got 1 rows (time spent 50423/51550 us)
(root@nebula) [nba]> MATCH (m)-[]-(n) WHERE id(m)=="Tim Duncan" OPTIONAL MATCH (n)-->(v) WHERE v.player.age < m.player.age RETURN n,v
+-------------------------------------------------------------------+----------+
| n                                                                 | v        |
+-------------------------------------------------------------------+----------+
| ("Tiago Splitter" :player{age: 34, name: "Tiago Splitter"})       | __NULL__ |
| ("Marco Belinelli" :player{age: 32, name: "Marco Belinelli"})     | __NULL__ |
| ("Tony Parker" :player{age: 36, name: "Tony Parker"})             | __NULL__ |
| ("Tony Parker" :player{age: 36, name: "Tony Parker"})             | __NULL__ |
| ("Tony Parker" :player{age: 36, name: "Tony Parker"})             | __NULL__ |
| ("Tony Parker" :player{age: 36, name: "Tony Parker"})             | __NULL__ |
| ("Shaquile O'Neal" :player{age: 47, name: "Shaquile O'Neal"})     | __NULL__ |
| ("LaMarcus Aldridge" :player{age: 33, name: "LaMarcus Aldridge"}) | __NULL__ |
| ("LaMarcus Aldridge" :player{age: 33, name: "LaMarcus Aldridge"}) | __NULL__ |
| ("Boris Diaw" :player{age: 36, name: "Boris Diaw"})               | __NULL__ |
| ("Manu Ginobili" :player{age: 41, name: "Manu Ginobili"})         | __NULL__ |
| ("Manu Ginobili" :player{age: 41, name: "Manu Ginobili"})         | __NULL__ |
| ("Manu Ginobili" :player{age: 41, name: "Manu Ginobili"})         | __NULL__ |
| ("Manu Ginobili" :player{age: 41, name: "Manu Ginobili"})         | __NULL__ |
[...]

PR #4670, opened 2022-09-22
Fix incorrect result of optional match statement

czpmango:
## What type of PR is this?
- [x] bug
- [ ] feature
- [ ] enhancement

#### Issue(s) number: 
#4168 

## How do you solve it?
The fix is complex and may require refactoring the code related to how the execution plan was generated. 
Give the user an error message for now.

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....


jievince (approved):

Shylock-Hg:
Can we resolve it by apply filter later.

codesigner:
LGTM

codesigner (approved):

czpmango:
> Can we resolve it by apply filter later?

The current implementation is to apply filter after the left join, which is the crux of the problem.

Issue #4649, opened 2022-09-15
getNeighbors return error


nevermore3:
**Please check the FAQ documentation before raising an issue**

**Describe the bug (__required__)**

```
go from "Tony Parker" over like Where $^.player.age > 37 yield $^ as src

Filter $^.player.age > 37  pushdown to storage
```

getNeighbor's result is 
```
 _vid|_stats|_tag:team:_tag:name|_tag:player:_tag:age:name|_tag:bachelor:_tag:name:speciality|_edge:+like:_dst|_expr|
 "Tony Parker"|__EMPTY__|__EMPTY__|[10,36,"Tony Parker"]|__EMPTY__|__EMPTY__|__EMPTY__|

```

```
go from "Tony Parker" over like Where like.likeness > 100 yield $^ as src

Filter like.likeness>100 pushdown to storage
```

getNeighbor's result is 
```
 _vid|_stats|_tag:team:_tag:name|_tag:player:_tag:age:name|_tag:bachelor:_tag:name:speciality|_edge:+like:_dst|_expr|
 "Tony Parker"|__EMPTY__|__EMPTY__|[10,36,"Tony Parker"]|__EMPTY__|__EMPTY__|__EMPTY__|

```
There will be an error when taking the vertex (should not be return)

Expect
`go from "Tony Parker" over like Where $^.player.age > 36 yield $^ as src`

getNeighbor's result is 
```
 _vid|_stats|_tag:team:_tag:name|_tag:player:_tag:age:name|_tag:bachelor:_tag:name:speciality|_edge:+like:_dst|_expr|

```

**Your Environments (__required__)**

* OS: `uname -a`
* Compiler: `g++ --version` or `clang++ --version`
* CPU: `lscpu`
* Commit id (e.g. `a3ffc7d8`)

**How To Reproduce(__required__)**

1. Step 1
2. Step 2
3. Step 3

**Expected behavior**

**Additional context**

critical27:
Em... GetNeighbors's filter is designed for edge's filter only at the first time...
```
go from "Tony Parker" over like Where $^.player.age > 37 yield $^ as src
```
We could fix it in storage indeed. But how do we handle an expression when contains both tag and edge?

I think there is a way which won't introduce too much changes: only push down edge's filter...

nevermore3:
Logically, we should distinguish whether the tag is not satisfied the condition or the edge is not satisfied the condition, if the tag is not satisfied return nothing else return a record

PR #4671, opened 2022-09-22
add vertex filter for GetNeighbors

nevermore3:
## What type of PR is this?
- [ ] bug
- [ ] feature
- [x] enhancement

#### Issue(s) number: 
close https://github.com/vesoft-inc/nebula/issues/4649

#### Description:

Add vertexFilter to getNeighbor interface, when the filter condition of the  vertex is not satisfied, do not return the data for this vertex and corresponding edge 
 

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....

### DIFF
--- a/src/storage/exec/FilterNode.h
+++ b/src/storage/exec/FilterNode.h
@@ -39,8 +39,13 @@ class FilterNode : public IterateNode<T> {
   FilterNode(RuntimeContext* context,
              IterateNode<T>* upstream,
              StorageExpressionContext* expCtx = nullptr,
-             Expression* exp = nullptr)
-      : IterateNode<T>(upstream), context_(context), expCtx_(expCtx), filterExp_(exp) {
+             Expression* exp = nullptr,
+             Expression* tagFilterExp = nullptr)
+      : IterateNode<T>(upstream),
+        context_(context),
+        expCtx_(expCtx),
+        filterExp_(exp),
+        tagFilterExp_(tagFilterExp) {
     IterateNode<T>::name_ = "FilterNode";
   }
 
@@ -70,10 +75,6 @@ class FilterNode : public IterateNode<T> {
     mode_ = mode;
   }
 
-  void setVertexFilter(Expression* vertexFilter) {
-    vertexFilterExp_ = vertexFilter;
-  }
-
  private:
   bool check() override {
     if (filterExp_ == nullptr) {
@@ -99,8 +100,8 @@ class FilterNode : public IterateNode<T> {
   // return true when the value iter points to a value which can filter
   bool checkTagAndEdge() {
     expCtx_->reset(this->reader(), this->key().str());
-    if (vertexFilterExp_ != nullptr) {
-      auto res = vertexFilterExp_->eval(*expCtx_);
+    if (tagFilterExp_ != nullptr) {
+      auto res = tagFilterExp_->eval(*expCtx_);
       if (!res.isBool() || !res.getBool()) {
         context_->resultStat_ = ResultStatus::TAG_FILTER_OUT;
         return false;
@@ -117,7 +118,7 @@ class FilterNode : public IterateNode<T> {
   RuntimeContext* context_;
   StorageExpressionContext* expCtx_;
   Expression* filterExp_{nullptr};
-  Expression* vertexFilterExp_{nullptr};
+  Expression* tagFilterExp_{nullptr};
   FilterMode mode_{FilterMode::TAG_AND_EDGE};
   int32_t callCheck{0};
 };
--- a/src/storage/exec/FilterNode.h
+++ b/src/storage/exec/FilterNode.h
@@ -55,7 +55,9 @@ class FilterNode : public IterateNode<T> {
         break;
       }
       if (this->valid() && !check()) {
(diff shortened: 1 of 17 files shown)


codesigner:
Please add some description of what this pr for, from code it seems pushdown vertex filter to storage, while graph side logic seems not changed, only pass nullptr to the new added vertex_filte.

nevermore3:
> Please add some description of what this pr for, from code it seems pushdown vertex filter to storage, while graph side logic seems not changed, only pass nullptr to the new added vertex_filte.

Yes, the getNeightbor interface of the graph side behaves the same as before， it is only used for subgraphs

src/storage/query/QueryBaseProcessor-inl.h
```diff
@@ -152,6 +152,13 @@ nebula::cpp2::ErrorCode QueryBaseProcessor<REQ, RESP>::buildFilter(
     if (filter_ == nullptr) {
       return nebula::cpp2::ErrorCode::E_INVALID_FILTER;
     }
+    const auto* vertexFilterStr = getFilter(req, true);
```
critical27, line 155: Be careful here, if the filter is vertex-only, both filter_ and vertexFilter_ will be set, the Filter will eval more than once?

critical27 (commented):
Is it a temporary solution for https://github.com/vesoft-inc/nebula/issues/4649?

nevermore3:
> Is it a temporary solution for #4649?

After discussing with Peng Wei, this is the final solution

src/storage/query/QueryBaseProcessor-inl.h
nevermore3: Yes, it will be executed twice if vertexFilter_ and filter_ are the same And the result is true

critical27: Is `filter` and `vertex-filter` mutually exclusive? (can't set them both at the same request)? If is, we only need to eval once I think~

BTW, where do we extract vertex filter? I didn't see the code to set it.

src/storage/exec/FilterNode.h
```diff
@@ -68,6 +70,10 @@ class FilterNode : public IterateNode<T> {
     mode_ = mode;
   }
 
+  void setVertexFilter(Expression* vertexFilter) {
```
pengweisong, line 73: why not set it through constructor?

src/storage/exec/GetNeighborsNode.h
```diff
@@ -50,6 +50,10 @@ class GetNeighborsNode : public QueryNode<VertexID> {
       return nebula::cpp2::ErrorCode::E_INVALID_DATA;
     }
 
+    if (context_->resultStat_ == ResultStatus::TAG_FILTER_OUT) {
```
pengweisong, line 53: could add some comments.

src/storage/exec/FilterNode.h
```diff
@@ -93,6 +99,13 @@ class FilterNode : public IterateNode<T> {
   // return true when the value iter points to a value which can filter
   bool checkTagAndEdge() {
     expCtx_->reset(this->reader(), this->key().str());
+    if (vertexFilterExp_ != nullptr) {
```
pengweisong, line 102: should this be also evaluated in `checkTagOnly`?

src/storage/query/GetNeighborsProcessor.cpp
```diff
@@ -313,13 +316,16 @@ nebula::cpp2::ErrorCode GetNeighborsProcessor::checkAndBuildContexts(
...
   }
-  code = buildFilter(req, [](const cpp2::GetNeighborsRequest& r) -> const std::string* {
-    if (r.get_traverse_spec().filter_ref().has_value()) {
-      return r.get_traverse_spec().get_filter();
-    } else {
-      return nullptr;
-    }
-  });
+  code =
+      buildFilter(req, [](const cpp2::GetNeighborsRequest& r, bool isVertex) -> const std::string* {
```
pengweisong, line 320: `onlyVertex` may be better then `isVertex`

src/storage/query/QueryBaseProcessor-inl.h
```diff
@@ -152,6 +152,13 @@ nebula::cpp2::ErrorCode QueryBaseProcessor<REQ, RESP>::buildFilter(
...
+    if (vertexFilterStr != nullptr && !vertexFilterStr->empty()) {
+      vertexFilter_ = Expression::decode(pool, *vertexFilterStr);
```
pengweisong, line 157: should `vertexFilter_` be check in `checkExp`?

src/storage/exec/FilterNode.h
nevermore3: fixed

src/storage/query/QueryBaseProcessor-inl.h
nevermore3: vertexfilter is a subset of filter_ and does not need to be checked repeatedly

critical27 (approved):

pengweisong (approved):
Good job

PR #4675, opened 2022-09-23
fix bug of tagless

cangfengzhs:
## What type of PR is this?
- [ ] bug
- [ ] feature
- [ ] enhancement

#### Description:
1. when use_vertex_key=true,not save tagless key 
2. The value of Vertices in show stats needs to be deduplicated

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....


critical27 (approved):
My bad, thx~

panda-sheep (approved):

PR #4688, opened 2022-09-29
Zhanfu/csr Framework for the CSR structure

peter-rich:
## What type of PR is this?
- [ ] bug
- [x] feature
- [ ] enhancement

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....


src/graph/executor/algo/IsomorExecutor.cpp
```diff
@@ -12,77 +12,61 @@ namespace graph {
...
+   unsigned int e_count = iterDE->size();
 
-   ResultBuilder builder;
-   builder.value(iter->valuePtr());
+   unsigned int *offset = new unsigned int[v_count + 2];
+   unsigned int *neighbors = new unsigned int[e_count * 2];
+   unsigned int *labels = new unsigned int[l_count];
+   // load data vertices id and tags
+   while (iterDV->valid()) {
+     const auto v = iterDV->getColumn(nebula::kVid);  // check if v is a vertex
```
wenhaocs, line 33: what are v and v1. Please use meaningful names.

src/graph/executor/algo/IsomorExecutor.cpp
```diff
@@ -12,77 +12,61 @@ namespace graph {
...
 
-   auto iter = ectx_->getResult(subgraph->inputVar()).iter();
-   auto gnSize = iter->size();
+   unsigned int v_count = iterDV->size();
+   unsigned int l_count = iterDV->size();
+   unsigned int e_count = iterDE->size();
 
-   ResultBuilder builder;
-   builder.value(iter->valuePtr());
+   unsigned int *offset = new unsigned int[v_count + 2];
```
wenhaocs, line 28: (1) Please explain via comments why the size is +2
(2) Please use vector. Don't use new.

src/graph/executor/algo/IsomorExecutor.cpp
```diff
@@ -12,77 +12,61 @@ namespace graph {
...
-     }
+
+   // load data edges
+   offset[0] = 0;
+   iterDE = ectx_->getResult(isomor->getdScanEOut()).iter();
+   while (iterDE->valid()) {
+     unsigned int src = iterDE->getEdgeProp("*", kSrc).getInt();
+     unsigned int dst = iterDE->getEdgeProp("*", kDst).getInt();
+
+     neighbors[offset[src + 1]] = dst;
```
wenhaocs, line 57: Do not understand this. Please explain via comment.

src/graph/executor/algo/IsomorExecutor.cpp
```diff
@@ -12,77 +12,61 @@ namespace graph {
...
-   auto iter = ectx_->getResult(subgraph->inputVar()).iter();
-   auto gnSize = iter->size();
+   unsigned int v_count = iterDV->size();
+   unsigned int l_count = iterDV->size();
+   unsigned int e_count = iterDE->size();
 
-   ResultBuilder builder;
-   builder.value(iter->valuePtr());
+   unsigned int *offset = new unsigned int[v_count + 2];
+   unsigned int *neighbors = new unsigned int[e_count * 2];
```
wenhaocs, line 29: Remember that for every edge A-B, we actually store 2 copies, A->B and B->A。 You need to take this into consideration. They can be distinguished by edge type signs. Positive and Negative represent them respectively. Please think about it.

src/graph/executor/algo/IsomorExecutor.cpp
peter-rich: I change it to the new version in the code, v is the content of the vertex id while v2 is the content of the label id.

src/graph/executor/algo/IsomorExecutor.cpp
peter-rich: (1) as the offset is recording the starting position of the node 0 which suppose to be 0, and offset[1] should be the end of position node 0 and start of position node 1. The size of the offset should be at least v_count + 1;
(2) Ok

src/graph/executor/algo/IsomorExecutor.cpp
peter-rich: Ok

src/graph/executor/algo/IsomorExecutor.cpp
peter-rich: (1) src and dst is the source and destination information.
(2) offset[src + 1] will be the position of the array.

peter-rich: I can change it and comment it in the new version

src/graph/executor/algo/IsomorExecutor.cpp
```diff
@@ -1,51 +1,46 @@
...
-     const auto v = iterDV->getColumn(nebula::kVid);  // check if v is a vertex
-     auto v_id = v.getInt();
-     const auto v2 = iterDV->getColumn(1);  // get label
-     auto l_id = v2.getInt();
+     const auto vertex = iterDV->getColumn(nebula::kVid);  // check if v is a vertex
+     auto v_id = vertex.getInt();
+     const auto label = iterDV->getColumn(1);  // get label by index
+     auto l_id = label.getInt();
      // unsigned int v_id = (unsigned int)v.getInt(0);
      labels[v_id] = l_id;  // Tag Id
```
wenhaocs, line 31: Can vID used as the index of the `labels` array? There is no restriction on what vID will look like, e.g., may be from 1000 to 100000.

wenhaocs: Therefore, it's better to use map here.

src/graph/executor/algo/IsomorExecutor.cpp
```diff
@@ -1,88 +1,152 @@
...
+  }
+  // load edges degree
+  while (iterDE->valid()) {
+    auto s = iterDE->getEdgeProp("*", kSrc);
+    unsigned int src = s.getInt();
+    offset[src]++;
+    iterDE->next();
+  }
+  for (unsigned int i = 0; i < v_count; i++) {
+    offset[i + 1] = offset[i];
```
wenhaocs, line 53: Is this wrong?

src/graph/executor/CMakeLists.txt
```diff
@@ -90,4 +93,4 @@ nebula_add_library(
     mutate/UpdateExecutor.cpp
 )
 
-nebula_add_subdirectory(test)
```
wenhaocs, line 93: why is this line deleted

src/graph/executor/algo/IsomorExecutor.cpp
```diff
@@ -50,13 +66,13 @@ folly::Future<Status> IsomorExecutor::execute() {
...
     unsigned int src = s.getInt();
-    offset[src]++;
+    degree[src]++;
     iterDE->next();
   }
 
   // caldulate the start position of each vertex in the neighborhood array
   for (unsigned int i = 0; i < v_count; i++) {
-    offset[i + 1] += offset[i];
+    offset[i + 1] += degree[i] + offset[i];
```
wenhaocs, line 75: +=？

Issue #4665, opened 2022-09-21
 The `IF NOT EXISTS` in create index statement


cooper-lzy:
The expected result should be `Execution succeeded`.
![image](https://user-images.githubusercontent.com/78672629/191429161-5ab35fd3-861f-4440-a5ea-65b888d404f4.png)

PR #4705, opened 2022-10-10
Fix if-not-exists behavior of create index statements

czpmango:
## What type of PR is this?
- [x] bug
- [ ] feature
- [ ] enhancement

#### Issue(s) number: 
https://github.com/vesoft-inc/nebula/issues/4665

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....


src/meta/processors/index/CreateEdgeIndexProcessor.cpp
```diff
@@ -96,7 +97,15 @@ void CreateEdgeIndexProcessor::process(const cpp2::CreateEdgeIndexReq& req) {
     }
 
     if (checkIndexExist(fields, item)) {
-      resp_.code_ref() = nebula::cpp2::ErrorCode::E_EXISTED;
+      if (ifNotExists) {
+        resp_.code_ref() = nebula::cpp2::ErrorCode::SUCCEEDED;
+        cpp2::ID thriftID;
```
SuperYoko, line 102: maybe use tools like  `resp_.id_ref() = to(edgeIndex, EntryType::INDEX);` ?

src/meta/processors/index/CreateEdgeIndexProcessor.cpp
```diff
@@ -96,7 +97,15 @@ void CreateEdgeIndexProcessor::process(const cpp2::CreateEdgeIndexReq& req) {
...
+        // Fill index id to avoid broken promise
+        thriftID.index_id_ref() = item.get_index_id();
+        resp_.id_ref() = thriftID;
```
panda-sheep, line 105: It feels a little weird this way
1. The first `ifNotExists` in line 45 is to judge whether the index name exists or not.

2 This place is to judge whether there are other indexes that contain this fields
for example
```
create edge e(a int, b int);
create edge index e_index on e(a, b)
create edge index if not exists e_index_2 on e(a);
```
This returns a successful creation and returns the indexId of the first index.
There is actually only one index e_index

panda-sheep: It shows that the index creation is successful, but when show, the index name does not exist

czpmango: Indeed, this is the point just as you said. 
IMO, index name is a redundant design. 
A better syntax might be this: 
```
create edge index on e(a);
create edge index if not exists on e(a);
rebuild edge index on e(a);
```
In this way, the implementation layer does not need to verify the index name(line 45).
Considering compatibility, this pr is not going to do major refactoring.

src/meta/processors/index/CreateEdgeIndexProcessor.cpp
czpmango: Just extract the id from the index item.

src/meta/processors/index/CreateEdgeIndexProcessor.cpp
panda-sheep: @czpmango @MuYiYong How is our GQL version index implemented?

czpmango: We should try to avoid similar problems in GQL version.

panda-sheep: 5.0 index design needs to consider this problem, thx @MuYiYong

panda-sheep (approved):

SuperYoko (approved):

jinyingsunny:
checked on 3.3-rc

PR #4716, opened 2022-10-12
Use vertex key only when use_vertex_key is on

critical27:
## What type of PR is this?
- [ ] bug
- [ ] feature
- [X] enhancement

#### Description:
PRD of vertex key changed again... Here is the new version of logic:
1. No matter use_vertex_key is on or off, the query result should be consistent with 3.1.
2. If use_vertex_key is on, we will use the vertex key to detect whether the vertex exists, which is the logic in 3.X version.
3. If use_vertex_key is off, we will prefix scan by vertex to detect whether the vertex exists, **which will bring performance regression**
4. We will insert the vertex only when use_vertex_key is on

> By default, use_vertex_key is off

## How do you solve it?
This PR reverts all related changes before (#4685 #4680 #4675 #4629), and apply the logic above. 

Related changes:
1. So the upgrade tool is necessary again, so bring it back.
2. The TCK cases keeps same as before #4629

> Maybe review the last commit is enough 

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....


Shylock-Hg:
Add test
```
match (v:player)-[:like]->(t) return v.player.name, t.team.name
```
and 
```
match (v) return v.bachelor.name limit 3
```

critical27:
> match (v:player)-[:like]->(t) return v.player.name, t.team.name
passed

> match (v) return v.bachelor.name limit 3
need to test in ent version

nevermore3 (approved):

PR #4717, opened 2022-10-12
fix job status when all finished

SuperYoko:
## What type of PR is this?
- [x] bug
- [ ] feature
- [ ] enhancement

#### Issue(s) number: 

https://github.com/vesoft-inc/nebula-ent/issues/1367

#### Description:

1. job name should be data_balance
removed:
_2. job should be finished when all task is done and no error_ 

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....


src/meta/processors/job/BalancePlan.cpp
```diff
@@ -63,9 +63,8 @@ folly::Future<meta::cpp2::JobStatus> BalancePlan::invoke() {
           CHECK_EQ(j, buckets_[i].size() - 1);
           saveInStore();
           VLOG(4) << "BalancePlan::invoke finish";
-          auto result =
-              stopped ? meta::cpp2::JobStatus::STOPPED
-                      : failed_ ? meta::cpp2::JobStatus::FAILED : meta::cpp2::JobStatus::FINISHED;
+          // The task is either finished or failed here.
+          auto result = failed_ ? meta::cpp2::JobStatus::FAILED : meta::cpp2::JobStatus::FINISHED;
```
SuperYoko, line 67: @critical27  @panda-sheep  please help me to confirm when to set stoped. 
Now I keeped set stop logic in onError callback. I intend to change only here.

panda-sheep:
@SuperYoko I think vesoft-inc/nebula-ent in issue#1367
The job status should be always in the `stop` status.
Because when the job is stopped, all the tasks of the job are running. But it's not over.
Finally, all the tasks of the job are completed. The job is still in the stop state, which is more reasonable.

SuperYoko:
> Because when the job is stopped, all the tasks of the job are running. But it's not over.

It's trule annoying... Maybe add check that you can't stop a job all task is dispatched?

panda-sheep:
> > Because when the job is stopped, all the tasks of the job are running. But it's not over.
> 
> It's truly annoying... Maybe add check that you can't stop a job all tasks are dispatched?

This is the definition of product behavior.

SuperYoko:
> > > Because when the job is stopped, all the tasks of the job are running. But it's not over.
> > 
> > 
> > It's truly annoying... Maybe add check that you can't stop a job all tasks are dispatched?
> 
> The job is stopped, and the status of the tasks of the job can be ignored.

Make sense...  @jinyingsunny  what do you think? If you're okay with it, I'll only change this Job name.

src/graph/executor/admin/SubmitJobExecutor.cpp
```diff
@@ -124,7 +124,7 @@ nebula::DataSet SubmitJobExecutor::buildShowResultData(
     uint32_t total = paras.size() - index - 1, succeeded = 0, failed = 0, inProgress = 0,
              invalid = 0;
     v.emplace_back(Row({jd.get_job_id(),
-                        apache::thrift::util::enumNameSafe(meta::cpp2::JobType::ZONE_BALANCE),
```
liwenhui-soul, line 127: not only this, show_all_jobs also need to change

src/meta/processors/job/JobManager.cpp
```diff
@@ -643,6 +643,10 @@ ErrorOr<nebula::cpp2::ErrorCode, std::vector<cpp2::JobDesc>> JobManager::showJob
         continue;
       }
       auto jobDesc = optJob.toJobDesc();
+      if (jobDesc.get_type() == cpp2::JobType::ZONE_BALANCE) {
```
liwenhui-soul, line 646: this way is also correct, but I prefer just modify this place SubmitJobExecutor.cpp:82,  what do you think

SuperYoko: Emm, it's better to not change jobdesc but only show name. Thanks.

liwenhui-soul (approved):

panda-sheep (approved):

critical27:
Wait a moment

PR #4722, opened 2022-10-13
remove rebuild index guard when job finished

SuperYoko:
## What type of PR is this?
- [x] bug
- [ ] feature
- [ ] enhancement

#### Issue(s) number: 

https://github.com/vesoft-inc/nebula-ent/issues/1410

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....

### DIFF
--- a/src/storage/admin/RebuildIndexTask.cpp
+++ b/src/storage/admin/RebuildIndexTask.cpp
@@ -18,10 +18,12 @@ bool RebuildIndexTask::check() {
   return env_->kvstore_ != nullptr;
 }
 void RebuildIndexTask::finish(nebula::cpp2::ErrorCode rc) {
-  auto space = *ctx_.parameters_.space_id_ref();
-  for (auto it = env_->rebuildIndexGuard_->begin(); it != env_->rebuildIndexGuard_->end(); ++it) {
-    if (std::get<0>(it->first) == space) {
-      env_->rebuildIndexGuard_->insert_or_assign(it->first, IndexState::FINISHED);
+  if (changedSpaceGuard_) {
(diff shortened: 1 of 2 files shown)


critical27 (approved):
Can't believe we have so many problem in rebuild index ......... Good job

SuperYoko:
> Can't believe we have so many problem in rebuild index ......... Good job

don't merge now, wait tests result later.

SuperYoko:
> > Can't believe we have so many problem in rebuild index ......... Good job
> 
> don't merge now, wait tests result later.

Tested locally , no job will be blocked if prev job failed now.

SuperYoko:
As you can see, failed job (caused by stop) will not affect new job, and recover is okay now.

image

liwenhui-soul (approved):

pengweisong:
mark for review

src/storage/admin/RebuildIndexTask.cpp
```diff
@@ -17,6 +17,17 @@ const int32_t kReserveNum = 1024 * 4;
 bool RebuildIndexTask::check() {
   return env_->kvstore_ != nullptr;
 }
+void RebuildIndexTask::finish(nebula::cpp2::ErrorCode rc) {
```
pengweisong, line 20: why just add check if it is IndexState::FAILED in line 76? If it is, then could skip it.

pengweisong:
Talked offline,  LGTM now, but could be improved later: use taskId as guard instead of IndexState. IndexState is only as lock instead of other usage.

pengweisong (approved):

PR #4727, opened 2022-10-14
forbid insert vertex when vertex_key flag is off

critical27:
## What type of PR is this?
- [ ] bug
- [ ] feature
- [X] enhancement

#### Description:
As title, when vertex_key is not used, we forbid it. Otherwise, it will report succeeded, but no data is inserted.

Fix https://github.com/vesoft-inc/nebula-ent/issues/1420, it seems we can't simply prefix scan, we need to check if any tag is valid indeed. And valid means: tag schema exists and no ttl expired.

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....

### DIFF
--- a/tests/tck/features/ttl/TTL.feature
+++ b/tests/tck/features/ttl/TTL.feature
@@ -394,21 +394,18 @@ Feature: TTLTest
       """
     Then the result should be, in any order, with relax comparison:
       | node  |
-      | ("1") |
     When executing query:
       """
       FETCH PROP ON person "1" YIELD person.id as id
       """
     Then the result should be, in any order:
       | id    |
-      | EMPTY |
     When executing query:
       """
       FETCH PROP ON * "1" YIELD person.id, career.id
       """
     Then the result should be, in any order:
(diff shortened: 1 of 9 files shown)


critical27:
Don't merge yet, there is another tagless related issue

critical27:
Ready to review now

tests/tck/features/ttl/TTL.feature
```diff
@@ -394,21 +394,18 @@ Feature: TTLTest
       """
     Then the result should be, in any order, with relax comparison:
       | node  |
-      | ("1") |
```
Shylock-Hg, line 397: Why got empty set when vertex inserted.

critical27: It has ttl column, and ttl expired

Shylock-Hg (approved):

nevermore3 (approved):

PR #4460, opened 2022-07-25
make 'stop job' work as expected

critical27:
## What type of PR is this?
- [ ] bug
- [ ] feature
- [X] enhancement

#### Description:
We support `stop job`, but some kind of jobs are not actually stopped, this PR will report if it is not supported and make the stoppable one work as expected.

**Before**:
image

**After**:
image

## Checklist:
Tests:
- [X] Unit test(positive and negative cases)
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:
- [X] Documentation affected (Please add the label if documentation needs to be modified.)
- [ ] Incompatibility (If it breaks the compatibility, please describe it and add the label.）
- [ ] If it's needed to cherry-pick (If cherry-pick to some branches is required, please label the destination version(s).)
- [ ] Performance impacted: Consumes more CPU/Memory

## Release notes:

Please confirm whether to be reflected in release notes and how to describe:
> ex. Fixed the bug .....

PR #4742, opened 2022-10-18
JobManager related fix

critical27:
## What type of PR is this?
- [X] bug
- [ ] feature
- [ ] enhancement

#### Description:
Two bugs are fixed:
1. Some job do not support stop, e.g. COMPACT/DOWNLOAD/INGEST and so on. Previously in #4460, we have already report error. However, sadly, when we `show job`, the job is marked as STOPPED, which is not expected. It looks like as  below:

image

After this PR, these unstoppable job will still keep in RUNNING state.
2. Some job need to run in storage leader. There is a corner case that, when we create a new space, before storage report leader distribution to meta, we start a job, e.g. STATS. For now JobManager can't select a storage to run the job, and the task won't be sent to storage for execution. Since no storage ever run it,  the job will always stay in RUNNING.

## How do you solve it?
1. Only save STOPPED when it really stopped
2. If no leader found, report error.

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....

### DIFF
--- a/src/meta/processors/job/StorageJobExecutor.cpp
+++ b/src/meta/processors/job/StorageJobExecutor.cpp
@@ -81,6 +81,12 @@ ErrOrHosts StorageJobExecutor::getLeaderHost(GraphSpaceID space) {
       it->second.emplace_back(partId);
     }
   }
+  // If storage has not report leader distribution to meta and we don't report error here,
+  // JobMananger will think of the job consists of 0 task, and the task will not send to any
+  // storage. And the job will always be RUNNING.
+  if (hosts.empty()) {
+    return nebula::cpp2::ErrorCode::E_JOB_HAS_NO_TARGET_STORAGE;
+  }
   return hosts;
 }
 
--- a/src/meta/test/JobManagerTest.cpp
+++ b/src/meta/test/JobManagerTest.cpp
@@ -41,7 +41,19 @@ class JobManagerTest : public ::testing::Test {
     kv_ = cluster.initMetaKV(rootPath_->path());
     GraphSpaceID spaceId = 1;
     int32_t partitionNum = 1;
+    PartitionID partId = 1;
     ASSERT_TRUE(TestUtils::createSomeHosts(kv_.get()));
+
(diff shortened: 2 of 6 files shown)


src/meta/processors/job/StorageJobExecutor.cpp
```diff
@@ -81,6 +81,12 @@ ErrOrHosts StorageJobExecutor::getLeaderHost(GraphSpaceID space) {
       it->second.emplace_back(partId);
     }
   }
+  // If storage has not report leader distribution to meta and we don't report error here,
+  // JobMananger will think of the job consists of 0 task, and the task will not send to any
+  // storage. And the job will always be RUNNING.
```
SuperYoko, line 86: Here I got an issue for @liangliangc . That Explorer will  execute "SUBMIT JOB STATS" after create space and wait a heartbeat, and job is always running, what do you think?

critical27: BTW, I am not sure whether we should check leader count same as partition count here... WDYT?

panda-sheep (approved):

src/meta/test/JobManagerTest.cpp
```diff
@@ -853,6 +871,15 @@ TEST_F(JobManagerTest, StoppableJob) {
     code = jobMgr->stopJob(spaceId, jobId);
     ASSERT_EQ(code, nebula::cpp2::ErrorCode::SUCCEEDED);
 
+    // check job status again, it still should be running
```
panda-sheep, line 874: Good job, the comments here should be `be stopped`.

src/meta/processors/job/StorageJobExecutor.cpp
SuperYoko: Maybe a flag to identify a space is ready to supply survice.
Maybe use leader count == partition,both okay for me.

foesa-yang:
https://github.com/vesoft-inc/nebula-docs-cn/pull/2234

Issue #3513, opened 2021-12-21
function to parse json format string prop


wey-gu:
context: https://discuss.nebula-graph.com.cn/t/topic/6991

jievince:
Return a map?

akshatakulkarni25:
Hello. I am Akshata Kulkarni. I am interested in working on this issue. Is this issue still open to work on??It would be my first issue. Could you also please guide through this issue.Thanks

wey-gu:
Dear @akshatakulkarni25 ,

Welcome to the community, this issue is still open!

I think a function similar to [JSON_EXTRACT](https://dev.mysql.com/doc/refman/5.7/en/json-search-functions.html#function_json-extract) could be introduced here.

@jievince do you see it as reasonable/feasible?

## why

The motivation of this issue was NebulaGraph doesn't support nested datatype, in some cases, user may like to serialize JSON data in a string as properties of Tag/edgeType, thus, this function would enable data extracting in return expression.
(We should consider stopping user from using it in filter expression, as this will lead to a data full scan. @jievince ? )


## what

- It takes a string and returns a [map](https://docs.nebula-graph.io/3.2.0/3.ngql-guide/3.data-types/8.map/):

```sql
YIELD JSON_EXTRACT('{"a": 1, "b": 2, "c": {"d": 4}}') AS map_foo
+-------------------------+
| map_foo                 |
+-------------------------+
| {a: 1, b: 2, c: {d: 4}} |
+-------------------------+
```
- It takes path as args, too (@jievince do you think this is needed?)
```sql
YIELD JSON_EXTRACT('{"a": 1, "b": 2, "c": {"d": 4}}', '$.c') AS map_foo_dot_c
+-----------+
| map_foo   |
+-----------+
| {d: 4}    |
+-----------+
```

jievince:
> Dear @akshatakulkarni25 ,
> 
> Welcome to the community, this issue is still open!
> 
> I think a function similar to [JSON_EXTRACT](https://dev.mysql.com/doc/refman/5.7/en/json-search-functions.html#function_json-extract) could be introduced here.
> 
> @jievince do you see it as reasonable/feasible?
> 
> ## why
> The motivation of this issue was NebulaGraph doesn't support nested datatype, in some cases, user may like to serialize JSON data in a string as properties of Tag/edgeType, thus, this function would enable data extracting in return expression. (We should consider stopping user from using it in filter expression, as this will lead to a data full scan. @jievince ? )
[...]

PR #4743, opened 2022-10-19
Introduce JSON_EXTRACT function

wey-gu:
close: #3513

Note, we don't support the path argument in this phase

## What type of PR is this?
- [ ] bug
- [x] feature
- [ ] enhancement

#### Description:

It takes a string and returns a [map](https://docs.nebula-graph.io/3.2.0/3.ngql-guide/3.data-types/8.map/):

```
YIELD JSON_EXTRACT('{"a": 1, "b": 2}') AS map_foo
+-------------+
| map_foo     |
+-------------+
| {a: 1, b: 2}|
+-------------+
```

**highlight** ( should be included in docs, too)
- "foo", "1", "true" are all valid JSON String though(RFC8259), they are treated as BAD_DATA here due to this function is trying to return either empty/null or Map() in its return value, thus "foo" will be considered to be dirty value.

```
  Scenario: Test Cases With Invalid JSON String
    When executing query:
      """
      YIELD JSON_EXTRACT('fuzz') AS result;
      """
      | result   |
      | BAD_DATA |
```

**limitation**
- Only Bool, Double, Int, String value and null are supported
- Only depth-1 nested Map is supported now.

Here nested Map depth is 1, it's parsed perfectly:

```sql
> YIELD JSON_EXTRACT('{"a": 1, "b": 0.2, "c": {"d": true}}')
+------------------------------------------------------+
| JSON_EXTRACT("{"a": 1, "b": 0.2, "c": {"d": true}}") |
+------------------------------------------------------+
| {a: 1, b: 0.2, c: {d: true}}                         |
+------------------------------------------------------+
```

Here nested Map depth is 2, the nested item is left as an empty Map():

```sql
> YIELD JSON_EXTRACT('{"a": "foo", "b": 0.2, "c": {"d": {"e": 0.1}}}')
+----------------------------------------------------------------+
| JSON_EXTRACT("{"a": "foo", "b": 0.2, "c": {"d": {"e": 0.1}}}") |
+----------------------------------------------------------------+
| {a: "foo", b: 0.2, c: {}}                                      |
+----------------------------------------------------------------+
```

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

Added function: `JSON_EXTRACT` to enable parsing JSON format string into Map.


src/common/datatypes/Map.h
```diff
@@ -22,6 +24,22 @@ struct Map {
...
+  // constructor to convert from json string
+  explicit Map(const std::string& json) {
+    auto obj = folly::parseJson(json);
+    for (auto& kv : obj.items()) {
+      kvs.emplace(kv.first.getString(), Value(kv.second));
+    }
+  }
+
+  // constructor to convert from Value::Type::STRING in JSON format
+  explicit Map(const Value& value) {
```
jievince, line 36: Add a Dcheck to check the value type is str, or just delete this interface.

wey-gu: Got it, done by removing it :)

jievince (approved):
LGTM

src/common/datatypes/Map.h
```diff
@@ -28,7 +29,23 @@ struct Map {
...
+      } else if (kv.second.isBool()) {
+        kvs.emplace(kv.first.asString(), Value(kv.second.asBool()));
+      } else if (kv.second.isNull()) {
+        kvs.emplace(kv.first.asString(), Value());
+      } else {
+        LOG(ERROR) << "Only string, int64, double, bool, null are supported in "
+                      "parseJson to construct Map, now trying to parse from: "
+                   << kv.second.typeName();
+      }
     }
```
jievince, line 49: It's better to move the implementation to .cpp file.

jievince (commented):
The ctest failed, please fix it.

src/common/datatypes/Map.h
wey-gu: > The ctest failed, please fix it.

working on it now,

> It's better to move the implementation to .cpp file.

will do so.

wey-gu: done :)

jievince (approved):
Lgtm

jievince (dismissed):
Lgtm

Shylock-Hg:
Plz add test in tck.

wey-gu:
@Shylock-Hg @jievince @Sophie-Xie it's now ready and rebased :)
Thanks!

jievince (approved):

src/common/function/FunctionManager.cpp
```diff
@@ -2766,6 +2771,34 @@ FunctionManager::FunctionManager() {
...
+    attr.body_ = [](const auto &args) -> Value {
+      if (!args[0].get().isStr()) {
+        return Value::kNullBadType;
+      }
+      auto json = args[0].get().getStr();
+
+      // invalid string to json will be caught and returned as null
+      try {
+        auto obj = folly::parseJson(json);
+        if (!obj.isObject()) {
```
Shylock-Hg, line 2789: Why check this?

wey-gu: Or invalid json string will end up abnormal here, led the connection hanging.

Shylock-Hg: `"string"` is not object but also valid json.

wey-gu: Oh, thanks @Shylock-Hg , now I know "foo" is indeed a valid JSON according to RFC 8259, I will revisit here and fix this case!

wey-gu: @Shylock-Hg , in SQL DB, "foo" was parsed as NULL, thus I think checking isObject() is still reasonable here, what do you think?

Screen Shot 2022-10-24 at 10 19 17

wey-gu: aligned with @Shylock-Hg to go with option-1, I have updated the highlight in this PR to ensure this behavior will be emphasized in docs.

Shylock-Hg (approved):

foesa-yang:
https://github.com/vesoft-inc/nebula-docs-cn/pull/2268

PR #1458, opened 2019-12-16
enhance show parts and show hosts

critical27:
1. Add `show part $id` and `show parts $id1, $id2, ..., $idn`
image

image

    close #1226 

2. Enhance `show hosts`, we can show total leader count
image

PR #4749, opened 2022-10-19
Fix 1458

xtcyclist:
… match.

## What type of PR is this?
- [x] bug
- [ ] feature
- [ ] enhancement

#### Issue(s) number:

[#1458](https://github.com/vesoft-inc/nebula-ent/issues/1458)

#### Description:

Didn't connect the necessary filter in optional match.

## How do you solve it?

Removed the condition that prevents the connection of a filter in the plan generated for optional match.

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [X] TCK

Affects:
- [X] If it's needed to cherry-pick (If cherry-pick to some branches is required, please label the destination version(s).)

The enterprise version, v3.3.

## Release notes:

> ex. Fixed the bug .....


xtcyclist:
After more investigations, this PR is not enough for this bug. Close it. Will start a new PR.

PR #4754, opened 2022-10-20
download job related fix

critical27:
## What type of PR is this?
- [X] bug
- [ ] feature
- [ ] enhancement

#### Issue(s) number: 
close https://github.com/vesoft-inc/nebula-ent/issues/1463
close https://github.com/vesoft-inc/nebula-ent/issues/1464

#### Description:
When exchange generate SST files, some part may do not have data, so there won't exist corresponding part directory. After we change the way to execute hdfs command, this will be an issue. Because there will be an error like `dir not exists`, so when storage really start download, first `ls`, if the directory exists then copy.

And we need to handle ingest no files into rocksdb.

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....

### DIFF
--- a/src/common/hdfs/HdfsCommandHelper.cpp
+++ b/src/common/hdfs/HdfsCommandHelper.cpp
@@ -23,6 +23,7 @@ Status HdfsCommandHelper::ls(const std::string& hdfsHost,
     folly::Subprocess proc(std::vector<std::string>({"/bin/bash", "-c", command}));
     auto result = proc.wait();
     if (!result.exited()) {
+      LOG(INFO) << "Failed to ls: " << result.str();
       return Status::Error("Failed to ls hdfs");
     } else if (result.exitStatus() != 0) {
       LOG(INFO) << "Failed to ls: " << result.str();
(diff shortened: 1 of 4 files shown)


SuperYoko (approved):
LGTM~

panda-sheep (approved):
Good job

src/common/hdfs/HdfsCommandHelper.cpp
```diff
@@ -49,6 +50,7 @@ Status HdfsCommandHelper::copyToLocal(const std::string& hdfsHost,
     folly::Subprocess proc(std::vector<std::string>({"/bin/bash", "-c", command}));
     auto result = proc.wait();
     if (!result.exited()) {
+      LOG(INFO) << "Failed to download: " << result.str();
```
SuperYoko, line 53: Here download error will failed this job too?

critical27: Yeah, if download occurs error we should fail it. (Exchange will not generate empty dir, if it indeed generate empty dir, ingest will handle it)

PR #4757, opened 2022-10-20
Fix fulltext stopped

SuperYoko:
## What type of PR is this?
- [x] bug
- [ ] feature
- [ ] enhancement

#### Description:

build ft index is not stopable. while it's status is set to stopped. 
and then job executor return "can not stop".
It's running and no new job is allowd. 

lQLPJxbLfF_FwEXNA7jNBO2wDWmaCRG22FADTnUfAoA2AA_1261_952

## How do you solve it?

add a check in set status to early return, keep status running.

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....


src/meta/processors/job/JobDescription.cpp
```diff
@@ -86,6 +86,9 @@ bool JobDescription::setStatus(Status newStatus, bool force) {
   if (JobStatus::laterThan(status_, newStatus) && !force) {
     return false;
   }
+  if (type_ == JobType::REBUILD_FULLTEXT_INDEX && newStatus == Status::STOPPED) {
```
SuperYoko, line 89: Wait, I'll add a stoppable judge for other type.

src/meta/processors/job/JobManager.cpp
```diff
@@ -396,6 +397,9 @@ nebula::cpp2::ErrorCode JobManager::jobFinished(
       }
       // job has been stopped successfully, so update the job status
       return save(jobKey, jobVal);
+    } else if (code == nebula::cpp2::ErrorCode::E_JOB_NOT_STOPPABLE) {
+      // rollback the job status
+      optJobDesc.setStatus(rollbackStatus, true);
```
SuperYoko, line 402: f***k there's risk stop block task report!

panda-sheep: This `optJobDesc` is a temporary variable. Where to save it?

SuperYoko: my bad.

SuperYoko:
closed since i fix it in wrong way.

src/meta/processors/job/JobManager.cpp
SuperYoko: fixed by https://github.com/vesoft-inc/nebula/pull/4742

Issue #4745, opened 2022-10-19
Missing prediccate in standalone optional match


nebula-bots:
**Please check the FAQ documentation before raising an issue**

**Describe the bug (__required__)**

**Your Environments (__required__)**

* OS: `uname -a`
* Compiler: `g++ --version` or `clang++ --version`
* CPU: `lscpu`
* Commit id (e.g. `a3ffc7d8`) 318f10db0

**How To Reproduce(__required__)**

```txt
(root@nebula) [nba]> match (v) where v.player.name == "ggg" return v limit 10;
+---+
| v |
+---+
+---+
Empty set (time spent 2562/2875 us)

Tue, 18 Oct 2022 17:50:20 CST

(root@nebula) [nba]>  optional match (v) where v.player.name == "ggg" return v limit 10;
+---------------------------------------------------------------+
| v                                                             |
+---------------------------------------------------------------+
| ("Dirk Nowitzki" :player{age: 40, name: "Dirk Nowitzki"})     |
| ("Chris Paul" :player{age: 33, name: "Chris Paul"})           |
| ("Dejounte Murray" :player{age: 29, name: "Dejounte Murray"}) |
| ("Grizzlies" :team{name: "Grizzlies"})                        |
| ("Klay Thompson" :player{age: 29, name: "Klay Thompson"})     |
| ("Nuggets" :team{name: "Nuggets"})                            |
| ("Suns" :team{name: "Suns"})                                  |
| ("Bulls" :team{name: "Bulls"})                                |
| ("Kings" :team{name: "Kings"})                                |
| ("Nobody" :player{age: 0, name: "Nobody"})                    |
+---------------------------------------------------------------+
Got 10 rows (time spent 2873/3254 us)

Tue, 18 Oct 2022 17:50:22 CST

(root@nebula) [nba]>

```

```txt
(root@nebula) [nba]> profile match (v) where v.player.name == "ggg" return v limit 10;
+---+
| v |
+---+
+---+
Empty set (time spent 1888/2217 us)

Execution Plan (optimize time 207 us)

-----+----------------+--------------+-----------------------------------------------------+------------------------------------------------------------------------------------------------------------------------
| id | name           | dependencies | profiling data                                      | operator info                                                                                                         |
-----+----------------+--------------+-----------------------------------------------------+------------------------------------------------------------------------------------------------------------------------
[...]

PR #4758, opened 2022-10-20
Partial fix of optional match, for bug #4745.

xtcyclist:
## What type of PR is this?
- [X] bug
- [ ] feature
- [ ] enhancement

#### Issue(s) number: 

#4745 

#### Description:

The where clause is omitted in a singular optional match, resulting in wrong responses.

## How do you solve it?

- Removed the condition that prevents the connection of a where clause to optional match.
- To prevent re-connecting this where clause in a left-join optional match, set the `isOptional` flag to flase in the match clause context, when a left-join has been generated in the plan.

- Needs to check whether this would cause other problems.
- There are other known problems with optional match, which are left out of this PR.

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....


xtcyclist:
This PR uses a wrong branch by mistake. Close.

PR #4767, opened 2022-10-21
Partial fix optional match 4745

xtcyclist:
## What type of PR is this?
- [X] bug
- [ ] feature
- [ ] enhancement

#### Issue(s) number: 

#4745 

#### Description:

The where clause is omitted in a singular optional match, resulting in wrong responses.

## How do you solve it?

- Added a seperate if-scope to prevent singular optional matches producing random results.
- Leave the other issues of optional matches for further discussions and fixes.

- Needs to check whether this would cause other problems.
- There are other known problems with optional match, which are left out of this PR.

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [X] TCK

Affects:
- [] Documentation affected (Please add the label if documentation needs to be modified.)
- [] Incompatibility (If it breaks the compatibility, please describe it and add the label.）
- [X] If it's needed to cherry-pick (If cherry-pick to some branches is required, please label the destination version(s).)

## Release notes:

> ex. Fixed the bug .....


xtcyclist:
After discussion, I close this PR because it only stops a singular `optional match` to produce random wrong results without completely fixing all problems around `optional match`. So it comes with a risk of sabotaging other queries, many of which may not have been covered by our current tcks.

We need to discuss a PR to fully address all issues around `optional match`.

Issue #4765, opened 2022-10-20
go statement, id($$) filter is incorrect


nebula-bots:
**Your Environments (__required__)**

**How To Reproduce(__required__)**

```

(root@nebula) [sf1]> show hosts graph
+-----------------+-------+----------+---------+--------------+---------+
| Host            | Port  | Status   | Role    | Git Info Sha | Version |
+-----------------+-------+----------+---------+--------------+---------+
| "192.168.8.224" | 10030 | "ONLINE" | "GRAPH" | "fe88a7f"    | "3.3.0" |
+-----------------+-------+----------+---------+--------------+---------+
Got 1 rows (time spent 1269/32906 us)

Thu, 20 Oct 2022 20:04:48 CST

(root@nebula) [sf1]> use sf1
Execution succeeded (time spent 1207/29005 us)

Thu, 20 Oct 2022 20:04:51 CST

(root@nebula) [sf1]> GO 1 STEP FROM 32985348833536,32985348833537 OVER LIKES,KNOWS where id($$)>1 YIELD distinct  id($$)  | yield count(*)
+----------+
| count(*) |
+----------+
| 0        |
+----------+
Got 1 rows (time spent 2609/31354 us)

Thu, 20 Oct 2022 20:04:54 CST

(root@nebula) [sf1]> GO 1 STEP FROM 32985348833536,32985348833537 OVER LIKES,KNOWS  YIELD distinct  id($$)  | yield count(*)
+----------+
| count(*) |
+----------+
| 389      |
+----------+
Got 1 rows (time spent 2382/32904 us)

Thu, 20 Oct 2022 20:04:56 CST
```

jinyingsunny:
fixed 3.3-rc
![image](https://user-images.githubusercontent.com/106510322/197515912-7d508112-ea90-4732-a7d9-b1029bba3e41.png)

PR #4768, opened 2022-10-21
fix issue 4765

jievince:
## What type of PR is this?
- [x] bug
- [ ] feature
- [ ] enhancement

#### Issue(s) number: 
fix https://github.com/vesoft-inc/nebula/issues/4765
Because GetDstBySrc doesn't support filter push-down,
so we don't replace `GetNeighbors` with `GetDstBySrc` in such case:
```
GO FROM "Tony Parker" OVER like WHERE like._dst != "Tim Duncan" YIELD DISTINCT id($$)
```

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....


tests/tck/features/go/SimpleCase.feature
```diff
@@ -426,30 +331,6 @@ Feature: Simple case
       | 2  | GetDstBySrc | 1            |                   |
       | 1  | Start       |              |                   |
       | 0  | Start       |              |                   |
-    When profiling query:
```
Shylock-Hg, line 429: Why delete tests?

nevermore3 (approved):

Shylock-Hg:
I think GetDstBySrc should support push-down filter too. But could add in future.

src/graph/util/ExpressionUtils.cpp
```diff
@@ -50,6 +50,30 @@ const Expression *ExpressionUtils::findAny(const Expression *self,
   return nullptr;
 }
 
+bool ExpressionUtils::findEdgeDstExpr(const Expression *expr) {
```
czpmango, line 53: Rename `hasEdgeDstExpr`.

jievince:
> I think GetDstBySrc should support push-down filter too. But could add in future.

Could add it in future. But I don't think it will gain much performance improvement against `GetNeighbors`

src/graph/util/ExpressionUtils.cpp
```diff
@@ -50,6 +50,30 @@ const Expression *ExpressionUtils::findAny(const Expression *self,
...
+  auto finder = [](const Expression *e) -> bool {
+    if (e->kind() == Expression::Kind::kEdgeDst) {
+      return true;
+    } else {
+      auto name = e->toString();
+      std::transform(name.begin(), name.end(), name.begin(), ::tolower);
+      if (name == "id($$)") {
```
Shylock-Hg, line 60: This is not always edge dst.

Shylock-Hg:
> > I think GetDstBySrc should support push-down filter too. But could add in future.
> 
> Could add it in future. But I don't think it will gain much performance improvement against `GetNeighbors`

Make sense.

src/graph/util/ExpressionUtils.cpp
jievince: Both is ok

czpmango (approved):

jievince:
Wait a moment

PR #4771, opened 2022-10-21
Fix/find start error

### LINKED ISSUE #4763: Optimizer error on correct match query

```
(root@nebula) [nba]> match (v1:player)-[e:serve]->(v2:team) with max(rank(e)) as max_rank, min(rank(e)) as min_rank match (v3:player)-[e2:serve]->(v4:team) where rank(e2) == max_rank or rank(e2) == min_rank return e2;
[ERROR (-1005)]: Optimizer error, when rewrite relational expression
```

* OS: `uname -a`
* CPU: `lscpu`
* Commit id (e.g. `a3ffc7d8`) `023420d0ce081cd687f6f167cdf9879da1ccb053`

**Expected behavior**

- This query shall be executed successfully.
- The optimizer shall not report fatal errors that would terminate queries.

**Additional context**

Shylock-Hg:
## What type of PR is this?
- [x] bug
- [ ] feature
- [ ] enhancement

#### Issue(s) number: 
Close #4763

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....

### DIFF
--- a/src/graph/planner/match/ArgumentFinder.h
+++ b/src/graph/planner/match/ArgumentFinder.h
@@ -26,6 +26,10 @@ class ArgumentFinder final : public StartVidFinder {
 
   StatusOr<SubPlan> transformEdge(EdgeContext* edgeCtx) override;
 
+  const char* name() const override {
+    return "ArgumentFinder";
+  }
+
  private:
   ArgumentFinder() = default;
 };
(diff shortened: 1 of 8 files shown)


jievince (approved):

xtcyclist (approved):
This PR looks good. OK to be merged. But the error reported in the optimizer has not been removed. In the end, I think the `optimizer` module shall not report any error that would terminate a query. In the worst case when no optimization could be made by the optimizer, at least the basleine plan shall be executed. This may require another PR.

Issue #4770, opened 2022-10-21
[subgraph]whether exists `with prop` will affect the number of rows returned


nebula-bots:
**Your Environments (__required__)**
```
(root@nebula) [sf1]> show hosts graph
+-----------------+-------+----------+---------+--------------+---------+
| Host            | Port  | Status   | Role    | Git Info Sha | Version |
+-----------------+-------+----------+---------+--------------+---------+
| "192.168.8.224" | 10030 | "ONLINE" | "GRAPH" | "fe88a7f"    | "3.3.0" |
+-----------------+-------+----------+---------+--------------+---------+
Got 1 rows (time spent 1893/56744 us)

Fri, 21 Oct 2022 16:30:49 CST
```
**How To Reproduce(__required__)**

```
(root@nebula) [sf1]> GET SUBGRAPH   from 17592186046466 IN WORK_AT,LIKES,KNOWS  YIELD vertices as a
+------------------------------+
| a                            |
+------------------------------+
| [(17592186046466 :Person{})] |
+------------------------------+
Got 1 rows (time spent 1549/38858 us)

Fri, 21 Oct 2022 16:26:54 CST

(root@nebula) [sf1]> GET SUBGRAPH with prop   from 17592186046466 IN WORK_AT,LIKES,KNOWS  YIELD vertices as a
+---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| a                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                 |
+---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
[...]

Issue #4764, opened 2022-10-20
[subgraph] return results are incomplete


nebula-bots:
**Your Environments (__required__)**
![image](https://user-images.githubusercontent.com/104358440/196925726-26010d1f-5d31-47d7-a661-37827647822f.png)

**How To Reproduce(__required__)**
```
(root@nebula) [nba]> use nba
Execution succeeded (time spent 587/30360 us)

Thu, 20 Oct 2022 18:32:40 CST

(root@nebula) [nba]>  GET SUBGRAPH FROM 'Tim Duncan' both  serve,like Where like.likeness<90 YIELD vertices as a
+----------------------------------------+
| a                                      |
+----------------------------------------+
| [("Tim Duncan" :player{} :bachelor{})] |
+----------------------------------------+
Got 1 rows (time spent 997/30410 us)

Thu, 20 Oct 2022 18:32:48 CST

(root@nebula) [nba]>  GET SUBGRAPH FROM 'Tim Duncan' both  serve,like Where like.likeness<90 and $$.player.name!="1" YIELD vertices as a
+---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| a                                                                                                                                                                                                             |
+---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| [("Tim Duncan" :player{} :bachelor{})]                                                                                                                                                                        |
| [("Marco Belinelli" :player{}), ("Boris Diaw" :player{}), ("Shaquile O'Neal" :player{}), ("Danny Green" :player{}), ("Tiago Splitter" :player{}), ("Aron Baynes" :player{}), ("LaMarcus Aldridge" :player{})] |
+---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
Got 2 rows (time spent 1513/31079 us)

Thu, 20 Oct 2022 18:33:04 CST
```
**Expected behavior**
The results of the two statements are the same

shanlai:
checked on 3.3-rc

PR #4776, opened 2022-10-24
fix subgraph step

nevermore3:
## What type of PR is this?
- [x] bug
- [ ] feature
- [ ] enhancement

#### Issue(s) number: 
close https://github.com/vesoft-inc/nebula/issues/4770
https://github.com/vesoft-inc/nebula/issues/4764

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....


jievince (approved):

Shylock-Hg:
Could you explain this bug?

nevermore3:
> Member

In the case of only output vertex, also need to take one more step, because the attributes of the destination point need to be obtained by getneighbor in the second step.

Shylock-Hg (approved):

PR #4780, opened 2022-10-24
Fix mutil-match crash

czpmango:
## What type of PR is this?
- [x] bug
- [ ] feature
- [ ] enhancement

#### Issue(s) number: 
https://github.com/vesoft-inc/nebula-ent/issues/1489

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....

### DIFF
--- a/src/graph/planner/plan/PlanNode.cpp
+++ b/src/graph/planner/plan/PlanNode.cpp
@@ -423,7 +423,11 @@ BinaryInputNode::BinaryInputNode(QueryContext* qctx,
   }
 
   addDep(right);
-  readVariable(right->outputVarPtr());
+  if (right != nullptr) {
+    readVariable(right->outputVarPtr());
+  } else {
+    inputVars_.emplace_back(nullptr);
+  }
 }
 
 // It's used for clone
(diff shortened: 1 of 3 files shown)


jievince (approved):

src/graph/planner/plan/PlanNode.cpp
```diff
@@ -423,7 +423,11 @@ BinaryInputNode::BinaryInputNode(QueryContext* qctx,
   }
 
   addDep(right);
-  readVariable(right->outputVarPtr());
+  if (right != nullptr) {
```
Shylock-Hg, line 426: Why right will be null?

jievince (approved):

src/graph/planner/plan/PlanNode.cpp
czpmango: Just like BinaryInputNode(qctx, kind, nullptr, nullptr)?

czpmango: BiLeftJoin::clone()

codesigner (approved):
LGTM

Shylock-Hg (approved):
LGTM, maybe need a constructor for clone.

Issue #4760, opened 2022-10-20
Unable to filter on path pattern


kikimo:
**Please check the FAQ documentation before raising an issue**

**Describe the bug (__required__)**

Perhaps a similar problem to https://github.com/vesoft-inc/nebula/issues/4714 , what we want is to filter out the end vertex with tag `player` of an edge ：

` match p = (v)-[e]->(t) where id(v) == "Tim Duncan" and not (v)-[]->(t:player)  return  t` did not work as expected:

```txt
(root@nebula) [nba]> match p = (v)-[e]->(t) where id(v) == "Tim Duncan" and not (v)-[]->(t:player)  return  t
+-------------------------------------------------------------------+
| t                                                                 |
+-------------------------------------------------------------------+
| ("Spurs" :team{name: "Spurs"})                                    |
| ("LaMarcus Aldridge" :player{age: 33, name: "LaMarcus Aldridge"}) |
| ("Tony Parker" :player{age: 36, name: "Tony Parker"})             |
| ("Tony Parker" :player{age: 36, name: "Tony Parker"})             |
| ("Danny Green" :player{age: 31, name: "Danny Green"})             |
| ("Manu Ginobili" :player{age: 41, name: "Manu Ginobili"})         |
| ("Manu Ginobili" :player{age: 41, name: "Manu Ginobili"})         |
+-------------------------------------------------------------------+
Got 7 rows (time spent 2866/3218 us)

Thu, 20 Oct 2022 14:54:47 CST

(root@nebula) [nba]>
```

while this clause work as expected `match p = (v)-[e]->(t) where id(v) == "Tim Duncan" and not (v)-[e]->(t:player)  return  t`

```txt
(root@nebula) [nba]> match p = (v)-[e]->(t) where id(v) == "Tim Duncan" and not (v)-[e]->(t:player)  return  t
+--------------------------------+
| t                              |
+--------------------------------+
| ("Spurs" :team{name: "Spurs"}) |
+--------------------------------+
Got 1 rows (time spent 2766/3095 us)

Thu, 20 Oct 2022 14:57:08 CST

(root@nebula) [nba]>
```

**Your Environments (__required__)**

* OS: `uname -a`
* Compiler: `g++ --version` or `clang++ --version`
* CPU: `lscpu`
* Commit id (e.g. `a3ffc7d8`)

**How To Reproduce(__required__)**

1. Step 1
2. Step 2
3. Step 3

**Expected behavior**

**Additional context**

[...]

PR #4788, opened 2022-10-26
Fix wrong output when using pattern expression as the filter in MATCH

Aiee:
## What type of PR is this?
- [x] bug
- [ ] feature
- [ ] enhancement

#### Issue(s) number: 
Close https://github.com/vesoft-inc/nebula/issues/4760

#### Description:
The `roolUp` executor produces wrong results.

After fix:
```
(root@nebula) [nba]> match p = (v:player{name:"Tim Duncan"})-[e]->(t) where not (v)-[]->(t:player) return t
+--------------------------------+
| t                              |
+--------------------------------+
| ("Spurs" :team{name: "Spurs"}) |
+--------------------------------+
Got 1 rows (time spent 22.629ms/23.301331ms)
```

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....

### DIFF
--- a/src/graph/executor/query/RollUpApplyExecutor.cpp
+++ b/src/graph/executor/query/RollUpApplyExecutor.cpp
@@ -42,18 +42,11 @@ void RollUpApplyExecutor::buildHashTable(const std::vector<Expression*>& compare
                                          Iterator* iter,
                                          std::unordered_map<List, List>& hashTable) const {
   QueryExpressionContext ctx(ectx_);
-  std::vector<Expression*> collectColsCopy;
-  collectColsCopy.reserve(compareCols.size());
-
-  // Copy the collectCols to make sure the propIndex_ is not cached in the expr
-  for (auto* col : compareCols) {
-    collectColsCopy.emplace_back(col->clone());
-  }
 
   for (; iter->valid(); iter->next()) {
     List list;
-    list.values.reserve(collectColsCopy.size());
-    for (auto& col : collectColsCopy) {
+    list.values.reserve(compareCols.size());
+    for (auto& col : compareCols) {
       Value val = col->eval(ctx(iter));
       list.values.emplace_back(std::move(val));
     }
@@ -122,25 +115,14 @@ DataSet RollUpApplyExecutor::probeSingleKey(Expression* probeKey,
 DataSet RollUpApplyExecutor::probe(std::vector<Expression*> probeKeys,
                                    Iterator* probeIter,
                                    const std::unordered_map<List, List>& hashTable) {
-  std::vector<Expression*> probeKeysCopy;
-  probeKeysCopy.reserve(probeKeys.size());
-
-  // Copy the collectCols to make sure the propIndex_ is not cached in the expr
-  for (auto* col : probeKeys) {
-    probeKeysCopy.emplace_back(col->clone());
-  }
-
   DataSet ds;
   ds.rows.reserve(probeIter->size());
   QueryExpressionContext ctx(ectx_);
   for (; probeIter->valid(); probeIter->next()) {
     List list;
-    list.values.reserve(probeKeysCopy.size());
-    for (auto& col : probeKeysCopy) {
-      // Clone the expression so when evaluating the InputPropertyExpression, the propIndex_ will
-      // not be cached
-      auto* colCopy = col->clone();
-      Value val = colCopy->eval(ctx(probeIter));
+    list.values.reserve(probeKeys.size());
+    for (auto& col : probeKeys) {
+      Value val = col->eval(ctx(probeIter));
       list.values.emplace_back(std::move(val));
     }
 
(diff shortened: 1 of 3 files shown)


jievince (dismissed):

Shylock-Hg (approved):
Good job.

src/graph/executor/query/RollUpApplyExecutor.cpp
```diff
@@ -121,7 +121,10 @@ DataSet RollUpApplyExecutor::probe(std::vector<Expression*> probeKeys,
     List list;
     list.values.reserve(probeKeys.size());
     for (auto& col : probeKeys) {
-      Value val = col->eval(ctx(probeIter));
+      // Clone the expression so when evaluating the InputPropertyExpression, the propIndex_ will
```
Shylock-Hg, line 124: I think you don't need to clone for each row, just clone once for `build` and `probe`.

src/graph/executor/query/RollUpApplyExecutor.cpp
```diff
@@ -121,7 +121,10 @@ DataSet RollUpApplyExecutor::probe(std::vector<Expression*> probeKeys,
...
+      // not be cached
+      auto* colCopy = col->clone();
```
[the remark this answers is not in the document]
Aiee: For example, the column `t` is the 3rd column in the left iterator and is the 2nd column in the right iterator.

```
I20221026 16:00:58.585834 2228773 RollUpApplyExecutor.cpp:39] left iter size: 7
I20221026 16:00:58.585844 2228773 RollUpApplyExecutor.cpp:40] left iter value: v|e|t|p|

I20221026 16:00:58.586164 2228773 RollUpApplyExecutor.cpp:42] right iter size: 6
I20221026 16:00:58.586175 2228773 RollUpApplyExecutor.cpp:43] right iter value: v|t|(v)-->(t:player)|
```

Aiee:
> `build` and `probe` will eval the shared expression object on different input tables.

This is correct, but the `propIndex_ ` is cached in the expression, and the same column name may have a different column index in different input tables.

Shylock-Hg:
> > `build` and `probe` will eval the shared expression object on different input tables.
> 
> This is correct, but the `propIndex_ ` is cached in the expression, and the same column name may have a different column index in different input tables.

Yes, that's the point of this bug, so don't need to clone expr in each row.

src/graph/executor/query/RollUpApplyExecutor.cpp
Aiee: You are right, I'll fix the copy part.

src/graph/executor/query/RollUpApplyExecutor.cpp
```diff
@@ -114,14 +122,25 @@ DataSet RollUpApplyExecutor::probeSingleKey(Expression* probeKey,
 DataSet RollUpApplyExecutor::probe(std::vector<Expression*> probeKeys,
                                    Iterator* probeIter,
                                    const std::unordered_map<List, List>& hashTable) {
+  std::vector<Expression*> probeKeysCopy;
```
Shylock-Hg, line 125: `singleKeyProbe` and `singleKeyBuild` also need.

Shylock-Hg:
I think you better to clone in `rollUpApply` function.

src/graph/executor/query/RollUpApplyExecutor.cpp
Aiee: The expr was cloned when pass to the function:
`    result = probeSingleKey(rollUpApplyNode->compareCols()[0]->clone(), lhsIter_.get(), hashTable);
`

Aiee:
> I think you better clone in `rollUpApply` function.

Moved the copy info `rollUpApply()`

src/graph/executor/query/RollUpApplyExecutor.cpp
```diff
@@ -143,25 +144,33 @@ folly::Future<Status> RollUpApplyExecutor::rollUpApply() {
   DataSet result;
   mv_ = movable(node()->inputVars()[0]);
 
-  if (rollUpApplyNode->compareCols().size() == 0) {
+  auto compareCols = rollUpApplyNode->compareCols();
```
Shylock-Hg, line 147: LGTM, but this copy is useless.

Shylock-Hg (approved):

src/graph/executor/query/RollUpApplyExecutor.cpp
Aiee: I'll remove this in another bug fix PR.

jievince (approved):

xtcyclist (approved):
Looks good.

foesa-yang:
https://github.com/vesoft-inc/nebula-docs-cn/pull/2519

PR #4817, opened 2022-11-02
only create 1 part in tck

critical27:
## What type of PR is this?
- [ ] bug
- [ ] feature
- [X] enhancement

#### Description:

Backport from https://github.com/vesoft-inc/nebula-ent/pull/1364.  Only create space with 1 part in all tck/pytest cases. (There are even space with 1000 parts before... This will make ent version ci quite unstable, because we need to create RocksDB for each part) Hope ci could handle it now, especially in ent version.

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....

### DIFF
--- a/tests/data/ldbc_v0_3_3/config.yaml
+++ b/tests/data/ldbc_v0_3_3/config.yaml
@@ -1,6 +1,6 @@
 space:
   name: ldbc_v0_3_3
-  partitionNum: 8
+  partitionNum: 1
   replicaFactor: 1
   vidType: FIXED_STRING(32)
   charset: utf8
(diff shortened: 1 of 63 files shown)


tests/data/ldbc_v0_3_3/config.yaml
```diff
@@ -1,6 +1,6 @@
 space:
   name: ldbc_v0_3_3
-  partitionNum: 8
```
Shylock-Hg, line 3: Why change this? I think we should test one and more partitions.

Shylock-Hg: Test code logic of partitions.

pengweisong (dismissed):

pengweisong (approved):

yixinglu (approved):

PR #4822, opened 2022-11-03
Fix/create variable map

Shylock-Hg:
## What type of PR is this?
- [ ] bug
- [ ] feature
- [ ] enhancement

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....

### DIFF
--- a/tests/tck/features/bugfix/MTSafeConcurrencyVariables.feature
+++ b/tests/tck/features/bugfix/MTSafeConcurrencyVariables.feature
@@ -1,0 +1,14 @@
+# Copyright (c) 2022 vesoft inc. All rights reserved.
+#
+# This source code is licensed under Apache 2.0 License.
+Feature: Test MT-safe varaibles
+
+  Scenario: Binary plan of minus
+    # It's not stable to reproduce the bug, so we run it 100 times
+    When executing query 100 times:
+      """
+      YIELD 1 AS number MINUS YIELD 2 AS number
+      """
+    Then the result should be, in any order:
+      | number |
+      | 1      |
--- a/src/graph/context/ExecutionContext.cpp
+++ b/src/graph/context/ExecutionContext.cpp
@@ -26,6 +26,7 @@ void ExecutionContext::setResult(const std::string& name, Result&& result) {
 }
 
 void ExecutionContext::dropResult(const std::string& name) {
+  DCHECK_EQ(valueMap_.count(name), 1);
   auto& val = valueMap_[name];
   if (FLAGS_enable_async_gc) {
     GC::instance().clear(std::move(val));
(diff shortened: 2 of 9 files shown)


tests/tck/features/bugfix/MTSafeConcurrencyVariables.feature
```diff
@@ -0,0 +1,14 @@
+# Copyright (c) 2022 vesoft inc. All rights reserved.
+#
+# This source code is licensed under Apache 2.0 License.
+Feature: Test MT-safe varaibles
+
+  Scenario: Binary plan of minus
+    # It's not stable to reproduce the bug, so we run it 100 times
```
yixinglu, line 7: So, what's the real reason about the crash? not thread-safe data structure usage?

Shylock-Hg: Yes, all variable should insert to map in building stage, and executor is running on MT-environment, so if executor insert new variable will lead to condition race.

Shylock-Hg:
> Anything related to [vesoft-inc/nebula-ent#1576](https://github.com/vesoft-inc/nebula-ent/issues/1576) ? And please take care similar issue introduced in #4498

Yes I created a issue related to #4498 in https://github.com/vesoft-inc/nebula/issues/4843

tests/tck/features/bugfix/MTSafeConcurrencyVariables.feature
jievince: Which line of code inserts new variabes to map in runtime?

Shylock-Hg: https://github.com/vesoft-inc/nebula/blob/419172474fc5cb3b822b640710faa6db6d38fc2e/src/common/expression/ListComprehensionExpression.cpp#L34

jievince: > https://github.com/vesoft-inc/nebula/blob/419172474fc5cb3b822b640710faa6db6d38fc2e/src/common/expression/ListComprehensionExpression.cpp#L34

Does it leads the query `YIELD 1 AS number MINUS YIELD 2 AS number` to crash?

Shylock-Hg: https://github.com/vesoft-inc/nebula/blob/419172474fc5cb3b822b640710faa6db6d38fc2e/src/graph/validator/YieldValidator.cpp#L183

Shylock-Hg: https://github.com/vesoft-inc/nebula/blob/419172474fc5cb3b822b640710faa6db6d38fc2e/src/graph/validator/YieldValidator.cpp#L187

jievince (approved):

PR #4846, opened 2022-11-09
fix wrong expr evaluation on bool-type property in with

xtcyclist:
## What type of PR is this?
- [X] bug
- [ ] feature
- [ ] enhancement

#### Issue(s) number: 

Close https://github.com/vesoft-inc/nebula-ent/issues/1574

#### Description:

An `e.non-exist-prop == true` expr is rewritten to `e.non-exist-prop` without checking whether `e.non-exist-prop` exists. This is fine by itself.

And, while the filter operator evaluates the `e.non-exist-prop` condition, it finds it to be of the `Value::kNullUnknownProp` type. This causes unexpected error messages that terminte the query.

After discussion, we've decided that `e.non-exist-prop == true` shall return a plain null value.

## How do you solve it?

`e.non-exist-prop == true` returns plain null.

It may be the case that all nulls of the unkown_prop type shall be changed to plain ones. But I want to keep the impact of a bug fix to its minimal level. So I only made this change for edges, excluding tags' properties. This may cause inconsistent nulls in the results, which I think is probabaly a minor issue.

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [X] TCK

Affects:

## Release notes:

> ex. Fixed the bug .....

### DIFF
--- a/src/common/expression/AttributeExpression.cpp
+++ b/src/common/expression/AttributeExpression.cpp
@@ -51,7 +51,7 @@ const Value &AttributeExpression::eval(ExpressionContext &ctx) {
       }
       auto iter = lvalue.getEdge().props.find(rvalue.getStr());
       if (iter == lvalue.getEdge().props.end()) {
-        return Value::kNullUnknownProp;
+        return Value::kNullValue;
       }
       return iter->second;
     }
--- a/tests/tck/features/optimizer/PrunePropertiesRule.feature
+++ b/tests/tck/features/optimizer/PrunePropertiesRule.feature
@@ -845,10 +845,10 @@ Feature: Prune Properties rule
       match (src_v)-[e:like|serve]->(dst_v)-[e2]-(dst_v2) where  id(src_v)=="Rajon Rondo" return properties(e).degree1,properties(e).degree1,e2.a,dst_v.p.name,dst_v.player.sex1,properties(src_v).name2 limit 5;
       """
     Then the result should be, in order, with relax comparison:
-      | properties(e).degree1 | properties(e).degree1 | e2.a         | dst_v.p.name | dst_v.player.sex1 | properties(src_v).name2 |
-      | UNKNOWN_PROP          | UNKNOWN_PROP          | UNKNOWN_PROP | NULL         | NULL              | UNKNOWN_PROP            |
-      | UNKNOWN_PROP          | UNKNOWN_PROP          | UNKNOWN_PROP | NULL         | NULL              | UNKNOWN_PROP            |
-      | UNKNOWN_PROP          | UNKNOWN_PROP          | UNKNOWN_PROP | NULL         | NULL              | UNKNOWN_PROP            |
-      | UNKNOWN_PROP          | UNKNOWN_PROP          | UNKNOWN_PROP | NULL         | NULL              | UNKNOWN_PROP            |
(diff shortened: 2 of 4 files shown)


Aiee (approved):

tests/tck/features/optimizer/PrunePropertiesRule.feature
```diff
@@ -845,10 +845,10 @@ Feature: Prune Properties rule
...
       """
     Then the result should be, in order, with relax comparison:
-      | properties(e).degree1 | properties(e).degree1 | e2.a         | dst_v.p.name | dst_v.player.sex1 | properties(src_v).name2 |
-      | UNKNOWN_PROP          | UNKNOWN_PROP          | UNKNOWN_PROP | NULL         | NULL              | UNKNOWN_PROP            |
-      | UNKNOWN_PROP          | UNKNOWN_PROP          | UNKNOWN_PROP | NULL         | NULL              | UNKNOWN_PROP            |
-      | UNKNOWN_PROP          | UNKNOWN_PROP          | UNKNOWN_PROP | NULL         | NULL              | UNKNOWN_PROP            |
-      | UNKNOWN_PROP          | UNKNOWN_PROP          | UNKNOWN_PROP | NULL         | NULL              | UNKNOWN_PROP            |
-      | UNKNOWN_PROP          | UNKNOWN_PROP          | UNKNOWN_PROP | NULL         | NULL              | UNKNOWN_PROP            |
+      | properties(e).degree1 | properties(e).degree1 | e2.a | dst_v.p.name | dst_v.player.sex1 | properties(src_v).name2 |
+      | UNKNOWN_PROP          | UNKNOWN_PROP          | NULL | NULL         | NULL              | UNKNOWN_PROP            |
```
jievince, line 849: Why properties(). degree still returns unkow_prop?

Shylock-Hg:
I think `NullUnkownProp` is also treated as null in filter.

tests/tck/features/optimizer/PrunePropertiesRule.feature
xtcyclist: Didn't change this part, as it is not part of this bug and unkown_prop may be useful. Maybe we should keep the change to its minimal level.

xtcyclist:
> I think `NullUnkownProp` is also treated as null in filter.



> I think `NullUnkownProp` is also treated as null in filter.

It is null, but it is bad null which would trigger a query-terminating error message.

src/common/expression/AttributeExpression.cpp
```diff
@@ -51,7 +51,7 @@ const Value &AttributeExpression::eval(ExpressionContext &ctx) {
       }
       auto iter = lvalue.getEdge().props.find(rvalue.getStr());
       if (iter == lvalue.getEdge().props.end()) {
-        return Value::kNullUnknownProp;
+        return Value::kNullValue;
```
yixinglu, line 54: make sense for edge value, but we can NOT do the same thing for vertex type since we allow the same name property defined in different tags.

yixinglu: https://github.com/vesoft-inc/nebula-ent/issues/1574#issuecomment-1316201288

Aiee (approved):

yixinglu (approved):
LGTM

PR #4859, opened 2022-11-11
fix some rocksdb api will fail in PlainTable

critical27:
## What type of PR is this?
- [X] bug
- [ ] feature
- [ ] enhancement

#### Description:
Legacy PR from #4254. (Legacy problem has been addressed in this PR)

This problem is found in nebula-ng. Some of our rocksdb api is not well verified in PlainTable, for example, whether the data is in sst will effect the api result. The main problem is that:
1. Since PlainTable only support prefix-based Seek, we need to specify prefix_same_as_start not only in prefix, but also range and rangeWithPrefix.
2. And the length of prefix_extractor need to be modified. Because within PlainTable, if the prefix bloom filter is not inserted, you can't be read by prefix any more. So to make sure every data could be read, we use the minimum length we use in prefix, which is 4.

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....

### DIFF
--- a/src/kvstore/RocksEngineConfig.cpp
+++ b/src/kvstore/RocksEngineConfig.cpp
@@ -292,7 +292,6 @@ rocksdb::Status initRocksdbOptions(rocksdb::Options& baseOpts,
     baseOpts.rate_limiter = rate_limiter;
   }
 
-  size_t prefixLength = sizeof(PartitionID) + vidLen;
   if (FLAGS_rocksdb_table_format == "BlockBasedTable") {
     // BlockBasedTableOptions
     std::unordered_map<std::string, std::string> bbtOptsMap;
@@ -330,6 +329,7 @@ rocksdb::Status initRocksdbOptions(rocksdb::Options& baseOpts,
           baseOpts.compaction_style == rocksdb::CompactionStyle::kCompactionStyleLevel;
     }
     if (FLAGS_enable_rocksdb_prefix_filtering) {
+      size_t prefixLength = sizeof(PartitionID) + vidLen;
       baseOpts.prefix_extractor.reset(rocksdb::NewCappedPrefixTransform(prefixLength));
     }
     bbtOpts.whole_key_filtering = FLAGS_enable_rocksdb_whole_key_filtering;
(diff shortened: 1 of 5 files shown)


SuperYoko (dismissed):
LGTM. I remember there used to be similar fix before.

critical27:
> LGTM. I remember there used to be similar fix before.

Yes, but that PR has one issue and I didn't fix it, it was fixed now

src/kvstore/RocksEngineConfig.cpp
```diff
@@ -346,6 +346,11 @@ rocksdb::Status initRocksdbOptions(rocksdb::Options& baseOpts,
     if (!FLAGS_enable_rocksdb_prefix_filtering) {
       return rocksdb::Status::InvalidArgument("PlainTable should use prefix bloom filter");
     }
+    // PlainTable only support prefix-based seek, which means if the prefix is not inserted into
+    // rocksdb, we can't read them from "prefix" api anymore. For simplarity, we just set the length
```
wenhaocs, line 350: `simplarity` or `simplicity`?

src/kvstore/test/RocksEngineTest.cpp
```diff
@@ -19,23 +19,30 @@ namespace kvstore {
...
+ protected:
+  bool flush_;
 };
 
 TEST_P(RocksEngineTest, SimpleTest) {
   fs::TempDir rootPath("/tmp/rocksdb_engine_SimpleTest.XXXXXX");
   auto engine = std::make_unique<RocksEngine>(0, kDefaultVIdLen, rootPath.path());
   EXPECT_EQ(nebula::cpp2::ErrorCode::SUCCEEDED, engine->put("key", "val"));
   std::string val;
+  if (flush_) {
```
wenhaocs, line 43: Why is flush needed?

src/kvstore/RocksEngineConfig.cpp
critical27: My bad, fixed~

src/kvstore/test/RocksEngineTest.cpp
critical27: Because, we need to verify the result is correct no matter data is in memtable or sst

wenhaocs (dismissed):
Great job!

panda-sheep (approved):
Good job

wenhaocs (approved):

PR #4861, opened 2022-11-11
catch up ent

cangfengzhs:
## What type of PR is this?
- [ ] bug
- [ ] feature
- [ ] enhancement

Modified some code generation to ensure that it is the same as the enterprise version. Avoid additional trouble caused by different codes.

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....


critical27 (approved):

panda-sheep (approved):
Good！

PR #4864, opened 2022-11-11
Fix two crash issues caused by the optimizer's filter push-down rules

xtcyclist:
## What type of PR is this?
- [X] bug
- [ ] feature
- [ ] enhancement

#### Issue(s) number: 
Close https://github.com/vesoft-inc/nebula-ent/issues/1580
Close https://github.com/vesoft-inc/nebula-ent/issues/1683

#### Description:
- `colNames[i]` runs out-of-range in `PrunPropertiesVisitor.cpp`, caused by errors in PushFilterDownAggregateRule & PushFilterDownFilterRule.
- `colNames` becomes larger than `groupItems` in PushFilterDownAggregateRule, causing `groupItems[i]` to run out-of-range.

## How do you solve it?
- Check range sizes in for loop in `PrunPropertiesVisitor.cpp`.
- Found the reason why the sizes differ in `PushFilterDownAggregateRule` and `PushFilterDownProjectRule`. Try fixing the exchange of operators.

Design-wise, these two structures that shall have the same size belong to different classes: one in the base class, the other in a derived class. They are manipulated independently in many places, directly or indirecly, causing their sizes to differ. If these structures have to be declared in different classes, maybe we could try grouping modifications on them together in the same method. This may prevent such issues at the first place.

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [X] TCK

Affects:

## Release notes:

> ex. Fixed the bug .....

### DIFF
--- a/src/graph/optimizer/rule/PushFilterDownAggregateRule.cpp
+++ b/src/graph/optimizer/rule/PushFilterDownAggregateRule.cpp
@@ -59,6 +59,7 @@ StatusOr<OptRule::TransformResult> PushFilterDownAggregateRule::transform(
   }
   std::unordered_map<std::string, Expression*> rewriteMap;
   auto colNames = newAggNode->colNames();
+  DCHECK_EQ(newAggNode->colNames().size(), newAggNode->groupItems().size());
   for (size_t i = 0; i < colNames.size(); ++i) {
     auto& colName = colNames[i];
     auto iter = std::find_if(propNames.begin(), propNames.end(), [&colName](const auto& name) {
@@ -89,11 +90,17 @@ StatusOr<OptRule::TransformResult> PushFilterDownAggregateRule::transform(
   newFilterNode->setCondition(newCondition);
 
   // Exchange planNode
+  // newAggNode shall inherit the output of the oldFilterNode
   newAggNode->setOutputVar(oldFilterNode->outputVar());
+  // as the new agg node now inherits the output var ptr from a filter node, the action of
+  // which alters its own colNames, its colNames need to be explicitly preserved.
+  newAggNode->setColNames(oldAggNode->colNames());
+  // newFilterNode shall inherit the input of the oldAggNode
   newFilterNode->setInputVar(oldAggNode->inputVar());
   DCHECK_EQ(oldAggNode->outputVar(), oldFilterNode->inputVar());
-  newAggNode->setInputVar(oldAggNode->outputVar());
-  newFilterNode->setOutputVar(oldAggNode->outputVar());
+  // newAggNode shall inherit oldFilterNode's inputs
+  newAggNode->setInputVar(oldFilterNode->inputVar());
+  newFilterNode->setOutputVar(newAggNode->inputVar());
 
   // Push down filter's optGroup and embed newAggGroupNode into old filter's
   // Group
--- a/src/graph/optimizer/rule/PushFilterDownProjectRule.cpp
+++ b/src/graph/optimizer/rule/PushFilterDownProjectRule.cpp
@@ -136,13 +136,18 @@ StatusOr<OptRule::TransformResult> PushFilterDownProjectRule::transform(
     newAboveFilterNode->setInputVar(newProjNode->outputVar());
     result.newGroupNodes.emplace_back(newAboveFilterGroupNode);
   } else {
+    // newProjNode shall inherit the output from oldFilterNode
+    // which is the output of the opt group
     newProjNode->setOutputVar(oldFilterNode->outputVar());
+    // newProjNode's col names, on the hand, should inherit those of the oldProjNode
+    // since they are the same project.
+    newProjNode->setColNames(oldProjNode->outputVarPtr()->colNames);
     auto newProjGroupNode = OptGroupNode::create(octx, newProjNode, filterGroupNode->group());
     newProjGroupNode->setDeps({newBelowFilterGroup});
     newProjNode->setInputVar(newBelowFilterNode->outputVar());
     result.newGroupNodes.emplace_back(newProjGroupNode);
   }
-
+  DCHECK_EQ(newProjNode->colNames().size(), newProjNode->columns()->columns().size());
   return result;
 }
 
--- a/src/graph/visitor/PrunePropertiesVisitor.cpp
+++ b/src/graph/visitor/PrunePropertiesVisitor.cpp
@@ -57,7 +57,8 @@ void PrunePropertiesVisitor::visitCurrent(Project *node) {
     rootNode_ = false;
     return;
   }
-  for (auto i = 0u; i < columns.size(); ++i) {
+  DCHECK_EQ(columns.size(), colNames.size());
+  for (auto i = 0u; i < columns.size() && i < colNames.size(); ++i) {
     auto *col = DCHECK_NOTNULL(columns[i]);
     auto *expr = col->expr();
     auto &alias = colNames[i];
--- a/src/graph/planner/plan/PlanNode.cpp
+++ b/src/graph/planner/plan/PlanNode.cpp
@@ -396,6 +396,12 @@ SingleInputNode::SingleInputNode(QueryContext* qctx, Kind kind, const PlanNode* 
   }
 }
 
+void SingleInputNode::copyInputColNames(const PlanNode* input) {
+  if (input != nullptr) {
+    setColNames(input->colNames());
+  }
+}
+
 std::unique_ptr<PlanNodeDescription> SingleDependencyNode::explain() const {
   auto desc = PlanNode::explain();
   DCHECK(desc->dependencies == nullptr);
--- a/src/graph/planner/plan/PlanNode.h
+++ b/src/graph/planner/plan/PlanNode.h
@@ -376,11 +376,7 @@ class SingleInputNode : public SingleDependencyNode {
     SingleDependencyNode::cloneMembers(node);
   }
 
-  void copyInputColNames(const PlanNode* input) {
-    if (input != nullptr) {
-      setColNames(input->colNames());
-    }
-  }
+  void copyInputColNames(const PlanNode* input);
 
   SingleInputNode(QueryContext* qctx, Kind kind, const PlanNode* dep);
 };
--- a/src/graph/planner/plan/Query.cpp
+++ b/src/graph/planner/plan/Query.cpp
@@ -519,6 +519,15 @@ void Sample::cloneMembers(const Sample& l) {
   count_ = l.count_->clone();
 }
 
+Aggregate::Aggregate(QueryContext* qctx,
+                     PlanNode* input,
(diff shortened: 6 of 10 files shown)


src/graph/visitor/PropertyTrackerVisitor.h
```diff
@@ -30,6 +30,8 @@ struct PropertyTracker {
   void insertVertexProp(const std::string& name, TagID tagId, const std::string& propName);
   void insertEdgeProp(const std::string& name, EdgeType type, const std::string& propName);
   void insertCols(const std::string& name);
+  // for debug only
```
Shylock-Hg, line 33: Please remove it.

src/graph/visitor/PrunePropertiesVisitor.cpp
```diff
@@ -57,7 +57,7 @@ void PrunePropertiesVisitor::visitCurrent(Project *node) {
     rootNode_ = false;
     return;
   }
-  for (auto i = 0u; i < columns.size(); ++i) {
```
Shylock-Hg, line 60: This don't fix the bug.

Shylock-Hg, line 60: Add `DCHECK_EQ(colNames.size(), columns.size())`

src/graph/visitor/PropertyTrackerVisitor.h
xtcyclist: ok

src/graph/visitor/PrunePropertiesVisitor.cpp
xtcyclist: this check may not be true in the current implementation. there are cases that their sizes differ. this current pr would fix the crash problem first.

Shylock-Hg: The correctness is guaranteed by constraint which size of `colName` is equal to `columns`, or some unexpected error will be reported.

Shylock-Hg: Should fix the size of colName or columns, or the result is incorrect.

src/graph/visitor/PropertyTrackerVisitor.cpp
```diff
@@ -99,6 +100,37 @@ bool PropertyTracker::hasAlias(const std::string &name) const {
...
+        LOG(INFO) << "\t\t\t" << elem3;
+      }
+    }
+  }
+  LOG(INFO) << "colsSet size: " << colsSet.size();
+  for (auto &elem : colsSet) {
+    LOG(INFO) << "\t" << elem;
+  }
+}
+*/
```
yixinglu, line 132: cleanup

src/graph/visitor/PrunePropertiesVisitor.cpp
xtcyclist: This constraint may not be valid. There may be valid columns without names. And, the current design in general does not enforce this. These two are different contatiners. So I think it's an issue larger than this crash bug. I opened a new issue to track this https://github.com/vesoft-inc/nebula/issues/4881.

xtcyclist: If we want to enforce their sizes to be the same, better to organize them as AoS. If necessary, that better be left as another issue to fix.

src/graph/visitor/PrunePropertiesVisitor.cpp
```diff
@@ -57,7 +57,7 @@ void PrunePropertiesVisitor::visitCurrent(Project *node) {
...
+  for (auto i = 0u; i < columns.size() && i < colNames.size(); ++i) {
```
yixinglu, line 60: AFAIK, the columns size should always be same as the colNames, maybe we have to check what causes the different result here.

xtcyclist: I wasn't sure if these two shall always have equal sizes. If so, let me check it then.

src/graph/visitor/PrunePropertiesVisitor.cpp
xtcyclist: ok, let me check it then. any clue why they were not the same?

xtcyclist: > If we want to enforce their sizes to be the same, better to organize them as AoS. If necessary, it better be left as another issue to fix.

Ignore this, checked, not feasible.

xtcyclist:
```
  // last vertex
  DCHECK(!nodeInfos.empty());
  addNode(nodeInfos.back());
```

`MatchSolver::buildProjectColumns` didn't set col name for the last vertex in the path.

xtcyclist: With more investigation, the error is expected to come from `CollapseProjectRule::transform`.

xtcyclist: Update the cause of the problem to `PushFilterDownProjectRule::transform`.

xtcyclist: A `filter->project-filter` group was optimized into `filter->project` after pushing the last filter completely down to the first one. The new project's output variable was set to the old and disappeared filter's output variable, which have different column names, cauding the columns and column names in the new project to differ.

This is the direct cause of the size differ that crashes graphd while prunning the properties, which does not check for size differences before this PR.

xtcyclist: The filter was improperly setup by the `PushFilterDownAggregateRule`. This shall be the final root cause.

src/graph/visitor/PrunePropertiesVisitor.cpp
Shylock-Hg, line 60: revert this, And add a DCHECK.

xtcyclist: I think this shall be kept, because these two structures may indeed have different lengths that would crash here. Our current tests are not be able to detect all risks. DCHECK if not enough.

xtcyclist: Could add a DCHECK though.

src/graph/optimizer/rule/PushFilterDownAggregateRule.cpp
```diff
@@ -89,11 +89,13 @@ StatusOr<OptRule::TransformResult> PushFilterDownAggregateRule::transform(
...
   // Exchange planNode
-  newAggNode->setOutputVar(oldFilterNode->outputVar());
-  newFilterNode->setInputVar(oldAggNode->inputVar());
+  newAggNode->setOutputVar(oldFilterNode->outputVar());  // ~ok
+  newFilterNode->setInputVar(oldAggNode->inputVar());    // ok
   DCHECK_EQ(oldAggNode->outputVar(), oldFilterNode->inputVar());
-  newAggNode->setInputVar(oldAggNode->outputVar());
-  newFilterNode->setOutputVar(oldAggNode->outputVar());
+  // newAggNode->setInputVar(oldAggNode->outputVar());
+  newAggNode->setInputVar(oldAggNode->inputVar());
```
Shylock-Hg, line 96: I think 
```
newAggNode->setInputVar(newFilterNode->outputVar());
```
 is enough.

xtcyclist: Agree. But that would be a little bit difficult to understand. Let me add some comments here.

xtcyclist: `newAggNode->setInputVar(newFilterNode->outputVar());` This seems not ok. I've reverted it. Check the reason later.

yixinglu: > I think
> 
> ```
> newAggNode->setInputVar(newFilterNode->outputVar());
> ```
> 
> is enough.

@xtcyclist I also want to know why this fix could not solve the issue. Maybe there exists other bugs if it still reports error.

xtcyclist: Yes, there is another bug. -1 * -1 = 1. 1 * -1 = -1. Ha, ha, ha.

xtcyclist: Updated PushFilterDownProjectRule as well to fix this issue. Please check the codes and the added tcks.

xtcyclist:
I added the fix to another issue (https://github.com/vesoft-inc/nebula-ent/issues/1683) in this PR, since they are concerning the same code segment in `PushFilterDownAggregateRule.cpp`.

src/graph/optimizer/rule/PushFilterDownProjectRule.cpp
```diff
@@ -136,13 +136,18 @@ StatusOr<OptRule::TransformResult> PushFilterDownProjectRule::transform(
     newAboveFilterNode->setInputVar(newProjNode->outputVar());
     result.newGroupNodes.emplace_back(newAboveFilterGroupNode);
   } else {
+    // newProjNode shall inherit the output from oldFilterNode
+    // which is the output of the opt group
     newProjNode->setOutputVar(oldFilterNode->outputVar());
+    // newProjNode's col names, on the hand, should inherit those of the oldProjNode
+    // since they are the same project.
+    newProjNode->setColNames(oldProjNode->outputVarPtr()->colNames);
```
Shylock-Hg, line 144: This is useless, `column names` is a member of output variable.

src/graph/optimizer/rule/PushFilterDownAggregateRule.cpp
```diff
@@ -89,11 +90,17 @@ StatusOr<OptRule::TransformResult> PushFilterDownAggregateRule::transform(
...
   newAggNode->setOutputVar(oldFilterNode->outputVar());
+  // as the new agg node now inherits the output var ptr from a filter node, the action of
+  // which alters its own colNames, its colNames need to be explicitly preserved.
+  newAggNode->setColNames(oldAggNode->colNames());
+  // newFilterNode shall inherit the input of the oldAggNode
   newFilterNode->setInputVar(oldAggNode->inputVar());
   DCHECK_EQ(oldAggNode->outputVar(), oldFilterNode->inputVar());
-  newAggNode->setInputVar(oldAggNode->outputVar());
-  newFilterNode->setOutputVar(oldAggNode->outputVar());
+  // newAggNode shall inherit oldAggnode's inputs
```
Shylock-Hg, line 101: This comment is different with code.

xtcyclist: Fixed.

yixinglu (approved):
Excellent!!

czpmango (approved):

PR #4868, opened 2022-11-14
allow redefined aliases in with.

xtcyclist:
## What type of PR is this?
- [X] bug
- [ ] feature
- [ ] enhancement

#### Issue(s) number: 
Close https://github.com/vesoft-inc/nebula-ent/issues/1579

#### Description:
Redefined aliases shall be allowed in the with clause.

## How do you solve it?
- Removed the guard on checking redefined aliases. Without the `with` clause, redefined aliases work fine, except that `return *` would report redudant errors.
- Remove redudant columns in the return clause. For example, the return as in `match (v)-[e]-(v) where .... return *` shall not contain two `v`s.

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [X] TCK

Affects:

## Release notes:

> ex. Fixed the bug .....

### DIFF
--- a/src/graph/validator/MatchValidator.cpp
+++ b/src/graph/validator/MatchValidator.cpp
@@ -555,13 +555,15 @@ Status MatchValidator::validateWith(const WithClause *with,
       auto label = static_cast<const LabelExpression *>(col->expr())->name();
       auto found = withClauseCtx.yield->aliasesAvailable.find(label);
       DCHECK(found != withClauseCtx.yield->aliasesAvailable.end());
-      if (!withClauseCtx.aliasesGenerated.emplace(col->alias(), found->second).second) {
+      /*if (!withClauseCtx.aliasesGenerated.emplace(col->alias(), found->second).second) {
         return Status::SemanticError("`%s': Redefined alias", col->alias().c_str());
-      }
+      }*/
+      withClauseCtx.aliasesGenerated.emplace(col->alias(), found->second);
     } else {
-      if (!withClauseCtx.aliasesGenerated.emplace(col->alias(), aliasType).second) {
+      /*if (!withClauseCtx.aliasesGenerated.emplace(col->alias(), aliasType).second) {
         return Status::SemanticError("`%s': Redefined alias", col->alias().c_str());
-      }
+      }*/
+      withClauseCtx.aliasesGenerated.emplace(col->alias(), aliasType);
     }
 
     if (!withClauseCtx.yield->hasAgg_ &&
--- a/src/graph/validator/MatchValidator.cpp
+++ b/src/graph/validator/MatchValidator.cpp
@@ -389,18 +389,28 @@ Status MatchValidator::buildColumnsForAllNamedAliases(const std::vector<QueryPar
     }
   }
 
+  std::set<std::string> visitedAliases;
   for (auto &match : currQueryPart.matchs) {
     for (auto &path : match->paths) {
       for (size_t i = 0; i < path.edgeInfos.size(); ++i) {
         if (!path.nodeInfos[i].anonymous) {
-          columns->addColumn(makeColumn(path.nodeInfos[i].alias));
+          if (visitedAliases.find(path.nodeInfos[i].alias) == visitedAliases.end()) {
+            columns->addColumn(makeColumn(path.nodeInfos[i].alias));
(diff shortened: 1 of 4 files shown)


jievince (dismissed):

src/graph/validator/MatchValidator.cpp
```diff
@@ -384,18 +384,28 @@ Status MatchValidator::buildColumnsForAllNamedAliases(const std::vector<QueryPar
     }
   }
 
+  std::set<std::string> visitedAliases;
   for (auto &match : currQueryPart.matchs) {
     for (auto &path : match->paths) {
```
Shylock-Hg, line 389: I think you should create a new `visitedAliases` for each query part.

src/graph/validator/MatchValidator.cpp
```diff
@@ -550,13 +560,15 @@ Status MatchValidator::validateWith(const WithClause *with,
       auto label = static_cast<const LabelExpression *>(col->expr())->name();
       auto found = withClauseCtx.yield->aliasesAvailable.find(label);
       DCHECK(found != withClauseCtx.yield->aliasesAvailable.end());
-      if (!withClauseCtx.aliasesGenerated.emplace(col->alias(), found->second).second) {
+      /*if (!withClauseCtx.aliasesGenerated.emplace(col->alias(), found->second).second) {
         return Status::SemanticError("`%s': Redefined alias", col->alias().c_str());
-      }
+      }*/
+      withClauseCtx.aliasesGenerated.emplace(col->alias(), found->second);
```
yixinglu, line 566: We should differentiate the variable aliases defined in with clause and inherited from the former clause, maybe MATCH, YIELD whatever. We only need to check whether the alias defined in with clause has been defined in current clause or former clause.

src/graph/validator/MatchValidator.cpp
xtcyclist: This is to avoid duplicate aliases in the `return *`. For return, only one query part, the last query part, is to be concerned. In this method, this query part is `currQueryPart`.

Within this `currQueryPart`, if there are duplicates of the same alias in multiple matches, they shall be removed, leaving only one alias in `return *`. Thus, this visitedAliases shall be declared outside this outer for loop.

Shylock-Hg: Got.

src/graph/validator/MatchValidator.cpp
```diff
@@ -550,13 +560,15 @@ Status MatchValidator::validateWith(const WithClause *with,
...
     } else {
-      if (!withClauseCtx.aliasesGenerated.emplace(col->alias(), aliasType).second) {
+      /*if (!withClauseCtx.aliasesGenerated.emplace(col->alias(), aliasType).second) {
         return Status::SemanticError("`%s': Redefined alias", col->alias().c_str());
-      }
+      }*/
+      withClauseCtx.aliasesGenerated.emplace(col->alias(), aliasType);
```
Shylock-Hg, line 571: DCHECK to ensure that the alias to be inserted.

Shylock-Hg (dismissed):

src/graph/validator/MatchValidator.cpp
xtcyclist: Understood. My changes here are wrong. Reverted.

src/graph/validator/MatchValidator.cpp
xtcyclist: My changes here are wrong. Reverted.

Shylock-Hg (dismissed):

tests/tck/features/match/RedefinedNode.feature
```diff
@@ -192,3 +192,11 @@ Feature: Redefined symbols
       MATCH (v:player{name:"abc"})-[e:like]->(v1)-[e:like]->(v2) RETURN *
       """
     Then a SemanticError should be raised at runtime: `e': Redefined alias
+
+  Scenario: Redefined alias in with
+    When executing query:
+      """
+      MATCH (n0)-[e0]->(n0) WHERE (id(n0) IN [7,10]) with * RETURN *
```
Aiee, line 199: Maybe change the query using the schema in the basketball player dataset.

Shylock-Hg (approved):

jievince (approved):

Issue #4818, opened 2022-11-02
should reset session space after `Space not existed`


nebula-bots:
there're two session, sessionA and sessionB

1. sessionA: create spaceA, and then use space.
2. sessionB: drop the spaceA
3. sessionA: create tag in spaceA

expected result:
1. raise a `Space not existed` error.
2. the space name in resultSet should be ""

actual result:
1. raise a `Space not existed` error.
2. the space name in resultSet is still "spaceA"

![image](https://user-images.githubusercontent.com/1726587/199410738-d7531336-3e80-45fa-94d7-415cb22cbb30.png)

space name is still "spaceA"

Aiee:
We might need to add a new error code to distinguish if the SPACE_NOT_FOUND is returned by a USE statement or other statements.

For example, for query like use space_1; create tag t1(), if space_1 was dropped before executing create tag t1(), the session space should be reset. If space_1 was dropped before executing use space_1, the space should remain the same.

This will be regarded as a known issue for now.

Sophie-Xie:
Discuss it with @MuYiYong @HarrisChu: we will not fix it.

PR #4870, opened 2022-11-14
[DO NOT MERGE] Reset space info of the session after drop a space

Aiee:
## What type of PR is this?
- [x] bug
- [ ] feature
- [ ] enhancement

#### Issue(s) number: 
Close https://github.com/vesoft-inc/nebula/issues/4818

#### Description:
image

Seems some further discussion on this issue is needed:

Should we reset the space to `null` after the space bond to the session is dropped? 
```
// session 1
create space space-1; use space-1;
// session 2
drop space space-1;
// session 1
create tag t1(); This query should cause an error says SPACE_NOT_FOUND, and we should reset the session space
```
What if the user uses a nonexistent space in a query? Should we reset it as well?
```
// session 1
create space space-1; use space-1;
use nonexistent_space; // This query should cause an error says SPACE_NOT_FOUND, but should we reset the session to NONE?
```

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....

### DIFF
--- a/tests/Makefile
+++ b/tests/Makefile
@@ -104,7 +104,7 @@ sess: currdir
 jobs: currdir
 	$(test_without_skip) tck/steps/test_jobs.py
 
-test: sess
+test:
 	$(test_j) --dist=loadfile -k "not tck" $(TEST_DIR)
 
 slow-query: currdir
--- a/src/graph/service/QueryInstance.cpp
+++ b/src/graph/service/QueryInstance.cpp
@@ -153,13 +153,21 @@ void QueryInstance::onError(Status status) {
     case Status::Code::kPermissionError:
       rctx->resp().errorCode = ErrorCode::E_BAD_PERMISSION;
       break;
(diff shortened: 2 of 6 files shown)


tests/Makefile
```diff
@@ -104,7 +104,7 @@ sess: currdir
 jobs: currdir
 	$(test_without_skip) tck/steps/test_jobs.py
 
-test: sess
+test:
```
HarrisChu, line 107: maybe a mistake

Aiee:
We might need to add a new error code to distinguish if the `SPACE_NOT_FOUND` is returned by a `USE` statement or other statements.

For example, for query like `use space_1; create tag t1()`, if space_1 was dropped before executing `create tag t1()`, the session space should be reset. If space_1 was dropped before executing `use space_1`, the space should remain the same.

This will be regarded as a known issue for now.

PR #4871, opened 2022-11-14
Evaluate expressions in with clauses that do not have other inputs

### LINKED ISSUE #4854: Property needed by return * may has been prunned.
### Query:
```
(root@nebula) [nebula_huskie]> MATCH (n0)-[e0]-()-[]->(n1:Label_1{labelProp_1_3: 120, labelProp_1_4: true}) WHERE (id(n0) IN [3,13]) AND (93 <= n1.Label_1.labelProp_1_0) WITH MIN(87) AS a0 return *
+----------+
| a0       |
+----------+
| __NULL__ |
+----------+
```

### Expected result:
```
+----------+
| a0       |
+----------+
| 87       |
+----------+
```

xtcyclist:
## What type of PR is this?
- [X] bug
- [ ] feature
- [ ] enhancement

#### Issue(s) number: 
Close https://github.com/vesoft-inc/nebula/issues/4854

#### Description:
Expressions in with clause that does not have inputs from its previous clauses haven't been evaluated.

## How do you solve it?
Replace the for loop that checks whether there are inputs to the with clause with a do-while loop that only check this after evaluating expressions.

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [X] TCK

Affects:

## Release notes:

> ex. Fixed the bug .....


xtcyclist:
After checking with openCyper standards, this is an non-issue. Close it.

PR #4872, opened 2022-11-14
unify time format in show job

SuperYoko:
## What type of PR is this?
- [x] bug
- [ ] feature
- [ ] enhancement

#### Issue(s) number: 

https://github.com/vesoft-inc/nebula-ent/issues/1553

#### Description:

As you can see, time show in this table is not in the same format, since a toString() method,  I unify them to version without toString, so that it can look the same.

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....


liwenhui-soul (approved):

pengweisong (approved):

foesa-yang:
https://github.com/vesoft-inc/nebula-docs-cn/pull/2497

Issue #4862, opened 2022-11-11
The length() function returns NULLs instead of the length of a path


abby-cyber:
<img width="876" alt="image" src="https://user-images.githubusercontent.com/78209557/201275023-1b14b24a-011d-4c75-94a2-948bb4524b5b.png">

The length() function returns NULLs instead of the path length when I want to query the name in returned paths.

czpmango:
Just need throw runtime error for incompatible types, this is my test in neo4j 4.2.4:
```
neo4j@neo4j> create (v:player{name:"czp"})-[:like]->();
Added 2 nodes, Created 1 relationships, Set 1 properties, Added 1 labels

neo4j@neo4j> match p=(v:player)--() return length(p),size(v.name),length(v.name);
LENGTH cannot be used on strings

neo4j@neo4j> match p=(v:player)--() return length(p),size(v.name);
+--------------------------+
| length(p) | size(v.name) |
+--------------------------+
| 1         | 3            |
+--------------------------+

neo4j@neo4j> match p=(v:player)--() return length(p),size(v.prop);
+--------------------------+
| length(p) | size(v.prop) |
+--------------------------+
| 1         | NULL         |
+--------------------------+

```

Shylock-Hg:
A bug of properties prune
```
(root@nebula) [nba]> explain match (v) where id(v) == 'Tim Duncan' return length(v.player.name)
Execution succeeded (time spent 4791/6365 us)

Execution Plan (optimize time 2369 us)

-----+----------------+--------------+----------------+--------------------------------
| id | name           | dependencies | profiling data | operator info                 |
-----+----------------+--------------+----------------+--------------------------------
|  6 | Project        | 7            |                | outputVar: {                  |
|    |                |              |                |   "colNames": [               |
|    |                |              |                |     "length(v.player.name)"   |
|    |                |              |                |   ],                          |
|    |                |              |                |   "type": "DATASET",          |
|    |                |              |                |   "name": "__Project_6"       |
|    |                |              |                | }                             |
|    |                |              |                | inputVar: __Project_5         |
|    |                |              |                | columns: [                    |
|    |                |              |                |   "length(v.player.name)"     |
|    |                |              |                | ]                             |
-----+----------------+--------------+----------------+--------------------------------
|  7 | AppendVertices | 2            |                | outputVar: {                  |
|    |                |              |                |   "colNames": [               |
|    |                |              |                |     "v"                       |
|    |                |              |                |   ],                          |
[...]

PR #4884, opened 2022-11-16
Fix wrong properties prune in `length` function.

Shylock-Hg:
## What type of PR is this?
- [x] bug
- [ ] feature
- [ ] enhancement

#### Issue(s) number: 
Close https://github.com/vesoft-inc/nebula/issues/4862

#### Description:
`length` function also could accept STRING input but not only PATH. So for compatibility, I disable this optimization. It lead to performance regression when `return length(p)` only.

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....


src/graph/visitor/PropertyTrackerVisitor.cpp
```diff
@@ -279,8 +279,10 @@ void PropertyTrackerVisitor::visit(AttributeExpression *expr) {
 }
 
 void PropertyTrackerVisitor::visit(FunctionCallExpression *expr) {
+  // length function support `STRING` input too, so we can't ignore it directly
+  // TODO add type info to variable to help optimize it.
   static const std::unordered_set<std::string> ignoreFuncs = {
-      "src", "dst", "type", "typeid", "id", "rank", "length"};
+      "src", "dst", "type", "typeid", "id", "rank" /*, "length"*/};
```
jievince, line 285: Could we check if the argument type is path？

Shylock-Hg: Could do it in later optimization. This PR just fix bug.

jievince: I remember It was to optimze length(path) related query in ldbc. @nevermore

codesigner (approved):
LGTM

src/graph/visitor/PropertyTrackerVisitor.cpp
nevermore3: yes, we can add the type of the variable to the operator later, which can be more precise when clipping attributes

nevermore3 (approved):

PR #4887, opened 2022-11-16
Profile storage GetDstBySrcProcessor

yixinglu:
## What type of PR is this?
- [ ] bug
- [ ] feature
- [X] enhancement

#### Description:

Replace std::vector<Value> with std::deque<Value> to improve the memory allocation.

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....


critical27 (approved):
I'll do some optimization in `GetDstBySrc` later. The dedup is quite slow even with multi thread. Need to check FLAGS_concurrent_dedup_threshold to see whether it really dedup in multi thread.

PR #4888, opened 2022-11-16
remove job entries when drop space && check space exist when check jo…

pengweisong:
## What type of PR is this?
- [ ] bug
- [ ] feature
- [ ] enhancement

#### Issue(s) number: 
https://github.com/vesoft-inc/nebula-ent/issues/1643

#### Description:
When drop the space, the job entries have not been dropped. Then we create backup, it may have some job entries left which are rebuilding jobs. In this case, there is no rebuilding job really, but the backup creation is blocked.

## How do you solve it?
1. drop the job entries when drop spaces.
2. skip the spaces not exist when check rebuilding jobs.

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....

### DIFF
--- a/src/meta/processors/job/JobManager.cpp
+++ b/src/meta/processors/job/JobManager.cpp
@@ -66,6 +66,16 @@ JobManager::~JobManager() {
   shutDown();
 }
 
+bool JobManager::spaceExist(GraphSpaceID spaceId) {
+  auto spaceKey = MetaKeyUtils::spaceKey(spaceId);
+  std::string val;
+  auto retCode = kvStore_->get(kDefaultSpaceId, kDefaultPartId, spaceKey, &val);
+  if (retCode == nebula::cpp2::ErrorCode::E_KEY_NOT_FOUND) {
+    return false;
+  }
+  return true;
+}
+
 nebula::cpp2::ErrorCode JobManager::handleRemainingJobs() {
   std::unique_ptr<kvstore::KVIterator> iter;
   auto retCode =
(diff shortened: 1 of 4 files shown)


critical27 (approved):

SuperYoko (approved):

src/meta/processors/job/JobManager.cpp
```diff
@@ -66,6 +66,16 @@ JobManager::~JobManager() {
...
 }
 
+bool JobManager::spaceExist(GraphSpaceID spaceId) {
+  auto spaceKey = MetaKeyUtils::spaceKey(spaceId);
+  std::string val;
+  auto retCode = kvStore_->get(kDefaultSpaceId, kDefaultPartId, spaceKey, &val);
+  if (retCode == nebula::cpp2::ErrorCode::E_KEY_NOT_FOUND) {
+    return false;
+  }
+  return true;
```
SuperYoko, line 76: This judge looks dangerous, but since true do nothing, I think it's okay.

pengweisong: OK, I'll make it explicit.

SuperYoko: > 

I'm okay with current code. If you modify, then better is better ovo

pengweisong: I've changed it, but forget to push it before it was merged..

SuperYoko: > 

Don't worry, I don't think it must be changed.

PR #4899, opened 2022-11-17
fix nightly coverage oom

cangfengzhs:
## What type of PR is this?
- [ ] bug
- [ ] feature
- [ ] enhancement

#### Description:
Nightly's coverage step will fail frequently. It is speculated that when ENABLE_COVERAGEis set to on, the number of concurrent links is large, leading to OOM.

Try to reduce the number of compilation threads to avoid this problem.

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -158,7 +158,7 @@ jobs:
       - name: Make
         run: |
           ccache -z
-          ninja -j $(nproc)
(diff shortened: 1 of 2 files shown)


codesigner (approved):
LGTM

xtcyclist (approved):

PR #4907, opened 2022-11-18
Remove all UNKNOWN_PROP as a type of null.

xtcyclist:
## What type of PR is this?
- [X] bug
- [ ] feature
- [ ] enhancement

#### Issue(s) number: 
Close https://github.com/vesoft-inc/nebula-ent/issues/1648
Close https://github.com/vesoft-inc/nebula-ent/issues/1670

#### Description:
`UNKNOWN_PROP` as a type of Null is quite troublesome. It may block normal valid queries to produce valid results.

## How do you solve it?
Removed it completely from the codebase.
I think we shall not store a type of abnormality as data in a value.

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....


czpmango (approved):
LGTM. 
Fwiw, are there other bad null types that need to delegate to null behavior?

yixinglu (approved):
LGTM

czpmango (approved):

yixinglu (approved):

abby-cyber:
https://github.com/vesoft-inc/nebula-docs-cn/pull/2493

Issue #4844, opened 2022-11-09
inner variable read/write conflict when run executor in parallel.


Shylock-Hg:
**Please check the FAQ documentation before raising an issue**

**Describe the bug (__required__)**
Wen set max_job_size > 1, the inner variable write/read will lead to condition race.
```
E20221109 18:01:45.060799 2658775 QueryInstance.cpp:139] SyntaxError: syntax error near `M MATCH('
E20221109 18:02:03.555559 2658775 QueryInstance.cpp:139] SyntaxError: syntax error near `M MATCH('
E20221109 18:02:18.386580 2658775 QueryInstance.cpp:139] SemanticError: `product': Unknown tag
E20221109 18:02:30.745103 2658775 QueryInstance.cpp:139] SemanticError: `e': Redefined alias
=================================================================
==2659462==ERROR: AddressSanitizer: heap-use-after-free on address 0x61d000114e80 at pc 0x0000022b7655 bp 0x7fede8338e00 sp 0x7fede8338df0
READ of size 1 at 0x61d000114e80 thread T16 (executor-pri3-4)
    #0 0x22b7654 in nebula::graph::Result::Core::Core(nebula::graph::Result::Core&&) src/graph/context/Result.h:85
    #1 0x22b8225 in nebula::graph::Result::Result(nebula::graph::Result&&) src/graph/context/Result.h:20
    #2 0x22b8225 in void __gnu_cxx::new_allocator<nebula::graph::Result>::construct<nebula::graph::Result, nebula::graph::Result>(nebula::graph::Result*, nebula::graph::Result&&) /usr/include/c++/9/ext/new_allocator.h:146
    #3 0x22b8225 in void std::allocator_traits<std::allocator<nebula::graph::Result> >::construct<nebula::graph::Result, nebula::graph::Result>(std::allocator<nebula::graph::Result>&, nebula::graph::Result*, nebula::graph::Result&&) /usr/include/c++/9/bits/alloc_traits.h:483
    #4 0x22b8225 in void std::__relocate_object_a<nebula::graph::Result, nebula::graph::Result, std::allocator<nebula::graph::Result> >(nebula::graph::Result*, nebula::graph::Result*, std::allocator<nebula::graph::Result>&) /usr/include/c++/9/bits/stl_uninitialized.h:912
[...]

PR #4913, opened 2022-11-22
make evaluation of inner var expr thread-safe

jievince:
## What type of PR is this?
- [x] bug
- [ ] feature
- [ ] enhancement

#### Issue(s) number: 
Fix https://github.com/vesoft-inc/nebula/issues/4844
## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....

### DIFF
--- a/src/graph/visitor/ValidatePatternExpressionVisitor.cpp
+++ b/src/graph/visitor/ValidatePatternExpressionVisitor.cpp
@@ -34,13 +34,16 @@ void ValidatePatternExpressionVisitor::visit(MatchPathPatternExpression *expr) {
         // TODO we should check variable is Node type
         // from local variable
         node->setVariableDefinedSource(MatchNode::VariableDefinedSource::kExpression);
-        auto *listElement = VariableExpression::make(pool_, listElementVar);
+        auto *listElement = VariableExpression::makeInner(pool_, listElementVar);
         // Note: this require build path by node pattern order
         auto *listElementId = FunctionCallExpression::make(
             pool_,
             "_nodeid",
             {listElement, ConstantExpression::make(pool_, static_cast<int64_t>(i))});
-        auto *nodeValue = VariableExpression::make(pool_, node->alias());
+        // The alias of node is converted to a inner variable.
+        // e.g. MATCH (v:player) WHERE [t in [v] | (v)-[:like]->(t)] RETURN v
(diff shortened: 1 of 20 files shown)


src/graph/visitor/ValidatePatternExpressionVisitor.cpp
```diff
@@ -34,13 +34,16 @@ void ValidatePatternExpressionVisitor::visit(MatchPathPatternExpression *expr) {
...
         // Note: this require build path by node pattern order
         auto *listElementId = FunctionCallExpression::make(
             pool_,
             "_nodeid",
             {listElement, ConstantExpression::make(pool_, static_cast<int64_t>(i))});
-        auto *nodeValue = VariableExpression::make(pool_, node->alias());
+        // The alias of node is converted to a inner variable.
+        // e.g. MATCH (v:player) WHERE [t in [v] | (v)-[:like]->(t)] RETURN v
+        // More cases could be found in PathExprRefLocalVariable.feature
+        auto *nodeValue = VariableExpression::makeInner(pool_, node->alias());
```
jievince, line 46: node->alias() is converted to a inner var expr here.

src/storage/context/StorageExpressionContext.h
```diff
@@ -316,6 +324,9 @@ class StorageExpressionContext final : public ExpressionContext {
 
   // name -> Value with multiple versions
   std::unordered_map<std::string, std::vector<Value>> valueMap_;
+
+  // Expression value map that stores the value of innerVar
+  std::unordered_map<std::string, Value> exprValueMap_;
```
Shylock-Hg, line 329: Why is it MT-safe? And I don't see you create different `ExprCtx` for each thread.

jievince: See https://github.com/vesoft-inc/nebula/blob/f2cee664e78e1f855e3f5cec0ad60797291bdc53/src/graph/executor/query/ProjectExecutor.cpp#L28
Every thread has its own QueryExpressionContext.

nevermore3 (approved):

yixinglu (commented):
LGTM

yixinglu (approved):

PR #4916, opened 2022-11-23
Enhancement/pattern expression

czpmango:
## What type of PR is this?
- [ ] bug
- [ ] feature
- [x] enhancement

#### Description:
This pr is optimized for the case of pattern expression as a predicate. In this case, the runtime doesn't actually have to construct the path. The principle is similar to sql semi apply related optimization. 

Compared to previous implementations, this pr has three main optimization points:
- The runtime does not construct paths to pattern predicate(overhead of Project(pathBuild expression)).
- Avoid the overhead of RollUpJoin, which is likely to have high cpu overhead(hash build and probe)and the large table being constructed will also have high memory overhead.
- Avoid the overhead of Filter executor.

For example:
Query statement: `MATCH (v:player)-->(n) where (v)-[:like]->() and size((v)-[:serve]-())>2 and v.age>3 and not (v)-[:teammate]-(n) RETURN id(v)`
Old plan: 
graphviz (3)
Optimized plan:
graphviz (4)

## Special notes for your reviewer
I will fix some FIXMEs related to handling of expression later. Expression visitor needs to cover all expressions, and currently supports only simple expressions.

## Checklist:
Tests:
- [ ] Function test
- [x] Performance test
- [ ] N/A

Affects:

## Release notes:
Optimize the implementation of pattern predicate.


jievince (approved):
Well done!

czpmango:
We should add PushFilterDownRollUpApply and PushFilterDownPatternApply rules.

czpmango:
> Please add test.

This pr can be considered as an optimization or refactoring, and all the statement behavior is guaranteed by previous tests related to pattern expression. 
If the test case coverage is not complete, it will be evaluated and completed in the future planning.

jievince (approved):

src/graph/planner/match/MatchPathPlanner.cpp
```diff
@@ -97,7 +97,11 @@ StatusOr<SubPlan> MatchPathPlanner::transform(
                             subplan,
                             nodeAliasesSeenInPattern));
 
-  MatchSolver::buildProjectColumns(qctx, path, subplan);
+  // No need to actually build path if the path is just a predicate
+  if (!path.isPred) {
```
Shylock-Hg, line 101: Better to do it in optimize rule.

czpmango: Generally ACK.
But IMHO, pattern expression and pattern predicate are semantically different, so it is also reasonable to generate different execution plans in the planner.

Shylock-Hg: Not, in general, it's properties prune problem. e.g. `where (v)-[]->()` means `List<Path>` is not empty.

Shylock-Hg: Planner should generate the most generic plan, and keep simple. Too many branches are evil.

czpmango: This pr does trigger some optimization related to property pruning, but does it have anything to do with pattern predicate. I don't get you about "in general, it's properties prune problem".

czpmango: > Planner should generate the most generic plan, and keep simple. Too many branches are evil.

Just one more if-block. Is it necessary to implement it in the optimizer phase and add a few hundred lines of code?

czpmango: I also approve of doing that in optimizer, but I still stand by my opinion.
It is true that planner should generate the most general plan, but it does not mean that a designed operator should never be generated in planner. For example, if the pattern(like (v)-[e]-(v)) is already a ExpandInto pattern, is it necessary to convert ExpandAll into ExpandInto in optimizer phase when all information is already obvious in the AST.

jievince (approved):

src/graph/planner/match/MatchPathPlanner.cpp
Shylock-Hg: ok

Shylock-Hg (approved):

yixinglu:
Now we don't hurry to tuning the LDBC according to the PR, you could transform the implementation to the optimizer rule.

czpmango:
> Now we don't hurry to tuning the LDBC according to the PR, you could transform the implementation to the optimizer rule.

IMHO, implementing this in the optimizer is not necessarily better. 
https://github.com/vesoft-inc/nebula/pull/4916#discussion_r1032056777

Sophie-Xie:
Pls don‘t merge, wait for @yixinglu .

yixinglu (dismissed):

jievince (dismissed):

jievince (approved):

yixinglu (commented):
Need some tck cases

czpmango:
> Need some tck cases

As it says here: https://github.com/vesoft-inc/nebula/pull/4916#issuecomment-1325870778

yixinglu (approved):

jievince (approved):

PR #4931, opened 2022-11-24
Minor enhancement about listener

critical27:
## What type of PR is this?
- [ ] bug
- [ ] feature
- [X] enhancement

#### Description:
Also a prepare work:
1. Only pass necessary parameter to Listener's constructor, other's could be passed to their inherited classes.
2. Modify the comment about Listener to keep up with latest changes.

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....

### DIFF
--- a/src/kvstore/Listener.h
+++ b/src/kvstore/Listener.h
@@ -77,14 +77,16 @@ using RaftClient = thrift::ThriftClientManager<raftex::cpp2::RaftexServiceAsyncC
  *   // extra initialize work could do here
  *   void init()
  *
+ *   // Main interface to process logs, listener need to apply the committed log entry to their
+ *   // state machine. Once apply succeeded, user should call persist() to make their progress
+ *   // persisted.
+ *   virtual void processLogs() = 0;
+ *
  *   // read last commit log id and term from external storage, used in initialization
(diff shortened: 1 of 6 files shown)


liwenhui-soul (dismissed):

src/kvstore/Listener.h
```diff
@@ -77,15 +77,17 @@ using RaftClient = thrift::ThriftClientManager<raftex::cpp2::RaftexServiceAsyncC
  *   // extra initialize work could do here
  *   void init()
  *
+ *   // Main interface to process logs, listener need to apply the committed log entry to their
+ *   // state machine. Once apply succeeded, user should call persist() to make their progress
+ *   // persisted.
+ *   virtual void processLogs() = 0;
```
panda-sheep, line 83: I have a small question:
Why does the function comment here exist? Can it be removed?
Otherwise, there are two copies of these comments, which feels a bit redundant?

panda-sheep (commented):
Good job

pengweisong (approved):

liwenhui-soul (approved):

Issue #4945, opened 2022-11-29
Profile shows arguments plan node has two same input variable


jievince:
**Please check the FAQ documentation before raising an issue**

**Describe the bug (__required__)**

e.g. 
```
explain match (v:player) where id(v) == "Tony Parker" match (v)-[e:like]->(v2) return *
```
![image](https://user-images.githubusercontent.com/38901892/204447143-e2793941-7cbb-4708-86ab-6c683ed528a4.png)

**Your Environments (__required__)**

* OS: `uname -a`
* Compiler: `g++ --version` or `clang++ --version`
* CPU: `lscpu`
* Commit id (e.g. `a3ffc7d8`)

**How To Reproduce(__required__)**

1. Step 1
2. Step 2
3. Step 3

**Expected behavior**

**Additional context**

jievince:
Close this because https://github.com/vesoft-inc/nebula/pull/4950 has fixed it.

PR #4946, opened 2022-11-29
fix explain of argument

jievince:
## What type of PR is this?
- [ ] bug
- [ ] feature
- [ ] enhancement

#### Issue(s) number: 
Fix https://github.com/vesoft-inc/nebula/issues/4945
#### Description:
Because `Argument` inherited from `SingleInputNode`, so it doesn't need to override the `explain` interface
Related pr: https://github.com/vesoft-inc/nebula/pull/4939/files#.

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....

### DIFF
--- a/src/graph/planner/plan/Logic.cpp
+++ b/src/graph/planner/plan/Logic.cpp
@@ -86,10 +86,5 @@ void Argument::cloneMembers(const Argument& arg) {
   PlanNode::cloneMembers(arg);
 }
 
-std::unique_ptr<PlanNodeDescription> Argument::explain() const {
-  auto desc = SingleInputNode::explain();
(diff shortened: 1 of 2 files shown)


Shylock-Hg (approved):

Shylock-Hg:
`Argument` has a member `alias_`

jievince:
> `Argument` has a member `alias_`

It's a little redundant... Because alias is always the same as the column name of the argument output

jievince:
Close this because https://github.com/vesoft-inc/nebula/pull/4950 has fixed it.

PR #4950, opened 2022-11-29
Fix argument

czpmango:
## What type of PR is this?
- [ ] bug
- [ ] feature
- [ ] enhancement

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....


src/graph/planner/plan/Logic.h
```diff
@@ -147,10 +147,10 @@ class PassThroughNode final : public SingleInputNode {
 };
 
 // This operator is used for getting a named alias from another executed operator.
-class Argument final : public SingleInputNode {
+class Argument final : public PlanNode {
```
Shylock-Hg, line 150: Why change this?

czpmango: Treat `Argument` as a leaf node and delete the useless `StartNode`.

yixinglu (approved):
Good Job!

src/graph/planner/plan/Logic.h
Shylock-Hg: Emm. A little wired, node with one input should inherit it.

czpmango: Just ok.

jievince (approved):

Aiee (approved):

PR #4961, opened 2022-11-30
fix ttl not work when ttl column has default value

critical27:
## What type of PR is this?
- [X] bug
- [ ] feature
- [ ] enhancement

#### Issue(s) number: 
close https://github.com/vesoft-inc/nebula/issues/4962

#### Description:
As title, the reason is that we only read the ttl property from RowReader, we need to use latest schema as well for data with older schema version.

## How do you solve it?
`QueryUtils::readValue` already handle this.

Add some comments.

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....

### DIFF
--- a/src/storage/CommonUtils.cpp
+++ b/src/storage/CommonUtils.cpp
@@ -16,6 +16,7 @@ bool CommonUtils::checkDataExpiredForTTL(const meta::SchemaProviderIf* schema,
                                          int64_t ttlDuration) {
   auto v = QueryUtils::readValue(reader, ttlCol, schema);
   if (!v.ok()) {
+    VLOG(3) << "failed to read ttl property";
     return false;
   }
   return checkDataExpiredForTTL(schema, v.value(), ttlCol, ttlDuration);
--- a/src/storage/CommonUtils.cpp
+++ b/src/storage/CommonUtils.cpp
@@ -5,6 +5,8 @@
 
 #include "storage/CommonUtils.h"
 
+#include "storage/exec/QueryUtils.h"
+
 namespace nebula {
 namespace storage {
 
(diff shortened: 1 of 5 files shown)


panda-sheep (approved):
LGTM

pengweisong:
could refer an issue in the community?

src/storage/CommonUtils.cpp
```diff
@@ -5,15 +5,20 @@
...
 namespace storage {
 
 bool CommonUtils::checkDataExpiredForTTL(const meta::SchemaProviderIf* schema,
                                          RowReader* reader,
                                          const std::string& ttlCol,
                                          int64_t ttlDuration) {
-  auto v = reader->getValueByName(ttlCol);
-  return checkDataExpiredForTTL(schema, v, ttlCol, ttlDuration);
+  auto v = QueryUtils::readValue(reader, ttlCol, schema);
+  if (!v.ok()) {
```
pengweisong, line 18: need here log the bad status for debug? or the info will be lost.

pengweisong:
> could refer an issue in the community?

I have fixed it.

tests/tck/features/ttl/TTL.feature
```diff
@@ -490,3 +490,38 @@ Feature: TTLTest
     Then the result should be, in any order:
       | age |
     And drop the used space
+
+  @wtf
```
HarrisChu, line 494: delete the label..

critical27: Sorry... My bad

src/storage/CommonUtils.cpp
critical27: 👌

panda-sheep (approved):

yixinglu (approved):

Issue #4960, opened 2022-11-30
LeftJoin executor outputs the joined columns multi times


jievince:
**Please check the FAQ documentation before raising an issue**

**Describe the bug (__required__)**

Assume the join key of left join is column `a`, it will output two columns called `a`, one comes from the left child, and the other comes from the right child. And because the left join will add a null value for the right side, the two-column `a` may contain a different value.

The inner join will also output two columns `a`, although there is no correctness problem here.

**Your Environments (__required__)**

* OS: `uname -a`
* Compiler: `g++ --version` or `clang++ --version`
* CPU: `lscpu`
* Commit id (e.g. `a3ffc7d8`)

**How To Reproduce(__required__)**

1. Step 1
2. Step 2
3. Step 3

**Expected behavior**

**Additional context**

HarrisChu:
has tck tests in pr now

PR #4965, opened 2022-11-30
fix duplicate output column names of join

jievince:
## What type of PR is this?
- [x] bug
- [ ] feature
- [ ] enhancement

#### Issue(s) number: 
Fix https://github.com/vesoft-inc/nebula/issues/4960
## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....

### DIFF
--- a/src/graph/executor/query/JoinExecutor.cpp
+++ b/src/graph/executor/query/JoinExecutor.cpp
@@ -60,11 +60,10 @@ Status JoinExecutor::checkBiInputDataSets() {
 
   DCHECK_LE(colSize_, lhsColSize + rhsColSize);
   if (colSize_ < lhsColSize + rhsColSize) {
-    rhsOutputColIdxs_ = std::make_unique<std::vector<size_t>>();
     for (size_t i = 0; i < rhsColSize; ++i) {
       auto it = std::find(lhsColNames.begin(), lhsColNames.end(), rhsColNames[i]);
       if (it == lhsColNames.end()) {
-        rhsOutputColIdxs_->push_back(i);
+        rhsOutputColIdxs_.push_back(i);
       }
     }
   }
@@ -108,8 +107,8 @@ Row JoinExecutor::newRow(Row left, Row right) const {
   row.values.insert(row.values.end(),
                     std::make_move_iterator(left.values.begin()),
                     std::make_move_iterator(left.values.end()));
-  if (rhsOutputColIdxs_) {
-    for (auto idx : *rhsOutputColIdxs_) {
+  if (!rhsOutputColIdxs_.empty()) {
+    for (auto idx : rhsOutputColIdxs_) {
       row.values.emplace_back(std::move(right.values[idx]));
     }
   } else {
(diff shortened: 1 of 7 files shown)


src/graph/executor/query/JoinExecutor.h
```diff
@@ -33,7 +33,13 @@ class JoinExecutor : public Executor {
 
   std::unique_ptr<Iterator> lhsIter_;
   std::unique_ptr<Iterator> rhsIter_;
+  // colSize_ is the size of the output columns of join executor.
+  // If the join is natural join, which means the left and right columns have duplicated names,
+  // the output columns will be a intersection of the left and right columns.
   size_t colSize_{0};
+  // If the join is natural join, rhsOutputColIdxs_ is used to record the output column index of the
+  // right
+  std::unique_ptr<std::vector<size_t>> rhsOutputColIdxs_;
```
Shylock-Hg, line 42: `std::vector<size_t>>` directly.

src/graph/executor/query/JoinExecutor.cpp
```diff
@@ -83,15 +103,21 @@ void JoinExecutor::buildSingleKeyHashTable(
...
+  Row row;
+  row.reserve(left.size() + right.size());
+  row.values.insert(row.values.end(),
+                    std::make_move_iterator(left.values.begin()),
+                    std::make_move_iterator(left.values.end()));
+  if (rhsOutputColIdxs_) {
+    for (auto idx : *rhsOutputColIdxs_) {
+      row.values.emplace_back(std::move(right.values[idx]));
+    }
+  } else {
```
Shylock-Hg, line 115: Wired logic. Insert right columns when `rhsOutputColIdxs_` is null?

yixinglu (commented):
Need some TCK cases, such as: `MATCH .. MATCH ... RETURN *`

src/graph/executor/query/JoinExecutor.h
jievince: If it is null, we should directly output the entire row of the right-side input. Otherwise, we should only output the non-duplicated columns of right-side input. And the non-duplicatd column index are recorded in the rhsOutputColIdxs_.

tests/tck/features/match/MultiQueryParts.feature
```diff
@@ -1,6 +1,7 @@
 # Copyright (c) 2021 vesoft inc. All rights reserved.
 #
 # This source code is licensed under Apache 2.0 License.
+@jie
```
yixinglu, line 4: cleanup

src/graph/executor/query/JoinExecutor.h
yixinglu: it seems you could do this by checking whether the vector object is empty

jievince: ok

tests/tck/features/match/MultiQueryParts.feature
jievince: ok

yixinglu (approved):
Good Job!

Shylock-Hg (approved):

PR #4969, opened 2022-12-01
Fix missing columns in orderby

### LINKED ISSUE #4930: order by report column not found while it does exists

Look at the query in Nebula below, Nebula report column `a0` not found, while it does exists:

```txt
(root@nebula) [gdlancer]> MATCH (v)-->(t) WHERE id(v) in [1, 2, 3, 4, 5, 6, 7] WITH v, t, count(t) as a0 WHERE a0 > 1 return id(v) AS r0, id(t) AS r1 ORDER BY a0
[ERROR (-1009)]: SemanticError: Column `a0' not found

Thu, 24 Nov 2022 12:00:26 CST

(root@nebula) [gdlancer]>
```

In contrast, Neo4j return expected result:

```txt
$ MATCH (v)-->(t) WHERE v.id in [1, 2, 3, 4, 5, 6, 7] WITH v, t, count(t) as a0 WHERE a0 > 1 return v.id AS r0, t.id AS r1 ORDER BY a0

╒════╤════╕
│"r0"│"r1"│
╞════╪════╡
│3   │6   │
├────┼────┤
│5   │24  │
├────┼────┤
│6   │23  │
├────┼────┤
│4   │13  │
├────┼────┤
│4   │20  │
├────┼────┤
│2   │2   │
└────┴────┘
```

* OS: `uname -a`
* CPU: `lscpu`
* Commit id (e.g. `a3ffc7d8`) be2c29400

1. Step 1
2. Step 2
3. Step 3

**Expected behavior**

**Additional context**

xtcyclist:
## What type of PR is this?
- [X] bug
- [ ] feature
- [ ] enhancement

#### Issue(s) number: 
Close #4930 

#### Description:
To comply with the openCyper, the scope of columns that an order-by can see depends on the whether the projection before it is changing the cardinality of the data set. In cases when the projection is neither aggregating or going with distinct, the ordery-by shall see columns before the projection. Or, in other words, users can sort the returned columns using a column that is not explicited used in the return clause.

This beheviour has not been implemented before this PR. Currently, order-by can only see columns that are either explicitly declared in the return clause or return *.

## How do you solve it?

- Adapted order-by's validator in the MatchValidator, expanding its scope when needed and feasible.
- When expanded, needed columns are added to the projection.
- These columns shall be invisible to the user. Thus, they are removed from the result dataset after executing Sort or TopN.

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [X] TCK

Affects:

## Release notes:

> ex. Fixed the bug .....

### DIFF
--- a/tests/tck/features/match/OrderBy.feature
+++ b/tests/tck/features/match/OrderBy.feature
@@ -1,0 +1,139 @@
+# Copyright (c) 2020 vesoft inc. All rights reserved.
+#
+# This source code is licensed under Apache 2.0 License.
+Feature: OrderBy in Match
+
+  Background:
+    Given a graph with space named "nba"
+
+  Scenario: orderby hidden column
+    When executing query:
+      """
+      match (v:player)-->(v1)
+      with v, v1, count(v1) as a0
+      where a0 > 1
+      return sum(v.player.age)
+      order by a0
+      limit 5
+      """
+    Then a SemanticError should be raised at runtime: Column `a0' not found
+    When executing query:
+      """
+      match (v:player)-->(v1)
+      with v, v1, count(v1) as a0
+      where a0 > 1
+      return distinct v
+      order by a0
+      limit 5
+      """
+    Then a SemanticError should be raised at runtime: Column `a0' not found
+    When executing query:
+      """
+      match (v:player)-->(v1:player)
+      with v, v1, v.player.name as name, v1.player.name as name2, count(v) as a0
+      where a0 > 1
+      return v, v1
+      order by a0, name, name2
+      """
+    Then the result should be, in order:
+      | v                                                                                                           | v1                                                                                                          |
+      | ("Manu Ginobili" :player{age: 41, name: "Manu Ginobili"})                                                   | ("Tim Duncan" :player{age: 42, name: "Tim Duncan"} :bachelor{name: "Tim Duncan", speciality: "psychology"}) |
+      | ("Tim Duncan" :player{age: 42, name: "Tim Duncan"} :bachelor{name: "Tim Duncan", speciality: "psychology"}) | ("Manu Ginobili" :player{age: 41, name: "Manu Ginobili"})                                                   |
+      | ("Tim Duncan" :player{age: 42, name: "Tim Duncan"} :bachelor{name: "Tim Duncan", speciality: "psychology"}) | ("Tony Parker" :player{age: 36, name: "Tony Parker"})                                                       |
+      | ("Tony Parker" :player{age: 36, name: "Tony Parker"})                                                       | ("LaMarcus Aldridge" :player{age: 33, name: "LaMarcus Aldridge"})                                           |
+      | ("Tony Parker" :player{age: 36, name: "Tony Parker"})                                                       | ("Manu Ginobili" :player{age: 41, name: "Manu Ginobili"})                                                   |
+      | ("Tony Parker" :player{age: 36, name: "Tony Parker"})                                                       | ("Tim Duncan" :player{age: 42, name: "Tim Duncan"} :bachelor{name: "Tim Duncan", speciality: "psychology"}) |
+
+  Scenario: topn hidden column
+    When executing query:
+      """
+      match (v:player)-->(v1:player)
+      with v, v1, v.player.name as name, v1.player.name as name2, count(v) as a0
+      where a0 > 0
+      return name
+      order by a0 desc, name desc, name2 desc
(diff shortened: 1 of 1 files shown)


src/common/datatypes/Value.cpp
```diff
@@ -678,6 +678,11 @@ const DataSet& Value::getDataSet() const {
   return *(value_.gVal);
 }
 
+DataSet& Value::getMutableDataSet() {
+  CHECK_EQ(type_, Type::DATASET);
```
czpmango, line 682: DCHECK

tests/tck/features/match/OrderBy.feature
```diff
@@ -0,0 +1,139 @@
...
+      | ("Tony Parker" :player{age: 36, name: "Tony Parker"})                                                       | ("Tim Duncan" :player{age: 42, name: "Tim Duncan"} :bachelor{name: "Tim Duncan", speciality: "psychology"}) |
+
+  Scenario: topn hidden column
+    When executing query:
+      """
+      match (v:player)-->(v1:player)
+      with v, v1, v.player.name as name, count(v) as a0
+      where a0 > 0
+      return name
+      order by a0 desc, name desc
```
yixinglu, line 54: it seems the `a0` variable is also invisible, why not to report the semantic error?

src/graph/executor/query/SortExecutor.cpp
```diff
@@ -44,6 +44,9 @@ folly::Future<Status> SortExecutor::execute() {
 
   auto seqIter = static_cast<SequentialIter *>(iter);
   std::sort(seqIter->begin(), seqIter->end(), comparator);
+  for (auto &idx : sort->inputVars()[0]->invisibleColIndicies) {
+    result.valuePtr()->getMutableDataSet().removeColumn(idx);
```
czpmango, line 48: Why do this in sort executor？

tests/tck/features/match/OrderBy.feature
xtcyclist: It's not an error. According to cyper's standard, this case is valid. Please this the issue this PR links to, I pasted the standard there.

src/parser/Clauses.h
```diff
@@ -285,9 +286,21 @@ class YieldColumn final {
...
+
+  void setVisible(bool visible) {
+    visible_ = visible;
+  }
+
  private:
   Expression *expr_{nullptr};
   std::string alias_;
+  // If some columns are required by order-by and are not explicitly
+  // declared in the return caluse, it is an invisible column and not to be
```
czpmango, line 301: IMHO, this is not a good fix.  Better to do that in optimizer,such as column pruning rule.

src/common/datatypes/Value.cpp
xtcyclist: All checks in this Value.cpp are `CHECK_EQ`. I discussed with @jievince, it seems `DCHECK_EQ` is enough here. Let's leave it for another PR.

xtcyclist: https://github.com/vesoft-inc/nebula/issues/4971

src/parser/Clauses.h
xtcyclist: I don't think so. It is not an optimization issue to me. It's a functional issue. Because this type of query shall be executable without the optimizer optimizing the plan.

Previously the validator reports column-not-found. Now it expands the scope for order-by when needed and possible by selecting required columns and adding them in the query plan. These columns are not to be expected in the result set. So I mark them to be invisible in their respective YieldColumn.

czpmango: You just need to make the filter execute ahead of time so that it can see all columns and the unexpected columns will not be projected in result set, which is also consistent with cypher's semantic definition of `with-where clause`:

image

src/common/datatypes/Value.cpp
czpmango: ACK.

src/parser/Clauses.h
xtcyclist: Discussed offline. To summarize, we'd better prepare all data for potential aggregation and orderby before the final projection, instead of passing hidden column beneath the projection.

xtcyclist:
We have decided not to address this compatibility (with openCypher) issue. Order-by won't support sorting by columns not listed in the return clause. If a query wants to sort its results, it has to do so on a column listed in the return clause. BTW, there is no correctness issues on order-by.

PR #4993, opened 2022-12-06
fix no print to console

panda-sheep:
## What type of PR is this?
- [x] bug
- [ ] feature
- [ ] enhancement

#### Description:
fix the enterprise version problem(close https://github.com/vesoft-inc/nebula-ent/issues/1813 ), the community version also exists, just submit it in the community version:

1 Before the `FLAGS_log_dir` directory is generated, you can use `LOG(ERROR)` to print errors to standard output, but you cannot use `vlog` to print log information. The reasons are as follows:

Calling `vlog` before creating the `FLAGS_log_dir` directory will cause the following error message:
```
Could not create logging file: No such file or directory
COULD NOT CREATE A LOGGINGFILE 20221205-164038.3542522!
```
This information is not what we want to output.

Calling `LOG(ERROR)` before creating the `FLAGS_log_dir` directory will cause an error message similar to the following:
```
WARNING: Logging before InitGoogleLogging() is written to STDERR
E20221205 17:31:35.309763 3455851 StorageDaemon.cpp:71] pids_listener/nebula-storaged.pid: Success
```
Here, because there is no `FLAGS_log_dir` directory, it is reasonable to print the error to the standard err.

2 `vlog` Information in `FileUtils::fileType` is redundant and meaningless, so remove them.

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....

### DIFF
--- a/src/common/fs/FileUtils.cpp
+++ b/src/common/fs/FileUtils.cpp
@@ -158,12 +158,9 @@ FileType FileUtils::fileType(const char* path) {
   struct stat st;
   if (lstat(path, &st)) {
     if (errno == ENOENT) {
-      VLOG(3) << "The path \"" << path << "\" does not exist";
       return FileType::NOTEXIST;
     } else {
-      // Failed o get file stat
-      VLOG(3) << "Failed to get information about \"" << path << "\" (" << errno
-              << "): " << strerror(errno);
+      // Failed to get file stat
       return FileType::UNKNOWN;
     }
   }
--- a/src/daemons/MetaDaemon.cpp
+++ b/src/daemons/MetaDaemon.cpp
@@ -79,10 +79,6 @@ int main(int argc, char* argv[]) {
   if (FLAGS_enable_ssl || FLAGS_enable_meta_ssl) {
     folly::ssl::init();
   }
-  if (FLAGS_data_path.empty()) {
-    LOG(ERROR) << "Meta Data Path should not empty";
-    return EXIT_FAILURE;
-  }
 
   // Setup logging
   status = setupLogging(argv[0]);
@@ -104,6 +100,11 @@ int main(int argc, char* argv[]) {
     return EXIT_FAILURE;
   }
 #endif
+
+  if (FLAGS_data_path.empty()) {
+    LOG(ERROR) << "Meta Data Path should not empty";
+    return EXIT_FAILURE;
+  }
 
   // Init stats
   nebula::initMetaStats();


src/common/fs/FileUtils.cpp
```diff
@@ -158,12 +158,9 @@ FileType FileUtils::fileType(const char* path) {
   struct stat st;
   if (lstat(path, &st)) {
     if (errno == ENOENT) {
-      VLOG(3) << "The path \"" << path << "\" does not exist";
       return FileType::NOTEXIST;
     } else {
-      // Failed o get file stat
-      VLOG(3) << "Failed to get information about \"" << path << "\" (" << errno
```
liwenhui-soul, line 165: it's just of different levels, why do you think it's redundant?

panda-sheep: The reasons are as follows:
1) Every call of FileUtils::fileType in the project will make a judgment and then output an error message.

2) This function can be regarded as the most basic function, it is best not to print error messages in it, and print them at the calling place.
For example, the current problem to be solved is that when creating a directory in setupLogging, it is normal that the directory does not exist, and the information should not be printed.

liwenhui-soul: this is expectant behavior, not redundant, you can't remove a low level log just because there's a similar one in a high level.

panda-sheep: I do not think so.

SuperYoko:
Maybe clone an issue here to make all your links accessable.

src/daemons/MetaDaemon.cpp
```diff
@@ -105,6 +101,11 @@ int main(int argc, char* argv[]) {
   }
 #endif
 
+  if (FLAGS_data_path.empty()) {
```
SuperYoko, line 104: Maybe give a check on standalone daemon.

src/common/fs/FileUtils.cpp
SuperYoko: I do think it's okay to let users print its error. This do not leak error info, if someone use this function, and leak this info, that's the problem. (And as you can see there's some path that error info is lost midway...).

SuperYoko: Talked offline with @liwenhui-soul , if this message explain this return value , I'd like it to be reserved.

liwenhui-soul: > I do not think so.

if you remove the log, which place can distinguish errno == ENOENT or not?

panda-sheep: Please take a look at the function `FileUtils::makeDir` and this issue to understand.

src/daemons/MetaDaemon.cpp
panda-sheep: The standalone daemon has been checked.

panda-sheep:
> Maybe clone an issue here to make all your links accessable.

okey

src/common/fs/FileUtils.cpp
liwenhui-soul: please don't resolve conversation before it is really solved

SuperYoko (approved):

panda-sheep:
close https://github.com/vesoft-inc/nebula/issues/4994

pengweisong (approved):

liwenhui-soul:
don't merge plz, the log could not be removed directly. have talked with qa, this issue don't need to fix.

critical27:
Not a big issue, print some unexpected log before setting up glog is reasonable. And this log is a v3 log, we won't use it for any production env.

PR #4995, opened 2022-12-06
Feature/test data

Shylock-Hg:
## What type of PR is this?
- [ ] bug
- [ ] feature
- [ ] enhancement

#### Description:
Generated by https://github.com/vesoft-inc/nebula-test/tree/app-fuzzer/functional/GDLancer . If you have some suggestions, submit issue in https://github.com/vesoft-inc/nebula-test

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....


tests/tck/features/optimizer/PrunePropertiesRule.feature
```diff
@@ -19,12 +19,12 @@ Feature: Prune Properties rule
...
-      | 8  | Project        | 4            |                                                                                                  |
-      | 4  | AppendVertices | 3            | {  "props": "[{\"props\":[\"age\"],\"tagId\":9}]" }                                              |
-      | 3  | Traverse       | 7            | {  "vertexProps": "", "edgeProps": "[{\"props\":[\"_dst\", \"_rank\", \"_type\"],\"type\":3}]" } |
-      | 7  | IndexScan      | 2            |                                                                                                  |
-      | 2  | Start          |              |                                                                                                  |
+      | id | name           | dependencies | operator info                                                                                     |
+      | 8  | Project        | 4            |                                                                                                   |
+      | 4  | AppendVertices | 3            | {  "props": "[{\"props\":[\"age\"],\"tagId\": 3}]" }                                              |
```
yixinglu, line 24: maybe we should remove the `tagId` from the tests since it depends on the runtime environment.

Shylock-Hg: Agree, better display name instead of Id. Could do it in later PR.

tests/common/types.py
```diff
@@ -98,7 +98,7 @@ class Prop(Column):
     def __init__(self, index: int, name: str, ptype: str):
         super().__init__(index)
         self._name = name
-        if ptype not in ['string', 'int', 'double']:
+        if ptype not in ['string', 'int', 'double', 'bool']:
```
codesigner, line 101: So, there is no time props?

codesigner: Or any complex props, like List, Map?

Shylock-Hg: Yes, the generator doesn't support temporal types, you could submit an issue. Nebula doesn't support list/map as property.

codesigner: > Yes, the generator doesn't support temporal types, you could submit an issue. Nebula doesn't support list/map as property.

This is not what we required previously, this need further refinement in the fufure.

codesigner (approved):

yixinglu (approved):

PR #4956, opened 2022-11-29
Push filter down HashInnerJoin rule

yixinglu:
## What type of PR is this?
- [X] bug
- [ ] feature
- [X] enhancement

#### Issue(s) number: 

fix #4979 

## Checklist:
Tests:
- [ ] Unit test(positive and negative cases)
- [ ] Function test
- [X] Performance test
- [ ] N/A

Affects:
- [ ] Documentation affected (Please add the label if documentation needs to be modified.)
- [ ] Incompatibility (If it breaks the compatibility, please describe it and add the label.）
- [ ] If it's needed to cherry-pick (If cherry-pick to some branches is required, please label the destination version(s).)
- [ ] Performance impacted: Consumes more CPU/Memory

## Release notes:

Please confirm whether to be reflected in release notes and how to describe:
> ex. Fixed the bug .....

PR #5008, opened 2022-12-06
Fix label tag expression toString function

yixinglu:
## What type of PR is this?
- [ ] bug
- [ ] feature
- [X] enhancement

#### Description:

It's confusing for the following expression string, we should not delete the leading character for all expression.

```
-.n1.player.age
```

## How do you solve it?

keep the `$` character for the `$-`/`$$`/`$^` expressions.

In addition, more tests for PushFilterDownHashInnerJoinRule introduced in #4956 

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [x] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....


Aiee (approved):

Sophie-Xie:
@yixinglu plz fix the error of format.

src/graph/service/QueryInstance.cpp
```diff
@@ -93,7 +93,8 @@ Status QueryInstance::validateAndOptimize() {
   // Optimize the query, and get the execution plan. We should not pass the optimizer errors to user
   // since the message is often not easy to understand. Logging them is enough.
   if (auto status = findBestPlan(); !status.ok()) {
-    LOG(ERROR) << "Error found in optimization stage: " << status.message();
+    LOG(ERROR) << "Error found in optimization stage for query: " << rctx->query()
+               << ", error: " << status.message();
     return Status::Error(
         "There are some errors found in optimizer, "
         "please contact to the admin to learn more details");
```
czpmango, line 100: The error message on line 100 is not appropriate, it does not prevent the error message from being passed to the user, and it is more confusing than the original error message.

src/common/expression/PropertyExpression.cpp
```diff
@@ -209,8 +209,20 @@ void LabelTagPropertyExpression::accept(ExprVisitor* visitor) {
...
+    labelStr = label_->toString();
+    // Remove the leading '$' character for variable except '$-/$$/$^'
+    if (labelStr.find(kInputRef) != 0 && labelStr.find(kSrcRef) != 0 &&
+        labelStr.find(kDstRef) != 0) {
+      labelStr.erase(0, 1);
+    }
+    labelStr += ".";
+  }
+  if (!sym_.empty()) {
+    labelStr += sym_ + ".";
```
czpmango, line 223: Better to use `:`.

czpmango (approved):

src/graph/service/QueryInstance.cpp
yixinglu: Actually, I am not sure what's the better way to handle the optimizer errors. I need to open an issue about this and research how to do that in other db.

src/common/expression/PropertyExpression.cpp
yixinglu: like the following format?

```
$-:player.age
$$:player.name
v:player.name
```

src/graph/service/QueryInstance.cpp
yixinglu: #5033

jievince (approved):

nevermore3 (approved):

Issue #5012, opened 2022-12-07
Balance data proceeds when there are running br backup job


nebula-bots:
**Please check the FAQ documentation before raising an issue**

**Describe the bug (__required__)**

1 We first start a br backup, and before it finish
2. we start a balance data job migrating data from to another two new nodes

We expect the balance data jobs to report an error before start, but it did proceeds, the backup data might be inconsistent in this situation.

**Your Environments (__required__)**

* OS: `uname -a`
* Compiler: `g++ --version` or `clang++ --version`
* CPU: `lscpu`
* Commit id (e.g. `a3ffc7d8`) https://github.com/vesoft-inc/nebula-ent/commit/af513f9621154870e02ee865e67bf5bed5326b77 release-3.3

**How To Reproduce(__required__)**

1. Step 1
2. Step 2
5. Step 3

**Expected behavior**

**Additional context**

Issue #4975, opened 2022-12-02
BR proceeds when there are running jobs


nebula-bots:
**Please check the FAQ documentation before raising an issue**

**Describe the bug (__required__)**

We use br to backup the cluster when there are running balance data jobs(we ensure that the job is started before br backup and still running after backup finished), we expecte br to report error and exit, but it just proceeds and continue backup, the backup data might be inconsistent in this situation:

```txt
"time":"2022-12-02T15:29:46.055Z","uri":"s3://test/graph1/BACKUP_2022_12_02_15_29_16/data/71/103"}
{"file":"github.com/vesoft-inc/nebula-agent@v0.1.2-0.20220615054614-af5197e7d961/pkg/client/client.go:49","func":"github.com/vesoft-inc/nebula-agent/pkg/client.New","level":"debug","msg":"Dialing to address store5:8888.","time":"2022-12-02T15:29:46.055Z"}
{"file":"github.com/vesoft-inc/nebula-agent@v0.1.2-0.20220615054614-af5197e7d961/pkg/storage/s3.go:312","func":"github.com/vesoft-inc/nebula-agent/pkg/storage.(*S3).GetDir","level":"debug","msg":"Get backend for s3://test/graph1/BACKUP_2022_12_02_15_29_16/data/71/133 successfully.","time":"2022-12-02T15:29:46.055Z","uri":"s3://test/graph1/BACKUP_2022_12_02_15_29_16/data/71/133"}
{"file":"github.com/vesoft-inc/nebula-agent@v0.1.2-0.20220615054614-af5197e7d961/pkg/client/client.go:49","func":"github.com/vesoft-inc/nebula-agent/pkg/client.New","level":"debug","msg":"Dialing to address store1:8888.","time":"2022-12-02T15:29:46.055Z"}
{"file":"github.com/vesoft-inc/nebula-agent@v0.1.2-0.20220615054614-af5197e7d961/pkg/storage/s3.go:312","func":"github.com/vesoft-inc/nebula-agent/pkg/storage.(*S3).GetDir","level":"debug","msg":"Get backend for s3://test/graph1/BACKUP_2022_12_02_15_29_16/data/71/186 successfully.","time":"2022-12-02T15:29:46.056Z","uri":"s3://test/graph1/BACKUP_2022_12_02_15_29_16/data/71/186"}
[...]

PR #5013, opened 2022-12-07
Check running job before create backup

pengweisong:
## What type of PR is this?
- [x] bug
- [ ] feature
- [ ] enhancement

#### Issue(s) number: 
https://github.com/vesoft-inc/nebula/issues/5012
https://github.com/vesoft-inc/nebula/issues/4975
## How do you solve it?
1. add more job type to check
2. add some test

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....


src/meta/processors/admin/CreateBackupProcessor.cpp
```diff
@@ -84,7 +84,11 @@ void CreateBackupProcessor::process(const cpp2::CreateBackupReq& req) {
...
   std::unordered_set<cpp2::JobType> jobTypes{cpp2::JobType::REBUILD_TAG_INDEX,
-                                             cpp2::JobType::REBUILD_EDGE_INDEX};
+                                             cpp2::JobType::REBUILD_EDGE_INDEX,
+                                             cpp2::JobType::COMPACT,
+                                             cpp2::JobType::INGEST,
+                                             cpp2::JobType::DATA_BALANCE,
+                                             cpp2::JobType::LEADER_BALANCE};
   auto result = jobMgr->checkTypeJobRunning(jobTypes);
   if (!nebula::ok(result)) {
     LOG(INFO) << "Get Index status failed, not allowed to create backup.";
```
panda-sheep, line 94: Maybe this log need to be changed？

panda-sheep (commented):
Good job.
I have a little problem.
In the second step of `CreateBackupProcessor::process`, if a `blocking_` lock is added to the part of the storage engine, writing is not allowed, but it will not block the execution of the job? Such as compact job

src/interface/common.thrift
```diff
@@ -403,7 +403,7 @@ enum ErrorCode {
     E_INVALID_JOB                     = -2065,  // Invalid task
 
     // Backup Failure
-    E_BACKUP_BUILDING_INDEX           = -2066,  // Backup terminated (index being created)
+    E_BACKUP_RUNNING_JOBS             = -2066,  // Backup terminated (index being created)
```
yixinglu, line 406: plz update the error code comment and revisit the compatible issue caused by this thrift change

yixinglu (commented):
Good enhancement!

pengweisong:
> This PR is to prevent BR running when there are other jobs, which is definitely a must. But in the issue, it is the other way around: stop the Balance job running when there are backup jobs. Will it be solved somewhere else?

No, but since BR  use hard link instead of copy data, it will be very fast. But indeed, we may need sth to handle the corner case.

src/meta/processors/admin/CreateBackupProcessor.cpp
```diff
@@ -93,8 +97,9 @@ void CreateBackupProcessor::process(const cpp2::CreateBackupReq& req) {
     return;
   }
   if (nebula::value(result)) {
-    LOG(INFO) << "Index is rebuilding, not allowed to create backup.";
-    handleErrorCode(nebula::cpp2::ErrorCode::E_BACKUP_BUILDING_INDEX);
+    LOG(INFO) << "Index rebuilding/Compact/Ingest/Balace job is running or queue, not allowed to "
```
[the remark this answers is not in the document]
pengweisong: It's hard to tell which type it is here, I'll try to log the specified type in the deeper code.

src/meta/processors/admin/CreateBackupProcessor.cpp
pengweisong: indeed, thanks

pengweisong:
> Good job. I have a question. In the second step of `CreateBackupProcessor::process`, if a `blocking_` lock is added to the part of the storage engine, writing is not allowed, but it will not block the execution of the job? Such as compact job

Ineed, then we should:
1. check when we start backup, there is no these jobs.
2. in the backup process we should prevent relevant jobs starting to run.

For the 1, we already check. The latter is the same problem raised by wenhao. We may could use an lock on space.

src/interface/common.thrift
pengweisong: good catch, thanks.

src/meta/processors/admin/CreateBackupProcessor.cpp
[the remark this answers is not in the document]
codesigner: plz try to print out which job is queued or running, block the backup, this LOG is not informative, do not let who read this log guess what is happening.

src/clients/meta/MetaClient.cpp
```diff
@@ -933,8 +933,8 @@ Status MetaClient::handleResponse(const RESP& resp) {
       return Status::Error("Task report is out of date!");
     case nebula::cpp2::ErrorCode::E_BACKUP_FAILED:
       return Status::Error("Backup failure!");
-    case nebula::cpp2::ErrorCode::E_BACKUP_BUILDING_INDEX:
-      return Status::Error("Backup building indexes!");
+    case nebula::cpp2::ErrorCode::E_BACKUP_RUNNING_JOBS:
+      return Status::Error("Backup encounter running or queue jobs!");
```
codesigner, line 937: "queued jobs" looks better?

src/meta/processors/admin/CreateBackupProcessor.cpp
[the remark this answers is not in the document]
pengweisong: OK, fixed it

critical27 (dismissed):

codesigner (approved):

critical27 (approved):

SuperYoko (approved):

yixinglu (dismissed):

critical27 (approved):

SuperYoko (approved):

PR #5018, opened 2022-12-07
Enhance the AppendVertices operator to avoid unnecessary rpc overhead

czpmango:
Do not make an rpc request if no need to fetch properties.

## What type of PR is this?
- [ ] bug
- [ ] feature
- [x] enhancement

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....

### DIFF
--- a/src/graph/planner/plan/Query.h
+++ b/src/graph/planner/plan/Query.h
@@ -346,6 +346,25 @@ class GetVertices : public Explore {
     return props_.get();
   }
 
+  bool noNeedFetchProp() const {
+    if (props_.get() == nullptr) {
+      return true;
+    }
+    auto& vprops = *props_;
+    for (const auto& vprop : vprops) {
+      auto& props = vprop.get_props();
+      if (props.size() > 1) {
+        return false;
+      }
+      DCHECK_EQ(props.size(), 1);
+      auto& prop = props.front();
+      if (prop.compare("_tag")) {
+        return false;
+      }
+    }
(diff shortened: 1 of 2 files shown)


yixinglu (dismissed):

src/graph/planner/plan/Query.h
```diff
@@ -346,6 +346,19 @@ class GetVertices : public Explore {
...
+  bool noNeedFetchProp() const {
+    if (props_.get() == nullptr) {
+      return true;
+    }
+    auto& props = *props_;
+    for (const auto& prop : props) {
+      if (!prop.get_props().empty()) {
+        return false;
+      }
+    }
```
jievince, line 358: _tag is also a prop

src/graph/planner/plan/Query.h
```diff
@@ -346,6 +346,25 @@ class GetVertices : public Explore {
...
   }
 
+  bool noNeedFetchProp() const {
+    if (props_.get() == nullptr) {
+      return true;
+    }
+    auto& vprops = *props_;
+    for (const auto& vprop : vprops) {
+      auto& props = vprop.get_props();
+      if (props.size() > 1) {
```
yixinglu, line 356: check whether props is empty.

yixinglu (approved):

nevermore3 (approved):

PR #5024, opened 2022-12-08
could not add a host that is not a storage

liwenhui-soul:
## What type of PR is this?
- [x] bug
- [ ] feature
- [ ] enhancement

#### Description:
start nebula graph with address 192.168.8.134:9669
add hosts 192.168.8.134:9669
create space
expected:
only store data in storage

actual:
partitions are wrong.
image

## How do you solve it?
when add a host, check if the host is storage, if not, it can't be added

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....

### DIFF
--- a/src/clients/meta/MetaClient.cpp
+++ b/src/clients/meta/MetaClient.cpp
@@ -965,6 +965,8 @@ Status MetaClient::handleResponse(const RESP& resp) {
       return Status::Error("There are still space on the host");
     case nebula::cpp2::ErrorCode::E_RELATED_FULLTEXT_INDEX_EXISTS:
       return Status::Error("Related fulltext index exists, please drop it first");
+    case nebula::cpp2::ErrorCode::E_HOST_CAN_NOT_BE_ADDED:
+      return Status::Error("Could not add a host, which is not a storage and not expired either");
     default:
       return Status::Error("Unknown error!");
   }
--- a/src/meta/processors/zone/AddHostsProcessor.cpp
+++ b/src/meta/processors/zone/AddHostsProcessor.cpp
@@ -50,6 +50,22 @@ void AddHostsProcessor::process(const cpp2::AddHostsReq& req) {
     spaceIter->next();
   }
   for (auto& host : hosts) {
+    auto hostKey = MetaKeyUtils::hostKey(host.host, host.port);
+    auto hostRes = doGet(hostKey);
+    // check if the host is storage, if it is not storage, it should be found in host table not
+    // expired, so it can't be added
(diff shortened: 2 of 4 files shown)


src/meta/processors/zone/AddHostsProcessor.cpp
```diff
@@ -50,6 +50,22 @@ void AddHostsProcessor::process(const cpp2::AddHostsReq& req) {
...
+    auto hostKey = MetaKeyUtils::hostKey(host.host, host.port);
+    auto hostRes = doGet(hostKey);
+    // check if the host is storage, if it is not storage, it should be found in host table not
+    // expired, so it can't be added
+    if (nebula::ok(hostRes)) {
+      auto now = time::WallClock::fastNowInMilliSec();
+      auto hostInfo = HostInfo::decode(nebula::value(hostRes));
+      if (hostInfo.role_ != cpp2::HostRole::STORAGE &&
+          now - hostInfo.lastHBTimeInMilliSec_ <
+              FLAGS_heartbeat_interval_secs * FLAGS_expired_time_factor * 1000) {
```
yixinglu, line 62: u also need to handle the host which is not storage and is expired.

yixinglu: what about add the graphd host ?

liwenhui-soul: if it's not expired, it can't be added. but if it's expired, it can be added, because it may be changed to storage

src/clients/meta/MetaClient.cpp
```diff
@@ -965,6 +965,8 @@ Status MetaClient::handleResponse(const RESP& resp) {
       return Status::Error("Related index exists, please drop index first");
     case nebula::cpp2::ErrorCode::E_RELATED_SPACE_EXISTS:
       return Status::Error("There are still space on the host");
+    case nebula::cpp2::ErrorCode::E_HOST_CAN_NOT_BE_ADDED:
+      return Status::Error("Host can not be added");
```
critical27, line 969: Could add some more explicitly error message about why host can not be added.

src/meta/processors/zone/AddHostsProcessor.cpp
liwenhui-soul: yes, if we don't allow to add a expired graphd, it would never be added even if user change it to storage

critical27 (approved):

yixinglu (approved):

Issue #5027, opened 2022-12-09
Use listenerId to distinguish multiple add listeners to one space


nebula-bots:
Use listenerId to distinguish multiple add listeners to one space
Can` add listener `immediately after `remove listener`, no need to wait for a while.

PR #5032, opened 2022-12-09
Use listenerId to distinguish multiple add listeners to one space

panda-sheep:
## What type of PR is this?
- [ ] bug
- [x] feature
- [ ] enhancement

#### Description:
1. Use listenerId to distinguish different add listeners.
Synchronized from the enterprise version https://github.com/vesoft-inc/nebula-ent/pull/1860.
solve the problem of adding listener immediately after remove listener.
close https://github.com/vesoft-inc/nebula/issues/5027

2. Modify the code to keep the community version and enterprise version consistent
and optimize some code to make the code cleaner.

3. Add UT and improve UT code

## How do you solve it?
Because remove listener and add listener are asynchronous operations.
In the past, after removing the listener, you need to wait for a while to add the listener, otherwise an error will occur.
After the current modification, after remove listener, you can add listener immediately.

Test case as follows:

step 1: 
create space space1(partition_num=1, replica_factor=1, vid_type = FIXED_STRING(30));
use space1
create tag player(name string, age int);
insert vertex player(name, age) values "11":("aaaa",1); 
insert vertex player(name, age) values "22":("aaaa",2); 
insert vertex player(name, age) values "33":("aaaa",3); 
insert vertex player(name, age) values "44":("aaaa",4); 
FETCH PROP ON player "11" yield player.name;
sign in text service(192.168.8.215:47500);
show text search clients;

step2:
add listener ELASTICSEARCH 192.168.8.211:56700;
show listener

step 3:
REMOVE LISTENER ELASTICSEARCH;
show listener 
add listener ELASTICSEARCH 192.168.8.211:56700;
show listener 

step 4:
The value in the ListenerId file is different, meet our expectations.

The first ListenerId：
:~$ ll
18:35 listnenerId
:~/nebula_test_tool$ ./readListenerId    xxxx/storage_listener/listener/1/1/wal/listnenerId
listenerId 3

The second ListenerId：
:~$ ll
18:38 listnenerId
:~/nebula_test_tool$ ./readListenerId  xxxxx/storage_listener/listener/1/1/wal/listnenerId
listenerId 4

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....


src/clients/meta/MetaClient.cpp
```diff
@@ -607,15 +607,15 @@ bool MetaClient::loadIndexes(GraphSpaceID spaceId, std::shared_ptr<SpaceInfoCach
 }
 
 bool MetaClient::loadListeners(GraphSpaceID spaceId, std::shared_ptr<SpaceInfoCache> cache) {
-  auto listenerRet = listListener(spaceId).get();
+  auto listenerRet = listListeners(spaceId).get();
```
panda-sheep, line 610: The modification here is to be consistent with the enterprise version.

src/common/utils/MetaKeyUtils.cpp
```diff
@@ -1190,10 +1192,16 @@ GraphSpaceID MetaKeyUtils::parseListenerSpace(folly::StringPiece rawData) {
 }
 
 PartitionID MetaKeyUtils::parseListenerPart(folly::StringPiece rawData) {
-  auto offset = kListenerTable.size() + sizeof(meta::cpp2::ListenerType) + sizeof(GraphSpaceID);
+  auto offset = kListenerTable.size() + sizeof(GraphSpaceID) + sizeof(meta::cpp2::ListenerType);
   return *reinterpret_cast<const PartitionID*>(rawData.data() + offset);
 }
 
+ListenerID MetaKeyUtils::parseListenerId(folly::StringPiece rawData) {
```
critical27, line 1199: Perhaps you need to handle the old existing data which does not have listener id.

critical27 (commented):
What will happen when this PR merged:
1. Does existing listener could start up by `NebulaStore::loadLocalListenerFromPartManager`? 
2. If 1 works, what is the listener id of existing listener? 

I didn't find any related logic, I guess we will crash.

critical27:
The listener id's problem (what you discussed that we need remove -> wait -> add), I think we could leave it alone and mark as known issue. My concern is that this PR only works for new cluster but not the existing one.

src/common/utils/MetaKeyUtils.cpp
panda-sheep: Good catch.  
Compatibility has already supported.

panda-sheep:
> What will happen when this PR merged:
> 
> 1. Does existing listener could start up by `NebulaStore::loadLocalListenerFromPartManager`?
> 2. If 1 works, what is the listener id of existing listener?
> 
> I guess we will crash.
 
1） Existing listener could start up by NebulaStore::loadLocalListenerFromPartManager.
2)   Before the upgrade, if there is listenr,  after the upgrade, use the default listenerId for compatibility.

panda-sheep:
> guess we will crash.

There is a listener before the upgrade, and it can continue to execute after the upgrade. already tested.@critical27

src/interface/meta.thrift
```diff
@@ -1271,9 +1272,9 @@ service MetaService {
     GetZoneResp    getZone(1: GetZoneReq req);
     ListZonesResp  listZones(1: ListZonesReq req);
 
-    ExecResp       addListener(1: AddListenerReq req);
-    ExecResp       removeListener(1: RemoveListenerReq req);
-    ListListenerResp listListener(1: ListListenerReq req);
+    ExecResp          addListener(1: AddListenerReq req);
+    ExecResp          removeListener(1: RemoveListenerReq req);
+    ListListenersResp listListeners(1: ListListenersReq req);
```
yixinglu, line 1277: Does the change of this interface affect the compatibility of different nebula version?

Issue #5034, opened 2022-12-09
Provides a startup script for the listener


nebula-bots:
**Is your feature request related to a problem? Please describe.**

Sometimes you encounter relative path problems, consider adding a listener startup script, similar to nebula-drainerd.service.

**Describe the solution you'd like**
nebula-listener.service start storaged
nebula-listener.service start metad

**Describe alternatives you've considered**

**Additional context**

PR #5035, opened 2022-12-09
support storaged-listener script startup

panda-sheep:
## What type of PR is this?
- [ ] bug
- [ ] feature
- [ ] enhancement

#### Description:
Sync pr from enterprise version pr https://github.com/vesoft-inc/nebula-ent/pull/1275
close https://github.com/vesoft-inc/nebula/issues/5034

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....

### DIFF
--- a/scripts/CMakeLists.txt
+++ b/scripts/CMakeLists.txt
@@ -98,3 +98,16 @@ install(
     COMPONENT
         meta
 )
+
+install(
+    FILES
+        nebula-storaged-listener.service
+    PERMISSIONS
+        OWNER_EXECUTE OWNER_WRITE OWNER_READ
+        GROUP_EXECUTE GROUP_READ
+        WORLD_EXECUTE WORLD_READ
+    DESTINATION
+        scripts
+    COMPONENT
+        storage
+)
--- a/scripts/nebula-storaged-listener.service
+++ b/scripts/nebula-storaged-listener.service
@@ -1,0 +1,6 @@
+#! /bin/bash
+
+SCRIPT_PATH=$(readlink -f $0)
+SCRIPT_DIR=$(dirname ${SCRIPT_PATH})
+export USAGE_INFO="${0} [-v] [-c /path/to/conf] <start|stop|restart|status|kill>"
+${SCRIPT_DIR}/nebula.service $@ storaged-listener
--- a/scripts/nebula.service
+++ b/scripts/nebula.service
@@ -95,6 +95,9 @@ case ${TARGET} in
     storaged)
         TARGETS=(${TARGET})
         ;;
+    storaged-listener)
+        TARGETS=(${TARGET})
+        ;;
     standalone)
         TARGETS=(${TARGET})
         ;;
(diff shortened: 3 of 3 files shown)


SuperYoko (approved):

scripts/nebula-storaged-listener.service
```diff
@@ -0,0 +1,6 @@
+#! /bin/bash
+
+SCRIPT_PATH=$(readlink -f $0)
```
liwenhui-soul, line 3: I think you'd better to check if SCRIPT_PATH is soft link in a while loop, for $(readlink -f $SCRIPT_PATH) may also be a soft link

liwenhui-soul: ok good

scripts/CMakeLists.txt
```diff
@@ -98,3 +98,16 @@ install(
     COMPONENT
         meta
 )
+
+install(
+    FILES
+        nebula-storaged-listener.service
```
pengweisong, line 104: will it have nebula-metad-listener.service?

scripts/nebula-storaged-listener.service
```diff
@@ -0,0 +1,6 @@
...
+SCRIPT_DIR=$(dirname ${SCRIPT_PATH})
+export USAGE_INFO="${0} [-v] [-c /path/to/conf] <start|stop|restart|status|kill>"
+${SCRIPT_DIR}/nebula.service $@ storaged-listener
```
liwenhui-soul, line 6: this way to handle listener is not necessary, checking whether the daemon is listener or not is the option listener_path, so we don't need the para storaged-listener, just  storaged is enough. and we also don't ned to handle storaged-listener in nebula.sevice.

scripts/CMakeLists.txt
panda-sheep: The enterprise version has it, but the community version does not have a meta listener

scripts/nebula-storaged-listener.service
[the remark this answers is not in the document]
panda-sheep: Be consistent with other scripts.

scripts/nebula-storaged-listener.service
panda-sheep: I think the scripted way is simpler.

scripts/nebula.service
```diff
@@ -119,6 +122,9 @@ function start_daemon {
     local daemon_name=nebula-${1}
     local config=${2}
     local executable=${INSTALL_ROOT}/bin/${daemon_name}
+     if [[ $daemon_name == *-listener ]]; then
+        executable=${INSTALL_ROOT}/bin/${daemon_name%-listener*}
```
liwenhui-soul, line 126: ditto, nebula-storaged-listener.service don't need to send the para storaged-listener, and this seems redundant.

pengweisong (approved):

Issue #4982, opened 2022-12-04
Restrict the max depth of plan tree in optimizer to avoid stack overflow


yixinglu:
**Introduction**

sometimes there is a too long query for optimizer to handle. because the optimizer of nebula use recursive functions to find the best plan currently, this maybe cause the stack overflow crash.

It's better to restrict the max depth of execution plan and response error if the plan tree exceed the threshold value. 

The long queries are often generated by some applications and that means the queries need some manual optimizations.

**Contents**

**Related work**

PR #5050, opened 2022-12-13
Restrict the max depth of plan tree in optimizer

jievince:
## What type of PR is this?
- [ ] bug
- [ ] feature
- [x] enhancement

#### Issue(s) number: 
Fix https://github.com/vesoft-inc/nebula/issues/4982.
Related https://github.com/vesoft-inc/nebula-ent/issues/1487.
#### Description:

In most cases, we can considier to constrict the depth of the plan tree to avoid stack overflow crash caused by some bad queries.
But for some queries such as `statement1 interest statement2 interest statement3 ...`,
which is a very normal statement, even if there are many statements to intersect,
we may need to consider the multi-way operator in planner to avoid the overflow.
e.g.

The binary intersect operator:
```
    intersect
    /       \
intersect   c 
  /     \
a       b
```

The multi-way intersect operator:

```
    intersect
    /   |    \
   a    b     c
```

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....

### DIFF
--- a/src/graph/optimizer/Optimizer.cpp
+++ b/src/graph/optimizer/Optimizer.cpp
@@ -22,6 +22,7 @@ using nebula::graph::Select;
 using nebula::graph::SingleDependencyNode;
 
 DEFINE_bool(enable_optimizer_property_pruner_rule, true, "");
+DEFINE_uint64(max_plan_depth, 512, "The max depth of plan tree");
 
 namespace nebula {
 namespace opt {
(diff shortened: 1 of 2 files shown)


yixinglu (approved):

xtcyclist (approved):
Nice to check the depth. Maybe we could factor in the depth of the tree as well as other factors altogether to prune the traversal of the optimization space in 5.0.

PR #5078, opened 2022-12-19
Replace LOG(FATAL) with LOG(DFATAL)

### LINKED ISSUE #5068: Reduce the usages of LOG(FATAL) or other self-descruction measures as many as possible.
RT.

To improve the stability and maximize the running hours of services as much as possible.

Reasons to self-destruct:

1. Data may be corrupted or lost, if the service continues to run.
2. (Add more reasons here, if any)

veezhang:
## What type of PR is this?
- [x] bug
- [ ] feature
- [ ] enhancement

#### Issue(s) number: 

fix #5068

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....

### DIFF
--- a/src/kvstore/raftex/RaftPart.cpp
+++ b/src/kvstore/raftex/RaftPart.cpp
@@ -95,7 +95,7 @@ class AppendLogsIterator final : public LogIterator {
         }
       }
       if (notFulfilledPromise > 0) {
-        LOG(FATAL) << "notFulfilledPromise == " << notFulfilledPromise;
+        LOG(DFATAL) << "notFulfilledPromise == " << notFulfilledPromise;
       }
     }
   }
@@ -394,7 +394,7 @@ const char* RaftPart::roleStr(Role role) const {
     case Role::LEARNER:
       return "Learner";
     default:
-      LOG(FATAL) << idStr_ << "Invalid role";
+      LOG(DFATAL) << idStr_ << "Invalid role";
   }
   return nullptr;
 }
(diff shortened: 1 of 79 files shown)


src/kvstore/raftex/RaftPart.cpp
```diff
@@ -1091,7 +1091,11 @@ void RaftPart::processAppendLogResponses(const AppendLogResponses& resps,
               [self = shared_from_this(), term = term_] { self->onLeaderReady(term); });
         }
       } else {
-        LOG(FATAL) << idStr_ << "Failed to commit logs";
+        LOG(DFATAL) << idStr_ << "Failed to commit logs";
+        checkAppendLogResult(code);
+        VLOG(3) << idStr_ << "commitLogs failed";
+        iter.commit(code);
```
xtcyclist, line 1097: Why add these three lines here? Why not simply return?

veezhang: Like here https://github.com/vesoft-inc/nebula/pull/5078/files/ddbb5a96236c6753494f533c74044b1dc8d33960#diff-f56618b12fe75ed2289df1551818fdda84ea75d2f5c8b4c406fc196e06e7768bL1062-L1067 

I think it's necessary to commit the error code too.

veezhang: @xtcyclist Of course, you are more familiar with it. If you think it is unnecessary, I can change it to return directly.

xtcyclist: Better remove these three lines. I think we'd better not adding new codes here.

xtcyclist (dismissed):

critical27 (commented):
Maybe you can revert the changes in dir `src/kvstore`. They rarely fail, and if it really happens, we'd better crash there.

veezhang:
> Maybe you can revert the changes in dir `src/kvstore`. They rarely fail, and if it really happens, we'd better crash there.

ok

critical27 (approved):

xtcyclist (approved):

Issue #5090, opened 2022-12-22
Incorrect match query result(2)


nebula-bots:
**Please check the FAQ documentation before raising an issue**

**Describe the bug (__required__)**

Look at the queries below:

1. The first query show that there are 24 rows in the result.
2. In the second query we change the last where predicate in the first query into its return result, all evaluated to be false, which conflict to the first query

```txt
(root@nebula) [gdlancer]> MATCH (v0:Label_4)-[e0:Rel_2]->()<-[e1*1..2]-() WHERE (id(v0) == 17) AND (e0.Rel_2_3_Bool) WITH (e0.Rel_2_1_Double / 0.793772) AS pa0, (- ((- 0.302276) * ((- e0.Rel_2_0_Int) + ((- (e0.Rel_2_5_Int * e0.Rel_2_0_Int)) * 0.686817)))) AS pa1, e0.Rel_2_6_String AS pa2, e0, (e0.Rel_2_1_Double / e0.Rel_2_5_Int) AS pa3, e0.Rel_2_1_Double AS pa4, e0.Rel_2_1_Double AS pa5 WHERE (NOT ((NOT e0.Rel_2_4_Bool) XOR e0.Rel_2_4_Bool)) RETURN count(*) as totalRows
+-----------+
| totalRows |
+-----------+
| 24        |
+-----------+
Got 1 rows (time spent 2.27ms/18.484875ms)

Mon, 05 Dec 2022 16:51:25 CST

(root@nebula) [gdlancer]> MATCH (v0:Label_4)-[e0:Rel_2]->()<-[e1*1..2]-() WHERE (id(v0) == 17) AND (e0.Rel_2_3_Bool) WITH (e0.Rel_2_1_Double / 0.793772) AS pa0, (- ((- 0.302276) * ((- e0.Rel_2_0_Int) + ((- (e0.Rel_2_5_Int * e0.Rel_2_0_Int)) * 0.686817)))) AS pa1, e0.Rel_2_6_String AS pa2, e0, (e0.Rel_2_1_Double / e0.Rel_2_5_Int) AS pa3, e0.Rel_2_1_Double AS pa4, e0.Rel_2_1_Double AS pa5 return (NOT ((NOT e0.Rel_2_4_Bool) XOR e0.Rel_2_4_Bool))
+---------------------------------------------+
| !((!(e0.Rel_2_4_Bool) XOR e0.Rel_2_4_Bool)) |
+---------------------------------------------+
| false                                       |
[...]

Issue #5089, opened 2022-12-22
Forget to handle Logical Xor expr


jievince:
**Please check the FAQ documentation before raising an issue**

**Describe the bug (__required__)**
![image](https://user-images.githubusercontent.com/38901892/209061135-5a50442c-d80f-49d5-8458-43c29773d770.png)

Forget to handle xor:
```
 if (condition->isLogicalExpr()) {
    // Case1: OR Expr
    if (conditionType == ExprKind::kLogicalOr) {
      return true;
    }
    // Case2: AND Expr
    if (conditionType == ExprKind::kLogicalAnd &&
        graph::ExpressionUtils::findAny(static_cast<LogicalExpression*>(condition),
                                        {ExprKind::kRelIn})) {
      return true;
    }
    // Check logical operands
    for (auto operand : static_cast<const LogicalExpression*>(condition)->operands()) {
      if (!operand->isRelExpr() || !operand->isLogicalExpr()) {
        return false;
      }
    }
  }
```

**Your Environments (__required__)**

* OS: `uname -a`
* Compiler: `g++ --version` or `clang++ --version`
* CPU: `lscpu`
* Commit id (e.g. `a3ffc7d8`)

**How To Reproduce(__required__)**

1. Step 1
2. Step 2
3. Step 3

**Expected behavior**

**Additional context**

PR #5091, opened 2022-12-22
fix bug of reduce unary not for logic xor

jievince:
## What type of PR is this?
- [x] bug
- [ ] feature
- [ ] enhancement

#### Issue(s) number: 
Fix https://github.com/vesoft-inc/nebula/issues/5090,
Fix https://github.com/vesoft-inc/nebula/issues/5089
## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....


src/graph/util/ExpressionUtils.cpp
```diff
@@ -544,7 +544,8 @@ Expression *ExpressionUtils::reduceUnaryNotExpr(const Expression *expr) {
   auto operandMatcher = [](const Expression *operandExpr) -> bool {
     return (operandExpr->kind() == Expression::Kind::kUnaryNot ||
             (operandExpr->isRelExpr() && operandExpr->kind() != Expression::Kind::kRelREG) ||
-            operandExpr->isLogicalExpr());
+            operandExpr->kind() == Expression::Kind::kLogicalAnd ||
+            operandExpr->kind() == Expression::Kind::kLogicalOr);
```
jievince, line 548: Fix Fix https://github.com/vesoft-inc/nebula/issues/5090

src/graph/optimizer/rule/UnionAllIndexScanBaseRule.cpp
```diff
@@ -48,19 +48,18 @@ bool UnionAllIndexScanBaseRule::match(OptContext* ctx, const MatchedResult& matc
...
     if (conditionType == ExprKind::kLogicalAnd &&
         graph::ExpressionUtils::findAny(static_cast<LogicalExpression*>(condition),
                                         {ExprKind::kRelIn})) {
       return true;
     }
     // Check logical operands
     for (auto operand : static_cast<const LogicalExpression*>(condition)->operands()) {
-      if (!operand->isRelExpr() || !operand->isLogicalExpr()) {
+      if (!operand->isRelExpr() || operand->kind() != Expression::Kind::kLogicalAnd ||
+          operand->kind() != Expression::Kind::kLogicalOr) {
```
jievince, line 62: Fix  https://github.com/vesoft-inc/nebula/issues/5089

Aiee (dismissed):

yixinglu (dismissed):

src/graph/optimizer/rule/UnionAllIndexScanBaseRule.cpp
jievince, line 62: @Aiee  Plz help make sure  if `||` or '&&`  is needed here.

tests/tck/features/expression/LogicalExpression.feature
```diff
@@ -18,4 +20,25 @@ Feature: Logical Expression
       """
     Then the result should be, in any order:
       | count(*) |
-      | 140      |
+      | 12       |
```
Aiee, line 23: Is the change in the result expected?

src/graph/optimizer/rule/UnionAllIndexScanBaseRule.cpp
Aiee: Should keep it as it was originally. Why add another condition here?

jievince: isLogicalExpr() contains LogicXor, but this rule doesn't expect logicXor.

jievince: I don't understand the logic here. Which kinds of exprs are expected here?

tests/tck/features/expression/LogicalExpression.feature
jievince: Yes, it's expected. The previous result was wrong.

src/graph/optimizer/rule/UnionAllIndexScanBaseRule.cpp
Aiee: As the comments in the code states:
```
// The matched expression should be either a OR expression or an expression that could be
// rewrote to a OR expression. There are 3 scenarios.
//
// 1. OR expr. If OR expr has an IN expr operand that has a valid index, expand it to OR expr.
//
// 2. AND expr such as A in [a, b] AND B when A has a valid index, because it can be transformed to
// (A==a AND B) OR (A==b AND B)
//
// 3. IN expr with its list size > 1, such as A in [a, b] since it can be transformed to (A==a) OR
// (A==b).
```

jievince: ok

yixinglu (approved):
LGTM

Aiee (approved):
Good job!

PR #5096, opened 2022-12-22
do not check term for leader info by default para

liwenhui-soul:
## What type of PR is this?
- [x] bug
- [ ] feature
- [ ] enhancement

#### Description:
our term is not persisted, so if the cluster restart, the term would be reset to 0, and the leader info would never be updated in meta, because the term in meta is greater than the reset term in storage.

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....


critical27 (approved):

kikimo (approved):
Tested, LGTM

liwenhui-soul:
https://github.com/vesoft-inc/nebula-ent/issues/1769

PR #5098, opened 2022-12-23
Restrict schema type conversion

wenhaocs:
## What type of PR is this?
- [x] bug
- [ ] feature
- [ ] enhancement

#### Issue(s) number: 
https://github.com/vesoft-inc/nebula-ent/issues/1786

#### Description:
We now restrict the type conversion in schema change. Only the following are allowed:
(1) fixed string -> larger fixed string
(2) fixed string -> string
(3) thinner int -> wider int
(4) float -> double
(5) no type change

All other conversions are not allowed.

E.g.,
Change string to fixed string
image
image

Change int8 to int16 and then back to int8
image

## Checklist:
Tests:
- [x] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....


src/meta/MetaServiceUtils.cpp
```diff
@@ -58,6 +99,13 @@ nebula::cpp2::ErrorCode MetaServiceUtils::alterColumnDefs(std::vector<cpp2::Colu
             LOG(INFO) << "Column: " << colName << " as ttl_col, change not allowed";
             return nebula::cpp2::ErrorCode::E_UNSUPPORTED;
           }
+          if (!isLegalTypeConversion(it->get_type(), col.get_type())) {
+            LOG(WARNING) << "Update colume type " << colName << " from "
```
yixinglu, line 103: LOG(ERROR)

src/meta/MetaServiceUtils.cpp
```diff
@@ -32,6 +32,47 @@ nebula::cpp2::ErrorCode backupTable(kvstore::KVStore* kvstore,
...
+  if (from.get_type() == nebula::cpp2::PropertyType::INT8) {
+    return to.get_type() == nebula::cpp2::PropertyType::INT64 ||
+           to.get_type() == nebula::cpp2::PropertyType::INT32 ||
+           to.get_type() == nebula::cpp2::PropertyType::INT16;
+  }
+  // Float is only allowed to convert to double
+  if (from.get_type() == nebula::cpp2::PropertyType::FLOAT) {
+    return to.get_type() == nebula::cpp2::PropertyType::DOUBLE;
+  }
+  // Forbid all the other conversion, as the old data are too different from the new data.
```
yixinglu, line 73: shall we forbid the type conversion from boolean to int?

wenhaocs: Data are still correct if we allow for bool <-> int conversion. But the data before conversion and after conversion will look different. In a strict manner, we can disallow any conversion that will make data look different. Or if we are not that strict, we can allow bool <-> int conversion. Please clarify. cc. @MuYiYong @critical27

HarrisChu: prefer strict

xtcyclist (dismissed):

src/meta/MetaServiceUtils.cpp
critical27: I think we could forbid it this version I think, it will introduce extra work in storage.

critical27 (dismissed):
LGTM, good job.

critical27:
It seems some tck test cases failed~

src/meta/test/ProcessorTest.cpp
```diff
@@ -1947,7 +2042,7 @@ TEST(ProcessorTest, AlterEdgeTest) {
     cpp2::Schema changeSch;
     cpp2::ColumnDef column;
     column.name = "edge_0_col_10";
-    column.type.type_ref() = PropertyType::DOUBLE;
+    column.type.type_ref() = PropertyType::INT64;
```
wenhaocs, line 2045: Note that the current column type of `col_10` is int64. The failure should come from changing col on ttl. So I keep the column type the same.

wenhaocs, line 2045: Note that the current column type of `col_10` is int64. The failure should come from changing col on ttl. So I keep the column type the same.

src/meta/test/ProcessorTest.cpp
```diff
@@ -1452,7 +1454,7 @@ TEST(ProcessorTest, AlterTagTest) {
     cpp2::Schema changeSch;
     cpp2::ColumnDef column;
     column.name = "tag_0_col_10";
-    column.type.type_ref() = PropertyType::DOUBLE;
+    column.type.type_ref() = PropertyType::INT64;
```
wenhaocs, line 1457: Note that the current column type of col_10 is int64. The failure should come from changing col with ttl. So I keep the column type the same, to avoid the error coming from illegal column type change.

tests/tck/features/ttl/TTL.feature
```diff
@@ -112,7 +112,7 @@ Feature: TTLTest
     Then the execution should be successful
     When executing query:
       """
-      CREATE TAG woman(name string, email string, age int, gender string, row_timestamp timestamp) ttl_duration = 0, ttl_col = "row_timestamp";
+      CREATE TAG woman(name string, email string, age int32, gender string, row_timestamp timestamp) ttl_duration = 0, ttl_col = "row_timestamp";
```
wenhaocs, line 115: Note I change from int to int32, so we will have room to "upgrade" it into `int64` later.

tests/tck/features/ttl/TTL.feature
```diff
@@ -340,27 +350,27 @@ Feature: TTLTest
       | "player" | 'CREATE EDGE `player` (\n `id` int64 NULL,\n `name` string NULL,\n `age` int64 NULL,\n `address` string NULL,\n `score` float NULL\n) ttl_duration = 0, ttl_col = ""' |
     When executing query:
       """
-      ALTER EDGE player change(name int), drop(name);
+      ALTER EDGE player change(name string), drop(name);
```
wenhaocs, line 353: The error should come from the intended, not from illegal type change. So here I keep type the same.

wenhaocs: The same for the followings.

critical27 (dismissed):

Sophie-Xie:
@wenhaocs pls fix the tck~

wenhaocs:
> @wenhaocs pls fix the tck~

Done!

xtcyclist (approved):

critical27 (approved):

foesa-yang:
https://github.com/vesoft-inc/nebula-docs-cn/pull/2516

PR #5099, opened 2022-12-23
add job permission check for data balance

SuperYoko:
## What type of PR is this?
- [x] bug
- [ ] feature
- [ ] enhancement

#### Description:

Users except GUEST can do data balance.

## How do you solve it?

Add data balance permission check for writespace.

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....

### DIFF
--- a/src/graph/service/PermissionCheck.cpp
+++ b/src/graph/service/PermissionCheck.cpp
@@ -109,8 +109,19 @@ namespace graph {
     case Sentence::Kind::kUpdateEdge:
     case Sentence::Kind::kDeleteVertices:
     case Sentence::Kind::kDeleteTags:
-    case Sentence::Kind::kDeleteEdges:
+    case Sentence::Kind::kDeleteEdges: {
+      return PermissionManager::canWriteData(session, vctx);
+    }
     case Sentence::Kind::kAdminJob: {
+      auto *adminJobSentence = dynamic_cast<AdminJobSentence *>(sentence);
+      if (adminJobSentence == nullptr) {
+        // should not happend.
+        LOG(WARNING) << "sentence is not AdminJobSentence";
+        return Status::PermissionError("No permission to write space.");
+      }
+      if (adminJobSentence->needWriteSpace()) {
+        return PermissionManager::canWriteSpace(session);
+      }
       return PermissionManager::canWriteData(session, vctx);
     }
     case Sentence::Kind::kDescribeTag:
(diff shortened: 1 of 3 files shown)


liwenhui-soul (approved):

src/graph/service/PermissionCheck.cpp
```diff
@@ -109,8 +109,19 @@ namespace graph {
...
-    case Sentence::Kind::kDeleteEdges:
+    case Sentence::Kind::kDeleteEdges: {
+      return PermissionManager::canWriteData(session, vctx);
+    }
     case Sentence::Kind::kAdminJob: {
+      auto *adminJobSentence = dynamic_cast<AdminJobSentence *>(sentence);
+      if (adminJobSentence == nullptr) {
+        // should not happend.
+        LOG(WARNING) << "sentence is not AdminJobSentence";
+        return Status::PermissionError("No permission to write space.");
```
pengweisong, line 120: should the message be sth like "bad format sentence"?

src/graph/service/PermissionCheck.cpp
```diff
@@ -109,8 +109,19 @@ namespace graph {
...
+      }
+      if (adminJobSentence->needWriteSpace()) {
```
pengweisong, line 122: what does  `needWriteSpace` mean?

SuperYoko: > what does `needWriteSpace` mean? Why only balance job need write space? others like compact will not write?

Yes, It's means space data distribuition level permission needed.( And it's the origin permission needed before refactor to jobmanager.

src/graph/service/PermissionCheck.cpp
SuperYoko: > should the message be sth like "bad format sentence"?

It's a path should never go in, I wan't use check not null first...  Make sence to me.

src/graph/service/PermissionCheck.cpp
pengweisong: you may need to update the comment before `PermissionCheck::permissionCheck` to explain this

SuperYoko: updated.

critical27 (approved):

liwenhui-soul (approved):

pengweisong (approved):

PR #5100, opened 2022-12-23
Fix the crash caused by a nullptr variable

xtcyclist:
## What type of PR is this?
- [X] bug
- [ ] feature
- [ ] enhancement

#### Issue(s) number: 
Close https://github.com/vesoft-inc/nebula-ent/issues/1971
Close https://github.com/vesoft-inc/nebula/issues/5101

#### Description:
ex->what() is called without checking whether ex is nullptr.

## How do you solve it?
For this else body in particular, no need to call ex->what().

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....

### DIFF
--- a/src/clients/storage/StorageClientBase-inl.h
+++ b/src/clients/storage/StorageClientBase-inl.h
@@ -198,14 +198,19 @@ folly::Future<StatusOr<Response>> StorageClientBase<ClientType, ClientManagerTyp
 
         using TransportException = apache::thrift::transport::TTransportException;
         auto ex = exWrapper.get_exception<TransportException>();
-        if (ex && ex->getType() == TransportException::TIMED_OUT) {
-          LOG(ERROR) << "Request to " << host << " time out: " << ex->what();
-          return Status::Error("RPC failure in StorageClient, probably timeout: %s", ex->what());
+        if (ex) {
+          if (ex->getType() == TransportException::TIMED_OUT) {
+            LOG(ERROR) << "Request to " << host << " time out: " << ex->what();
+            return Status::Error("RPC failure in StorageClient with timeout: %s", ex->what());
+          } else {
+            LOG(ERROR) << "Request to " << host << " failed: " << ex->what();
+            return Status::Error("RPC failure in StorageClient: %s", ex->what());
+          }
         } else {
           auto partsId = getReqPartsId(request);
           invalidLeader(spaceId, partsId);
           LOG(ERROR) << "Request to " << host << " failed.";
-          return Status::Error("RPC failure in StorageClient, probably timeout.");
+          return Status::Error("RPC failure in StorageClient.");
         }
       });
 }
--- a/src/clients/storage/StorageClientBase-inl.h
+++ b/src/clients/storage/StorageClientBase-inl.h
@@ -200,12 +200,13 @@ folly::Future<StatusOr<Response>> StorageClientBase<ClientType, ClientManagerTyp
         auto ex = exWrapper.get_exception<TransportException>();
         if (ex && ex->getType() == TransportException::TIMED_OUT) {
           LOG(ERROR) << "Request to " << host << " time out: " << ex->what();
+          return Status::Error("RPC failure in StorageClient, probably timeout: %s", ex->what());
         } else {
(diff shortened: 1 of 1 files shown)


src/clients/storage/StorageClientBase-inl.h
```diff
@@ -203,7 +203,7 @@ folly::Future<StatusOr<Response>> StorageClientBase<ClientType, ClientManagerTyp
         } else {
           auto partsId = getReqPartsId(request);
           invalidLeader(spaceId, partsId);
-          LOG(ERROR) << "Request to " << host << " failed: " << ex->what();
+          LOG(ERROR) << "Request to " << host << " failed.";
```
jievince, line 206: Could the ex not be null sometimes in the else branch?

src/clients/storage/StorageClientBase-inl.h
```diff
@@ -203,7 +203,7 @@ folly::Future<StatusOr<Response>> StorageClientBase<ClientType, ClientManagerTyp
...
         }
         return Status::Error("RPC failure in StorageClient, probably timeout: %s", ex->what());
```
kikimo, line 208: This line may also crash

kikimo (changes requested):

src/clients/storage/StorageClientBase-inl.h
kikimo: I mean here in the return statemetn`return Status::Error("RPC failure in StorageClient, probably timeout: %s", ex->what());`

xtcyclist: Fixed.

jievince (approved):

src/clients/storage/StorageClientBase-inl.h
xtcyclist: Seems like when there was an error but not an apache::thrift::transport::TTransportException, it would run into this else branch. Not sure what would cause this though.

src/clients/storage/StorageClientBase-inl.h
```diff
@@ -200,12 +200,13 @@ folly::Future<StatusOr<Response>> StorageClientBase<ClientType, ClientManagerTyp
         auto ex = exWrapper.get_exception<TransportException>();
         if (ex && ex->getType() == TransportException::TIMED_OUT) {
           LOG(ERROR) << "Request to " << host << " time out: " << ex->what();
+          return Status::Error("RPC failure in StorageClient, probably timeout: %s", ex->what());
         } else {
           auto partsId = getReqPartsId(request);
           invalidLeader(spaceId, partsId);
-          LOG(ERROR) << "Request to " << host << " failed: " << ex->what();
+          LOG(ERROR) << "Request to " << host << " failed.";
+          return Status::Error("RPC failure in StorageClient, probably timeout.");
```
Shylock-Hg, line 208: In `else` branch, it's not time out error.

src/clients/storage/StorageClientBase-inl.h
```diff
@@ -200,12 +200,13 @@ folly::Future<StatusOr<Response>> StorageClientBase<ClientType, ClientManagerTyp
         auto ex = exWrapper.get_exception<TransportException>();
         if (ex && ex->getType() == TransportException::TIMED_OUT) {
```
yixinglu, line 201: maybe better to handle the error as the following way:

```cpp
if (ex) {
  // Transport exception
  if (ex->getType() == TIMEOUT) {
    // report time out error
  } else {
    // report transport error
  }
} else {
  // report other exception
}
```

xtcyclist: Fixed.

src/clients/storage/StorageClientBase-inl.h
xtcyclist: Fixed.

codesigner (approved):

jievince (approved):

yixinglu (dismissed):

codesigner (approved):

kikimo (approved):
LGTM

Issue #5097, opened 2022-12-23
graphd report unexpected error `Wrong type result, the type should be NULL, EMPTY, BOOL`


nebula-bots:
**Please check the FAQ documentation before raising an issue**

**Describe the bug (__required__)**

Similar to #1574 but that's a fixed issued and we are using the latest master branch.
While executing the following query, graph report an unexpected error:

```txt
(root@nebula) [gdlancer]> MATCH p0 = (v0)-[e0]->()<-[e1:Rel_2]-(v1) WHERE (id(v0) in [13, 11, 20, 21, 2, 23, 28, 27, 9, 14, 19]) AND (e0.Rel_0_0_Bool) UNWIND nodes(p0) AS ua0 WITH e0, ua0.Label_0.Label_0_3_Bool AS pa0, ua0, 0.915546 AS pa1 WHERE ua0.Label_0.Label_0_3_Bool RETURN count(*) as totalRows
[ERROR (-1005)]: Wrong type result, the type should be NULL, EMPTY, BOOL

Mon, 12 Dec 2022 18:52:51 CST

(root@nebula) [gdlancer]>
```

In contrast the same query Neo4j reports no error:

```txt
$ MATCH p0 = (v0)-[e0]->()<-[e1:Rel_2]-(v1) WHERE (v0.id in [13, 11, 20, 21, 2, 23, 28, 27, 9, 14, 19]) AND (e0.Rel_0_0_Bool) UNWIND nodes(p0) AS ua0 WITH e0, ua0.Label_0_3_Bool AS pa0, ua0, 0.915546 AS pa1 WHERE ua0.Label_0_3_Bool RETURN count(*) as totalRows

╒═══════════╕
│"totalRows"│
╞═══════════╡
│1          │
└───────────┘
```

**Your Environments (__required__)**

* OS: `uname -a`
* Compiler: `g++ --version` or `clang++ --version`
* CPU: `lscpu`
* Commit id (e.g. `a3ffc7d8`) 573f7f414 (community edition)

**How To Reproduce(__required__)**

1. Step 1
2. Step 2
3. Step 3

**Expected behavior**

**Additional context**

xtcyclist:
This is another bug on property prunning.

xtcyclist:
This is related to the usage of `NullType::BAD_TYPE`, which we have decided to keep the status quo. I've improved the error message to remind users to write <boolean condition> == <true/false> to bypass this. Not going to apply further fixes.

PR #5102, opened 2022-12-23
revise error messages.

xtcyclist:
## What type of PR is this?
- [X] bug
- [ ] feature
- [ ] enhancement

#### Issue(s) number: 
Close https://github.com/vesoft-inc/nebula/issues/5097

#### Description:
NullType::BAD_TYPE triggers error in filter. We've decided to keep similar issues as the status quo.

## How do you solve it?
Improved the error message, telling users how to bypass this problem.

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....

### DIFF
--- a/src/graph/executor/query/FilterExecutor.cpp
+++ b/src/graph/executor/query/FilterExecutor.cpp
@@ -65,7 +65,10 @@ StatusOr<DataSet> FilterExecutor::handleJob(size_t begin, size_t end, Iterator *
   for (; iter->valid() && begin++ < end; iter->next()) {
     auto val = condition->eval(ctx(iter));
     if (val.isBadNull() || (!val.empty() && !val.isImplicitBool() && !val.isNull())) {
-      return Status::Error("Wrong type result, the type should be NULL, EMPTY, BOOL");
+      return Status::Error("Failed to evaluate condition: %s. %s%s",
+                           condition->toString().c_str(),
+                           "For boolean conditions, please write in their full forms like",
+                           " <condition> == <true/false> or <condition> IS NULL.");
     }
     if (!(val.empty() || val.isNull() || (val.isImplicitBool() && !val.implicitBool()))) {
       // TODO: Maybe we can move.
@@ -96,7 +99,10 @@ Status FilterExecutor::handleSingleJobFilter() {
     while (iter->valid()) {
       auto val = condition->eval(ctx(iter));
       if (val.isBadNull() || (!val.empty() && !val.isImplicitBool() && !val.isNull())) {
-        return Status::Error("Wrong type result, the type should be NULL, EMPTY, BOOL");
+        return Status::Error("Failed to evaluate condition: %s. %s%s",
+                             condition->toString().c_str(),
+                             "For boolean conditions, please write in their full forms like",
+                             " <condition> == <true/false> or <condition> IS NULL.");
       }
       if (val.empty() || val.isNull() || (val.isImplicitBool() && !val.implicitBool())) {
         if (UNLIKELY(filter->needStableFilter())) {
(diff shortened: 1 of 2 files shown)


src/graph/executor/query/FilterExecutor.cpp
```diff
@@ -65,7 +65,9 @@ StatusOr<DataSet> FilterExecutor::handleJob(size_t begin, size_t end, Iterator *
   for (; iter->valid() && begin++ < end; iter->next()) {
     auto val = condition->eval(ctx(iter));
     if (val.isBadNull() || (!val.empty() && !val.isImplicitBool() && !val.isNull())) {
-      return Status::Error("Wrong type result, the type should be NULL, EMPTY, BOOL");
+      return Status::Error("Failed to evaluate condition: %s. %s",
+                           condition->toString().c_str(),
+                           "For boolean conditions, please write <condition> == <true/false>.");
```
yixinglu, line 70: prefer the error message: `it is not a boolean condition.` sometimes the condition could be `X.xx IS NULL`, which is not the format such as `<condition> == <true/false>`

xtcyclist: Fixed. Need to explicitly recommend a way to write conditions here to avoid running into errors.

src/graph/executor/query/FilterExecutor.cpp
```diff
@@ -65,7 +65,10 @@ StatusOr<DataSet> FilterExecutor::handleJob(size_t begin, size_t end, Iterator *
   for (; iter->valid() && begin++ < end; iter->next()) {
     auto val = condition->eval(ctx(iter));
     if (val.isBadNull() || (!val.empty() && !val.isImplicitBool() && !val.isNull())) {
-      return Status::Error("Wrong type result, the type should be NULL, EMPTY, BOOL");
+      return Status::Error("Failed to evaluate condition: %s. %s%s",
+                           condition->toString().c_str(),
+                           "For boolean conditions, please write in their full forms like",
+                           " <condition> == <true/false> or <condition> IS NULL.");
```
codesigner, line 71: IS [NOT] NULL, add not as an optional

xtcyclist: Fixed.

src/graph/executor/query/FilterExecutor.cpp
```diff
@@ -96,7 +99,10 @@ Status FilterExecutor::handleSingleJobFilter() {
     while (iter->valid()) {
       auto val = condition->eval(ctx(iter));
       if (val.isBadNull() || (!val.empty() && !val.isImplicitBool() && !val.isNull())) {
-        return Status::Error("Wrong type result, the type should be NULL, EMPTY, BOOL");
+        return Status::Error("Failed to evaluate condition: %s. %s%s",
+                             condition->toString().c_str(),
+                             "For boolean conditions, please write in their full forms like",
+                             " <condition> == <true/false> or <condition> IS NULL.");
```
codesigner, line 105: also fix here

tests/tck/features/match/Base.feature
```diff
@@ -632,7 +632,7 @@ Feature: Basic match
       """
       MATCH (v:player) where v.player.name return v
       """
-    Then a ExecutionError should be raised at runtime: Wrong type result, the type should be NULL, EMPTY, BOOL
+    Then a ExecutionError should be raised at runtime: Failed to evaluate condition: v.player.name. For boolean conditions, please write <condition> == <true/false> or <condition> IS NULL.
```
codesigner, line 635: add here

xtcyclist: Fixed.

src/graph/executor/query/FilterExecutor.cpp
xtcyclist: Fixed.

codesigner (approved):

jievince (dismissed):

jievince (approved):

codesigner (approved):

PR #5122, opened 2022-12-27
Add more fulltext index tck case

cangfengzhs:
## What type of PR is this?
- [ ] bug
- [ ] feature
- [ ] enhancement

1. Add fulltext index ddl tck test case
2. fix ESListener crash when there is no elasticsearch

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....

### DIFF
--- a/src/kvstore/listener/elasticsearch/ESListener.cpp
+++ b/src/kvstore/listener/elasticsearch/ESListener.cpp
@@ -26,22 +26,6 @@ void ESListener::init() {
   }
   isIntVid_ = vidTypeRet.value() == nebula::cpp2::PropertyType::INT64;
 
-  auto cRet = schemaMan_->getServiceClients(meta::cpp2::ExternalServiceType::ELASTICSEARCH);
-  if (!cRet.ok() || cRet.value().empty()) {
-    LOG(FATAL) << "elasticsearch clients error";
-  }
-  std::vector<nebula::plugin::ESClient> esClients;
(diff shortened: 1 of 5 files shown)


SuperYoko (approved):

src/kvstore/listener/elasticsearch/ESListener.cpp
```diff
@@ -70,7 +54,13 @@ bool ESListener::apply(const BatchHolder& batch) {
     pickTagAndEdgeData(std::get<0>(log), std::get<1>(log), std::get<2>(log), callback);
   }
   if (!bulk.empty()) {
-    auto status = esAdapter_.bulk(bulk);
+    auto esAdapterRes = getESAdapter();
```
critical27, line 57: Why we need to retrieve adapter for each data?

cangfengzhs: Because the password of elasticsearch may be modified during the system operation. Therefore, you need to obtain the configuration of the elasticsearch service again.

In addition, there was a problem before that. If there was no sign in text service when adding listener, the nebula-listener would crash directly. Therefore, the address of elasticsearch cannot be obtained during initialization.

critical27: 👌

critical27 (approved):

Issue #5123, opened 2022-12-27
match can not filter by rank(e)


nebula-bots:
**Please check the FAQ documentation before raising an issue**
use rank(e)==2 in where statement, however return rank(e)=0;
```
 match ()-[e]->() where rank(e) == 2 return e limit 10
```

result is: 
```
(root@nebula) [backup_sf100]> profile match ()-[e]->() where rank(e) == 2 return e limit 10
+-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| e                                                                                                                                                                             |
+-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| [:IS_LOCATED_IN "c-10509641"->"l-1" @0 {}]                                                                                                                                    |
| [:COMMENT_HAS_CREATOR "c-10509641"->"p-41416" @0 {browserUsed: "Firefox", content: "thanks", creationDate: 2010-02-18T03:04:41.000000, length: 6, locationIP: "1.2.116.192"}] |
| [:REPLY_OF_COMMENT "c-10509641"->"c-10509627" @0 {}]                                                                                                                          |
| [:IS_LOCATED_IN "c-1051079"->"l-103" @0 {}]                                                                                                                                   |
| [:COMMENT_HAS_CREATOR "c-1051079"->"p-217287" @0 {browserUsed: "Firefox", content: "no", creationDate: 2010-02-19T20:22:43.000000, length: 2, locationIP: "27.112.114.114"}]  |
| [:REPLY_OF_POST "c-1051079"->"s-1051069" @0 {}]                                                                                                                               |
| [:IS_LOCATED_IN "c-10512839"->"l-0" @0 {}]                                                                                                                                    |
| [:COMMENT_HAS_CREATOR "c-10512839"->"p-16592" @0 {browserUsed: "Chrome", content: "thx", creationDate: 2010-02-27T05:36:54.000000, length: 3, locationIP: "49.249.123.82"}]   |
| [:REPLY_OF_COMMENT "c-10512839"->"c-10512837" @0 {}]                                                                                                                          |
| [:IS_LOCATED_IN "c-10514795"->"l-13" @0 {}]                                                                                                                                   |
+-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
Got 10 rows (time spent 5651/26492 us)
```

**Your Environments (__required__)**

nebula-ent，2022.12.22-nightly-ent
commit ID：8ac1a26

**Expected behavior**

[...]

Issue #5120, opened 2022-12-27
支持按边的rank作过滤


bocai3030:
当前版本3.4.2下，
match ()-[e]->() where rank(e) == 2 return e limit 10
会报错：-1005:Scan vertices or edges need to specify a limit number, or limit number can not push down.

xtcyclist:
Duplicate with https://github.com/vesoft-inc/nebula/issues/5123. Going to be fixed together.

PR #5127, opened 2022-12-27
Fix picker in PushFilterDownTraverseRule

xtcyclist:
## What type of PR is this?
- [x] bug
- [ ] feature
- [ ] enhancement

#### Issue(s) number: 
Close https://github.com/vesoft-inc/nebula/issues/5123
Close https://github.com/vesoft-inc/nebula/issues/5120
Close https://github.com/vesoft-inc/nebula-ent/issues/1996

#### Description:
`$-.e[0].likeness` can be pushed into AppendVertices and further pushed down into the storage. But `rank($-.e[0])` cannot.

## How do you solve it?
Use the same condition of rewriteEdgePropertyFilter to ban the pushdown of some expressions. Only isSingleLenExpandExpr ones pass through.

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [X] TCK

Affects:

## Release notes:

> ex. Fixed the bug .....

### DIFF
--- a/src/graph/optimizer/rule/PushFilterDownTraverseRule.cpp
+++ b/src/graph/optimizer/rule/PushFilterDownTraverseRule.cpp
@@ -75,10 +75,12 @@ StatusOr<OptRule::TransformResult> PushFilterDownTraverseRule::transform(
         return false;
       }
     }
-
     auto varProps = graph::ExpressionUtils::collectAll(
         e, {Expression::Kind::kInputProperty, Expression::Kind::kVarProperty});
     if (varProps.empty()) {
+      return false;
+    }
+    if (!graph::ExpressionUtils::isSingleLenExpandExpr(edgeAlias, e)) {
       return false;
     }
     for (auto* expr : varProps) {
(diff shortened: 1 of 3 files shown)


src/graph/optimizer/rule/PushFilterDownTraverseRule.cpp
```diff
@@ -75,12 +75,14 @@ StatusOr<OptRule::TransformResult> PushFilterDownTraverseRule::transform(
...
     }
-
     auto varProps = graph::ExpressionUtils::collectAll(
         e, {Expression::Kind::kInputProperty, Expression::Kind::kVarProperty});
     if (varProps.empty()) {
       return false;
     }
+    if (!graph::ExpressionUtils::isSingleLenExpandExpr(edgeAlias, e)) {
+      return false;
+    }
```
jievince, line 85: What if`$-.e[0].likeness > 78`?

xtcyclist:
Handed over to @jievince to fix these issues.

Issue #2671, opened 2021-03-22
case3 :alter schema with index


whitewum:
These are migration from internal jira reports.

## case 1: the same property name, drop it , then re-create it.

![image](https://user-images.githubusercontent.com/50101159/111953870-31778f00-8b22-11eb-988c-13025195a2be.png)

## case2: alter schema change property to a different type (string->int)

```

create tag test(col1 string , col2 string);

insert vertex test(col1,col2) values "11":("aa","bb");

alter tag test change (col2 int)

(root@nebula) [test2]> fetch prop on test "11"
+----------+-----------+-----------+
| VertexID | test.col1 | test.col2 |
+----------+-----------+-----------+
| "11" | "aa" | {color:#de350b}"bb"{color} |
+----------+-----------+-----------+

 

insert vertex test(col1,col2) values "22":("aa",99)
create tag index test_index on test(col2);

rebuild tag index test_index
+------------+
| New Job Id |
+------------+
| 49 |
+------------+
 show jobs;
+--------+---------------------+------------+------------+------------+
| Job Id | Command | Status | Start Time | Stop Time |
+--------+---------------------+------------+------------+------------+
| 49 | "REBUILD_TAG_INDEX" | "FINISHED" | 1607570799 | 1607570799 |

 

(root@nebula) [test2]> lookup on test where test.col2 < 99 yield test.col2;
+------+----------------------+
| _vid | col2 |
+------+----------------------+
| "11" | {color:#de350b}-2134143272594898944{color} |
+------+----------------------+
Got 1 rows (time spent 1957/3046 us)

```

## case 3: alter schema with index
```
create tag test4(col1 int,col2 int);

insert vertex test4(col1,col2) values "a":(1,2)

insert vertex test4(col1,col2) values "b":(2,3)

alter tag test4 add (col3 int)

create tag index test4_index on test4(col1,col2,col3);

rebuild tag index test4_index

(root@nebula) [test3]> lookup on test4 where test4.col1 == 2;
Empty set (time spent 2026/2789 us)
```

## case4:default value

```
# v2.0
(root@nebula) [test1]> create tag tag1(col1 timestamp default 0)
 Execution succeeded (time spent 18185/18647 us)
Tue, 26 Jan 2021 10:35:13 CST
(root@nebula) [test1]> insert vertex tag1() values "1":()
 Execution succeeded (time spent 2850/3427 us)
Tue, 26 Jan 2021 10:35:30 CST
[...]

PR #5130, opened 2022-12-28
Forbid adding schema properties if they existed before

wenhaocs:
## What type of PR is this?
- [x] bug
- [ ] feature
- [ ] enhancement

#### Issue(s) number: 
Case 1 in https://github.com/vesoft-inc/nebula/issues/2671
#### Description:
When we change schema, the existing data will not be changed. In case 1 above, if a user adds a column/property back (with the same name, but has different attributes) after dropping it, querying the existing data will show the properties under the old schema, not matching the new schema. Per discussion, we just forbid adding column/property back with the same name.

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> Adding column/property back with the same name is forbidden.


tests/query/stateless/test_schema.py
```diff
@@ -124,23 +124,15 @@ def test_alter_tag_succeed(self):
...
-        self.check_result(resp, expect)
-
-        resp = self.execute('DESC TAG student')
-        self.check_resp_succeeded(resp)
-        expect = [['name', 'string', 'YES', T_EMPTY, T_EMPTY],
-                  ['email', 'string', 'YES', T_EMPTY, T_EMPTY],
-                  ['birthday', 'timestamp', 'YES', T_EMPTY, T_EMPTY],
-                  ['age', 'string', 'YES', T_EMPTY, T_EMPTY]]
+                  ['age_str', 'string', 'YES', T_EMPTY, T_EMPTY]]
         self.check_result(resp, expect)
```
wenhaocs, line 136: These are redundant.

src/common/utils/MetaKeyUtils.cpp
```diff
@@ -573,16 +573,17 @@ SchemaVer MetaKeyUtils::parseEdgeVersion(folly::StringPiece key) {
          *reinterpret_cast<const SchemaVer*>(key.begin() + offset);
 }
 
-SchemaVer MetaKeyUtils::getLatestEdgeScheInfo(kvstore::KVIterator* iter, folly::StringPiece& val) {
+SchemaVer MetaKeyUtils::getLatestEdgeScheInfo(
+    kvstore::KVIterator* iter, std::unordered_map<SchemaVer, folly::StringPiece>& schemasRaw) {
```
critical27, line 577: Maybe rename to schemaHistory or something else?

critical27 (approved):
Good job~

SuperYoko (approved):

abby-cyber:
https://github.com/vesoft-inc/nebula-docs-cn/pull/2496

PR #5137, opened 2022-12-28
add tagFilter for traverse

nevermore3:
## What type of PR is this?
- [x] bug
- [ ] feature
- [ ] enhancement

#### Issue(s) number: 
close https://github.com/vesoft-inc/nebula-ent/issues/1668
close https://github.com/vesoft-inc/nebula-ent/issues/2022

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....

### DIFF
--- a/src/graph/planner/plan/Query.h
+++ b/src/graph/planner/plan/Query.h
@@ -1641,6 +1641,14 @@ class Traverse final : public GetNeighbors {
     firstStepFilter_ = filter;
   }
 
+  Expression* tagFilter() const {
+    return tagFilter_;
+  }
+
+  void setTagFilter(Expression* tagFilter) {
+    tagFilter_ = tagFilter;
+  }
+
  private:
   friend ObjectPool;
   Traverse(QueryContext* qctx, PlanNode* input, GraphSpaceID space)
@@ -1657,6 +1665,7 @@ class Traverse final : public GetNeighbors {
   bool trackPrevPath_{true};
   // Push down filter in first step
   Expression* firstStepFilter_{nullptr};
+  Expression* tagFilter_{nullptr};
 };
 
 // Append vertices to a path.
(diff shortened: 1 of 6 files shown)


xtcyclist (approved):

src/graph/planner/plan/Query.h
```diff
@@ -1641,6 +1641,14 @@ class Traverse final : public GetNeighbors {
     firstStepFilter_ = filter;
   }
 
+  Expression* tagFilter() const {
```
Shylock-Hg, line 1644: What's the difference from `vertex filter`?

nevermore3: same, but the vertexfilter will not be pushed down to the storage layer, tagFilter will

nevermore3: And  vertex filter will be moved to firstStepFilter in the optimizer

Shylock-Hg: Too much filter is really confusion.

nevermore3: vertexFilter and edgeFilter are in the graph layer. Under some conditions, they cannot be pushed to the storage layer, so they continue to filter at the graph layer. firstStepFilter is the first step version of the filter, which is pushed down to the storage layer, but they cannot filter vertex. So tagFilter is added, if the vertex's attribute does not satisfy tagFilter, it will be filtered out directly in the storage layer

xtcyclist (approved):

src/graph/planner/plan/Query.h
jievince: We really should not design such a large-grained plan node in the future.

jievince (approved):

PR #5139, opened 2022-12-28
Test/user

Shylock-Hg:
## What type of PR is this?
- [ ] bug
- [ ] feature
- [x] enhancement

#### Issue(s) number: 
Sub job about https://github.com/vesoft-inc/nebula-ent/issues/1754

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....


tests/tck/conftest.py
```diff
@@ -980,3 +980,19 @@ def switch_to_new_session(conn_pool, user, password, class_fixture_variables, ex
...
+
+@when(parse('login with user "{user}"'))
+def login_without_password(conn_pool, user):
+    sess = None
+    try:
+        sess = conn_pool.get_session(user, '')
+    except Exception as e:
+        assert e
+
+@when(parse('login with user "{user}" with "{password}"'))
```
yixinglu, line 992: `'login with user "{user}" and password "{password}"'`

tests/tck/conftest.py
```diff
@@ -980,3 +980,19 @@ def switch_to_new_session(conn_pool, user, password, class_fixture_variables, ex
     sess = conn_pool.get_session(user, password)
     class_fixture_variables["sessions"].append(sess)
     exec_ctx["current_session"] = sess
+
+@when(parse('login with user "{user}"'))
+def login_without_password(conn_pool, user):
+    sess = None
+    try:
+        sess = conn_pool.get_session(user, '')
```
yixinglu, line 988: why not to reset the current_session in exec_ctx?

Shylock-Hg: Just verify logging, avoid affecting other tests.

yixinglu: if you don't change the current session, how do you make sure the next test cases use the current login user?

tests/tck/conftest.py
```diff
@@ -980,3 +980,19 @@ def switch_to_new_session(conn_pool, user, password, class_fixture_variables, ex
...
+    except Exception as e:
+        assert e
+
+@when(parse('login with user "{user}" and "{password}"'))
```
yixinglu, line 992: plz given the password keyword to indicate what the following string is.

```
login with user "{user}" and password "{password}"
```

tests/tck/conftest.py
Shylock-Hg: Next test will use origin session.

tests/tck/conftest.py
Shylock-Hg: ok

tests/tck/conftest.py
yixinglu: if so, does this step just test whether the user exists?  why not to use `show users` to test it? if you want to switch user and test role, plz consider to switch the current session.

Shylock-Hg: This will be logged in by this user.

yixinglu: Just test login feature? what to do after the user login?  I don't find where to use the login user?

Shylock-Hg: Just test login.

yixinglu: why not to test this feature in role cases ? such as:

```
When executing query:
  """
  create user username...
  grant user some rule ..
  """
Then executing successfully
When switch to new session with user "{user}" and password "{password}"
Then executing successfully
When executing query:
  """
   test role of user
  """
Then ...
```

it's not necessary to create a new step for this feature and we have the related step for switching user:

```
switch to new session with user "{user}" and password "{password}"
```

Shylock-Hg: Don't want to change current session.

yixinglu (approved):

czpmango (approved):

PR #5082, opened 2022-12-21
Memory Tracker

codesigner:
## What type of PR is this?
- [ ] bug
- [ ] feature
- [x] enhancement

#### Issue(s) number: 

#5081

#### Description:

UPDATE: MemoryTracker is only turn on when jemalloc in enabled. because default malloc cannot get accurate size of free(ptr) (see untrackMemory in Memory.h, malloc_usable_size() is not accurate);

override the default c++ new/delete operator, the new added new/delete function do the memory quota counting before do the real alloc/free

- NewDelete.cpp take over the c++ global new/free operators, hand over control to Memory.h; 
- Memory.h defined the tracking/alloc/dealloc routines
     - using jemalloc to get the alignment and sizing staff;
     - use MemoryTracker to track the memory usage;
- MemoryTracker.cpp
   - A thread local MemoryStats to make thread can reserve some memory in local, avoid get memory from global on every (de)alloc;
   - A global used memory counter and A set-able limit, throw bad_alloc when used exceeded limit;
- Graph Executors
   - added thenError callback to capture the bad_alloc exception, wrap error in status, making the query hit the limit fail; 
- Storage Processors
   - added try catch, to catch exception in singleThread, (FLAGS_query_concurrently=false )
   - add thenError, to be able failed in multiple Thread(FLAGS_query_concurrently=true)

## Checklist:
Tests:
- [ ] Unit test(positive and negative cases)
- [ ] Function test
- [ ] Performance test
- [ ] N/A
- [x] Mannual
      
### Test
- DataSet: LDBC sf1, set cgroup to limit graphd to 500Mb, or storaged to 300Mb to let graphd/storaged hit the memory limit
- Query ` match (v:Person)-[e*3]->(x) where id(v)==933 return x` 
- Env: graphd, storaged's max memory controled by separate cgroup
- Before:  result in graphd/storaged killed by system's OOM killer, 
- After: 
   - query failed with error message
 ```
(root@nebula) [sf1]> match (v:Person)-[e*3]->(x) where id(v)==933 return x;
[ERROR (-1005)]: Graph Error: GRAPH_MEMORY_EXCEEDED(-2600)

(root@nebula) [sf1]> match (v:Person)-[e*3]->(x) where id(v)==933 return x;
[ERROR (-1005)]: Storage Error: STORAGE_MEMORY_EXCEEDED(-3600)
``` 
- graphd/storaged not killed,

- Debug LOG indicate it do not leak memory
 - graphd memory stats log
```
[...]

PR #5140, opened 2022-12-28
Enhance MemoryStats to avoid the false sharing effect on the CPU cache

czpmango:
## What type of PR is this?
- [ ] bug
- [ ] feature
- [x] enhancement

#### Description:
As discussed here: https://github.com/vesoft-inc/nebula/pull/5082/files#r1057092947
image

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

### DIFF
--- a/src/common/memory/MemoryTracker.h
+++ b/src/common/memory/MemoryTracker.h
@@ -13,9 +13,9 @@ namespace memory {
 
 constexpr size_t
 #if defined(__cpp_lib_hardware_interference_size)
-    L1_CACHE_LINE_SIZE = hardware_destructive_interference_size;
+    CACHE_LINE_SIZE = hardware_destructive_interference_size;
 #else
-    L1_CACHE_LINE_SIZE = 64;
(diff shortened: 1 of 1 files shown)


codesigner (dismissed):

Aiee (approved):

src/common/memory/MemoryTracker.h
```diff
@@ -11,6 +11,13 @@
 namespace nebula {
 namespace memory {
 
+constexpr size_t
+#if defined(__cpp_lib_hardware_interference_size)
+    L1_CACHE_LINE_SIZE = hardware_destructive_interference_size;
+#else
+    L1_CACHE_LINE_SIZE = 64;
```
codesigner, line 18: Minor：cache line size is same for all level of cpu caches,

czpmango: ACK.

codesigner (approved):

Aiee (approved):

Issue #5081, opened 2022-12-21
more precise momery quota control


codesigner:
**Introduction**

- add memory tracker to tracking memory allocation and deallocation, 
- kill query to avoid memory abuse usage, avoid process killed by os 

**Contents**

**Related work**

PR #5167, opened 2022-12-30
[memtracker] improve throw bad_alloc logic

codesigner:
## What type of PR is this?
- [ ] bug
- [ ] feature
- [x] enhancement

#### Issue(s) number: 
#5081 

#### Description:

- the overloaded new/delete operator will influence all thread, include third-party's, like rocksdb's flush/compact threads, thrift rpc threads, which is not what we want.
- introduce a throwOnMemoryExceeded in ThreadMemoryStats, it indicat whether throw on memory exceed, default to FALSE, then all thread is default will not get bad_alloc, unless a MemoryCheckGuard is create by that thread;
- MemoryCheckGuard:  used when some routine need mem check turn on, when this guard was added to our code path, will turn on that code scope's memory check, until it freed (turn off on destruction);

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....


src/graph/executor/algo/BFSShortestPathExecutor.cpp
```diff
@@ -29,18 +29,26 @@ folly::Future<Status> BFSShortestPathExecutor::execute() {
...
+    memory::MemoryCheckGuard guard;
+    return buildPath(true);
+  });
   futures.emplace_back(std::move(leftFuture));
   futures.emplace_back(std::move(rightFuture));
 
   return folly::collect(futures)
       .via(runner())
       .thenValue([this](auto&& status) {
+        memory::MemoryCheckGuard guard;
```
Shylock-Hg, line 46: I think you could enable it in `Schedueler::execute` and disable it in callback of `Future`.

codesigner: > I think you could enable it in `Schedueler::execute` and disable it in callback of `Future`.

`QueryInstance::execute` is the caller of `Schedueler::execute`, it already have MemoryCheckGuard on, see change in QueryInstance.cpp

The thread invoke thenValue's callback was another thread from folly, it should turn MemoryCheckGuard on by itself.

critical27:
Maybe need to consider adding`MemoryCheckGuard` to the `thenValue` in `StorageClientBase::collectResponse/getResponse`. That is where we may encounter oom when retrieving data from storage at the first place.

codesigner:
> Maybe need to consider adding `MemoryCheckGuard` to the `thenValue` in `StorageClientBase::collectResponse/getResponse`. That is where we may encounter oom when retrieving data from storage at the first place.

OK, let me check.

czpmango (dismissed):

codesigner:
After this, I will mannually add some throw bad_alloc logic, to verify code path we turned on should successfully failed that query.

> Maybe need to consider adding `MemoryCheckGuard` to the `thenValue` in `StorageClientBase::collectResponse/getResponse`. That is where we may encounter oom when retrieving data from storage at the first place.

added

czpmango (approved):

codesigner:
> Could this mechanism cover encode/decode process of thrift?

Not quite sure, is the decode/dencode you mean is code in like DataOps-inl.h, if the decode/encode is write by ourselfs, then yes.

PR #5175, opened 2022-12-31
modify jobmanager ut

SuperYoko:
## What type of PR is this?
- [ ] bug
- [ ] feature
- [x] enhancement

#### Description:

Add add/stop/recover/show job ut test.
And little fix.

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....

### DIFF
--- a/src/meta/processors/job/JobDescription.cpp
+++ b/src/meta/processors/job/JobDescription.cpp
@@ -83,12 +83,16 @@ cpp2::JobDesc JobDescription::toJobDesc() {
   return ret;
 }
 
+bool JobDescription::notSetable(Status status) {
+  return status == cpp2::JobStatus::FINISHED;
+}
+
 bool JobDescription::setStatus(Status newStatus, bool force) {
-  if (status_ == cpp2::JobStatus::FINISHED) {
-    // no one should change the status of a finished job
-    // in case that recovered job running again, but set finished to finished should be ok
-    return newStatus == cpp2::JobStatus::FINISHED;
+  if (notSetable(status_)) {
+    // no need to change time.
+    return status_ == newStatus;
   }
+
   if (JobStatus::laterThan(status_, newStatus) && !force) {
     return false;
   }
--- a/src/meta/processors/job/JobDescription.cpp
+++ b/src/meta/processors/job/JobDescription.cpp
@@ -11,6 +11,7 @@
 #include <stdexcept>
 
 #include "common/utils/MetaKeyUtils.h"
+#include "interface/gen-cpp2/meta_types.h"
 #include "kvstore/KVIterator.h"
 #include "meta/processors/Common.h"
 
@@ -83,6 +84,11 @@ cpp2::JobDesc JobDescription::toJobDesc() {
 }
 
 bool JobDescription::setStatus(Status newStatus, bool force) {
+  if (status_ == cpp2::JobStatus::FINISHED) {
+    // no one should change the status of a finished job
+    // in case that recovered job running again, but set finished to finished should be ok
+    return newStatus == cpp2::JobStatus::FINISHED;
+  }
   if (JobStatus::laterThan(status_, newStatus) && !force) {
     return false;
   }
--- a/src/meta/processors/job/JobDescription.cpp
+++ b/src/meta/processors/job/JobDescription.cpp
@@ -83,12 +83,8 @@ cpp2::JobDesc JobDescription::toJobDesc() {
   return ret;
 }
 
-bool JobDescription::notSetable(Status status) {
-  return status == cpp2::JobStatus::FINISHED;
-}
-
 bool JobDescription::setStatus(Status newStatus, bool force) {
-  if (notSetable(status_)) {
+  if (JobStatus::notSetable(status_)) {
     // no need to change time.
     return status_ == newStatus;
   }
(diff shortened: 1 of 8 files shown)


src/meta/processors/job/JobManager.cpp
```diff
@@ -869,7 +879,8 @@ ErrorOr<nebula::cpp2::ErrorCode, uint32_t> JobManager::recoverJob(
       JobID jid;
       bool jobExist = checkOnRunningJobExist(spaceId, job.getJobType(), job.getParas(), jid);
       if (!jobExist) {
-        job.setStatus(cpp2::JobStatus::QUEUE, true);
+        job.setStatus(cpp2::JobStatus::QUEUE, true);  // which cause the job execute again
+        job.setErrorCode(nebula::cpp2::ErrorCode::E_JOB_SUBMITTED);
```
liwenhui-soul, line 883: what's the usage of the error coed?

SuperYoko: > what's the usage of the error coed?

indicate a job is submitted, it's default errorcode.

src/meta/processors/job/JobDescription.cpp
```diff
@@ -83,6 +84,11 @@ cpp2::JobDesc JobDescription::toJobDesc() {
 }
 
 bool JobDescription::setStatus(Status newStatus, bool force) {
+  if (status_ == cpp2::JobStatus::FINISHED) {
+    // no one should change the status of a finished job
+    // in case that recovered job running again, but set finished to finished should be ok
+    return newStatus == cpp2::JobStatus::FINISHED;
```
pengweisong, line 90: 1. could add a log here?
2. could wrap the logic to a function:` JobStatus::isUnsettable(status)`  and comment on the function?
```
if ( JobStatus::isUnsettable(status)) {
  return newStatus == status_;
}
```

src/meta/processors/job/JobManager.cpp
```diff
@@ -682,6 +686,10 @@ bool JobManager::isExpiredJob(JobDescription& jobDesc) {
     return false;
   }
   auto jobStart = jobDesc.getStartTime();
+  if (jobStart == 0) {
+    // should not happend, but just in case keep this job
```
pengweisong, line 690: Then add some error log?

src/meta/processors/job/JobManager.cpp
```diff
@@ -42,7 +43,10 @@ JobManager* JobManager::getInstance() {
   return &inst;
 }
 
-bool JobManager::init(nebula::kvstore::KVStore* store, AdminClient* adminClient) {
+bool JobManager::init(nebula::kvstore::KVStore* store,
+                      AdminClient* adminClient,
+                      std::shared_ptr<JobExecutorFactory> factory) {
```
pengweisong, line 48: how about `JobExecutorFactory*` keep consistent with other parameter, besides raw pointer has better performance.

SuperYoko: > how about `JobExecutorFactory*` keep consistent with other parameter, besides raw pointer has better performance.

Better to make JobManager to control its life. it should not affect performance.

src/meta/processors/job/JobManager.cpp
SuperYoko: okay.

src/meta/processors/job/JobDescription.cpp
SuperYoko: Think it's small enough. better to make put all logic here.

src/meta/processors/job/JobManager.cpp
pengweisong: Then why not use an  unique_ptr in JobManager, which could make it explicit which lifecycle it belongs.

SuperYoko: > Then why not use an unique_ptr in JobManager, which could make it explicit which lifecycle it belongs.

I use it for test, test drives me use a shared pointer which I can better mock it's behavior, Maybe unique ptr is better.

SuperYoko: > Then why not use an unique_ptr in JobManager, which could make it explicit which lifecycle it belongs.

I use it for test, test drives me use a shared pointer which I can better mock it's behavior, Maybe unique ptr is better.

src/meta/processors/job/JobDescription.cpp
pengweisong: I think whether we wrap an function does not depends on how small it is. But about:
1. explicit, make the comment to an function name.
2. consistent, with the `JobStatus::laterThan`. Sentences should have similar abstract level.

SuperYoko: > I think whether we wrap an function does not depends on how small it is. But about:
> 
> 1. explicit, make the comment to an function name.
> 2. consistent, with the `JobStatus::laterThan`. Sentences should have similar abstract level.

make sence to me.

liwenhui-soul (approved):

src/meta/processors/job/JobDescription.cpp
```diff
@@ -82,7 +83,16 @@ cpp2::JobDesc JobDescription::toJobDesc() {
   return ret;
 }
 
+bool JobDescription::notSetable(Status status) {
```
pengweisong, line 86: move to JobStatus::notSetable ?

SuperYoko: mybad.

pengweisong (approved):

liwenhui-soul (approved):

critical27 (approved):
Great work~

PR #5176, opened 2023-01-02
optimize match node label

nevermore3:
## What type of PR is this?
- [ ] bug
- [ ] feature
- [x] enhancement

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....

### DIFF
--- a/src/graph/executor/query/TraverseExecutor.cpp
+++ b/src/graph/executor/query/TraverseExecutor.cpp
@@ -36,7 +36,6 @@ Status TraverseExecutor::buildRequestVids() {
   auto iter = static_cast<SequentialIter*>(inputIter);
   size_t iterSize = iter->size();
   vids_.reserve(iterSize);
-  initVids_.reserve(iterSize);
   auto* src = traverse_->src();
   QueryExpressionContext ctx(ectx_);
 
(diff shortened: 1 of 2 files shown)


src/graph/executor/query/TraverseExecutor.cpp
```diff
@@ -164,27 +160,18 @@ folly::Future<Status> TraverseExecutor::handleResponse(RpcResponse&& resps) {
...
-    existVids.reserve(iter->numRows());
+    initVertices_.reserve(iter->numRows());
     auto vertices = iter->getVertices();
+    // match (v)-[e:Rel]-(v1:Label1)-[e1*2]->() where id(v0) in [6, 23] return v1
+    // save the vertex that meets the filter conditions as the starting vertex of the current
+    // traverse
     for (auto& vertex : vertices.values) {
       if (vertex.isVertex()) {
-        existVids.emplace(vertex);
+        initVertices_.emplace_back(vertex);
```
yixinglu, line 170: Isn't it necessary to deduplicate the vids for next traverse ?

nevermore3: the first column vid in the result returned by getNeighbor is not repeated

yixinglu: make sense.

yixinglu (approved):
LGTM

codesigner (approved):

PR #5180, opened 2023-01-03
fix processor_test timeout

critical27:
## What type of PR is this?
- [X] bug
- [ ] feature
- [X] enhancement

#### Description:
Introduced in #5130, `processor_test` may timeout because history schemas will be scanned. And there is a bug in alter processor...

## How do you solve it?
Need to lower iteration times, it takes about 2 mins in my env with debug version, let's see if ci could survive.

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....


SuperYoko (approved):

yixinglu (approved):

xtcyclist:
shou dong dian zan

PR #5188, opened 2023-01-03
fix stderr  save error log

HarrisChu:
syntax error would be saved in both `nebula-graphd.ERROR` and `graphd-stderr.log`.
to save disk usage, improve the stderrthreshold level.


czpmango (approved):

abby-cyber:
https://github.com/vesoft-inc/nebula-docs-cn/pull/2495

PR #5199, opened 2023-01-04
improve memtracker, add missed check & remove unnecessary thenError&tryCatch check

codesigner:
## What type of PR is this?
- [x] bug
- [ ] feature
- [x] enhancement

#### Issue(s) number: 

https://github.com/vesoft-inc/nebula/issues/5205
https://github.com/vesoft-inc/nebula/issues/5206

#### Description:
- When memory tracker is turned on, we should make sure if some std::bac_alloc throws, the corresponding query should be killed with correct error message. For this purpose, I test with mannually add `throw std::bac_alloc` to check expected bahavior happens.
- Almost every memtracker turned on code block, tested with mannually throw bad_alloc, check query failed with except error, if `// MemoryTrackerVerified` added to that context, these places were manually test;

- graphd: 
    - some thenError is remove, which is not nessacery, future<Status> most of time can chined and returned to up callstack; (what @Shylock-Hg comment  in previouse pr it right,  but have no time to check every place, some place remains need thenError)
    - some code block missed turned on memtracker, added;

- storaged:
    -  remove XXXProcessor::process's try catch, added the root catch in  GraphStorageServiceHandler.cpp;
    -  added a MemoryCheckScope to ease wrap a function in memory tracker turned on context, used by async execution submitted to executor;
 
- rpc
    - added memroy tracker in DataSet::read that decode DataSet  from wire， every rpc response from stroage is defined as a DataSet, so I think this is enough; have tried add a throw in DataSet::read, it works as expected.

- fix a build issue when build using gcc with sanitize

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....

### DIFF
--- a/src/storage/CMakeLists.txt
+++ b/src/storage/CMakeLists.txt
@@ -109,4 +109,4 @@ nebula_add_library(
 )
 
 nebula_add_subdirectory(stats)
-#nebula_add_subdirectory(test)
+nebula_add_subdirectory(test)
--- a/src/storage/CMakeLists.txt
+++ b/src/storage/CMakeLists.txt
@@ -109,4 +109,4 @@ nebula_add_library(
 )
 
 nebula_add_subdirectory(stats)
-nebula_add_subdirectory(test)
+#nebula_add_subdirectory(test)
--- a/cmake/nebula/SanitizerConfig.cmake
+++ b/cmake/nebula/SanitizerConfig.cmake
@@ -15,6 +15,7 @@ if(ENABLE_ASAN)
     add_compile_options(-fsanitize=address)
     add_compile_options(-g)
     add_compile_options(-fno-omit-frame-pointer)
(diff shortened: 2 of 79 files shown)


xtcyclist (approved):

src/storage/CMakeLists.txt
```diff
@@ -109,4 +109,4 @@ nebula_add_library(
 )
 
 nebula_add_subdirectory(stats)
-nebula_add_subdirectory(test)
+#nebula_add_subdirectory(test)
```
yixinglu, line 112: do you forget to restore these tests?

codesigner: > do you forget to restore these tests?

Yes, good catch

critical27 (approved):

PR #5227, opened 2023-01-09
Update response message when adding schema historically existed

wenhaocs:
## What type of PR is this?
- [ ] bug
- [ ] feature
- [x] enhancement

#### Issue(s) number: 
Followup of Case 1 in https://github.com/vesoft-inc/nebula/issues/2671
Previous PR: #5130 
In this PR, we distinguish behaviors of adding column that is currently existed from historically existed. So users will get better hints.

image

## Checklist:
Tests:
- [x] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....

### DIFF
--- a/src/meta/MetaServiceUtils.cpp
+++ b/src/meta/MetaServiceUtils.cpp
@@ -87,11 +87,20 @@ nebula::cpp2::ErrorCode MetaServiceUtils::alterColumnDefs(
     bool isEdge) {
   switch (op) {
     case cpp2::AlterSchemaOp::ADD:
+      // Check the current schema first. Then check all schemas.
+      for (auto it = cols.begin(); it != cols.end(); ++it) {
+        if (it->get_name() == col.get_name()) {
+          LOG(INFO) << "Column existing: " << col.get_name();
+          return nebula::cpp2::ErrorCode::E_EXISTED;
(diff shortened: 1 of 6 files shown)


critical27 (dismissed):

SuperYoko (approved):

src/meta/MetaServiceUtils.cpp
```diff
@@ -87,11 +87,20 @@ nebula::cpp2::ErrorCode MetaServiceUtils::alterColumnDefs(
     bool isEdge) {
   switch (op) {
     case cpp2::AlterSchemaOp::ADD:
+      // Check the current schema first. Then check all schemas.
+      for (auto it = cols.begin(); it != cols.end(); ++it) {
+        if (it->get_name() == col.get_name()) {
+          LOG(INFO) << "Column existing: " << col.get_name();
```
yixinglu, line 93: LOG(ERROR)

SuperYoko: > LOG(ERROR)

Do we treat this as error ?

yixinglu (approved):

critical27 (approved):

Issue #5209, opened 2023-01-05
match query's result is incorrect


nevermore3:
**Please check the FAQ documentation before raising an issue**

**Describe the bug (__required__)**

**Your Environments (__required__)**

* OS: `uname -a`
* Compiler: `g++ --version` or `clang++ --version`
* CPU: `lscpu`
* Commit id (e.g. `a3ffc7d8`)

**How To Reproduce(__required__)**

 
 ```
     MATCH (v:player{name: 'Tim Duncan'})-[e:like*3]->(n), (t:team {name: "Spurs"})
      WITH v, e, collect(distinct n) AS ns
        UNWIND [n in ns | ()-[e*3]->(n:player)] AS p
        RETURN p
```

neo4j's result:
<img width="1272" alt="image" src="https://user-images.githubusercontent.com/6930445/210741383-9df929f3-32f1-4a2e-ae1e-e4681e149812.png">
    
our's result:
```
(root@nebula) [nba]>  MATCH (v:player{name: 'Tim Duncan'})-[e:like*3]->(n), (t:team {name: "Spurs"})  WITH v, e, collect(distinct n) AS ns  UNWIND [n in ns | ()-[e*3]->(n:player)] AS p  RETURN p 
+----+
| p  |
+----+
| [] |
| [] |
| [] |
| [] |
| [] |
+----+
Got 5 rows (time spent 125497/126082 us)
```

1. Step 1
2. Step 2
3. Step 3

**Expected behavior**

**Additional context**

**Comment by yixinglu** [comment]
It seems related to the issue #5221

**Comment by nevermore3** [comment]
```
(root@nebula) [nba]> MATCH (v:player{name: 'Tim Duncan'})-[e:like*3]->(n)   WITH  e,collect(distinct n) AS ns        unwind [a in ns | ()-[e*3]->(a:player)] AS p   return p
+----+
| p  |
+----+
| [] |
| [] |
| [] |
| [] |
| [] |
+----+
Got 5 rows (time spent 121702/122480 us)
```
when we modify the steps to 3, there is no result
```
(root@nebula) [nba]> MATCH (v:player{name: 'Tim Duncan'})-[e:like*2]->(n)   WITH  e,collect(distinct n) AS ns        unwind [a in ns | ()-[e*2]->(a:player)] AS p   return p
+-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
[...]

PR #5268, opened 2023-01-17
Disable edgelist join

nevermore3:
## What type of PR is this?
- [x] bug
- [ ] feature
- [ ] enhancement

#### Issue(s) number: 
close https://github.com/vesoft-inc/nebula/issues/5209

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....


src/parser/MatchPath.cpp
```diff
@@ -92,6 +92,9 @@ std::string MatchPath::toString() const {
   buf.reserve(256);
 
   if (alias_ != nullptr) {
+    if (alias_->front() == '(') {
```
yixinglu, line 95: this check maybe not always true, do you test the following case?

```
match (v)
return ((v)-[e]-()) as t
```

yixinglu (dismissed):
LGTM

src/graph/validator/MatchValidator.cpp
```diff
@@ -126,6 +126,9 @@ Status MatchValidator::validateImpl() {
 // \param matchClauseCtx the context of current match clause
 Status MatchValidator::validatePath(const MatchPath *path, MatchClauseContext &matchClauseCtx) {
   matchClauseCtx.paths.emplace_back();
+  if (path->alias() == nullptr) {
+    const_cast<MatchPath *>(path)->setPredicate();
+  }
```
czpmango, line 131: What does this mean, that paths without aliases are all predicates?

nevermore3: yes

src/graph/validator/MatchValidator.cpp
```diff
@@ -185,7 +188,7 @@ Status MatchValidator::buildPathExpr(const MatchPath *path,
   pathBuild->add(InputPropertyExpression::make(pool, nodeInfos.back().alias));
   pathInfo.pathBuild = std::move(pathBuild);
   pathInfo.anonymous = false;
-  pathInfo.alias = *pathAlias;
+  pathInfo.alias = pathAlias == nullptr ? path->toString() : *pathAlias;
   return Status::OK();
```
czpmango, line 192: These codes are confusing.
When the path has no alias, why does the `PathBuild` expression need to be built, and why is the anonymous flag not set to true?

nevermore3: match (a:player) return (a)-[]->() . the pattern in return has no alias, but needs to output path。the variable anonymous is useless

src/graph/validator/MatchValidator.cpp
```diff
@@ -1292,19 +1292,25 @@ Status MatchValidator::validatePathInWhere(
...
     // The inner variable of expression will be replaced by anno variable
     if (!node->alias().empty() && node->alias()[0] != '_') {
       pathInfo.compareVariables.emplace_back(node->alias());
     }
   }
   for (const auto &edge : path->edges()) {
-    if (edge->alias()[0] != '_') {
-      pathInfo.compareVariables.emplace_back(edge->alias());
+    const auto &edgeAlias = edge->alias();
+    if (!edgeAlias.empty() && edgeAlias.front() != '_') {
```
czpmango, line 1303: We should try to avoid this kind of code.
Maybe `edge->isAnonymous()` is more appropriate.

src/graph/validator/MatchValidator.cpp
```diff
@@ -1292,19 +1292,25 @@ Status MatchValidator::validatePathInWhere(
...
+      if (edge->range()) {
+        return Status::SemanticError(
+            "Variable '%s` 's type is list. not support used in multiple patterns "
```
czpmango, line 1306: Use a friendlier error message: `Repeating variable length edge variables not supported`.

nevermore3: i think it's ok

src/graph/validator/MatchValidator.cpp
nevermore3: fixed

src/graph/validator/MatchValidator.cpp
yixinglu: prefer some comments about this update

nevermore3: done

czpmango: `setPredicate()` this function is designed to mark a path pattern to the predicate not a path pattern without aliases.
I cannot get your point.

src/graph/validator/MatchValidator.cpp
czpmango: If an anonymous pattern is set to a non-anonymous pattern, or vice versa, how do you ensure that other code using that data structure doesn't misuse it?

src/graph/validator/MatchValidator.cpp
czpmango: Better to use `Edge List` instead of `list`.

src/parser/MatchPath.h
```diff
@@ -115,6 +115,10 @@ class MatchEdge final {
     return range_.get();
   }
 
+  bool anonymous() const {
+    return alias_.empty() || alias_.front() == '_';
```
czpmango, line 119: Better to use `AnonVarGenerator::isAnnoVar()`.
IMHO, we should manage all anonymous variables in query context.

src/graph/validator/MatchValidator.cpp
nevermore3: Yes, but I saw the use of the predicate function when generating an execution plan, which means the same as not generating a path structure

nevermore3: Annotations have been added where used

src/graph/validator/MatchValidator.cpp
nevermore3: fixed

src/graph/validator/MatchValidator.cpp
czpmango: > Yes, but I saw the use of the predicate function when generating an execution plan, which means the same as not generating a path structure

It is true that a pattern predicate does not produce a path data at runtime, but that does not mean something that doesn't produce a path data should be set as a path predicate.

nevermore3: what is your suggestion? Redefine a variable?

src/graph/planner/match/MatchSolver.cpp
```diff
@@ -278,7 +278,7 @@ void MatchSolver::buildProjectColumns(QueryContext* qctx, const Path& path, SubP
   DCHECK(!nodeInfos.empty());
   addNode(nodeInfos.back());
 
-  if (!path.anonymous) {
+  if (path.genPath) {
```
czpmango, line 281: Is that necessary?

nevermore3: `match (v) unwind (v)-[]->() AS p return p`
pattern in unwind statement has no path variable。but need to generate path struct

tests/tck/features/match/PathExpr.feature
```diff
@@ -413,37 +413,64 @@ Feature: Basic match
...
+    Then a SemanticError should be raised at runtime: Variable 'e` 's type is edge list. not support used in multiple patterns simultaneously.
     When executing query:
       """
       MATCH (v:player)-[e:like*3]->(n)
         WHERE (n)-[e*3]->(:player)
         RETURN v
       """
-    Then the result should be, in any order:
-      | v |
+    Then a SemanticError should be raised at runtime: Variable 'e` 's type is edge list. not support used in multiple patterns simultaneously.
```
czpmango, line 466: Add test cases:
```
MATCH (v:player)-[e:like]->(n) WHERE (n)-[e*1..4]->(:player) RETURN v
MATCH (v:player)-[e:like*3]->(n) WHERE (n)-[e*1]->(:player) RETURN v
MATCH (v:player)-[e:like*1..3]->(n) WHERE (n)-[e]->(:player) RETURN v
```
```
MATCH (v:player)-[e:like]->(n), (n)-[e*1..4]->(:player) RETURN v
MATCH (v:player)-[e:like*3]->(n) MATCH (n)-[e*1]->(:player) RETURN v
MATCH (v:player)-[e:like*1..3]->(n) MATCH (n)-[e]->(:player) RETURN v
```
```
MATCH (v:player)-[e:like]->(n) WITH (n)-[e*1..4]->(:player) AS v RETURN v
MATCH (v:player)-[e:like*3]->(n) UNWIND (n)-[e*1]->(:player) as v RETURN v
MATCH (v:player)-[e:like*1..3]->(n) UNWIND collect(n) as ns WITH ns[0] AS n,e AS e MATCH (n)-[e]->(v:player) RETURN v
```

Sophie-Xie:
@nevermore3 Did you discuss it with @MuYiYong?

tests/tck/features/match/PathExpr.feature
nevermore3: ack

nevermore3:
> @nevermore3 Did you discuss it with @MuYiYong?

opencypher has no clear definition of this behavior, so we  disable

yixinglu (approved):

jievince (approved):

Issue #5272, opened 2023-01-17
Report errors on where clauses in optional match queries


nebula-bots:
Optional match does not support where clauses for the time being. Queries like `... optional match (v) where v.player.age == 1000 ...` may very likely return wrong results. Thus, we need to ban it and report errors.

It is ok to use a singular optional match without where clauses, like `... optional match (v:player) match ...`.

PR #5273, opened 2023-01-17
report errors on where clauses in optional match queries.

xtcyclist:
## What type of PR is this?
- [X] bug
- [ ] feature
- [ ] enhancement

#### Issue(s) number: 
Close https://github.com/vesoft-inc/nebula/issues/5272

#### Description:
To support queries like `OPTIONAL MATCH <pattern> WHERE <condition>` has been a beta feature, in which the `<condition>` cannot be evaluated correctly in all cases in the current version. A simpler query like `OPTIONAL MATCH <pattern>` without a where clause, on the other hand, can be evaluated correctly.

## How do you solve it?
Report errors on the where clauses in optional match queries.

## Checklist:
Tests:
- [x] Function test
- [ ] Performance test
- [ ] N/A

Affects:

**If users have written queries like `optional match ... where...`, it may have appeared to be running ok, although it is not. With this PR, such queries would run into error messages.**

## Release notes:

> ex. Fixed the bug .....


src/parser/parser.yy
```diff
@@ -1682,7 +1682,11 @@ match_clause
         $$ = new MatchClause($2, $3, false/*optional*/);
     }
     | KW_OPTIONAL KW_MATCH match_path_list where_clause {
-        $$ = new MatchClause($3, $4, true);
+        if ($4 != nullptr) {
+            throw nebula::GraphParser::syntax_error(@4, "Where clause in optional match is not supported.");
```
yixinglu, line 1686: delete $3 before throwing error to avoid memory leak of $3.

yixinglu: and delete $4

xtcyclist: Fixed.

yixinglu (approved):
LGTM

codesigner (approved):

abby-cyber:
https://github.com/vesoft-inc/nebula-docs-cn/pull/2515

PR #5276, opened 2023-01-19
fix memtracker bugs during stress test on graphd and storaged

codesigner:
## What type of PR is this?
- [x] bug
- [ ] feature
- [ ] enhancement

##### graphd
- fix multiple not throw safe code path
   - fix Value's setXXX, should set type only when set ptr succedd, during request deserialize;
   - fix ObjectPool deadlock if throw on alloc memory
   - fix Symbol & ExecutionContext's valueMap_, not thread-safe when multiple Executor run concurrently

- fix scheduler
   - fix AsyncMsgNotifyBasedScheduler's scheduling logic, add wait mechinism to guarantee Executors all finish

  

##### storaged
- write operators:  turn off memtracker on stroage's write path
- read operators:  fix uncaught execption & memory leak for storage's read Operators

   
##### others
   - remove noexcept of functions those need alloc memory;
   - add MemoryCheckOffGuard to MetaClient's functions, meta need little memory, but they are important, turn off check on all these functions.

##### Tests
- graphd
   - running days on limited memory, triger memory exceeded for most queries, no crashed;
- storaged
  - run through three test scenario, not crashed,  in https://github.com/vesoft-inc/nebula-ent/issues/2221
     - python3 run.py stress run -scenario go.Go3StepEdge
     - python3 run.py stress run -scenario lookup.LookUp
     - python3 run.py stress run -scenario  find_path.FindShortestPath

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....


critical27 (approved):
Well done~ Hope we could survive long term test...

xtcyclist (dismissed):

critical27 (approved):

src/common/memory/MemoryUtils.cpp
```diff
@@ -118,18 +127,77 @@ StatusOr<bool> MemoryUtils::hitsHighWatermark() {
...
+      if (trackable > 0) {
+        MemoryStats::instance().setLimit(trackable * limitRatio);
+      } else {
+        LOG(ERROR) << "Total memory less than untracked " << untrackedMb << " Mib,"
+                   << " MemoryTracker limit is not set properly, "
+                   << " Current memory stats:" << MemoryStats::instance().toString();
+      }
+      LOG(INFO) << "MemoryTracker set static ratio: " << limitRatio;
+    }
+  } else if (limitRatio == 2.0) {
```
yixinglu, line 173: best not to use equals to compare floating point data types

src/common/memory/MemoryUtils.cpp
```diff
@@ -118,18 +127,77 @@ StatusOr<bool> MemoryUtils::hitsHighWatermark() {
...
+     *           limit = current_used_memory + (available - untracked) * availableRatio
+     */
+    double trackableAvailable = available - (untrackedMb * MiB);
+    if (trackableAvailable > 0) {
+      MemoryStats::instance().updateLimit(trackableAvailable * availableRatio);
+    } else {
+      MemoryStats::instance().updateLimit(0);
+    }
+    DLOG(INFO) << "MemoryTracker set to dynamic self adaptive";
+  } else if (limitRatio == 3.0) {
```
yixinglu, line 185: ditto

src/common/memory/MemoryUtils.cpp
```diff
@@ -118,18 +127,77 @@ StatusOr<bool> MemoryUtils::hitsHighWatermark() {
...
+    /**
+     * (== 3.0): Special value for disable memory_tracker;
+     *           limit is set to max
+     */
+    if (kCurrentLimitRatio_ != limitRatio) {
```
yixinglu, line 190: ditto

src/graph/scheduler/AsyncMsgNotifyBasedScheduler.cpp
```diff
@@ -225,17 +239,94 @@ void AsyncMsgNotifyBasedScheduler::notifyError(std::vector<folly::Promise<Status
...
 
+folly::Future<Status> AsyncMsgNotifyBasedScheduler::runExecute(Executor* executor) const {
+  // catch Executor::execute here, upward call stack should only get Status, no exceptions.
+  try {
+    addExecuting(executor);
+    if (hasFailStatus()) return failedStatus_.value();
+    folly::Future<Status> status = Status::OK();
+    {
+      memory::MemoryCheckGuard guard;
+      status = executor->execute();
```
yixinglu, line 270: plz NOTE: the `execute` is an async function, your memory check guard may be invalid when the tasks really run.

src/graph/scheduler/AsyncMsgNotifyBasedScheduler.cpp
```diff
@@ -225,17 +239,94 @@ void AsyncMsgNotifyBasedScheduler::notifyError(std::vector<folly::Promise<Status
...
+  return ss.str();
+}
+
+void AsyncMsgNotifyBasedScheduler::appendExecutor(size_t spaces,
+                                                  Executor* executor,
+                                                  std::stringstream& ss) {
+  ss << std::string(spaces, ' ') << formatPrettyId(executor) << std::endl;
+  for (auto depend : executor->depends()) {
+    appendExecutor(spaces + 1, depend, ss);
+  }
```
yixinglu, line 328: do the loop body executors need to be considered here?

src/graph/scheduler/AsyncMsgNotifyBasedScheduler.cpp
codesigner: > plz NOTE: the `execute` is an async function, your memory check guard may be invalid when the tasks really run.

This MemoryCheckGuard is only used by some Executor return a already satisfied future, like DedupExecutor::execute, Async future's MemoryCheckGuard is not here, they are declare inside, for example in StorageClientBase-inl.h:

```
  return folly::collectAll(respFutures)
      .deferValue([this, requests = std::move(requests), totalLatencies, hosts](
                      std::vector<folly::Try<StatusOr<Response>>>&& resps) {
        // throw in MemoryCheckGuard verified
        memory::MemoryCheckGuard guard;
        StorageRpcResponse<Response> rpcRes
```

src/graph/scheduler/AsyncMsgNotifyBasedScheduler.cpp
codesigner: > do the loop body executors need to be considered here?

no, just print a original dependency tree.

src/common/memory/MemoryUtils.cpp
codesigner: > best not to use equals to compare floating point data types

I had considered it as a issue, but I tried in real run environment, it works as expected.

yixinglu (approved):
LGTM

critical27 (approved):

Issue #5235, opened 2023-01-11
Handling ambiguities in the bison glr parser


czpmango:
**Describe the bug**
An ambiguity from bison glr parser occurred when parsing some statements.

**Your Environment**
nebula 3.3.0

**How To Reproduce**
```
(root@nebula) [nba]> MATCH (v:player{name: 'Tim Duncan'})-[e:like*0..2]-(v2) WHERE size([ii in e WHERE (v)-[ii]-(v2) | ii])>1  RETURN count(*) AS cnt 
[ERROR (-1004)]: SyntaxError: syntax is ambiguous near `(v)-[ii]-(v2)'
```

**Expected behavior**
Parse out the correct AST.

czpmango:
Already fixed.

Issue #5221, opened 2023-01-09
Rethink variable scope validation for path pattern 


yixinglu:
**Please check the FAQ documentation before raising an issue**

**Describe the bug (__required__)**

At present, the variable validation in path pattern only happens in match clause, but when the path pattern placed in the expression context, the variable's behavior will not be expected.

for example:

```sql
MATCH (v:player{name: 'Tim Duncan'})-[e:like*0..2]-(v2)
WHERE size([i in e WHERE (v)-[i]-(v2) | i])>1
RETURN count(*) AS cnt

//------

MATCH (v:player{name: 'Tim Duncan'})-[e:like*0..2]-(v2)-[i]-(v3)
WHERE size([i in e WHERE (v)-[i]-(v2) | i])>1
RETURN count(*) AS cnt
```

above tests are from the PR #5215 

**Your Environments (__required__)**

* OS: `uname -a`
* Compiler: `g++ --version` or `clang++ --version`
* CPU: `lscpu`
* Commit id (e.g. `a3ffc7d8`)

**How To Reproduce(__required__)**

1. Step 1
2. Step 2
3. Step 3

**Expected behavior**

**Additional context**

yixinglu:
@czpmango could you add more tests discussed offline for this topic?

czpmango:
Variable shadowing error is unexpected:
```
(root@nebula) [nba]> match (v:player)-[e*2]->(n) where size([n in e where (v{name:"Tim Duncan"})-[n]-()])>3 return v
[ERROR (-1009)]: SemanticError: `n' is defined with type Node, but referenced with type Edge
```

Making an implicit join on the inner variable `n` and the outer variable `n` by mistake leads to an incorrect result. The expected row number is 4945(test in neo4j 4.2.1).
```
(root@nebula) [nba]>  match (v:player)-[e*2]->()-[n]-() where size([n in e where (v)-[n]-()])>0 return count(*)
+----------+
| count(*) |
+----------+
| 837      |
+----------+
Got 1 rows (time spent 132190/132459 us)
```

xtcyclist:
Moved this issue to the 3.5 milestone. `path pattern placed in the expression context` does not seem to be an urgent issue for 3.4.

yixinglu:
related to #5235

PR #5290, opened 2023-01-31
Remove glr parser

czpmango:
The previous implementation of pattern expression introduce bison glr parser to handle conflicts, but that is actually unnecessary and harmful. This pr refactors the implementation and fix some related bugs.

## What type of PR is this?
- [x] bug
- [ ] feature
- [x] enhancement

#### Issue(s) number: 
https://github.com/vesoft-inc/nebula/issues/5235

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:


tests/tck/features/match/Unwind.feature
```diff
@@ -149,7 +149,7 @@ Feature: Unwind clause
     Then a SyntaxError should be raised at runtime: syntax error near `WITH'
     When executing query:
       """
-      MATCH (a:player {name:"Tim Duncan"}) - [e:like] -> (b)
+      MATCH (a:player {name:"Tim Duncan"})-[e:like]->(b)
```
Shylock-Hg, line 152: These are obviously unaccepted.

czpmango: I don't think it's a big deal.

czpmango: You also could write as this: MATCH (a:player {name:"Tim Duncan"}) -[e:like]-> (b)

src/parser/parser.yy
```diff
@@ -1797,9 +1797,15 @@ match_path_list
...
-    : L_PAREN match_alias R_PAREN {
-        $$ = new MatchNode(*$2, nullptr, nullptr);
-        delete $2;
+    : L_PAREN R_PAREN {
+        $$ = new MatchNode();
+    }
+    | parenthesized_expression {
+        auto& e = $1;
+        if (e->kind() != Expression::Kind::kLabel) {
+            throw nebula::GraphParser::syntax_error(@1, "Invalid node pattern");
```
yixinglu, line 1806: delete $1 before throw exception

yixinglu (commented):
LGTM, prefer to test the case in #5221 and following tests:

```
with [1,2] as a, 2 as b, [1] as c return (a)-[b]-(c);
with [1,2] as a, 2 as b, [1] as c return (a)- [b] -(c);
```

yixinglu (approved):
Good job!

jievince (approved):

codesigner (approved):

codesigner:
This fixed some patten expressibility broblems, but also introduce some compatibility(which is not very common), we should notice the change in release note when 3.5 released.

Sophie-Xie:
> This fixed some patten expressibility problems, but also introduce some compatibility(which is not very common), we should notice the change in release note when 3.5 released. @Sophie-Xie

@MuYiYong pls review this incompatible pr, thanks.

src/parser/parser.yy
czpmango: The expression memory is managed by ObjectPool, and there is no need for delete, which would result in double free.

PR #5291, opened 2023-01-31
Fix the crash caused by failed downcast of iter

xtcyclist:
## What type of PR is this?
- [X] bug
- [ ] feature
- [ ] enhancement

#### Issue(s) number: 
Close https://github.com/vesoft-inc/nebula-ent/issues/2336
Close https://github.com/vesoft-inc/nebula/issues/5292

#### Description:
The cast of iter to GetNeighborsIter using static_cast may fail and leave a non-accessible GetNeighborsIter*, which would later crash the graphd on seg fault.

## How do you solve it?
Tried using dynamic_cast. On the advice by @yixinglu, changed to protect the static_cast by checking the kind of iter.

## Checklist:
Tests:
- [x] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....

### DIFF
--- a/src/graph/executor/algo/SubgraphExecutor.cpp
+++ b/src/graph/executor/algo/SubgraphExecutor.cpp
@@ -100,17 +100,20 @@ folly::Future<Status> SubgraphExecutor::handleResponse(RpcResponse&& resps) {
 
 void SubgraphExecutor::filterEdges(int version) {
   auto iter = ectx_->getVersionedResult(subgraph_->outputVar(), version).iter();
-  auto* gnIter = static_cast<GetNeighborsIter*>(iter.get());
-  while (gnIter->valid()) {
-    const auto& dst = gnIter->getEdgeProp("*", nebula::kDst);
-    if (validVids_.find(dst) == validVids_.end()) {
-      auto edge = gnIter->getEdge();
-      gnIter->erase();
-    } else {
-      gnIter->next();
+  auto* iterPtr = iter.get();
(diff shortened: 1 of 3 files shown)


src/graph/executor/algo/SubgraphExecutor.cpp
```diff
@@ -100,17 +100,19 @@ folly::Future<Status> SubgraphExecutor::handleResponse(RpcResponse&& resps) {
...
   auto iter = ectx_->getVersionedResult(subgraph_->outputVar(), version).iter();
-  auto* gnIter = static_cast<GetNeighborsIter*>(iter.get());
-  while (gnIter->valid()) {
-    const auto& dst = gnIter->getEdgeProp("*", nebula::kDst);
-    if (validVids_.find(dst) == validVids_.end()) {
-      auto edge = gnIter->getEdge();
-      gnIter->erase();
-    } else {
-      gnIter->next();
+  auto* gnIter = dynamic_cast<GetNeighborsIter*>(iter.get());
```
yixinglu, line 103: Prefer to check the Kind of Iterator than using `dynamic_cast` to cast the type of object. dynamic_cast has runtime cost and maybe need to turn on the compile option `RTTI`, not recommend to use `dynamic_cast`.

xtcyclist: Used kind instead. Fixed.

nevermore3 (approved):

yixinglu (approved):

PR #5314, opened 2023-02-06
Traverse executor in parallel

yixinglu:
## What type of PR is this?
- [ ] bug
- [ ] feature
- [x] enhancement

#### Description:

run traverse executor in parallel when the number rows of response greater than the threshold value.

## How do you solve it?

Expand each response of RPC in parallel

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....

### DIFF
--- a/src/graph/executor/query/TraverseExecutor.cpp
+++ b/src/graph/executor/query/TraverseExecutor.cpp
@@ -11,6 +11,10 @@
 #include "graph/util/SchemaUtil.h"
 #include "graph/util/Utils.h"
 
+DEFINE_uint64(traverse_parallel_threshold_rows,
+              150000,
+              "threshold row number of traverse executor in parallel");
+
 using nebula::storage::StorageClient;
 using nebula::storage::StorageRpcResponse;
 using nebula::storage::cpp2::GetNeighborsResponse;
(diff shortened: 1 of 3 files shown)


src/graph/executor/query/TraverseExecutor.cpp
```diff
@@ -157,10 +163,98 @@ size_t TraverseExecutor::numRowsOfRpcResp(const RpcResponse& resps) const {
...
+        }
+        otherStats_.emplace("expandTaskRunTime", folly::toPrettyJson(taskRunTimeArray));
+
+        return Status::OK();
+      });
+}
+
+folly::Future<Status> TraverseExecutor::expandOneStep(RpcResponse&& resps) {
+  auto numRows = numRowsOfRpcResp(resps);
+  if (numRows < FLAGS_traverse_parallel_threshold_rows) {
```
jievince, line 253: `<` ???

yixinglu: When return too many vertices from storage, the collect task will speed a lot time. so not use the parallel execution in that case.

nevermore3 (approved):

Issue #5320, opened 2023-02-07
meta data inconsistent


nebula-bots:
**Your Environments (__required__)**
3.4.0-ent rc： f82007406455b8c3e54b9ec23b77238c2cc8f4a4
graph： nebula-console -addr 192.168.8.224  -port 10010 -u root -p 'nebula'
nebula： 192.168.8.224:/data/lb_test_3.4.0_rc

**problem description**
space 1、2、3 have been droped, but the space info still exists in the meta follower metad2
![image](https://user-images.githubusercontent.com/104358440/217184893-eb956c42-72fc-499c-8877-92cf427d1b2f.png)

PR #5321, opened 2023-02-07
Fix meta's cleanup not work as expected

critical27:
## What type of PR is this?
- [X] bug
- [ ] feature
- [ ] enhancement

#### Issue(s) number: 
Close https://github.com/vesoft-inc/nebula-ent/issues/2399
Close https://github.com/vesoft-inc/nebula/issues/5320

#### Description:
Meta does not clean up all data before receiving snapshot... The bug must have existed for quite a long time...

## How do you solve it?
Delete all data by delete range. The code is quite ugly, but I don't think of any better ways...

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....

### DIFF
--- a/src/kvstore/Part.cpp
+++ b/src/kvstore/Part.cpp
@@ -7,6 +7,7 @@
 
 #include "common/time/ScopedTimer.h"
 #include "common/utils/IndexKeyUtils.h"
+#include "common/utils/MetaKeyUtils.h"
 #include "common/utils/NebulaKeyUtils.h"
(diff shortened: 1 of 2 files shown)


wenhaocs:
> Weird... hope that clean up part is not called to many times.

Will the receiving snapshot happen very often? If not, I think that's fine. Or if it happens very often, the performance and availability will also be impacted even without this cleanup.

wenhaocs (approved):

PR #5338, opened 2023-02-14
Remove the lock from collectResponse in storage client

yixinglu:
## What type of PR is this?
- [ ] bug
- [ ] feature
- [X] enhancement

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....

### DIFF
--- a/src/clients/storage/StorageClientBase.h
+++ b/src/clients/storage/StorageClientBase.h
@@ -35,6 +35,7 @@ class StorageRpcResponse final {
 
   explicit StorageRpcResponse(size_t reqsSent) : totalReqsSent_(reqsSent) {
     lock_ = std::make_unique<std::mutex>();
+    responses_.reserve(reqsSent);
   }
 
   bool succeeded() const {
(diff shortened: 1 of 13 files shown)


src/clients/storage/StorageClientBase.h
```diff
@@ -47,46 +48,67 @@ class StorageRpcResponse final {
...
-    std::lock_guard<std::mutex> g(*lock_);
+  void setLatencyUnsafe(const HostAddr& host, int32_t latency, int32_t e2eLatency) {
     if (latency > maxLatency_) {
       maxLatency_ = latency;
     }
     hostLatency_.emplace_back(std::make_tuple(host, latency, e2eLatency));
   }
 
-  void markFailure() {
+  void setLatency(const HostAddr& host, int32_t latency, int32_t e2eLatency) {
```
codesigner, line 58: Seems there no usage of the previouse safe version? if we only call these functions with single thread, just delete the lock.

yixinglu: yeah. There's no usage for these functions in multiple thread scenario. I'll delete the lock from the class.

codesigner (approved):

critical27 (approved):

critical27:
We need the lock before is that, the responses are handled in different threads, see https://github.com/vesoft-inc/nebula/pull/3977

PR #5360, opened 2023-02-20
Support match from id Set and fix some related issues

czpmango:
## What type of PR is this?
- [x] bug
- [x] feature
- [ ] enhancement

#### Description:
https://discuss.nebula-graph.com.cn/t/topic/12208

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

### DIFF
--- a/src/common/datatypes/Set.cpp
+++ b/src/common/datatypes/Set.cpp
@@ -11,6 +11,20 @@
 
 namespace nebula {
 
+Set Set::set_intersection(const Set& lhs, const Set& rhs) {
+  if (lhs.size() <= rhs.size()) {
+    Set iset;
+    for (const Value& key : lhs.values) {
+      if (rhs.values.count(key) > 0) {
+        iset.values.insert(key);
+      }
+    }
+    return iset;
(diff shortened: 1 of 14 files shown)


src/common/datatypes/Set.cpp
```diff
@@ -11,6 +11,20 @@
 
 namespace nebula {
 
+Set Set::set_intersection(const Set& lhs, const Set& rhs) {
```
Aiee, line 14: Use camelcase naming.

czpmango: Consistent with std::set_intersection naming.

jievince (approved):

src/common/datatypes/Set.cpp
Aiee: Let's keep the consistence in the project

yixinglu (approved):

abby-cyber:
https://github.com/vesoft-inc/nebula-docs-cn/pull/2692

Issue #5358, opened 2023-02-20
storage raft election leader always failed and graph query storage rpc timeout when rocksdb write stall


shixiangz:
Bg: storage executes ingest, rocksdb triggers write stall, storage will always report master election failure, query storage, it will report rpc timeout
reason：GraphStorageServiceHandler, RaftexService,RaftPart shares the cpu threads pool
![image](https://user-images.githubusercontent.com/122794081/219998244-fd83a4f8-771f-4d2c-a522-279cb0ee9695.png)
![image](https://user-images.githubusercontent.com/122794081/219998373-61f5cad2-cdb0-463e-b339-2079de1abb0d.png)
![image](https://user-images.githubusercontent.com/122794081/219998504-2a95ddbb-68ad-4904-b850-1dcc327e9e91.png)

The first step of RaftPart's heartbeat will write rocksdb, if now rocksdb write stall
, a host has a large number of part leaders, which will cause all cpu thread pools to be blocked in write rocksdb (assuming the write stall exceeds 30s), at this time storage raft will fail, and storage query will also fail
I think raftPart should be created separately without affecting the call of rpc
![image](https://user-images.githubusercontent.com/122794081/219998768-6e0fbe7a-2148-45f1-bf66-f91cee3d4532.png)

HarrisChu:
https://discuss.nebula-graph.com.cn/t/topic/12133/22

PR #5377, opened 2023-03-01
RaftPart use a separate execution thread pool to prevent blocking rpc and raft call

shixiangz:
https://github.com/vesoft-inc/nebula/issues/5358

### DIFF
--- a/.github/actions/tagname-action/action.yml
+++ b/.github/actions/tagname-action/action.yml
@@ -18,7 +18,7 @@ runs:
         tag=$(echo ${{ github.ref }} | rev | cut -d/ -f1 | rev)
         tagnum=$(echo $tag | sed "s/^v//")
         majorver=$(echo $tag | cut -d "." -f 1)
-        echo "::set-output name=tag::$tag"
-        echo "::set-output name=tagnum::$tagnum"
-        echo "::set-output name=majorver::$majorver"
+        echo "tag=$tag" >> $GITHUB_OUTPUT
+        echo "tagnum=$tagnum" >> $GITHUB_OUTPUT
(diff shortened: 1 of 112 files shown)


critical27:
Thx for contribution~ Indeed, RPC and Raft are sharing the same thread pool. But I'am afraid the changes of this PR are too aggressive. Thread model may impact many aspects, including leader change you've encountered, strange deadlock/livelock, and extra context switch may hurt the performance.

If we want to separate raft from normal data thread pool, it could involve many changes (basically every path of RaftService + RaftPart need to be considered). I believe it would work and you could use it in your env. But adding a thread  pool merely in NebulaStore is not enough, this PR need to be hold for a while. If you are interested in details, could contact me later.

shixiangz:
Can you tell me about the problems that will arise after the modification? I'm worried that unexpected problems may arise

critical27:
> Can you tell me about the problems that will arise after the modification? I'm worried that unexpected problems may arise

If we modified in this way:
* Normal RPC and Raft RPC server still use the old thread pool
* RaftPart's write path, which will write data into RocksDB, will be accomplished in the new thread pool

However, the Raft RPC (which is RaftService) provides at least three interfaces: append logs, leader election and snapshot. For example, snapshot may trigger write stall as well, so the problem is not addressed totally I guess. Another concern is that, we have encounter lots of deadlock related to thread pool. (nebula's thread pool usage is quite subtle, I admit that it is quite error-prone). As for this PR, I think it will work well limited to most of normal read/write between graph/storage and the problem posted in your issue. But there may be some other operations that may trigger read/write, e.g., data balance and raft snapshot. I can't promise it will work, it needs quite a log test to verify it in our internal team.

PR #5381, opened 2023-03-03
Remove virtual function in RowReader

critical27:
## What type of PR is this?
- [ ] bug
- [ ] feature
- [X] enhancement

#### Description:
Since we don't support data of 1.x long ago, we don't need the virtual interface of `RowReader` anymore. So I removed `RowReader` and `RowReaderV1`.

`RowReaderWrapper` now only contains a pointer of `RowReaderV2`. No more virtual function called.

After this PR, the virtual interface of `SchemaProviderIf` could be removed as well, I'll handle it later

@HarrisChu, this PR may or may not improve the performance, could pay some attention to it later.

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....


cangfengzhs:
Since you modified it, why is it still called RowReaderV2? V2 can be deleted

critical27:
> Since you modified it, why is it still called RowReaderV2? V2 can be deleted

For now, just keep the old naming, there is also a `RowWriterV2`...
All code use `RowReaderWrapper`, we could remove it and rename `RowReaderV2` to `RowReader` together if necessary.

pengweisong (approved):
Good job

PR #4804, opened 2022-10-28
UDF based on existing function infra

zhaojunnana:
## What type of PR is this?
- [ ] bug
- [x] feature
- [ ] enhancement

#### Issue(s) number: 
https://github.com/vesoft-inc/nebula/discussions/4793
Close #5337

#### Description:
Users can define functions by implementing GraphFunction

## Checklist:
Tests:
- [ ] Unit test(positive and negative cases)
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:
- [x] Documentation affected (Please add the label if documentation needs to be modified.)
    see https://github.com/vesoft-inc/nebula/pull/4804/#issuecomment-1427318108
- [ ] Incompatibility (If it breaks the compatibility, please describe it and add the label.）
- [ ] If it's needed to cherry-pick (If cherry-pick to some branches is required, please label the destination version(s).)
- [ ] Performance impacted: Consumes more CPU/Memory

## Release notes:

Please confirm whether to be reflected in release notes and how to describe:
> ex. Fixed the bug .....

PR #5391, opened 2023-03-08
feat: optimized udf build

wey-gu:
The indent character was wrong, now it's fixed and tested.

## What type of PR is this?
- [x] bug
- [x] feature
- [ ] enhancement

## desc

- Fixed the indentation character from white space to tab character
- now, we could do `make UDF=foo` to build when we have `foo.cpp` and `foo.h` and it'll output `foo.so`

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [x] N/A

Affects:

## Release notes:

n/a


xtcyclist (approved):

Shinji-IkariG (approved):

abby-cyber:
ref: https://github.com/vesoft-inc/nebula/pull/4804

Issue #5392, opened 2023-03-09
Insert userIpSessionCount to local cache failed


Aiee:
**Please check the FAQ documentation before raising an issue**

**Describe the bug (__required__)**

`failed: Auth failed: Insert userIpSessionCount to local cache failed.`
Unstable to reproduce

Ref: https://discuss.nebula-graph.com.cn/t/topic/12196/5

**Your Environments (__required__)**

NebulaGraph v3.4

**How To Reproduce(__required__)**

**Expected behavior**

**Additional context**

PR #5393, opened 2023-03-09
Fix `addSessionCount()` in multu-thread case

Aiee:
## What type of PR is this?
- [x] bug
- [ ] feature
- [ ] enhancement

#### Issue(s) number: 
Close https://github.com/vesoft-inc/nebula/issues/5392

#### Description:
When multiple threads try to increment session counts at the same time, there is a chance that `emplace()` will return false. To avoid this issue, it's recommended to use insert_or_assign instead of emplace because the check for session counts doesn't need to be too strict.

## How do you solve it?
Use `insert_or_assign` instead of `emplace`.

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....


yixinglu (commented):
LGTM

yixinglu (approved):

xtcyclist (approved):

PR #5410, opened 2023-03-17
Remove virtual interface in SchemaProviderIf

critical27:
## What type of PR is this?
- [ ] bug
- [ ] feature
- [X] enhancement

#### Description:
Follow up of #5381, since `RowReader` has been removed, we could remove the virtual interface `SchemaProviderIf` further more. It has two derived class, `NebulaSchemaProvider` is the one we used, `ResultSchemaProvider` is the one that compatible with `RowReaderV1`, which is not used anymore. 

## How do you solve it?

In this PR, all schema is replaced with `NebulaSchemaProvider`, the base `SchemaProviderIf` and derived `ResultSchemaProvider` along with outdated test, `RowReaderV1` are removed.

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....


src/common/meta/NebulaSchemaProvider.h
```diff
@@ -89,34 +91,91 @@ class NebulaSchemaProvider : public SchemaProviderIf {
...
+        index_++;
+        field_ = schema_->field(index_);
+      }
+      return *this;
+    }
+
+    Iterator& operator+(uint16_t steps) {
+      if (field_) {
+        index_ += steps;
+        field_ = schema_->field(index_);
```
pengweisong, line 117: if index_ out of range, operator*() will raise exception. Is this expected?

pengweisong (approved):
Good job

cangfengzhs (approved):

PR #5422, opened 2023-03-21
Fix ExtractFilterExprVisitor.

xtcyclist:
## What type of PR is this?
- [X] bug
- [ ] feature
- [ ] enhancement

#### Issue(s) number: 
Fix https://github.com/vesoft-inc/nebula/issues/5423

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [X] TCK

Affects:

## Release notes:

> ex. Fixed the bug .....

### DIFF
--- a/src/graph/visitor/ExtractFilterExprVisitor.cpp
+++ b/src/graph/visitor/ExtractFilterExprVisitor.cpp
@@ -240,12 +240,26 @@ void ExtractFilterExprVisitor::ExtractRemainExpr(LogicalExpression *expr,
     extractedExpr_ = operands[0];
   }
 
+  LogicalExpression *remainedExpr = nullptr;
   if (remainedOperands.size() > 1) {
-    auto remainedExpr = LogicalExpression::makeAnd(pool_);
+    remainedExpr = LogicalExpression::makeAnd(pool_);
     remainedExpr->setOperands(std::move(remainedOperands));
-    remainedExpr_ = std::move(remainedExpr);
   } else {
-    remainedExpr_ = std::move(remainedOperands[0]);
+    remainedExpr = static_cast<LogicalExpression *>(std::move(remainedOperands[0]));
+  }
+  if (remainedExpr_ != nullptr) {
+    if (remainedExprFromAnd_) {
+      auto mergeExpr = LogicalExpression::makeAnd(pool_, remainedExpr_, std::move(remainedExpr));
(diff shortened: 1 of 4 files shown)


src/graph/visitor/ExtractFilterExprVisitor.cpp
```diff
@@ -240,12 +240,26 @@ void ExtractFilterExprVisitor::ExtractRemainExpr(LogicalExpression *expr,
...
-    auto remainedExpr = LogicalExpression::makeAnd(pool_);
+    remainedExpr = LogicalExpression::makeAnd(pool_);
     remainedExpr->setOperands(std::move(remainedOperands));
-    remainedExpr_ = std::move(remainedExpr);
   } else {
-    remainedExpr_ = std::move(remainedOperands[0]);
+    remainedExpr = static_cast<LogicalExpression *>(std::move(remainedOperands[0]));
+  }
+  if (remainedExpr_ != nullptr) {
+    if (remainedExprFromAnd_) {
```
nevermore3, line 251: a little confused, when remainedExpr_ is not equal to null

xtcyclist:
```
      MATCH (v0)<-[e0]-()<-[e1]-(v1)
      WHERE (id(v0) == 6)
      AND ((NOT (NOT ((v1.Label_6.Label_6_5_Int > (e1.Rel_0_5_Int - 0.300177))
      AND v1.Label_6.Label_6_1_Bool))))
      RETURN count(*)
```

In this case (added in the tck), for example, there are three conditions in total, forming a tree of expressions, where all index nodes are `AND` logics.

`v1.Label_6.Label_6_5_Int > (e1.Rel_0_5_Int - 0.300177)` can not be pushed down into AppendVertices. It is a leaf node in the expression tree. And, it is an `remainedExpr_`, while visiting its direct father node.

`id(v0) == 6` also cannot be pushed down into AppendVertices. It is an `remainedExpr_` either.

Without this bug fix, `v1.Label_6.Label_6_5_Int > (e1.Rel_0_5_Int - 0.300177)` as an `remainedExpr_` was thrown away, leaving only `id(v0) == 6` as an `remainedExpr_`, because it is visited later.

nevermore3 (approved):

Shylock-Hg (approved):

PR #5430, opened 2023-03-23
support ttl in ms

### LINKED ISSUE #5376: TTL in millisecond

critical27:
## What type of PR is this?
- [ ] bug
- [ ] feature
- [X] enhancement

#### Issue(s) number: 
Close #5376

#### Description:
Add a flag to control which unit that ttl uses, user **should not modify** it after one ttl column is specified. This behavior has been discussed with @MuYiYong. I think the flag should not be exposed as well. For those scenario who want to use ttl in ms, just set the flag to true once and for all before any ttl column is specified.

For example, if `ttl_use_ms` is false by default, set a ttl column as 500, so data of 500 seconds ago are expired. If user change `ttl_use_ms` from false too true, data of 500ms ago are expired. So again, don't modify it unless no more ttl columns are used.

## Checklist:
Tests:
- [X] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....

### DIFF
--- a/src/storage/CommonUtils.cpp
+++ b/src/storage/CommonUtils.cpp
@@ -7,6 +7,10 @@
 
 #include "storage/exec/QueryUtils.h"
 
+DEFINE_bool(ttl_use_ms,
+            false,
+            "whether the ttl is configured as milliseconds instead of default seconds");
+
 namespace nebula {
 namespace storage {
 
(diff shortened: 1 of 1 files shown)


SuperYoko (approved):

pengweisong (approved):

abby-cyber:
https://github.com/vesoft-inc/nebula-docs-cn/pull/2702

Issue #5432, opened 2023-03-23
Graph return incorrect query result on null value filter


nebula-bots:
**Please check the FAQ documentation before raising an issue**

**Describe the bug (__required__)**

Look at the query below:

```txt
(root@nebula) [gdlancer]> MATCH (v0:Label_0)-[e0]->()-[e1*1..1]->(v1) WHERE (id(v0) == 15) AND ((v1.Label_6.Label_6_4_Int == v1.Label_6.Label_6_5_Int)) return labels(v1),  v1.Label_6.Label_6_4_Int, v1.Label_6.Label_6_5_Int, ((v1.Label_6.Label_6_4_Int == v1.Label_6.Label_6_5_Int))
+-------------+--------------------------+--------------------------+------------------------------------------------------+
| labels(v1)  | v1.Label_6.Label_6_4_Int | v1.Label_6.Label_6_5_Int | (v1.Label_6.Label_6_4_Int==v1.Label_6.Label_6_5_Int) |
+-------------+--------------------------+--------------------------+------------------------------------------------------+
| ["Label_2"] | __NULL__                 | __NULL__                 | __NULL__                                             |
| ["Label_2"] | __NULL__                 | __NULL__                 | __NULL__                                             |
| ["Label_2"] | __NULL__                 | __NULL__                 | __NULL__                                             |
+-------------+--------------------------+--------------------------+------------------------------------------------------+
Got 3 rows (time spent 3.693ms/17.584958ms)

Thu, 12 Jan 2023 17:38:19 CST

(root@nebula) [gdlancer]>
```

` ((v1.Label_6.Label_6_4_Int == v1.Label_6.Label_6_5_Int))` evaluate to null but the filter in `WHERE (id(v0) == 15) AND ((v1.Label_6.Label_6_4_Int == v1.Label_6.Label_6_5_Int))` did not work and return three wrong result.

**Your Environments (__required__)**

* OS: `uname -a`
* Compiler: `g++ --version` or `clang++ --version`
* CPU: `lscpu`
* Commit id (e.g. `a3ffc7d8`) https://github.com/vesoft-inc/nebula-ent/commit/261433608ccb385339a610a3877014377defc379

**How To Reproduce(__required__)**

1. Step 1
2. Step 2
3. Step 3

**Expected behavior**

**Additional context**

jinyingsunny:
talk with @xtcyclist 
[...]

PR #5433, opened 2023-03-23
Fix the comparisions involving EMPTY.

xtcyclist:
## What type of PR is this?
- [X] bug
- [ ] feature
- [ ] enhancement

#### Issue(s) number: 
Fix https://github.com/vesoft-inc/nebula/issues/5432

#### Description:
`EMPTY == EMPTY` shall not return true.

## How do you solve it?

```
EMPTY == (x), return value: 
==      EMPTY		NULL		Others
EMPTY	NULL		FALSE		FALSE

EMPTY < (x), return value:
<	EMPTY		NULL		Others
EMPTY	NULL		NULL		NULL
```

Tried to return null in all cases if an EMPTY value is involved in comparisions. But some cases have relied on it to return FALSE. Thus, I applied the above truth table for `==` and `<`.

I think we currently haven't explained any truth table involving EMTPY or EMTPY-like values in our documents. Right? Not sure. 

Simlar to NULL (https://docs.nebula-graph.com.cn/3.4.0/3.ngql-guide/3.data-types/5.null/), we may need to explain the behaviours involving EMTPY or EMTPY-like values.

For example, queries like 
```
MATCH (v:player)-[]-(v2) WHERE v.player.nnnnnnnnnnnnnn == v2.player.nnnnnnnnnnnnnn RETURN *
```
would return nothing, as the property `nnnnnnnnnnnnnn` does not exist (EMPTY) and `EMTPY == EMPTY` = NULL.

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [X] TCK

Affects:

## Release notes:

> ex. Fixed the bug .....

### DIFF
--- a/src/common/datatypes/Value.cpp
+++ b/src/common/datatypes/Value.cpp
@@ -1777,7 +1777,7 @@ Value Value::toSet() const {
   }
 }
 Value Value::lessThan(const Value& v) const {
-  if UNLIKELY(empty() || v.empty()) {
+  if (UNLIKELY(empty() || v.empty())) {
     return Value::kNullValue;
   }
   auto vType = v.type();
@@ -1871,8 +1871,8 @@ Value Value::lessThan(const Value& v) const {
 }
 
 Value Value::equal(const Value& v) const {
-  if UNLIKELY(empty() || v.empty()) {
-    return !empty() || !v.empty() ? false: Value::kNullValue;
+  if (UNLIKELY(empty() || v.empty())) {
+    return !empty() || !v.empty() ? false : Value::kNullValue;
   }
   auto vType = v.type();
   auto hasNull = (type_ | vType) & Value::Type::NULLVALUE;
--- a/src/common/datatypes/Value.cpp
+++ b/src/common/datatypes/Value.cpp
@@ -1777,7 +1777,7 @@ Value Value::toSet() const {
   }
 }
 Value Value::lessThan(const Value& v) const {
-  if (empty() || v.empty()) {
+  if UNLIKELY(empty() || v.empty()) {
     return Value::kNullValue;
   }
   auto vType = v.type();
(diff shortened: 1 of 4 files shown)


src/common/datatypes/Value.cpp
```diff
@@ -1871,8 +1871,12 @@ Value Value::lessThan(const Value& v) const {
...
 Value Value::equal(const Value& v) const {
-  if (empty()) {
-    return v.isNull() ? Value::kNullValue : v.empty();
+  if (empty() || v.empty()) {
+    if (!empty() || !v.empty()) {
+      return false;
+    } else {
+      return Value::kNullValue;
+    }
   }
```
jievince, line 1880: This writing style is a bit strange. Could you replace it with the following?
```
if (empty() && v.empty()) return Value::kNullValue;
if (empty() || v.empty()) return false;
```

xtcyclist: That would leave two extra conditions for normal non-empty cases. Changed to another form instead.

src/graph/executor/logic/LoopExecutor.cpp
```diff
@@ -20,6 +20,9 @@ folly::Future<Status> LoopExecutor::execute() {
   QueryExpressionContext ctx(ectx_);
 
   auto value = expr->eval(ctx);
+  if (value.isNull()) {
+    value = Value(true);
```
yixinglu, line 24: it seems that u should assign false to the value. maybe you need not to check the result of value, just delete the following DCHECK.

xtcyclist: Removed.

xtcyclist: It won't pass the tck.

jievince (approved):

jievince (approved):

yixinglu (approved):

abby-cyber:
https://github.com/vesoft-inc/nebula-docs-cn/pull/2716

PR #5437, opened 2023-03-24
Generate an plan base on the output

nevermore3:
## What type of PR is this?
- [ ] bug
- [ ] feature
- [ ] enhancement

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....

### DIFF
--- a/src/graph/validator/GoValidator.cpp
+++ b/src/graph/validator/GoValidator.cpp
@@ -69,21 +69,8 @@ Status GoValidator::validateWhere(WhereClause* where) {
   }
 
   auto expr = where->filter();
-  auto filterExprs = ExpressionUtils::collectAll(
-      expr, {Expression::Kind::kVertex, Expression::Kind::kDstProperty});
-  for (auto filterExpr : filterExprs) {
-    if (filterExpr->kind() == Expression::Kind::kDstProperty) {
-      goCtx_->joinDst = true;
-    } else {
-      auto& name = static_cast<const VertexExpression*>(filterExpr)->name();
-      if (name == "VERTEX") {
-        return Status::SemanticError("`%s' is not support in go sentence.",
-                                     filterExpr->toString().c_str());
-      }
-      if (name == "$$") {
-        goCtx_->joinDst = true;
-      }
-    }
+  if (!checkDstPropOrVertexExist(expr)) {
+    return Status::SemanticError("`%s' is not support in go sentence.", expr->toString().c_str());
   }
 
   where->setFilter(ExpressionUtils::rewriteLabelAttr2EdgeProp(expr));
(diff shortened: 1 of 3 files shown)


yixinglu (dismissed):

src/graph/validator/GoValidator.cpp
```diff
@@ -69,6 +69,23 @@ Status GoValidator::validateWhere(WhereClause* where) {
   }
 
   auto expr = where->filter();
+  auto filterExprs = ExpressionUtils::collectAll(
```
czpmango, line 72: Better to add a expression util function to avoid repeating the same code logic elsewhere, maybe another xxxValidator?

src/graph/validator/GoValidator.cpp
```diff
@@ -69,6 +69,23 @@ Status GoValidator::validateWhere(WhereClause* where) {
...
+      expr, {Expression::Kind::kVertex, Expression::Kind::kDstProperty});
+  for (auto filterExpr : filterExprs) {
+    if (filterExpr->kind() == Expression::Kind::kDstProperty) {
+      goCtx_->joinDst = true;
+    } else {
+      auto& name = static_cast<const VertexExpression*>(filterExpr)->name();
+      if (name == "VERTEX") {
```
czpmango, line 79: Just return false if you dont care this scenario.

src/graph/validator/GoValidator.cpp
```diff
@@ -69,6 +69,23 @@ Status GoValidator::validateWhere(WhereClause* where) {
...
+        return Status::SemanticError("`%s' is not support in go sentence.",
+                                     filterExpr->toString().c_str());
+      }
+      if (name == "$$") {
```
czpmango, line 83: How about `properties($$).age` case?

src/graph/validator/GoValidator.cpp
czpmango: This is code logic you should check elsewhere than here, IMHO.

src/graph/validator/GoValidator.cpp
nevermore3: fixed

src/graph/validator/GoValidator.cpp
nevermore3: should report error if user input
`go from ... yield vertex as v`

src/graph/validator/GoValidator.cpp
nevermore3: what is checked is the extracted expression

src/graph/validator/GoValidator.cpp
czpmango: Didn't the previous implementation check that, I mean you shouldn't mess with the logic of checking if there is an endpoint.

src/graph/validator/GoValidator.cpp
czpmango: Please add the test case about this.

src/graph/validator/GoValidator.cpp
nevermore3: I think checking whether there is information about the destination vertex and whether to joinDst is logically consistent and does not need to be implemented separately

src/graph/validator/GoValidator.cpp
nevermore3: $$ in properties($$).age is a vertexExpression

src/graph/validator/GoValidator.cpp
nevermore3: got what's u mean

src/graph/validator/GoValidator.cpp
czpmango: Ok, I missed that.

src/graph/validator/GoValidator.cpp
```diff
@@ -261,5 +263,24 @@ Status GoValidator::buildColumns() {
...
+      if (name == "VERTEX") {
+        return false;
+      }
+      if (name == "$$") {
+        goCtx_->joinDst = true;
+      }
+    }
+  }
+  return true;
+}
```
czpmango, line 283: Maybe this is better:
```
bool ExpressionUtils::existAnyDstVertex(const Expression* expr) {
    auto finder = [](const Expression* expr) -> bool {
        return (e->kind() == Expression::Kind::kDstProperty) ||
               (e->kind == ExprKind::kVertex &&
                static_cast<const VertexExpression*>(filterExpr)->name() == "$$");
    }
    FindVisitor visitor(finder);
    const_cast<Expression*>(self)->accept(&visitor);
    return !visitor.results().empty();
}
```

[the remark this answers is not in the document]
czpmango: goCtx_->joinDst = ExpressionUtils::existAnyDstVertex(filter);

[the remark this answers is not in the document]
nevermore3: although the functions are separated, it also causes repeated verification

codesigner (approved):

Issue #5329, opened 2023-02-09
does nebula support outputting the attributes of a specified tag of a vertex?


may11544:
nebula version:2022.11.15-nightly

A simple example.I want to get all attributes with the tag name of default on vertex with the vid of 10.2.29.88.
Think according to inertia,I will write ngql statements like this

MATCH (v) WHERE id(v) IN ["10.2.29.88"] RETURN properties(v.default) 
or 
MATCH (v) WHERE id(v) IN ["10.2.29.88"] RETURN properties(v:default) 

But the above two statements will report errors

It is said that nebula already supports the writing method of <vertex >.<tag>

What I want to know is that there is something wrong with my writing method or the current version does not support

xtcyclist:
Bug:
```
(root@nebula) [nba]> match (v) where id(v) == "Tim Duncan" return v.player
+----------------------+
| v.player |
+----------------------+
| {}                   |
+----------------------+
Got 1 rows (time spent 2982/3311 us)
```

Expected:
```
Fri, 10 Feb 2023 10:31:45 CST

(root@nebula) [nba]> match (v) where id(v) == "Tim Duncan" return properties(v)
+---------------------------------------------------------+
| properties(v)                                           |
+---------------------------------------------------------+
| {age: 42, name: "Tim Duncan", speciality: "psychology"} |
+---------------------------------------------------------+
Got 1 rows (time spent 2943/3291 us)

Fri, 10 Feb 2023 10:31:48 CST

(root@nebula) [nba]> match (v) where id(v) == "Tim Duncan" return v.player
+-------------------------------------------------------------------------------------------------------------+
| v.player                                                                                                           |
+-------------------------------------------------------------------------------------------------------------+
| {age: 42, name: "Tim Duncan"} |
+-------------------------------------------------------------------------------------------------------------+
Got 1 rows (time spent 2863/3245 us)
```

Milittle:
> nebula version:2022.11.15-nightly
> 
> A simple example.I want to get all attributes with the tag name of default on vertex with the vid of 10.2.29.88. Think according to inertia,I will write ngql statements like this
> 
> MATCH (v) WHERE id(v) IN ["10.2.29.88"] RETURN properties(v.default) or MATCH (v) WHERE id(v) IN ["10.2.29.88"] RETURN properties(v:default)
> 
> But the above two statements will report errors
> 
> It is said that nebula already supports the writing method of .
> 
[...]

PR #5440, opened 2023-03-27
Allow the expressions like 'v.tag' in return

xtcyclist:
## What type of PR is this?
- [X] bug
- [ ] feature
- [ ] enhancement

#### Issue(s) number: 
Fix https://github.com/vesoft-inc/nebula/issues/5329

#### Description:
'v.tag' means all properties of that tag are needed. They shall not be pruned by the optimizer.

## How do you solve it?
Add a wildcard `"*"` into the VertexProp map in the tracker when encountered expressions like `v.tag` after the `tag` has been verified to be a real tag.

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [X] TCK

Affects:

## Release notes:

> ex. Fixed the bug .....

### DIFF
--- a/src/graph/visitor/PropertyTrackerVisitor.cpp
+++ b/src/graph/visitor/PropertyTrackerVisitor.cpp
@@ -218,7 +218,7 @@ void PropertyTrackerVisitor::visit(AttributeExpression *expr) {
       auto ret = qctx_->schemaMng()->toTagID(space_, tagName);
       if (!ret.ok()) {
         status_ = std::move(ret).status();
-        return;
+        break;
       }
       auto tagId = ret.value();
       propsUsed_.insertVertexProp(entityAlias, tagId, "*");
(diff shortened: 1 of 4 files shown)


src/graph/visitor/PropertyTrackerVisitor.cpp
```diff
@@ -208,10 +208,19 @@ void PropertyTrackerVisitor::visit(AttributeExpression *expr) {
...
+      auto &entityAlias = varPropExpr->prop();
+      propsUsed_.insertEdgeProp(entityAlias, unknownType_, propName);
+      // maybe: $v.tag
+      auto ret = qctx_->schemaMng()->toTagID(space_, propName);
+      if (!ret.ok()) {
+        status_ = std::move(ret).status();
+        break;
+      }
+      auto tagId = ret.value();
+      propsUsed_.insertVertexProp(entityAlias, tagId, "*");
```
yixinglu, line 223: Is it necessary to insert edge and vertex props at same time? maybe only need to handle one branch by checking where the prop is from?

xtcyclist: It goes into `Kind::kSubscript` as `$-.e[0]`. Maybe we can delete it already.

nevermore3 (approved):

Aiee (approved):

src/graph/visitor/PropertyTrackerVisitor.cpp
xtcyclist: Seems like we don't have sufficient information here to tell them apart. Any advise?

yixinglu (approved):

Issue #5187, opened 2023-01-03
Wrong comparisons of paths involving self-reflective edges


xtcyclist:
Wrong result:
```
(root@nebula) [test]> match p = (v:Label_5)-[e:Rel_0]->(v1:Label_1), p2 = (v)-[e1:Rel_0]-(v1) where id(v) == 47 and p != p2 and e == e1 and v == v and v1 == v1 return count(*)
+----------+
| count(*) |
+----------+
| 1        |
+----------+
```

Expected:
```
(root@nebula) [test]> match p = (v:Label_5)-[e:Rel_0]->(v1:Label_1), p2 = (v)-[e1:Rel_0]-(v1) where id(v) == 47 and p != p2 and e == e1 and v == v and v1 == v1 return count(*)
+----------+
| count(*) |
+----------+
| 0        |
+----------+
```

The first patten p asks for directed edges for e. The second one asks for edges of both directions for e1. They matched the same identical self-reflective edge. But the paths are considered not to be equal.

```
+---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| e                                                                                                                                                                                                                                                                                                                                                                                                                                                                                     | e1                                                                                                                                                                                                                                                                                                                                                                                                                                                                                    |
+---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
[...]

PR #5444, opened 2023-03-28
Fix the == operator of Path

xtcyclist:
## What type of PR is this?
- [X] bug
- [ ] feature
- [ ] enhancement

#### Issue(s) number: 
Fix https://github.com/vesoft-inc/nebula/issues/5187

#### Description:
The same self-reflective step with +/- edge types were treated as different steps.

## How do you solve it?
Consider them as equal.

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [X] TCK

Affects:

## Release notes:

> ex. Fixed the bug .....

### DIFF
--- a/src/common/datatypes/Path.h
+++ b/src/common/datatypes/Path.h
@@ -87,8 +87,8 @@ struct Step {
     if (dst != rhs.dst) {
       return dst < rhs.dst;
     }
-    if (type != rhs.type) {
-      return type < rhs.type;
+    if (type != rhs.type && type != -rhs.type) {
+      return abs(type) < abs(rhs.type);
     }
     if (ranking != rhs.ranking) {
       return ranking < rhs.ranking;
--- a/src/common/datatypes/Path.h
+++ b/src/common/datatypes/Path.h
@@ -79,8 +79,8 @@ struct Step {
   }
 
   bool operator==(const Step& rhs) const {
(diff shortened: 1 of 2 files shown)


src/common/datatypes/Path.h
```diff
@@ -79,8 +79,8 @@ struct Step {
   }
 
   bool operator==(const Step& rhs) const {
-    return dst == rhs.dst && type == rhs.type && name == rhs.name && ranking == rhs.ranking &&
-           props == rhs.props;
+    return dst == rhs.dst && (type == rhs.type || type == -rhs.type) && name == rhs.name &&
```
Shylock-Hg, line 82: Need to fix `operator<` too.

Shylock-Hg, line 82: Should we compare `dst == src` when type is negative?

xtcyclist: I think it's not needed. This is comparing two `STEPS`, not `EDGES`, so only the `dst` need to be compared. There is also no `src` member in STEP.

This diff avoids treating two physical versions of the same self-reflective edges as different.

xtcyclist: OK. Fixed.

Shylock-Hg (approved):

yixinglu (approved):

Issue #5438, opened 2023-03-25
[TTL] Historical data may never get the opportunity to get garbage collected


luyade:
**Please check the FAQ documentation before raising an issue**

**Describe the bug (__required__)**

By default, custom_filter_interval_secs is set as 24 * 3600, which means in 24 hours, we will only have one chance to do custom minor compaction. Otherwise, it will go to the default minor compaction.

<img width="736" alt="企业微信截图_107a2bbb-a440-4aa5-9634-7def95070245" src="https://user-images.githubusercontent.com/5063363/227717374-8046dc9d-c668-4e47-92ae-41485eb1a3e9.png">

For historical data, many data will reside in the bottommost level. For the expired data in the bottommost level, it doesn't have many chance to get GC except periodic compaction. However, during daily running, the only custom-compaction chance will most possibly be used by upper level data compaction, such as level0 => level1. So the default 30-days periodic compaction will go through the default minor compaction, without go through custom compaction filter. So the expired data will always be there.

Here is some log I print out in StorageIterator.h:

<img width="817" alt="企业微信截图_12cc53a8-4ba2-473a-8f10-f4438c027c9d" src="https://user-images.githubusercontent.com/5063363/227718071-3dca95f9-9bb8-4979-a257-ef8b04f50349.png">

As you can see, it will read a bunch of expired edges during edge traverse. 

After I fixed the compaction logic, the performance got extremely better.

<img width="1420" alt="企业微信截图_13ab7c92-681c-4d23-aa92-e242a316a553" src="https://user-images.githubusercontent.com/5063363/227718269-a111c27d-914f-42d9-9391-c4810cc8bf15.png">

**Your Environments (__required__)**

* OS: `uname -a`
* Compiler: `g++ --version` or `clang++ --version`
* CPU: `lscpu`
* Commit id (e.g. `a3ffc7d8`)

**How To Reproduce(__required__)**

1. Step 1
2. Step 2
3. Step 3

**Expected behavior**

**Additional context**

pengweisong:
In the design now, it indeed depends on user to trigger the compact job to GC the garbage in the bottom level.

luyade:
> In the design now, it indeed depends on user to trigger the compact job to GC the garbage in the bottom level.

As far as I can see, in the current design, the only way user can trigger the compaction of the bottommost level is to submit a compact job, which will then trigger a full compaction. Actually, full compaction is almost unacceptable in production environment.

pengweisong:
[...]

PR #5447, opened 2023-03-28
Change the compaction filter logic to let periodic compaction go through custom compaction filter, to gc expired data

luyade:
Fix the issue that, many expired data in the bottommost level don't get garbage collected

## What type of PR is this?
- [ ] bug
- [ ] feature
- [ ] enhancement

#### Issue(s) number: 
https://github.com/vesoft-inc/nebula/issues/5438
## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....

### DIFF
--- a/src/kvstore/CompactionFilter.h
+++ b/src/kvstore/CompactionFilter.h
@@ -34,7 +34,7 @@ class KVCompactionFilter final : public rocksdb::CompactionFilter {
   /**
    * @brief whether remove the key during compaction
    *
-   * @param level Levels of key in rocksdb, not used for now
+   * @param level Levels of key in rocksdb
    * @param key Rocksdb key
    * @param val Rocksdb val
    * @return true Key will not be removed
@@ -45,8 +45,8 @@ class KVCompactionFilter final : public rocksdb::CompactionFilter {
               const rocksdb::Slice& val,
               std::string*,
               bool*) const override {
-    UNUSED(level);
-    return kvFilter_->filter(spaceId_,
+    return kvFilter_->filter(level,
+                             spaceId_,
                              folly::StringPiece(key.data(), key.size()),
                              folly::StringPiece(val.data(), val.size()));
   }
--- a/src/kvstore/CompactionFilter.h
+++ b/src/kvstore/CompactionFilter.h
@@ -80,18 +80,12 @@ class KVCompactionFilterFactory : public rocksdb::CompactionFilterFactory {
     auto now = time::WallClock::fastNowInSec();
     if (context.is_full_compaction || context.is_manual_compaction) {
       LOG(INFO) << "Do full/manual compaction!";
-      lastRunCustomFilterTimeSec_ = now;
-      return std::make_unique<KVCompactionFilter>(spaceId_, createKVFilter());
     } else {
-      if (FLAGS_custom_filter_interval_secs >= 0 &&
-          now - lastRunCustomFilterTimeSec_ > FLAGS_custom_filter_interval_secs) {
-        LOG(INFO) << "Do custom minor compaction!";
-        lastRunCustomFilterTimeSec_ = now;
-        return std::make_unique<KVCompactionFilter>(spaceId_, createKVFilter());
-      }
-      LOG(INFO) << "Do default minor compaction!";
-      return std::unique_ptr<rocksdb::CompactionFilter>(nullptr);
+      // No worry, by default flush will not go through the custom compaction filter.
+      // See CompactionFilterFactory::ShouldFilterTableFileCreation.
+      LOG(INFO) << "Do automatic or periodic compaction!";
     }
+    return std::make_unique<KVCompactionFilter>(spaceId_, createKVFilter());
   }
 
   const char* Name() const override {
--- a/src/kvstore/CompactionFilter.h
+++ b/src/kvstore/CompactionFilter.h
@@ -95,7 +95,6 @@ class KVCompactionFilterFactory : public rocksdb::CompactionFilterFactory {
 
  private:
   GraphSpaceID spaceId_;
-  int32_t lastRunCustomFilterTimeSec_ = 0;
 };
 
 /**
--- a/src/kvstore/CompactionFilter.h
+++ b/src/kvstore/CompactionFilter.h
@@ -77,7 +77,6 @@ class KVCompactionFilterFactory : public rocksdb::CompactionFilterFactory {
    */
   std::unique_ptr<rocksdb::CompactionFilter> CreateCompactionFilter(
       const rocksdb::CompactionFilter::Context& context) override {
-    auto now = time::WallClock::fastNowInSec();
     if (context.is_full_compaction || context.is_manual_compaction) {
       LOG(INFO) << "Do full/manual compaction!";
     } else {
--- a/resources/gflags.json
+++ b/resources/gflags.json
@@ -7,7 +7,6 @@
         "clean_wal_interval_secs",
         "wal_ttl",
         "clean_wal_interval_secs",
-        "custom_filter_interval_secs",
         "accept_partial_success",
         "system_memory_high_watermark_ratio",
         "num_rows_to_check_memory",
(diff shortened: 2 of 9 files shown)


src/kvstore/CompactionFilter.h
```diff
@@ -80,18 +80,12 @@ class KVCompactionFilterFactory : public rocksdb::CompactionFilterFactory {
...
-        lastRunCustomFilterTimeSec_ = now;
-        return std::make_unique<KVCompactionFilter>(spaceId_, createKVFilter());
-      }
-      LOG(INFO) << "Do default minor compaction!";
-      return std::unique_ptr<rocksdb::CompactionFilter>(nullptr);
+      // No worry, by default flush will not go through the custom compaction filter.
+      // See CompactionFilterFactory::ShouldFilterTableFileCreation.
+      LOG(INFO) << "Do automatic or periodic compaction!";
     }
+    return std::make_unique<KVCompactionFilter>(spaceId_, createKVFilter());
```
critical27, line 88: The modification will make every compaction go through `CompactionFilter`, it will hurt performance a lot. As your issue, I suggest you modify the options `periodic_compaction_seconds` in RocksDB. 

See `kBottommostFiles` in RocksDB

critical27:
> However, during daily running, the only custom-compaction chance will most possibly be used by upper level data compaction, such as level0 => level1. So the default 30-days periodic compaction will go through the default minor compaction, without go through custom compaction filter. So the expired data will always be there.

I do agree what you said here, this is an issue. Maybe we could judge the input level and output level. For example, only >= L4, we use the custom compaction filter?

src/kvstore/CompactionFilter.h
luyade: > The modification will make every compaction go through `CompactionFilter`, it will hurt performance a lot. As your issue, I suggest you modify the options `periodic_compaction_seconds` in RocksDB.
> 
> See `kBottommostFiles` or `kPeriodicCompaction` in RocksDB, I don't recall it precisely

I don't think this is only my issue, but it should be a common issue as long as user has set ttl. 

"The modification will make every compaction go through CompactionFilter, it will hurt performance a lot"

> Do we have performance testing for such cases? Personally I doubt this is the fact according to my experience in the production environment. 

And also please trust me, changing periodic_compaction_seconds won't help at all. I tested this a lot of times.

luyade:
> > However, during daily running, the only custom-compaction chance will most possibly be used by upper level data compaction, such as level0 => level1. So the default 30-days periodic compaction will go through the default minor compaction, without go through custom compaction filter. So the expired data will always be there.
> 
> I do agree what you said here, this is an issue. In theory, we could judge the input level and output level. For example, only >= L4, we use the custom compaction filter? But I didn't find way to do it in `CompactionFilterFactory` for now.

Yes, we don't have a way to do compaction for only >=L4. 

For now, according to my experience, as time proceeds, the space data get more and more larger with lots of expired data, which makes the performance very bad and also higher risk for single rocksdb instance with large data.

src/kvstore/CompactionFilter.h
critical27: > Do we have performance testing for such cases? Personally I doubt this is the fact according to my experience in the production environment.

If you do so some profiling, you will find that custom compaction filter is way more slow that default one. Because for each key, it will check the schema is valid or not, parse the value and check ttl. Validation of schema is the bottleneck.

critical27: > changing periodic_compaction_seconds won't help at all.

How do you set it? And why that rocksdb didn't pick the outdate sst file

luyade: > > changing periodic_compaction_seconds won't help at all.
> 
> How do you set it? And why rocksdb didn't pick the outdate sst file at bottom level, the score is not big enough or what?
> 
> IIRC, if your file has been `marked_for_compaction`, it definitely will be picked. The question is that have you marked them

Yes, I changed the periodic_compaction_seconds and DO noticed that the bottom level files had all gone through compaction. But it won't help at all because it is just doing default compaction, without going through custom compaction filter. The default compaction doesn't have the logic to validate if the data is expired, so as you can see,  the data will still be there.

critical27: If the bottom sst are picked, I agree to remove `custom_filter_interval_secs`, but we need to address the performance issue as well.

Here is the way to leverage performance:
1. StorageCompactionFilter have constructor of SchemaManager and IndexManager
2. Whenever we check a key/value could pass the filter, we check whether the schema are valid. So for each key, we will check it like codes below
```
    auto tagId = NebulaKeyUtils::getTagId(vIdLen_, key);
    auto schema = schemaMan_->getTagSchema(spaceId, tagId);
    if (!schema) {
      VLOG(3) << "Space " << spaceId << ", Tag " << tagId << " invalid";
      return false;
    }
```
edges are similar.
3. What we need to do is that, when we construct StorageCompactionFilter, get every schema once and for all (by calling `getAllVerTagSchema` and `getAllVerEdgeSchema` in SchemaManager). When we valid schema later, we only need to check it locally instead of check it in SchemaManager.

critical27: See the latest comment, there is a simpler way to address the issue.

luyade:
> Actually, there indeed is a level, see the code in `KVCompactionFilter`
> 
> ```
>   bool Filter(int level,
>               const rocksdb::Slice& key,
>               const rocksdb::Slice& val,
>               std::string*,
>               bool*) const override {
>     UNUSED(level);
>     return kvFilter_->filter(spaceId_,
>                              folly::StringPiece(key.data(), key.size()),
>                              folly::StringPiece(val.data(), val.size()));
>   }
> ```
> 
> We can modify this way

updated

src/kvstore/CompactionFilter.h
```diff
@@ -80,18 +80,12 @@ class KVCompactionFilterFactory : public rocksdb::CompactionFilterFactory {
...
     } else {
-      if (FLAGS_custom_filter_interval_secs >= 0 &&
-          now - lastRunCustomFilterTimeSec_ > FLAGS_custom_filter_interval_secs) {
-        LOG(INFO) << "Do custom minor compaction!";
-        lastRunCustomFilterTimeSec_ = now;
-        return std::make_unique<KVCompactionFilter>(spaceId_, createKVFilter());
-      }
-      LOG(INFO) << "Do default minor compaction!";
-      return std::unique_ptr<rocksdb::CompactionFilter>(nullptr);
+      // No worry, by default flush will not go through the custom compaction filter.
```
pengweisong, line 84: 1. Maybe we should check `level` here. 
2. Besides, I think when level < 4, we could also do the custom compaction now and then, with low frequency.

luyade: > 1. Maybe we should check `level` here.
> 2. Besides, I think when level < 4, we should also do the custom compaction now and then, with low frequency.

But it is not possible to get the level here.

pengweisong: But in your code now, if the level < 4, filter will also be called for each key, which is not a good way.

critical27 (dismissed):
LGTM, thx~

critical27:
CI failed because of compile, please fix it~

luyade:
> CI failed because of compile, please fix it~

Fixed.

critical27:
The test case failed 😢 , probably related to the new gflag `min_level_for_custom_filter`, probably need to set it to 0 in test case.

luyade:
> The test case failed 😢 , probably related to the new gflag `min_level_for_custom_filter`, probably need to set it to 0 in test case.

Yes, I noticed that failure. Just updated again for the test case.

critical27 (approved):
Good job~~~ Thx again

src/kvstore/CompactionFilter.h
luyade: IMO, this is a tradeoff. Although this is not the perfect idea, it should be better to leave the expired data not-really handled. What do you think? If there is better idea, it is appreciated.

pengweisong (approved):

PR #5451, opened 2023-03-29
remove some raft lock for higer performance

cangfengzhs:
## What type of PR is this?
- [ ] bug
- [ ] feature
- [ ] enhancement

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....


critical27:
http://nmeter.vesoft-inc.com/perf/baseline/cmp/1056?label=1

critical27 (approved):
Although it is quite subtle, it is ok to remove the lock for now. @HarrisChu , would youplease help to verify it again before releasing this version

critical27:
Better to take a look at the assembly code whether the instructions are atomic

SuperYoko (approved):

PR #5458, opened 2023-03-30
fix path result without prop

nevermore3:
## What type of PR is this?
- [ ] bug
- [ ] feature
- [x] enhancement

#### Issue(s) number: 
close https://github.com/vesoft-inc/nebula-ent/issues/2530

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....

### DIFF
--- a/src/graph/planner/ngql/PathPlanner.cpp
+++ b/src/graph/planner/ngql/PathPlanner.cpp
@@ -183,9 +183,11 @@ StatusOr<SubPlan> PathPlanner::allPathPlan() {
   auto withProp = pathCtx_->withProp;
   auto* path = AllPaths::make(
       qctx, leftPlan.root, rightPlan.root, pathCtx_->space.id, steps, pathCtx_->noLoop, withProp);
-  auto vertexProp = SchemaUtil::getAllVertexProp(qctx, pathCtx_->space.id, withProp);
-  NG_RETURN_IF_ERROR(vertexProp);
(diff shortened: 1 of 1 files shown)


src/graph/planner/ngql/PathPlanner.cpp
```diff
@@ -183,9 +183,11 @@ StatusOr<SubPlan> PathPlanner::allPathPlan() {
   auto withProp = pathCtx_->withProp;
   auto* path = AllPaths::make(
       qctx, leftPlan.root, rightPlan.root, pathCtx_->space.id, steps, pathCtx_->noLoop, withProp);
-  auto vertexProp = SchemaUtil::getAllVertexProp(qctx, pathCtx_->space.id, withProp);
```
Shylock-Hg, line 186: Any tests?

nevermore3: already exist in allpath.feature

Shylock-Hg: But it doesn't detect these different of results.

nevermore3: the tester has its own comparison function, which is a per-character comparison

yixinglu (approved):
LGTM if tested in TCK!

Issue #5459, opened 2023-03-30
geo type insert illegal data success，but query report error 


nebula-bots:
as title 
```
(root@nebula) [ngdata]>  create tag t1(age int,geo geography)
Execution succeeded (time spent 1324/31109 us)

Tue, 28 Mar 2023 10:53:17 CST

(root@nebula) [ngdata]> insert vertex t1(age,geo) values 1:(30,"POINT(1 1)")
Execution succeeded (time spent 1773/27024 us)

Tue, 28 Mar 2023 10:54:00 CST

(root@nebula) [ngdata]>  fetch prop on t1 1 yield vertex as v
[ERROR (-1005)]: Storage Error: Tag prop not found.

Tue, 28 Mar 2023 10:54:15 CST

(root@nebula) [ngdata]> match (v:t1) return v limit 2
[ERROR (-1005)]: Storage Error: Tag prop not found.

Tue, 28 Mar 2023 10:54:26 CST
```
the error is because the geo value i insert is illegal，in our doc it should be：
```
 INSERT VERTEX any_shape(geo) VALUES "103":(ST_GeogFromText("POLYGON((0 1, 1 2, 2 3, 0 1))")); 
```
![image](https://user-images.githubusercontent.com/106510322/228119964-eafadad4-8f36-4387-a3c9-7b461b73d9a0.png)

after ST_GeogFromText() it process right 

**Your Environments (__required__)**

3.4.0-ent
commitid: 6b14be2

**Expected behavior**

1. please report error when insert 
2. our doc may do some change，because the example seems misleading

PR #5460, opened 2023-03-30
Add the type check when inserting geography value

jievince:
## What type of PR is this?
- [x] bug
- [ ] feature
- [ ] enhancement

#### Issue(s) number: 
Fix https://github.com/vesoft-inc/nebula/issues/5459
#### Description:

When inserting, the geography value is converted to a WKB string format. However, it is not possible to differentiate whether a string being inserted into a property of geography type is in fact a WKB string or just an ordinary string.

So I add an argument `isWKB` to distinguish it:
```
WriteResult RowWriterV2::write(ssize_t index, folly::StringPiece v, bool isWKB)
```
## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....

### DIFF
--- a/src/codec/RowWriterV2.h
+++ b/src/codec/RowWriterV2.h
@@ -254,7 +254,7 @@ class RowWriterV2 {
   WriteResult write(ssize_t index, uint64_t v);
 
   WriteResult write(ssize_t index, const std::string& v);
-  WriteResult write(ssize_t index, folly::StringPiece v);
+  WriteResult write(ssize_t index, folly::StringPiece v, bool isWKB = false);
   WriteResult write(ssize_t index, const char* v);
 
   WriteResult write(ssize_t index, const Date& v);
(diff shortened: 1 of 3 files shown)


src/codec/RowWriterV2.h
```diff
@@ -254,7 +254,7 @@ class RowWriterV2 {
   WriteResult write(ssize_t index, uint64_t v);
 
   WriteResult write(ssize_t index, const std::string& v);
-  WriteResult write(ssize_t index, folly::StringPiece v);
+  WriteResult write(ssize_t index, folly::StringPiece v, bool isWKB = false);
```
Shylock-Hg, line 257: I think you could add a new structure like `struct WKB {folly::StringPiece}`.

jievince: I want to reuse the code of `write(ssize_t index, folly::StringPiece v)`

yixinglu (approved):

xtcyclist (approved):

PR #5470, opened 2023-04-03
Split optimizer rules

czpmango:
## What type of PR is this?
- [ ] bug
- [ ] feature
- [x] enhancement

Split the optimizer rule `PushFilterDownTraverseRule` to `PushFilterThroughAppendVerticesRule` and `PushFilterDownTraverseRule`.

This pr is for refactoring purposes but also introduces some optimizations, such as non-endpoint predicates that can be pushed down before AppendVertices.

The purpose of refactoring is that the previous implementation was too ad-hoc for further optimization, and there were some conflicts between rules, so we should try to avoid similar code implementation from a more common perspective, especially for the development of optimizer.

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

### DIFF
--- a/src/common/expression/PropertyExpression.h
+++ b/src/common/expression/PropertyExpression.h
@@ -32,6 +32,10 @@ class PropertyExpression : public Expression {
  public:
   bool operator==(const Expression& rhs) const override;
 
+  bool isPropertyExpr() const override {
+    return true;
+  }
+
   const Value& eval(ExpressionContext& ctx) override;
 
   const std::string& ref() const {
--- a/src/graph/optimizer/rule/PushFilterThroughAppendVerticesRule.cpp
+++ b/src/graph/optimizer/rule/PushFilterThroughAppendVerticesRule.cpp
@@ -1,0 +1,123 @@
+/* Copyright (c) 2023 vesoft inc. All rights reserved.
+ *
+ * This source code is licensed under Apache 2.0 License.
+ */
+
+#include "graph/optimizer/rule/PushFilterThroughAppendVerticesRule.h"
+
+#include "common/expression/Expression.h"
+#include "graph/optimizer/OptContext.h"
+#include "graph/optimizer/OptGroup.h"
+#include "graph/planner/plan/PlanNode.h"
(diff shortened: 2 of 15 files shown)


yixinglu (dismissed):

src/common/expression/PropertyExpression.h
```diff
@@ -32,6 +32,10 @@ class PropertyExpression : public Expression {
  public:
   bool operator==(const Expression& rhs) const override;
 
+  bool isPropertyExpr() const override {
+    return false;
```
nevermore3, line 36: return true?

jievince (dismissed):

src/common/expression/PropertyExpression.h
czpmango: ACK.

jievince (approved):
LGTM

tests/tck/features/optimizer/CollapseProjectRule.feature
```diff
@@ -6,6 +6,7 @@ Feature: Collapse Project Rule
   Background:
     Given a graph with space named "nba"
 
+  @czp
```
yixinglu, line 9: cleanup

src/graph/optimizer/rule/PushFilterThroughAppendVerticesRule.cpp
```diff
@@ -0,0 +1,123 @@
...
+  auto* oldAvNode = static_cast<graph::AppendVertices*>(oldAvGroupNode->node());
+  auto& dstNode = oldAvNode->nodeAlias();
+
+  auto inputColNames = oldAvNode->inputVars().front()->colNames;
+  auto qctx = octx->qctx();
+
+  auto picker = [&inputColNames, &dstNode](const Expression* expr) -> bool {
+    auto finder = [&inputColNames, &dstNode](const Expression* e) -> bool {
+      if (e->isPropertyExpr() &&
+          std::find(inputColNames.begin(),
```
yixinglu, line 54: format

yixinglu (approved):

PR #5489, opened 2023-04-12
fix ContainsFilter random fail

codesigner:
## What type of PR is this?
- [x] bug
- [ ] feature
- [ ] enhancement

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....

### DIFF
--- a/tests/tck/features/bugfix/ContainsFilter.feature
+++ b/tests/tck/features/bugfix/ContainsFilter.feature
@@ -10,7 +10,7 @@ Feature: contains filter
   Scenario: contains filter
     When executing query:
       """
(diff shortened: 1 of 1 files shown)


Aiee (approved):

czpmango:
Can you explain what is the root cause for this random failure?

codesigner:
> Can you explain what is the root cause for this random failure?

sorted on the column that the result of two rows has the same value, the result depends on the input to the sort operator, seems the input is not alway the same.

PR #5493, opened 2023-04-13
Fixed graphd startup issue

dutor:
When thrift server fails to start, e.g., port conflicts, there are chances that the whole daemon gets stuck in `GraphServer::waitUntilStop` and cannot be stopped via signal.

This PR trivially fixes the issue.


Aiee (approved):
Good job!

Issue #5524, opened 2023-04-23
Prefix performance get very bad after clear data


luyade:
**Introduction**

If we have a large amount of data in the backend rocksdb, the prefix performance will get extremely bad after the data is cleared, such as clear space or clear part before sending snapshot. And also this will impact the startup of storaged (see RocksEngine::allParts() implementation), which will even need several hours to finish the startup process.

**Contents**

Set iterate upper bound for read options of prefix operation. I did the testing, after the change, the performance will get almost normal as usual.

**Related work**

critical27:
You'll set upper bound for each interface?

critical27:
BTW, a recent version of RocksDB has claimed to improve performance after DeleteRange, this is a big pitfalls before. But I haven't test it yet.

PR #5525, opened 2023-04-23
Optimize the prefix performance of RocksDB iterator after delete range

luyade:
## What type of PR is this?
- [ ] bug
- [ ] feature
- [ ] enhancement

#### Issue(s) number: 
https://github.com/vesoft-inc/nebula/issues/5524

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....


src/kvstore/RocksEngine.h
```diff
@@ -70,6 +85,11 @@ class RocksPrefixIter : public KVIterator {
   RocksPrefixIter(std::unique_ptr<rocksdb::Iterator> iter, rocksdb::Slice prefix)
       : iter_(std::move(iter)), prefix_(prefix) {}
 
+  RocksPrefixIter(rocksdb::Slice prefix) : prefix_(prefix) {
```
wey-gu, line 88: `Single-parameter constructors should be marked explicit.  [runtime/explicit] [5]`

[src/kvstore/RocksEngine.h:88](https://github.com/vesoft-inc/nebula/actions/runs/4778193665/jobs/8494456333?pr=5525#step:6:9)

src/kvstore/RocksEngine.h
```diff
@@ -92,9 +117,26 @@ class RocksPrefixIter : public KVIterator {
...
+  void setIterator(std::unique_ptr<rocksdb::Iterator> iter) {
+    iter_ = std::move(iter);
+  }
+
  protected:
-  std::unique_ptr<rocksdb::Iterator> iter_;
+  std::unique_ptr<rocksdb::Iterator> iter_{nullptr};
   rocksdb::Slice prefix_;
+  // temporary key to make sure data_ of iterateUpperBound_ valid during its lifetime
+  std::string iterateUpperBoundKey_;
```
wenhaocs, line 136: Since it's temporary, I prefer keeping its lifetime temp as well, that is, not setting as protected/private variable.

src/kvstore/RocksEngine.h
```diff
@@ -70,6 +90,11 @@ class RocksPrefixIter : public KVIterator {
   RocksPrefixIter(std::unique_ptr<rocksdb::Iterator> iter, rocksdb::Slice prefix)
       : iter_(std::move(iter)), prefix_(prefix) {}
 
+  explicit RocksPrefixIter(rocksdb::Slice prefix) : prefix_(prefix) {
+    iterateUpperBoundKey_ = NebulaKeyUtils::lastKey(prefix_.ToString(), 128);
```
wenhaocs, line 94: Here you are padding fixed 128 EOF. Is there any performance impact when the prefix length is already very long?

wenhaocs (commented):
Good job. Thanks for the contribution.

src/kvstore/RocksEngine.h
luyade: 128 is referred to the existing code in Part::cleanup, as far as I can see, there is no performance degradation.

src/kvstore/RocksEngine.h
luyade: iterateUpperBoundKey_ is not *really* temporary. This string should be valid during the whole lifetime of this iterator. Otherwise, if the string got destroyed before iterator, the iterate_upper_bound of rocksdb::ReadOptions will become invalid, which will then result in crash.

src/kvstore/RocksEngine.h
```diff
@@ -54,10 +59,25 @@ class RocksRangeIter : public KVIterator {
     return folly::StringPiece(iter_->value().data(), iter_->value().size());
   }
 
+  rocksdb::Slice* iterateUpperBound() const {
+    return iterateUpperBound_.get();
+  }
+
+  void setIterator(rocksdb::Iterator* iter) {
```
pengweisong, line 66: just `reset` will be more concision? since the class itself is an Iter, need not repeat it in the function.

src/kvstore/RocksEngine.h
```diff
@@ -54,10 +59,25 @@ class RocksRangeIter : public KVIterator {
...
+  }
+
  private:
-  std::unique_ptr<rocksdb::Iterator> iter_;
+  std::unique_ptr<rocksdb::Iterator> iter_{nullptr};
   rocksdb::Slice start_;
   rocksdb::Slice end_;
+  // Make sure that the lifetime of iterate_upper_bound in rocksdb::ReadOptions
+  // is the same as RocksRangeIter object
+  std::unique_ptr<rocksdb::Slice> iterateUpperBound_;
```
pengweisong, line 80: just `uppperBound_` is ok, same reason as above. Besides, could keep consist type with start and end, that it, just use `rocksdb::Slice`.

src/kvstore/RocksEngine.h
```diff
@@ -54,10 +59,25 @@ class RocksRangeIter : public KVIterator {
     return folly::StringPiece(iter_->value().data(), iter_->value().size());
   }
 
+  rocksdb::Slice* iterateUpperBound() const {
```
pengweisong, line 62: ditto

src/kvstore/RocksEngine.h
```diff
@@ -92,9 +117,26 @@ class RocksPrefixIter : public KVIterator {
...
+    return iterateUpperBound_.get();
+  }
+
+  void setIterator(rocksdb::Iterator* iter) {
```
pengweisong, line 124: ditto

src/kvstore/RocksEngine.h
```diff
@@ -92,9 +117,26 @@ class RocksPrefixIter : public KVIterator {
     return folly::StringPiece(iter_->value().data(), iter_->value().size());
   }
 
+  rocksdb::Slice* iterateUpperBound() const {
```
pengweisong, line 120: ditto

src/kvstore/RocksEngine.h
pengweisong: then the field `iterateUpperBound_` is no needed, you could construct `rocksdb::Slice` from upperBoundKey_ every time when needed. Since rocksdb::Slice is an string view, and really lightweight.

pengweisong (commented):
Good job, thanks for contribution.

src/kvstore/RocksEngine.h
luyade: iterateUpperBound_ is only constructed the first time when set ReadOptions, not every time..

luyade: updated. We passed a pointer to ReadOptions and we should make sure the pointer valid before the iterator is destructed, so the Slice also should not be a local temporary variable.

pengweisong (approved):
Good job

wenhaocs (approved):

Issue #5094, opened 2022-12-22
Session expire on graphd will end up floating session being released


wey-gu:
This is an issue talked about in WeChat Group:Earth, synced msg was pasted in the end of the issue.

It seems in current idle-release session mechanism was monitored and triggered in each graphD, that is, when a session(session-id:foo) was floated from graphd0 to graphd1, and will be actively used in graphd1 and not caming back to graphd0, after idle session timer expires in graphd0, foo will be released from graphd0.

I am on GitHub mobile(don't have access to the code to verify this)due to Covid leave, so just quickly track the issue here.

Plz forgive my unchecking of the code before raising this.

cc @Aiee 

ref:

- https://github.com/vesoft-inc/nebula-community/discussions/152#discussioncomment-4473597
- https://github.com/vesoft-inc/nebula-community/discussions/152#discussioncomment-4473690
- https://github.com/vesoft-inc/nebula-community/discussions/152#discussioncomment-4474626

HarrisChu:
steps:

1. create a session in graphA.
2. the session connection is broke, then create the connection with graphB.
3. graphB find the session in meta, and it exists, so could query with this session.
4. but session in graphA local would never be updated, and then would be expired by graphA.
5. after delete the session, use cannot use this session in graphB. (as session has been delete in metad)

more info

1. create in graph2 (port 10010)
<img width="1439" alt="image" src="https://user-images.githubusercontent.com/1726587/209643884-ba1c452b-7689-4caf-bfa0-a66c7395a664.png">

2. connect with graph1 (port 10005)
<img width="1418" alt="image" src="https://user-images.githubusercontent.com/1726587/209643913-9f290d64-d27e-4102-98a4-dc5e14235f0e.png">

3. session is delete by graph2
<img width="1439" alt="image" src="https://user-images.githubusercontent.com/1726587/209644102-a7c5dd77-6d59-4eee-8e80-9d47db699965.png">

HarrisChu:
we can simply compare the latest graphAddr in meta, if not in the same, could not delete the session.

not sure if there's any other problems.

Aiee:
This is a rare case and it won't be fixed in 3.4.
cc @xtcyclist

xtcyclist:
> we can simply compare the latest graphAddr in meta, if not in the same, could not delete the session.
> 
> not sure if there's any other problem.

Need to make sure the timeout mechanism still works, if we do this.

xtcyclist:
[...]

PR #5541, opened 2023-05-09
bugfix: only expire local graph sessions

wey-gu:
close: #5094

## What type of PR is this?
- [x] bug
- [ ] feature
- [ ] enhancement

#### Description:

Previously, the session will be expired from a graphd even if it was actively used in other graphd(s), this fixed it. 

## How do you solve it?

Just do reclaim only it's on the current graphd, while now sessions last being active on dead graphds won't be expired anymore.

## Checklist:
Tests:
- [x] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

Fixed an issue when graph session will be expired by one graphd if it is only actively used on other graphd(s), #5094.

> ex. Fixed the bug .....


wey-gu:
(sorry for spam on CI, for some reason, I cannot build on the latest version of the nebula-dev container, with either make or ninja :(, )

lee-qiu:
when graph  load session from meta, it will update related graph address as itself, so I guess this fix seems makes no influence.

wey-gu:
> > > @wey-gu when graph load session from meta, it will update related graph address as itself, so I guess this fix seems makes no sense.
> > 
> > 
> > Thanks @lee-qiu and I agree that this change should be revisited carefully, but the issue that it is trying to fix exist, would you mind coming with fix on it ? :-p
> 
> @wey-gu let me have a try. I wanna add a new rpc in meta to transfer session for graphA to graphB and update related graph address, then I will make sure that graph address must matched with session's address in store for remove/update session rpc. what about this proposal?

I tended to reduce the change surface as much as possible as RPC changes may require major version changes (although Metad's interface is normally not exposed externally, there is still cases doing that).

Would you mind starting a RFC in an issue on your proposal, not necessarily to be in detail, so we could involve more guys from the community to help advice?

Thanks!

lee-qiu:
> > > > @wey-gu when graph load session from meta, it will update related graph address as itself, so I guess this fix seems makes no sense.
> > > 
> > > 
> > > Thanks @lee-qiu and I agree that this change should be revisited carefully, but the issue that it is trying to fix exist, would you mind coming with fix on it ? :-p
> > 
> > 
> > @wey-gu let me have a try. I wanna add a new rpc in meta to transfer session for graphA to graphB and update related graph address, then I will make sure that graph address must matched with session's address in store for remove/update session rpc. what about this proposal?
> 
> I tended to reduce the change surface as much as possible as RPC changes may require major version changes (although Metad's interface is normally not exposed externally, there is still cases doing that).
> 
> Would you mind starting a RFC in an issue on your proposal, not necessarily to be in detail, so we could involve more guys from the community to help advice?
> 
> And I am closing this humble/lazy/dirty change. Thanks!

ok, I will raise a issue for this bug

PR #5584, opened 2023-06-08
modify ft search

cangfengzhs:
## What type of PR is this?
- [ ] bug
- [ ] feature
- [ ] enhancement

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....

### DIFF
--- a/src/parser/test/ParserTest.cpp
+++ b/src/parser/test/ParserTest.cpp
@@ -2929,7 +2929,7 @@ TEST_F(ParserTest, Zone) {
 
 TEST_F(ParserTest, FullText) {
   {
-    std::string query = "LOOKUP ON t1 WHERE ES_QUERY(\"abc\", \"qwerty\")";
+    std::string query = "LOOKUP ON t1 WHERE ES_QUERY(abc, \"qwerty\")";
     auto result = parse(query);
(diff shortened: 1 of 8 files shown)


src/parser/test/ParserTest.cpp
```diff
@@ -2929,22 +2929,7 @@ TEST_F(ParserTest, Zone) {
...
-    EXPECT_TRUE(result.ok()) << result.status();
-  }
-  {
-    std::string query = "LOOKUP ON t1 WHERE ES_QUERY(\"index1\", \"qwerty\")";
-    auto result = parse(query);
-    EXPECT_TRUE(result.ok()) << result.status();
-  }
-  {
-    std::string query = "LOOKUP ON t1 WHERE ES_QUERY(\"qwerty\")";
+    std::string query = "LOOKUP ON t1 WHERE ES_QUERY(\"abc\", \"qwerty\")";
```
yixinglu, line 2932: the first parameter of es_query is better to be a label rather than string

cangfengzhs: done

critical27 (approved):

PR #5609, opened 2023-06-22
Increment ActiveZones only once

wenhaocs:
## What type of PR is this?
- [x] bug
- [ ] feature
- [ ] enhancement

#### Description:
In `CreateSpaceProcessor`, we will check that number of replicas not greater than number of zones and number of active zones. By definition, a zone is active as long as there is an active host in the zone. However, the current code would increment `activeZoneSize` whenever there is an active host in the zone, leading to `activeZoneSize` way more than number of zones. This may cause bugs. 

In this PR, we increment `activeZoneSize` only once for each zone as long as there is an active host in the zone.

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....

### DIFF
--- a/src/meta/processors/parts/CreateSpaceProcessor.cpp
+++ b/src/meta/processors/parts/CreateSpaceProcessor.cpp
@@ -191,6 +191,7 @@ void CreateSpaceProcessor::process(const cpp2::CreateSpaceReq& req) {
 
     auto now = time::WallClock::fastNowInMilliSec();
     auto hosts = MetaKeyUtils::parseZoneHosts(std::move(nebula::value(zoneValueRet)));
+    bool zoneActive = false;
     for (auto& host : hosts) {
       auto key = MetaKeyUtils::hostKey(host.host, host.port);
       auto ret = doGet(key);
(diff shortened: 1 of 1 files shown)


SuperYoko:
Please check logic in ent repo, there's some difference. Zones in this repo only contain one host.

SuperYoko:
> > Please check the logic in ent repo, there's some difference. Zones in this repo only contain one host.
> 
> It's true that it's correct for now. But the way of how the code is written will lead to potential bugs (e.g., when adding more hosts into a zone). In another word, the current result is correct only based on the assumption of one host per zone. Tomorrow, if I want to add more hosts into a zone (what I am exactly doing in the ent version), the current code will be wrong. I can fix it only in the ent version. But I want to make it consistent between ent version and open source version.

Zone has been decided to be removed in open souce version. Here we use zone to adapt "data balance" code to existing code with zone. Better do it in ent version.

wenhaocs:
> > > Please check the logic in ent repo, there's some difference. Zones in this repo only contain one host.
> > 
> > 
> > It's true that it's correct for now. But the way of how the code is written will lead to potential bugs (e.g., when adding more hosts into a zone). In another word, the current result is correct only based on the assumption of one host per zone. Tomorrow, if I want to add more hosts into a zone (what I am exactly doing in the ent version), the current code will be wrong. I can fix it only in the ent version. But I want to make it consistent between ent version and open source version.
> 
> Zone has been decided to be removed in open souce version. And there's no route to add one host into a zone. Here we use zone to adapt "data balance" code to the existing code with zone. Better do it in ent version.

Fair enough.

PR #5651, opened 2023-07-20
replace collect with collectAll

nevermore3:
## What type of PR is this?
- [x] bug
- [ ] feature
- [ ] enhancement

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....

### DIFF
--- a/src/graph/executor/algo/ProduceAllPathsExecutor.cpp
+++ b/src/graph/executor/algo/ProduceAllPathsExecutor.cpp
@@ -42,9 +42,18 @@ folly::Future<Status> ProduceAllPathsExecutor::execute() {
 
   return folly::collectAll(futures)
       .via(runner())
-      .thenValue([this](auto&& status) {
+      .thenValue([this](std::vector<folly::Try<Status>>&& resps) {
+        for (auto& respVal : resps) {
+          if (respVal.hasException()) {
+            auto ex = respVal.exception().get_exception<std::bad_alloc>();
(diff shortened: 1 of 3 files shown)


src/graph/executor/algo/ProduceAllPathsExecutor.cpp
```diff
@@ -40,7 +40,7 @@ folly::Future<Status> ProduceAllPathsExecutor::execute() {
   futures.emplace_back(std::move(leftFuture));
   futures.emplace_back(std::move(rightFuture));
 
-  return folly::collect(futures)
+  return folly::collectAll(futures)
       .via(runner())
       .thenValue([this](auto&& status) {
```
yixinglu, line 45: Are the stats here not of type std::vector<folly::Try<>>? Can you simply ignore the handling of Try?

nevermore3: The type of the parameter is std::vector<folly::Try<Status>>
The buildPath function always returns Status::OK() unless the memory tracker runs out of exception（ will be caught by then Error）

nevermore3: fixed

yixinglu (approved):

codesigner (approved):

PR #5670, opened 2023-08-10
Improve host's leader number bounds when doing leader balance

songqing:
## What type of PR is this?
- [ ] bug
- [ ] feature
- [x] enhancement

#### Description:
When doing leader balance, we can choose a leader number bound for each host, however, the current method is a little simple. 

## How do you solve it?
The gflags "leader_balance_deviation" has been imported, and we can use it to set the host bound dynamically.

## Checklist:
Tests:
- [x] Function test
- [x] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....


songqing:
CI failed, but it seems the test is not stable as the same test can be passed in other test env, besides, the failed test seems have nothing to do with my change.

critical27 (approved):
Good job~

cangfengzhs (approved):

songqing:
@Sophie-Xie Could you help to trigger the CI test again?

PR #5701, opened 2023-09-01
fix concurrency bug

nevermore3:
## What type of PR is this?
- [x] bug
- [ ] feature
- [ ] enhancement

#### Issue(s) number: 
close https://github.com/vesoft-inc/nebula-ent/issues/3006
#### Description:

collectAll holds an iterable collection type whose element type is Future<T>, and any error in the component Future will not cause the process to terminate prematurely

collect() is  similar to collectAll(), the only difference is that if any of the input Futures throws an exception, the Future will be terminated early

so when we set memory tracker, if we use folly::collect , when a thread triggers the limit of the memory threshold，the program may end and there are other threads running.

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....

### DIFF
--- a/src/graph/executor/Executor.h
+++ b/src/graph/executor/Executor.h
@@ -174,7 +174,7 @@ auto Executor::runMultiJobs(ScatterFunc &&scatter, GatherFunc &&gather, Iterator
   }
 
   // Gather all results and do post works
-  return folly::collect(futures).via(runner()).thenValue(std::move(gather));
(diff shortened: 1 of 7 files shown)


Sophie-Xie:
fixed in https://github.com/vesoft-inc/nebula/pull/5702

PR #5734, opened 2023-10-08
each group applies a rule only once

nevermore3:
## What type of PR is this?
- [ ] bug
- [ ] feature
- [x] enhancement

#### Issue(s) number: 
close https://github.com/vesoft-inc/nebula-ent/issues/3052
#### Description:

The execution plan tree is a DAG
In the optimizer phase, checkPlanDepth, explore, getPlan, pruneProperties and analyzeLifetime in the scheduling phase will recursively traverse the execution plan tree, and nodes that have been visited will be visited again.

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....

### DIFF
--- a/src/graph/optimizer/Optimizer.cpp
+++ b/src/graph/optimizer/Optimizer.cpp
@@ -52,7 +52,8 @@ StatusOr<const PlanNode *> Optimizer::findBestPlan(QueryContext *qctx) {
 
 // Just for Properties Pruning
 Status Optimizer::postprocess(PlanNode *root, graph::QueryContext *qctx, GraphSpaceID spaceID) {
-  NG_RETURN_IF_ERROR(rewriteArgumentInputVar(root));
+  std::unordered_set<const PlanNode *> visitedPlanNode;
+  NG_RETURN_IF_ERROR(rewriteArgumentInputVar(root, visitedPlanNode));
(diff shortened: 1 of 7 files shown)


src/graph/optimizer/Optimizer.cpp
```diff
@@ -206,37 +215,47 @@ Status Optimizer::rewriteArgumentInputVarInternal(PlanNode *root,
...
+        rewriteArgumentInputVar(const_cast<PlanNode *>(sel->otherwise()), visitedPlanNode));
   }
 
   return Status::OK();
 }
 
 // static
-Status Optimizer::rewriteArgumentInputVar(PlanNode *root) {
+Status Optimizer::rewriteArgumentInputVar(PlanNode *root,
+                                          std::unordered_set<const PlanNode *> &visitedPlanNode) {
```
yixinglu, line 235: `visitedPlanNode` only needs to be defined inside this function, and does not need to be passed in as a parameter

nevermore3: rewriteArgumentInputVar & rewriteArgumentInputVarInternal. is static function
so visitedPlanNode is defined outside the function

yixinglu (approved):

codesigner (approved):

Issue #5772, opened 2023-11-27
round() causes the graphd to crash


QingZ11:
As titled, the `round()` function causes the graphd to crash.

* nebula version: v3.4.0 
* error infos:

<img width="450" alt="image" src="https://github.com/vesoft-inc/nebula/assets/38887077/e2b8ede8-5b1f-44e6-94fa-6d188808345f">

<img width="555" alt="image" src="https://github.com/vesoft-inc/nebula/assets/38887077/1c1b5cf5-d039-444f-b4af-330c0a537158">

PR #5773, opened 2023-11-27
fix round crash

HarrisChu:
fix https://github.com/vesoft-inc/nebula/issues/5772

### DIFF
--- a/src/common/function/FunctionManager.cpp
+++ b/src/common/function/FunctionManager.cpp
@@ -584,7 +584,7 @@ FunctionManager::FunctionManager() {
         case Value::Type::FLOAT: {
           if (args.size() >= 2) {
             if (args[1].get().type() == Value::Type::INT) {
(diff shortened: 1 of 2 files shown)


jievince (approved):

Aiee (approved):

src/common/function/FunctionManager.cpp
```diff
@@ -584,7 +584,7 @@ FunctionManager::FunctionManager() {
         case Value::Type::FLOAT: {
           if (args.size() >= 2) {
             if (args[1].get().type() == Value::Type::INT) {
-              auto val = args[0].get().getFloat();
+              auto val = args[0].get().isInt() ? args[0].get().getInt() : args[0].get().getFloat();
```
yixinglu, line 587: why args[0] is not float number when its type is float?

HarrisChu: because of the previous case ...

PR #5775, opened 2023-11-29
feat: add IPv6 support

wey-gu:
## What type of PR is this?
- [ ] bug
- [x] feature
- [ ] enhancement

#### Description:

Now it should support dual stack for some friends in the community. This rough change is targeted to create a branch for their quick testing on IPv6-only env(that unfortunately doesn't come with a DNS)

Note intToIPv4 is not handled due to it's probably for the metad of NebulaGraph v1 only.

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

IPv6 is supported but not fully tested.


wey-gu:
ping @congguosn

PR #5790, opened 2023-12-25
check max validate depth

nevermore3:
## What type of PR is this?
- [ ] bug
- [ ] feature
- [x] enhancement

#### Issue(s) number: 
close https://github.com/vesoft-inc/nebula-ent/issues/3403
## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....

### DIFF
--- a/src/graph/validator/Validator.cpp
+++ b/src/graph/validator/Validator.cpp
@@ -40,7 +40,7 @@
 #include "graph/visitor/EvaluableExprVisitor.h"
 #include "parser/Sentence.h"
 DEFINE_uint32(max_statements,
-              1024,
+              512,
               "threshold for maximun number of statements that can be validate");
 namespace nebula {
 namespace graph {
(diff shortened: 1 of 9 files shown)


src/graph/validator/Validator.h
```diff
@@ -203,6 +204,7 @@ class Validator {
   std::set<std::string> userDefinedVarNameList_;
   // vid's Type
   Value::Type vidType_;
+  static thread_local uint32_t maxStatements_;
```
jievince, line 207: Why it need to be thread_local? Validator is single-thread.

nevermore3: makeValidator is a static function，maxStatement_ is also static member, so use thread_local

jievince (dismissed):

src/graph/validator/Validator.cpp
```diff
@@ -273,22 +281,27 @@ std::unique_ptr<Validator> Validator::makeValidator(Sentence* sentence, QueryCon
...
-  if (!root) {
-    return Status::SemanticError("Get null plan from sequential validator");
+    auto root = validator->root();
+    if (!root) {
+      return Status::SemanticError("Get null plan from sequential validator");
+    }
+    qctx->plan()->setRoot(root);
+    // reset maxStatements
+    maxStatements_ = 0;
+  } catch (const std::runtime_error& error) {
```
yixinglu, line 301: catch std::exception

nevermore3: done

yixinglu (approved):

jievince (approved):

Issue #5764, opened 2023-11-16
The param: ”noloop“ in find path does not exclude ”self-loop“


lishihao97:
**Please check the FAQ documentation before raising an issue**

**Describe the bug (__required__)**
在find path语句中使用noloop参数未排除自环情况
NGQL：
find noloop path from "nodea" to "nodea" over e1 YIELD PATH as p 

![image](https://github.com/vesoft-inc/nebula/assets/40377824/c6bc304b-b6cb-4536-8a7b-c4c07feb3b89)

The param: ”noloop“ in find path does not exclude ”Loop“

**Your Environments (__required__)**
nebula 3.6
* OS: `linux 3.10.0-1062.9.1.el7.x86_64`
* Compiler: `g++ --version` or `clang++ --version`
* CPU: `lscpu`
* Commit id (e.g. `a3ffc7d8`)

ref: [https://discuss.nebula-graph.com.cn/t/topic/14370](https://discuss.nebula-graph.com.cn/t/topic/14370)

AntiTopQuark:
Is the issue definitively buggy, please?
If it needs to be fixed, can I be assigned this task?

wey-gu:
Wow, thanks so much @AntiTopQuark !!!

Sure! I assigned this to you, let us know what help from us could be provided any time, and take your time!

cc @MuYiYong @yixinglu

PR #5805, opened 2024-01-13
fix:a single node self-loop when executing NOLOOP

AntiTopQuark:
## What type of PR is this?
- [x] bug
- [ ] feature
- [ ] enhancement

#### Issue(s) number: 
fix https://github.com/vesoft-inc/nebula/issues/5764
#### Description:
Use of the noloop parameter in the find path statement does not exclude the self-loop case

## How do you solve it?
New logical judgement on self-loops in the AllPathsExecutor::doBuildPath() function

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....

### DIFF
--- a/src/graph/executor/algo/AllPathsExecutor.cpp
+++ b/src/graph/executor/algo/AllPathsExecutor.cpp
@@ -331,6 +331,12 @@ folly::Future<std::vector<AllPathsExecutor::NPath*>> AllPathsExecutor::doBuildPa
         continue;
       }
       for (auto& edge : adjEdges) {
+        if (noLoop_) {
+          if (edge.getEdge().dst == edge.getEdge().src) {
+            continue;
+          }
+        }
+        auto path = NPath(src, edge);
         threadLocalPtr_->emplace_back(NPath(src, edge));
         newPathsPtr->emplace_back(&threadLocalPtr_->back());
       }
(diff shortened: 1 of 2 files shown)


tests/tck/features/path/NoLoop.feature
```diff
@@ -242,3 +242,40 @@ Feature: NoLoop Path
...
+      """
+    And wait 1 seconds
+    Then the execution should be successful
+    When executing query:
+      """
+      FIND NOLOOP PATH FROM "nodea" TO "nodea" OVER Link YIELD PATH AS p;
+      """
+    Then the result should be, in any order:
+      | p     |
+      | EMPTY |
```
yixinglu, line 269: The result seems an empty table rather than a table contains EMPTY value, such as: 

```feature
      | p     |
```

src/graph/executor/algo/AllPathsExecutor.cpp
```diff
@@ -331,6 +331,12 @@ folly::Future<std::vector<AllPathsExecutor::NPath*>> AllPathsExecutor::doBuildPa
         continue;
       }
       for (auto& edge : adjEdges) {
+        if (noLoop_) {
+          if (edge.getEdge().dst == edge.getEdge().src) {
+            continue;
+          }
+        }
+        auto path = NPath(src, edge);
```
Salieri-004, line 339: This code seems redundant.

AntiTopQuark: done

AntiTopQuark:
image
The two results look the same, but pytest thinks they are different and reports an error?

yixinglu (approved):

Salieri-004:
> image The two results look the same, but pytest thinks they are different and reports an error?

This is not as expected and may be a problem with our testing framework. Thanks for your help.

PR #5816, opened 2024-02-01
fix compare node

Salieri-004:
## What type of PR is this?
- [x] bug
- [ ] feature
- [ ] enhancement

#### Description:
In pr https://github.com/vesoft-inc/nebula/pull/5805, results were the same but the test framework reported an error.
image

## How do you solve it?

In compare_node(lhs, rhs),If lhs.tags == [] and rhs.tags is None, the two tags should be equal.

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....

### DIFF
--- a/tests/common/comparator.py
+++ b/tests/common/comparator.py
@@ -295,21 +295,20 @@ class DataSetComparator:
         return False
 
     def compare_node(self, lhs: Vertex, rhs: Vertex):
-        rtags = []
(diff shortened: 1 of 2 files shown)


codesigner (approved):

yixinglu (approved):
Good job!

Issue #5840, opened 2024-03-21
Add an MD5() function to NebulaGraph


QingZ11:
The idea was provided by @AntiTopQuark , thank you.

## Requirement Description

MD5 (Message Digest Algorithm 5) is a widely used cryptographic hash function that produces a 128-bit hash value, commonly used to ensure the integrity and consistency of information transmission. In graph databases, MD5 can be used for rapidly comparing node or edge attribute values, generating unique identifiers, etc. Therefore, implementing the MD5 function will enhance NebulaGraph's data processing capabilities and provide more flexibility in data operations.

NebulaGraph already supports a variety of functions and expressions, such as `split` and `concat`. The implementations of these functions provide us with a reference framework for writing custom functions, which includes but is not limited to:

* **Function registration**: Understanding how to register new functions into NebulaGraph's function library.

* **Argument handling**: Becoming familiar with how to process input arguments, including type checking and conversion.

* **Result returning**: Mastering how to compute results and return them to the caller in the appropriate data type.

## Implementation Approach

1. Refer to Existing Functions

By consulting the official documentation, we can find that NebulaGraph has implemented a variety of functions and expressions, such as string functions. These existing implementations provide us with valuable references, allowing us to quickly locate the corresponding implementations and commit records in the codebase by searching for specific function names, like `json_extract`. For example, we can find commit records and merge request information related to `json_extract`.

By analyzing these historical commits, we learn that implementing a new function involves the following steps:

* Add the registration and computational logic of the function in the `src/common/function/FunctionManager.cpp` file.

* Add corresponding unit tests in the `src/common/function/test/FunctionManagerTest.cpp` file.

* For new features, add TCK integration tests.

* (Optional) Update the documentation in the [nebula-docs](https://github.com/vesoft-inc/nebula-docs) repository, providing explanations and examples for the new feature to help users better understand and use the new functionality.

![image](https://github.com/vesoft-inc/nebula/assets/38887077/970ed353-2955-4779-b454-7f19d301f2cc)

2. MD5 Function Implementation

Implementing the MD5 function may not require us to handle the details of the function from scratch; we can start by conducting a global search in the codebase to see if there are any existing implementations for reference. A simple search can easily locate some available function instances.

![image](https://github.com/vesoft-inc/nebula/assets/38887077/4a373c58-9222-4e04-973b-4bd4afbba288)

3. Related Code Implementation

[...]

PR #5846, opened 2024-04-02
feat(function): support md5

fansehep:
## What type of PR is this?
- [ ] bug
- [ ] feature
- [ ] enhancement

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....


QingZ11:
The automated lint test has failed. You can click on this [link](https://github.com/vesoft-inc/nebula/actions/runs/8586763639/job/23529873358?pr=5846) or check the automated test results below to see where the issue is and fix it. Once fixed, you can commit the changes.

CMakeLists.txt
```diff
@@ -30,6 +30,8 @@ option(ENABLE_PACKAGE_TAR "Enable package artifacts to tar." OFF)
 option(ENABLE_CREATE_GIT_HOOKS "Enable create git hooks." ON)
 option(ENABLE_INCLUDE_WHAT_YOU_USE "Enable include-what-you-use find  nouse include files" OFF)
 
+SET(CMAKE_EXPORT_COMPILE_COMMANDS TRUE)
```
yixinglu, line 33: This `CMAKE_EXPORT_COMPILE_COMMANDS` option could be specified in cmake configuration step and need not to always turn it on. please cleanup it.

yixinglu (commented):
LGTM, but need to fix the lint error

CMakeLists.txt
fansehep: please let the CI run.

QingZ11: @fansehep hi, according to yixinglu's response, you can simply remove the added `CMAKE_EXPORT_COMPILE_COMMANDS` configuration statement. Currently, it is set to true, but as per yixinglu's suggestion, it is already configurable elsewhere, so there is no need to hardcode it here.

fansehep: thanks for your reminders, but from the CI compile output, the mold linker link error :(

QingZ11: Nebula CI sometimes needs to run a few more times to pass. I will rerun it a few more times. Wait for my good news 🥺.

Salieri-004: > thanks for your reminders, but from the CI compile output, the mold linker link error :(

Perhaps you should add some link dependencies, as commented in this issue. #5840

yixinglu: @fansehep To fix the failure of CI, u'd better to check where the target `$<TARGET_OBJECTS:function_manager_obj>` is used in the whole code base, and add the the `${PROXYGEN_LIBRARIES}` dependency for there.

zhanghuidinah: @fansehep CI has been fixed, shall we go on?

PR #5863, opened 2024-04-18
fix  ifa->ifa_addr 空指针

qingyan01:
## What type of PR is this?
 bug

image

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....

### DIFF
--- a/src/common/network/NetworkUtils.cpp
+++ b/src/common/network/NetworkUtils.cpp
@@ -66,7 +66,7 @@ StatusOr<std::vector<std::pair<std::string, std::string>>> NetworkUtils::listDev
   }
   for (auto* ifa = iflist; ifa != nullptr; ifa = ifa->ifa_next) {
     // Skip non-IPv4 devices
-    if (ifa->ifa_addr->sa_family != AF_INET) {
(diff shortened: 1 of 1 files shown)


src/common/network/NetworkUtils.cpp
```diff
@@ -66,7 +66,7 @@ StatusOr<std::vector<std::pair<std::string, std::string>>> NetworkUtils::listDev
   }
   for (auto* ifa = iflist; ifa != nullptr; ifa = ifa->ifa_next) {
     // Skip non-IPv4 devices
-    if (ifa->ifa_addr->sa_family != AF_INET) {
+    if (nullptr == ifa->ifa_addr || ifa->ifa_addr->sa_family != AF_INET) {
```
yixinglu, line 69: prefer to move the following `auto *addr` above the `if` and check whether the `addr` is nullptr

qingyan01: ok

yixinglu (approved):

Salieri-004 (approved):

Salieri-004 (approved):

PR #5905, opened 2024-07-05
fix. raft follower will rollback itself when it misses a certain log

UndefinedSy:
## What type of PR is this?
- [x] bug
- [ ] feature
- [ ] enhancement

#### Description:
    if a leader fails to update its last_log_id_sent (maybe caused be
    follower fail or network issue), it will then try to send some
    duplicated logs in the next appendLog request.
    
    ```
                          commitId_: 99
                              |
                              v
    local wal: |-------------------|
    req.append_entries:        |----------------|
    ```
    
    in this case, follower will do rollback even if the logs are the same.

## How do you solve it?
add a bool variable to track if the log is inconsistent and only do rollback if the log has a conflict.

with this fix, followers will not do unnecessary rollback in network issues.  
raft rollback is not a lightweight operation, especially if we have a large number of parts, which may may lead to cascade problems.

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [x] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....


UndefinedSy:
> Sorry that we are not working on this repo for a long time... Good job~

The CI workflow doesn't seem to work. Actually, I've some other PRs also failed in the build check. 
Will this be fixed or this repo will no longer be maintained. If so, let's just leave it here so that others can find a solution to this problem.

PR #5914, opened 2024-07-31
Added support for the List/Set type

YZW00:
For one-dimensional data of type List/Set:
List_string,  List_int,  List_float,
Set_string,  Set_int,   Set_float:

1. Support attributes defined as List /Set data types.

2. Use the List /Set data type for storage.

3. The List /Set data type can be filtered during query.

4. Attributes of the List /Set data type can be returned when the query is supported.

5. Added UPDATE statement to modify (REPLACE) and remove (ERASE) specified elements in List/Set types.

6. Added UPDATE statement to insert new data into List/Set: List = List + value, Set = SADD(Set, value).


Salieri-004:
Thanks for your help. You can add some tests for your new features according to the following articles.
https://discuss.nebula-graph.com.cn/t/topic/4172
https://discuss.nebula-graph.com.cn/t/topic/4594

src/graph/validator/MutateValidator.cpp
```diff
@@ -714,12 +714,24 @@ Status UpdateValidator::getUpdateProps() {
       symNames.emplace(name_);
     }
     if (item->getFieldExpr() != nullptr) {
-      DCHECK(item->getFieldExpr()->kind() == Expression::Kind::kLabelAttribute);
-      auto laExpr = static_cast<const LabelAttributeExpression *>(item->getFieldExpr());
-      symNames.emplace(laExpr->left()->name());
-      symName = laExpr->left()->name();
-      const auto &value = laExpr->right()->value();
-      fieldName = value.getStr();
+      if (item->getFieldExpr()->kind() == Expression::Kind::kLabelAttribute) {
```
flymysql, line 717: 这段逻辑的变动确实会影响到单元测试
/home/nebula/tests/tck/features/bugfix/SubscriptInUpdate.feature

image

src/graph/executor/algo/AllPathsExecutor.cpp
```diff
@@ -562,38 +562,38 @@ folly::Future<Status> AllPathsExecutor::conjunctPath(std::vector<NPath*>& leftPa
   }
   return folly::collectAll(futures)
       .via(runner())
-      .thenValue([this,
```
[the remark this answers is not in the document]
YZW00: Sorry, the code is the same as the original code, but I modified part of the format, which caused the system to misjudge. Now I have modified back to the original code, but the system still shows that I modified, need to trouble the community to review again

src/common/meta/NebulaSchemaProvider.cpp
```diff
@@ -167,6 +167,20 @@ std::size_t NebulaSchemaProvider::fieldSize(PropertyType type, std::size_t fixed
       return 8;  // as same as STRING
     case PropertyType::DURATION:
       return sizeof(int64_t) + sizeof(int32_t) + sizeof(int32_t);
+    case PropertyType::LIST_STRING:
+    case PropertyType::LIST_INT:
+    case PropertyType::LIST_FLOAT:
+    case PropertyType::LIST_LIST_STRING:
```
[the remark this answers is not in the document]
YZW00: Ok, I've removed all the 2D data types that I added myself

src/codec/RowWriterV2.cpp
```diff
@@ -820,6 +829,130 @@ WriteResult RowWriterV2::write(ssize_t index, const Geography& v) {
...
+
+  if (isSet_[index]) {
+    outOfSpaceStr_ = true;
+  }
+
+  int32_t listSize = list.size();
+  int32_t listOffset = buf_.size();
+  buf_.append(reinterpret_cast<char*>(&listSize), sizeof(int32_t));
+
+  for (const auto& item : list.values) {
```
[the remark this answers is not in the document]
YZW00: Ok, I've rewritten a template function called writeContainer that supports both List and Set

src/codec/RowWriterV2.h
```diff
@@ -263,6 +263,8 @@ class RowWriterV2 {
   WriteResult write(ssize_t index, const Duration& v);
 
   WriteResult write(ssize_t index, const Geography& v);
+  WriteResult write(ssize_t index, const List& list);
```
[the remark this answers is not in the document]
YZW00: Of course, I've added comments to both write()

src/common/function/FunctionManager.cpp
```diff
@@ -1388,43 +1388,41 @@ FunctionManager::FunctionManager() {
...
+        std::string &origStr = const_cast<std::string &>(args[0].get().getStr());
+        std::string search = args[1].get().getStr();
+        std::string newStr = args[2].get().getStr();
+        boost::replace_all(origStr, search, newStr);
+        return origStr;
       }
+
       if (args[0].get().isList()) {
-        auto list = args[0].get().getList();
+        auto &list = const_cast<List &>(args[0].get().getList());
```
Salieri-004, line 1402: You might misunderstand my comment. `const_cast` is risky; we should not modify the value of the input parameter. Instead, we should copy it.

src/common/function/FunctionManager.cpp
```diff
@@ -1388,43 +1388,41 @@ FunctionManager::FunctionManager() {
       if (args[0].get().isNull() || args[1].get().isNull() || args[2].get().isNull()) {
         return Value::kNullValue;
       }
+
       if (args[0].get().isStr() && args[1].get().isStr() && args[2].get().isStr()) {
-        std::string origStr(args[0].get().getStr());
-        std::string search(args[1].get().getStr());
-        std::string newStr(args[2].get().getStr());
-        return boost::replace_all_copy(origStr, search, newStr);
+        std::string &origStr = const_cast<std::string &>(args[0].get().getStr());
```
Salieri-004, line 1394: DITTO

src/common/function/FunctionManager.cpp
```diff
@@ -1388,43 +1388,41 @@ FunctionManager::FunctionManager() {
...
+            item = newValue;
           }
         }
-        return result;
+        return list;
       }
+
       if (args[0].get().isSet()) {
-        auto set = args[0].get().getSet();
+        auto &set = const_cast<Set &>(args[0].get().getSet());
```
Salieri-004, line 1414: DITTO

src/codec/RowWriterV2.cpp
```diff
@@ -16,6 +16,61 @@ namespace nebula {
 
 using nebula::cpp2::PropertyType;
 
+// Template functions for handling different types (e.g. strings, ints, floats) in lists and
+// collections
+template <typename Container>
+WriteResult writeContainer(const Container& container,
+                           Value::Type valueType,
+                           bool isSetType,
+                           std::unordered_set<std::string>& serializedStrings,
```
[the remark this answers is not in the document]
YZW00: Thank you for your suggestion. I have further improved it, but our implementation can only distinguish at runtime and cannot be done during static compilation.

src/common/function/FunctionManager.cpp
YZW00: Ok, I have removed the const_cast and changed it to directly copy the value of the input parameter for operation. At the same time, the list is modified in place directly instead of creating an additional result variable.

src/codec/RowWriterV2.cpp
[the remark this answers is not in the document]
Salieri-004: > Thank you for your suggestion. I have further improved it, but our implementation can only distinguish at runtime and cannot be done during static compilation.

We can determine its type outside the loop, instead of checking its type each time.
```
template <typename Type>
WriteResult write(...) {
    std::unordered_set<Type> serialized;
    if constexpr (std::is_same_v<Type, int>) {
        ...
    } else if constexpr (std::is_same_v<Type, std::string>) {
        ...
    } else if constexpr (std::is_same_v<Type, float>) {
        ...
    }
}

    switch (propType) {
        case LIST_STRING: {
            write<std::string>();
        }
        case LIST_INT: {
            write<int64_t>();
        }
    }
```

src/codec/RowWriterV2.cpp
```diff
@@ -16,6 +16,50 @@ namespace nebula {
 
 using nebula::cpp2::PropertyType;
 
+// Template functions implement support for list and set data types
+template <typename Container>
+WriteResult writeContainer(const Container& container, Value::Type valueType, std::string& buffer) {
+  std::unordered_set<Value> serialized;
```
Salieri-004, line 22: If the container is a List, we don't need the`unordered_set`. Can we be sure that the compiler can ignore it without affecting performance?

YZW00: OK, I have modified the implementation of two functions corresponding to List and Set respectively. Unordered_set is only used when necessary: ​​writeList() does not use unordered_set, writeSet() uses it to ensure data uniqueness

Salieri-004:
LGTM. We need to add some doc for this feature. cc. @ChrisChen2023 @abby-cyber

Salieri-004 (dismissed):

tests/tck/features/update/Update.feature
```diff
@@ -1207,3 +1207,135 @@ Feature: Update string vid of vertex and edge
...
+    And having executed:
+      """
+      CREATE TAG player(name string, age int, hobby List< string >, ids List< int >, score List< float >);
+      """
+    And wait 3 seconds  # Wait for the index to be created
+    # Insert vertex data for player100
+    When try to execute query:
+      """
+      INSERT VERTEX player(name, age, hobby, ids, score) VALUES "player100":(
+          "Tim Duncan", 42, ["Basketball", "Swimming"], [100, 528], [50.0, 22.0]
```
codesigner, line 1226: Is there testcase cover some corner cases, e.g.
1. empty list/set, insert or update.
2. Fire some error conditions: for example: 
   - type mismatched, like insert List[?] to Set[?] , or insert List[int] to List[string];
   - invoke new added function with wrong type, e.g. add string to List[int] 
   - what happends if try to create index on list/set column?
3. very large list/set, according to your implementation, there is seem no limit (or more accurate max of int32) on the size of list or set, right? Should we support that large size list/set?

src/common/function/FunctionManager.cpp
```diff
@@ -1377,6 +1395,65 @@ FunctionManager::FunctionManager() {
...
+        auto set = args[0].get().getSet();
+        auto target = args[1].get();
+        set.values.erase(target);
+        return set;
+      }
+      return Value::kNullBadType;
+    };
+  }
+  {
+    auto &attr = functions_["sadd"];
```
codesigner, line 1444: likes like sadd is short for set add, maybe setadd is more readable?

codesigner (commented):
Thinks for the contirbution.

Salieri-004:
@YZW00 We need to add some restrictions to disallow `create index` on set or list types.

tests/tck/features/update/Update.feature
YZW00: 1. I have added test cases: Insert and update empty list/set
2. If the inserted data type is inconsistent, an error will be reported: Storage Error: The data type does not meet the requirements. Use the correct type of data.
    I also restricted the creation of list/set indexes and added test cases.
3. I have added a restriction on the array length

codesigner (approved):

Salieri-004 (approved):

flymysql (commented):
LGTM

flymysql (approved):
LGTM

PR #5963, opened 2024-10-23
feat:add transfer Leader cmd

E2ern1ty:
## What type of PR is this?
- [ ] bug
- [x] feature
- [ ] enhancement

#### Description:
I think this feature will be more universal if placed in the nebula kernel
https://github.com/vesoft-inc/nebula-operator/issues/270

## How do you solve it?
add new transfer cmd for nebula, remove leader before change storaged group.
Maybe this function needs some guidance to make it better.

test in 3 replica,2 shards,just as the picture:
企业微信截图_31c1c9b2-9f8c-4454-aed5-87cbc76ade5e

## Checklist:
Tests:
- [x] Function test
- [ ] Performance test
- [ ] N/A

Affects:

## Release notes:

> ex. Fixed the bug .....

### DIFF
--- a/src/clients/meta/MetaClient.cpp
+++ b/src/clients/meta/MetaClient.cpp
@@ -2840,6 +2840,26 @@ folly::Future<StatusOr<std::vector<cpp2::RoleItem>>> MetaClient::getUserRoles(st
   return future;
 }
 
+folly::Future<StatusOr<bool>> MetaClient::transferLeader(const HostAddr& host,
+                                                         GraphSpaceID spaceId,
+                                                         int32_t concurrency) {
+  cpp2::LeaderTransferReq req;
+  req.space_id_ref() = spaceId;
+  req.host_ref() = host;
+  req.concurrency_ref() = concurrency;
+  folly::Promise<StatusOr<bool>> promise;
+  auto future = promise.getFuture();
+  getResponse(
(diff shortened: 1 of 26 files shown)


critical27:
Thx for contribution. But I'm wondering what's the purpose of this command? Is it used for `data balance remove ...`, as described in https://github.com/vesoft-inc/nebula-operator/issues/270? If yes, what happened here is that before we remove a part from the leader, the leadership has already been transferred.

E2ern1ty:
> Thx for contribution. But I'm wondering what's the purpose of this command? Is it used for `data balance remove ...`, as described in [vesoft-inc/nebula-operator#270](https://github.com/vesoft-inc/nebula-operator/issues/270)? If yes, what happened here is that before we remove a part from the leader, the leadership has already been transferred. So the command is not quite necessary.



> Thx for contribution. But I'm wondering what's the purpose of this command? Is it used for `data balance remove ...`, as described in [vesoft-inc/nebula-operator#270](https://github.com/vesoft-inc/nebula-operator/issues/270)? If yes, what happened here is that before we remove a part from the leader, the leadership has already been transferred. So the command is not quite necessary.


This command is used to migrate the leaders on a storage node before repairing the machine. It does not involve any data migration, aiming to minimize the impact on upper-layer services.

Issue #6042, opened 2025-04-10
Support vector value type and data type


czpmango:

tom-jerr:
**Feature Request: Add Data Type Specification for Vector Values**
1. We need a new value type VECTOR, like List, Map, etc.. This type is for vector data.
2. VECTOR type has original vector data and metadata of data, like dimension.
3. We also need a property type for vector property.
4. For better performance, we need to storage other properties and vector property separately, so we use multiple column family rocksdb.

PR #6060, opened 2025-06-23
lzy: Support vector value type and data type

tom-jerr:
## What type of PR is this?
- [ ] bug
- [x] feature
- [ ] enhancement

#### Issue(s) number: 
Support vector value type and data type #6042
#### Description:
**Feature Request: Add Data Type Specification for Vector Values**
1. We need a new value type VECTOR, like List, Map, etc.. This type is for vector data.
2. VECTOR type has original vector data and metadata of data, like dimension.
3. We also need a property type for vector property.
4. For better performance, we need to storage other properties and vector property separately, so we use multiple column family rocksdb.

## How do you solve it?
1. Implement vector value type and property type and do the unit test for vector type
	- Add new value type: VECTOR
	- Modify RocksEngine to support multiple column family of RocksDB
2. Implement serialization and deserialization for vector type: Graphd to Storaged by thrift
	- Add thrift operation for vector type
	- Encoding of vector type key
	- RowWriter and RowReader for vector type

### Vector Value Type
Vector type is consist of original vector data and dimension of vector data. 
VectorType

### Vector Property Type
The VECTOR type can be used as a property type. For storage, the actual data content of the vector is placed at the end of the encoded data block. We then use an offset and a length to specify the exact memory region where the vector's content is located. 

###  Storage for Vector Data
Specifically, Vertex data is stored in RocksDB in the form of KV pairs. We use RocksDB's Column Family to separate storage of VECTOR attributes and other graph attributes.  We should add read and write method for RowReader and RowWriter.

#### Key of Vector Type 
The key of vector attributes is PartitionID + VertexID + TagID + PropID, Value is serialized vector data.
vector_key

#### Encoded Value
The encoded value of vector is consist of metadata and original data. Metadata is offset and length of original data, and original data is at last of whole value.
encode_value

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

TCK Tests:
tck_tests

Affects:

## Release notes:

> ex. Fixed the bug .....

### DIFF
--- a/src/common/datatypes/Vector.h
+++ b/src/common/datatypes/Vector.h
@@ -1,0 +1,171 @@
+/* Copyright (c) 2020 vesoft inc. All rights reserved.
+ *
+ * This source code is licensed under Apache 2.0 License.
+ */
+
+#ifndef COMMON_DATATYPES_VECTOR_H_
+#define COMMON_DATATYPES_VECTOR_H_
+#include <folly/dynamic.h>
+
+#include <cassert>
+#include <cmath>
+#include <iostream>
+#include <sstream>
+#include <vector>
+namespace nebula {
+/*
+ * Vector Type is consist of a vector of values and dimension of vector.
+ * The element type in vector is float.
+ */
+struct Vector {
+  int64_t dimension;
+  std::vector<float> values;
+
+  Vector() : dimension(0) {}
+  Vector(const Vector& rhs) = default;
+  Vector(Vector&& rhs) noexcept = default;
+  explicit Vector(std::vector<float> vals, int64_t dim) : dimension(dim), values(std::move(vals)) {}
+  explicit Vector(std::vector<float> vals) : dimension(vals.size()), values(std::move(vals)) {}
+
+  Vector& operator=(const Vector& rhs) {
+    if (this == &rhs) {
+      return *this;
+    }
+    values = rhs.values;
+    dimension = rhs.dimension;
+    return *this;
+  }
+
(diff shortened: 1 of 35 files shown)


src/common/datatypes/Vector.h
```diff
@@ -0,0 +1,171 @@
...
+#include <iostream>
+#include <sstream>
+#include <vector>
+namespace nebula {
+/*
+ * Vector Type is consist of a vector of values and dimension of vector.
+ * The element type in vector is float.
+ */
+struct Vector {
+  int64_t dimension;
```
czpmango, line 21: The field `dimension` will introduce additional memory overhead, which is unnecessary.
```
auto dim() const {
     return values.size();
}
```

src/common/datatypes/Vector.h
```diff
@@ -0,0 +1,171 @@
...
+    return result;
+  }
+
+  Vector operator-(const Vector& rhs) const {
+    Vector result;
+    result.dimension = dimension;
+    result.values.resize(dimension);
+    for (size_t i = 0; i < static_cast<size_t>(dimension); ++i) {
+      result.values[i] = values[i] - rhs.values[i];
+    }
```
czpmango, line 112: Comparison operations on vector data should not be allowed. We should limit its operation semantics, which is also the key reason why vector data type cannot reuse list type.

tom-jerr: Ok, I will update the code and document

Issue #6067, opened 2025-07-01
Implement the vector value type and serialization and deserialization for it


tom-jerr:
# Vector Value Type
We need to support vector data in nebula graph. So we should a new type called VECTOR in type system, which must be different from List. It contains original vector data (a set of float) and metadata (dimension).

PR #6068, opened 2025-07-01
Implement vector value type

tom-jerr:
## What type of PR is this?
- [ ] bug
- [x] feature
- [ ] enhancement

#### Issue(s) number: 
Support vector value type and data type #6042
Implement the vector value type and serialization and deserialization for it  #6067

#### Description:
**Feature Request: Add Data Type Specification for Vector Values**
1. We need a new value type VECTOR, like List, Map, etc.. This type is for vector data.
2. VECTOR type has original vector data and from original data we can get dimension of a vector.

## How do you solve it?
1. Implement vector value type and do the unit test for vector type
	- Add new value type: VECTOR
	- Add serialization and deserialization method for vector type.
### Vector Value Type
Vector type is consist of original vector data. 
We also offered a dim() method to get the dimension of vector data, which is from vector data size.

image

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

TCK Tests:

image

Affects:

## Release notes:

> ex. Fixed the bug .....

### DIFF
--- a/src/common/datatypes/Value.cpp
+++ b/src/common/datatypes/Value.cpp
@@ -19,6 +19,7 @@
 #include "common/datatypes/Map.h"
 #include "common/datatypes/Path.h"
 #include "common/datatypes/Set.h"
+#include "common/datatypes/Vector.h"
 #include "common/datatypes/Vertex.h"
 
 namespace nebula {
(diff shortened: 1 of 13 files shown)


src/common/datatypes/Value.cpp
```diff
@@ -32,6 +33,7 @@ const Value Value::kNullOverflow(NullType::ERR_OVERFLOW);
 const Value Value::kNullUnknownProp(NullType::UNKNOWN_PROP);
 const Value Value::kNullDivByZero(NullType::DIV_BY_ZERO);
 const Value Value::kNullOutOfRange(NullType::OUT_OF_RANGE);
+const Value Value::kVectorDimNotMatch(NullType::VEC_DIM_NOT_MATCH);
```
czpmango, line 36: Why we need this runtime error value?

czpmango (approved):
LGTM

src/common/datatypes/Value.cpp
tom-jerr: When inserting, the vector dimensions do not match, or when comparing vectors, the dimensions do not match, a separate error needs to be handled, so I added VEC_DIM_NOT_MATCH as a runtime error.

codesigner (approved):

Issue #6075, opened 2025-07-06
Implement schema with vector columns for DDL


tom-jerr:
# Schema Modification
To simply support original column and vector column, we need to modify the schema as follows:

1. Original columns (for other types except VECTOR type)
2. Vector columns (just for VECTOR type)
3. Schema Properties Options (TTL, TTL_COL, etc.)

PR #6076, opened 2025-07-06
Implement CREATE TAG statement for vector type

tom-jerr:
## What type of PR is this?
- [ ] bug
- [x] feature
- [ ] enhancement

#### Issue(s) number: 

Create tag with vector type properties
 #6040
Implement schema with vector columns for DDL #6075
Implement CREATE TAG statement for vector type #6078

#### Description:
1. Schema Provider Modification 

To simply support original column and vector column, we need to modify the schema provider as follows:

- Original columns (for other types except VECTOR type)
- Vector columns (just for VECTOR type)
- Schema Properties Options (TTL, TTL_COL, etc.)

2. RowWriter and RowReader

Then, we also need to modify RowWriter and RowReader to support vector column. The RowWriter will write the vector data into the RocksDB "vector" column family, and the RowReader will read the vector data from the RocksDB "vector" column family.

### Schema Provider Modification

To simply support original column and vector column, we need to modify the schema as follows:

1.  Original columns (for other types except VECTOR type)
2.  Vector columns (just for VECTOR type)
3.  Schema Properties Options (TTL, TTL_COL, etc.)

image

#### Vector type syntax
We've designed a dedicated syntax for vector types, representing a 3-dimensional vector as (0.1, 0.2, 0.3). Specifically, we've imposed a grammatical restriction that a vector type must have at least two dimensions. 

Internally, vector types are recognized as ConstantExpression for subsequent validation, planning, and execution.

#### Vector Property

- Use `type_lenght` to indicate the dimension of the vector type.

```cpp
struct ColumnTypeDef {
    1: required common.PropertyType    type,
    // type_length is valid for fixed_string type and vector type for dimension
    2: optional i16                    type_length = 0,
    // geo_shape is valid for geography type
    3: optional GeoShape               geo_shape,
}
```

We will validate that the default value matches the dimension of the vector property when we prepare to store the schema in the metadata (or metastore). 

### Modify the processor of graphd and metad
- Check vector property of schema utils
- Implement the description of schema consist of vector columns

### Modify the python test
- Add comparator for vector property
- Use new common generated python code
- Modify new parse of tck for vector property

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

TCK Tests:
Add new schema for vector property.

ddl_tck

Affects:

## Release notes:

> ex. Fixed the bug .....

### DIFF
--- a/src/common/datatypes/Vector.h
+++ b/src/common/datatypes/Vector.h
@@ -23,7 +23,8 @@ struct Vector {
   Vector() = default;
   Vector(const Vector& rhs) = default;
   Vector(Vector&& rhs) noexcept = default;
-  explicit Vector(std::vector<float> vals) : values(std::move(vals)) {}
+  explicit Vector(const std::vector<float>& vals) : values(vals) {}
+  explicit Vector(std::vector<float>&& vals) : values(std::move(vals)) {}
 
   Vector& operator=(const Vector& rhs) {
     if (this == &rhs) {
--- a/src/common/datatypes/VectorOps-inl.h
+++ b/src/common/datatypes/VectorOps-inl.h
@@ -45,8 +45,14 @@ uint32_t Cpp2Ops<nebula::Vector>::write(Protocol* proto, nebula::Vector const* o
   xfer += proto->writeStructBegin("Vector");
 
   xfer += proto->writeFieldBegin("values", apache::thrift::protocol::T_LIST, 1);
+  // Convert float vector to double vector for serialization
+  std::vector<double> doubleValues;
+  doubleValues.reserve(obj->values.size());
+  for (const auto& val : obj->values) {
+    doubleValues.emplace_back(static_cast<double>(val));
+  }
   xfer += detail::pm::protocol_methods<type_class::list<type_class::floating_point>,
-                                       std::vector<float>>::write(*proto, obj->values);
+                                       std::vector<double>>::write(*proto, doubleValues);
   xfer += proto->writeFieldEnd();
 
   xfer += proto->writeFieldStop();
@@ -67,9 +73,15 @@ void Cpp2Ops<nebula::Vector>::read(Protocol* proto, nebula::Vector* obj) {
   }
 
 _readField_values : {
-  obj->values = std::vector<float>();
+  // Read as double vector and convert to float vector
+  std::vector<double> doubleValues;
   detail::pm::protocol_methods<type_class::list<type_class::floating_point>,
-                               std::vector<float>>::read(*proto, obj->values);
+                               std::vector<double>>::read(*proto, doubleValues);
+  obj->values.clear();
+  obj->values.reserve(doubleValues.size());
+  for (const auto& val : doubleValues) {
+    obj->values.emplace_back(static_cast<float>(val));
+  }
 }
 
   if (UNLIKELY(!readState.advanceToNextField(proto, 1, 0, protocol::T_STOP))) {
@@ -110,9 +122,15 @@ uint32_t Cpp2Ops<nebula::Vector>::serializedSize(Protocol const* proto, nebula::
   xfer += proto->serializedStructSize("Vector");
 
   xfer += proto->serializedFieldSize("values", apache::thrift::protocol::T_LIST, 2);
-  xfer +=
-      detail::pm::protocol_methods<type_class::list<type_class::floating_point>,
-                                   std::vector<float>>::serializedSize<false>(*proto, obj->values);
+  // Convert float vector to double vector for size calculation
+  std::vector<double> doubleValues;
+  doubleValues.reserve(obj->values.size());
+  for (const auto& val : obj->values) {
+    doubleValues.emplace_back(static_cast<double>(val));
+  }
+  xfer += detail::pm::protocol_methods<type_class::list<type_class::floating_point>,
+                                       std::vector<double>>::serializedSize<false>(*proto,
+                                                                                   doubleValues);
   xfer += proto->serializedSizeStop();
   return xfer;
 }
@@ -124,9 +142,15 @@ uint32_t Cpp2Ops<nebula::Vector>::serializedSizeZC(Protocol const* proto,
   xfer += proto->serializedStructSize("Vector");
 
   xfer += proto->serializedFieldSize("values", apache::thrift::protocol::T_LIST, 2);
-  xfer +=
-      detail::pm::protocol_methods<type_class::list<type_class::floating_point>,
-                                   std::vector<float>>::serializedSize<false>(*proto, obj->values);
+  // Convert float vector to double vector for size calculation
+  std::vector<double> doubleValues;
+  doubleValues.reserve(obj->values.size());
+  for (const auto& val : obj->values) {
+    doubleValues.emplace_back(static_cast<double>(val));
+  }
+  xfer += detail::pm::protocol_methods<type_class::list<type_class::floating_point>,
+                                       std::vector<double>>::serializedSize<false>(*proto,
+                                                                                   doubleValues);
   xfer += proto->serializedSizeStop();
   return xfer;
 }
--- a/src/common/meta/NebulaSchemaProvider.h
+++ b/src/common/meta/NebulaSchemaProvider.h
@@ -141,9 +141,10 @@ class NebulaSchemaProvider {
   };
 
  public:
-  explicit NebulaSchemaProvider(SchemaVer ver) : ver_(ver), numNullableFields_(0) {}
+  explicit NebulaSchemaProvider(SchemaVer ver)
+      : ver_(ver), numNullableFields_(0), numVectorNullableFields_(0) {}
 
-  NebulaSchemaProvider() : ver_(0), numNullableFields_(0) {}
+  NebulaSchemaProvider() : ver_(0), numNullableFields_(0), numVectorNullableFields_(0) {}
 
   SchemaVer getVersion() const noexcept;
   // Returns the size of fields_
@@ -177,6 +178,26 @@ class NebulaSchemaProvider {
     return Iterator(this, getNumFields());
   }
 
+  size_t getVectorNumFields() const noexcept;
+  size_t getVectorNumNullableFields() const noexcept;
+  size_t vectorSize() const noexcept;
+
+  int64_t getVectorFieldIndex(const std::string& name) const;
+  const char* getVectorFieldName(int64_t index) const;
+
+  nebula::cpp2::PropertyType getVectorFieldType(int64_t index) const;
+  nebula::cpp2::PropertyType getVectorFieldType(const std::string& name) const;
+
+  const SchemaField* vectorField(int64_t index) const;
+  const SchemaField* vectorField(const std::string& name) const;
+
+  void addVectorField(const std::string& name,
+                      nebula::cpp2::PropertyType type,
+                      size_t fixedStrLen,
+                      bool nullable,
+                      std::string defaultValue,
+                      cpp2::GeoShape geoShape);
+
   void setProp(cpp2::SchemaProp schemaProp);
 
   const cpp2::SchemaProp getProp() const;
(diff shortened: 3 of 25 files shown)


czpmango:
Better to split this PR into two parts: DDL and DML, which are orthogonal functions and can be tested independently.

src/kvstore/test/RocksEngineMultiCFTest.cpp
```diff
@@ -0,0 +1,566 @@
+/* Copyright (c) 2018 vesoft inc. All rights reserved.
```
czpmango: 2018 -> 2025

tom-jerr:
# Vector Float Precision Discrepancy

I have an issue where Python natively only supports 64-bit floats. However, our Vector type is actually a collection of 32-bit floating-point numbers. So the serialization and deserialization method are set for 32-bit floating-point.

However some languages do not have native 32-bit float, so we can not serialized vector data as a list of 32-bit float-point.

So, my question is: Do we need to modify our Vector type to be a collection of 64-bit floating-point numbers, or how should we handle the serialization and deserialization process instead?

tests/tck/features/vector/Schema.feature
```diff
@@ -0,0 +1,38 @@
...
+  Scenario: describe vector tag
+    Given an empty graph
+    And create a space with following options:
+      | partition_num  | 1                |
+      | replica_factor | 1                |
+      | vid_type       | FIXED_STRING(20) |
+    # empty prop
+    When executing query:
+      """
+      CREATE TAG vectag1(id int, vec vector(3))
```
czpmango: add more test cases:
- default value
- nullable 
- primary key check
- comment
- ...

tests/tck/features/vector/Schema.feature
```diff
@@ -0,0 +1,38 @@
...
+      """
+    # desc tag
+    When executing query:
+      """
+      DESCRIBE TAG vectag1
+      """
+    Then the result should be, in any order:
+      | Field | Type     | Null  | Default | Comment |
+      | "id"  | "int64"  | "YES" | EMPTY   | EMPTY   |
+      | "vec" | "vector" | "YES" | EMPTY   | EMPTY   |
```
czpmango: add dimension

tests/tck/features/vector/Schema.feature
```diff
@@ -0,0 +1,38 @@
...
+      """
+    Then the execution should be successful
+    # if not exists
+    When executing query:
+      """
+      CREATE TAG IF NOT EXISTS vectag1(id int, vec vector(3))
+      """
+    Then the execution should be successful
+    # check result
+    When executing query:
```
czpmango, line 25: add more ddl:
- alter tag
- create edge
- ...

src/parser/parser.yy
```diff
@@ -1321,6 +1340,11 @@ type_spec
     | KW_SET LT KW_FLOAT GT {
         $$ = new meta::cpp2::ColumnTypeDef();
         $$->type_ref() = nebula::cpp2::PropertyType::SET_FLOAT;
+    } 
+    | KW_VECTOR L_PAREN INTEGER R_PAREN {
```
czpmango, line 1344: float

src/parser/parser.yy
```diff
@@ -749,6 +752,22 @@ constant_expression
...
         $$ = ConstantExpression::make(qctx->objPool(), $1);
     }
+    | L_PAREN vector_element_list R_PAREN {
+        $$ = ConstantExpression::make(qctx->objPool(), nebula::Vector(*$2));
+        delete $2;
+    }
+    ;
+
+vector_element_list
+    : DOUBLE COMMA DOUBLE {
```
czpmango: : DOUBLE COMMA DOUBLE -> : DOUBLE

tom-jerr: If changed to: DOUBLE, it would be impossible to distinguish between vectors (0.2)and bracketed operation expressions (0.2). For this reason, I changed the syntax like this, maybe we can also change the syntax for vectors to {double, double, double}

tests/tck/features/vector/Schema.feature
tom-jerr: How can we test primary key check?

src/parser/parser.yy
czpmango: try this:
```
double_literal
    : DOUBLE {
        $$ = $1;
    }

double_literal_list
    : double_literal {
        $$ = new std::vector<double>();
        $$->push_back($1);
    }
    | double_literal_list COMMA double_literal {
        $1->push_back($3);
        $$ = $1;
    }
    ;

vector_literal
    : KW_VECTOR L_BRACKET double_literal_list R_BRACKET {
        $$ = $3;
    }
    | KW_VECTOR L_PAREN double_literal_list R_PAREN {
        $$ = $3;
    }
```

tests/tck/features/vector/Schema.feature
czpmango: > How can we test primary key check?

vector type property cannot be a primary key, u can check this in `CreateTagValidator`

src/parser/parser.yy
tom-jerr: OK，I will try.

tests/tck/features/vector/Schema.feature
tom-jerr: According to the document and the code, the constraint of the index primary key should be handled in CreateTagIndex, there is no relevant description in CreateTag.

czpmango: ACK

src/common/datatypes/Vector.h
```diff
@@ -24,6 +24,12 @@ struct Vector {
   Vector(const Vector& rhs) = default;
   Vector(Vector&& rhs) noexcept = default;
   explicit Vector(std::vector<float> vals) : values(std::move(vals)) {}
+  explicit Vector(std::vector<double> vals) {
```
czpmango: const & and &&

src/common/datatypes/VectorOps-inl.h
```diff
@@ -67,9 +73,15 @@ void Cpp2Ops<nebula::Vector>::read(Protocol* proto, nebula::Vector* obj) {
   }
 
 _readField_values : {
-  obj->values = std::vector<float>();
+  // Read as double vector and convert to float vector
+  std::vector<double> doubleValues;
```
czpmango, line 77: Why cant use float?

src/common/meta/NebulaSchemaProvider.h
```diff
@@ -140,10 +140,60 @@ class NebulaSchemaProvider {
...
+   public:
+    const SchemaField& operator*() const {
+      return *field_;
+    }
+
+    const SchemaField* operator->() const {
+      return field_;
+    }
+
+    VectorIterator& operator++() {
```
czpmango: What the VectorIterator for?

src/interface/meta.thrift
```diff
@@ -90,9 +90,10 @@ struct SchemaProp {
     3: optional binary   comment,
 }
 
-struct Schema {
+struct Schema {  
     1: list<ColumnDef> columns,
     2: SchemaProp schema_prop,
+    3: optional list<ColumnDef> vector_columns,
```
czpmango: why we need extra field here?

src/kvstore/test/RocksEngineMultiCFTest.cpp
```diff
@@ -1,4 +1,4 @@
-/* Copyright (c) 2018 vesoft inc. All rights reserved.
+/* Copyright (c) 2025 vesoft inc. All rights reserved.
```
czpmango, line 1: revert this

src/interface/meta.thrift
tom-jerr: I think separating vector columns and other columns makes the logic clearer, simplifies modifications to RowReader and RowWriter, and makes DML processing easier.

src/kvstore/test/RocksEngineMultiCFTest.cpp
tom-jerr: OK

src/common/meta/NebulaSchemaProvider.h
tom-jerr: This is not in use for now, I will delete it first.

src/common/datatypes/VectorOps-inl.h
tom-jerr: In the original code generated by Thrift, serialization and deserialization for floating-point numbers all uses 64-bit floating-point numbers, and the Python client does not support 32-bit floating-point numbers, so here I have used 64-bit floating-point numbers for serialization and deserialization.

src/common/datatypes/Vector.h
tom-jerr: I will modify.

src/common/datatypes/VectorOps-inl.h
czpmango: ok

src/interface/meta.thrift
czpmango: Better to unify to columns, u can handle in metad side codes

src/common/meta/NebulaSchemaProvider.h
czpmango: ok

src/common/datatypes/Vector.h
```diff
@@ -23,7 +23,20 @@ struct Vector {
   Vector() = default;
   Vector(const Vector& rhs) = default;
   Vector(Vector&& rhs) noexcept = default;
-  explicit Vector(std::vector<float> vals) : values(std::move(vals)) {}
+  explicit Vector(const std::vector<float>& vals) : values(std::move(vals)) {}
```
czpmango: line 26 `values(std::move(vals))` takes no effect, modify it to `values(vals)`

src/interface/meta.thrift
tom-jerr: 1: list<ColumnDef> columns,
    2: SchemaProp schema_prop,
Do I need to keep this structure unchanged?

src/common/datatypes/Vector.h
tom-jerr: OK

src/common/datatypes/Vector.h
```diff
@@ -23,7 +23,20 @@ struct Vector {
...
+  explicit Vector(std::vector<float>&& vals) : values(std::move(vals)) {}
+  explicit Vector(const std::vector<double>& vals) {
```
czpmango: why we need a double ctor, let caller do this cast

src/meta/processors/schema/AlterEdgeProcessor.cpp
```diff
@@ -56,10 +56,19 @@ void AlterEdgeProcessor::process(const cpp2::AlterEdgeReq& req) {
   auto schema = MetaKeyUtils::parseSchema(schemasRaw[latestVersion]);
   auto columns = schema.get_columns();
   auto prop = schema.get_schema_prop();
-
+  bool hasVectorColumns = schema.get_vector_columns() != nullptr;
```
czpmango: can we avoid these special case codes？

src/common/datatypes/Vector.h
tom-jerr: This is for the parser to use, the parser receives double type data and then converts it to std::vector<double>. I will modify it to let parser cast double to float.

src/interface/meta.thrift
czpmango: Try not to modify thrift if not necessary, as it may cause some compatibility issues

tom-jerr: I understand, I will try by other way.

src/meta/processors/schema/AlterEdgeProcessor.cpp
tom-jerr: I will try.

Issue #6086, opened 2025-07-24
Implement DML statement  and process for vector type


tom-jerr:
# DML Syntax Design

We should offer insert, delete and update statement for vertexes and edges with vector property.

### Insert Vertex Sentence For Vector Type

To insert a vertex with a vector type property of the tag, we can use the following sentence:

```cypher
INSERT VERTEX tag1(id, vec1, vec2) VALUES 'v5':(5, vector (0.4, 0.5, 0.6), vector (0.1, 0.2, 0.3, 0.4)), 'v6':(6, vector (0.7, 0.8, 0.9), vector (0.2, 0.4, 0.6, 0.8));
```
### Delete Vertex Sentence For Vector Type

```cypher
Delete VERTEX on tag1 "v5" [with edge]
```

### Update Vertex Sentence For Vector Type

```cypher
UPDATE VERTEX on tag1 'v5', 'v6' SET vec1 = vector (0.1, 0.2, 0.3), vec2 = vector (0.4, 0.5, 0.6, 0.7);
```
### Insert Edge Sentence For Vector Type

```cypher
INSERT EDGE e1 (ev1) VALUES "v5"->"v6":(vector(1.1,1.2,1.3)); 
```

#### Delete Edge Sentence For Vector Type

```cypher
DELETE EDGE e1 "v5" -> "v6";
```

### Update Edge Sentence For Vector Type

```cypher
update EDGE on e1 "v5"->"v6" set ev1 = vector(1.1,1.2,1.3);
```

Issue #6082, opened 2025-07-13
Implement RowReader and RowWriter for vector property


tom-jerr:
# RowReader and RowWriter for vector property 
We modified the Schema to add a column for the vector property, so we need to provide an adapted RowReader and RowWriter for writing vector data to RocksDB.

PR #6087, opened 2025-07-30
Implement DML statement and process for vector type

tom-jerr:
## What type of PR is this?
- [ ] bug
- [x] feature
- [ ] enhancement

#### Issue(s) number: 
Support DML for vector type #6032 (Stage 2)
Implement RowReader and RowWriter for vector property #6082
Implement DML statement and process for vector type #6086

#### Description:
We added the ability to execute DML statements, including insert, delete, update, and upsert, for properties of VECTOR type.

### Storage
Storaged and Metad regularly use heartbeats to update the local Schema cache, thereby obtaining more recent Schema information. We need to modify the method for building Schema caches, adding vector columns to the SchemaProvider.

Storaged executes the `process` method of the mutate Processor to start handling DML requests, which involves the `doPut()` method of BaseProcessor, actually dealing with the writing to Raft-wal. We modify the process of handling logs in Raft.

image

#### Storaged Schema Cache Update
1. Storaged has the meta client, when metad is ready, storaged will get the schema from metad.
2. So if we want to insert vertices in a cluster, we need to add vector columns for the schema in the metaClient not only the metad.

image

#### Write data by Raft-wal
To ensure consistency in DML operations, we modify the system under the condition that each request uses only one asynchronous operation. We actually make these modifications during the Raft commitLog processing, primarily by checking the first byte of the key (**Type**) to determine whether to write to the vector column family.
    image

image

### RowReader and RowWriter for SchemaProvider

Actually, the manipulation of the schema is done in the `SchemaProvider`. So we need to add new data structure and iterator for vector columns in the `SchemaProvider`.

Input Data will be stored by RowWriter, which contains a `SchemaProvider*` pointer and a `isVector` flag to indicate whether the column is a vector type. By this way, we can encode the vector data and other data separately.

Reading data will be done by RowReader, which also contains a `SchemaProvider*` pointer and a `isVector` flag. The RowReader will read the vector data from the RocksDB "vector" column family if the `isVector` flag is true. Otherwise, it will read the data from the default column family.

image

### Storage Plan for Update
We have added management for the key, value, and RowReader of each Vertex/Edge VECTOR property in TagNode/EdgeNode, and increased the execution to return vector type data and RowReader to upstream nodes.

image

## Checklist:
Tests:
- [ ] Function test
- [ ] Performance test
- [ ] N/A

TCK Test:
- We do not add new tck test for vector dml statements recently.

dml_tck

Affects:

## Release notes:

> ex. Fixed the bug .....

### DIFF
--- a/src/codec/RowReaderV2.cpp
+++ b/src/codec/RowReaderV2.cpp
@@ -40,7 +40,10 @@ Value extractIntOrFloat(const folly::StringPiece& data, size_t& offset) {
   return Value(std::move(container));
 }
 
-bool RowReaderV2::resetImpl(meta::NebulaSchemaProvider const* schema, folly::StringPiece row) {
+bool RowReaderV2::resetImpl(meta::NebulaSchemaProvider const* schema,
+                            folly::StringPiece row,
+                            bool hasVectorCol,
+                            int32_t vectorIndex) {
   schema_ = schema;
   data_ = row;
 
@@ -59,11 +62,21 @@ bool RowReaderV2::resetImpl(meta::NebulaSchemaProvider const* schema, folly::Str
 #endif
 
   // Null flags
-  size_t numNullables = schema_->getNumNullableFields();
-  if (numNullables > 0) {
-    numNullBytes_ = ((numNullables - 1) >> 3) + 1;
+  if (!hasVectorCol) {
+    size_t numNullables = schema_->getNumNullableFields();
+    if (numNullables > 0) {
+      numNullBytes_ = ((numNullables - 1) >> 3) + 1;
+    } else {
+      numNullBytes_ = 0;
+    }
   } else {
-    numNullBytes_ = 0;
+    size_t numVecNullables = schema_->vectorField(vectorIndex)->nullable();
+    CHECK_LE(numVecNullables, 1);
+    if (numVecNullables > 0) {
+      numNullBytes_ = ((numVecNullables - 1) >> 3) + 1;
+    } else {
+      numNullBytes_ = 0;
+    }
   }
 
   return true;
@@ -74,6 +87,12 @@ bool RowReaderV2::isNull(size_t pos) const {
 
   size_t offset = headerLen_ + (pos >> 3);
   int8_t flag = data_[offset] & bits[pos & 0x0000000000000007L];
+  return flag != 0;
+}
+
+bool RowReaderV2::isVecNull() const {
+  size_t offset = headerLen_;
+  int8_t flag = data_[offset] & 0x80;
   return flag != 0;
 }
 
@@ -301,11 +320,49 @@ Value RowReaderV2::getValueByIndex(const int64_t index) const {
       return nebula::extractIntOrFloat<int32_t, Set>(data_, offset);
     case PropertyType::SET_FLOAT:
       return nebula::extractIntOrFloat<float, Set>(data_, offset);
-      case PropertyType::VECTOR:
-      // TODO(LZY)
-      break;
+    case PropertyType::VECTOR:
     case PropertyType::UNKNOWN:
       break;
+  }
+  LOG(FATAL) << "Should not reach here, illegal property type: " << static_cast<int>(field->type());
+  return Value::kNullBadType;
+}
+
+Value RowReaderV2::getVectorValueByName(const std::string& prop) const {
+  int64_t index = schema_->getVectorFieldIndex(prop);
+  LOG(ERROR) << "LZY getVectorValueByName() prop: " << prop << ", index: " << index;
+  return getVectorValueByIndex(index);
+}
+
+Value RowReaderV2::getVectorValueByIndex(const int64_t index) const {
+  if (index < 0 || static_cast<size_t>(index) >= schema_->getVectorNumFields()) {
+    return Value(NullType::UNKNOWN_PROP);
+  }
+
+  auto field = schema_->vectorField(index);
+  size_t offset = headerLen_ + numNullBytes_;
+
+  if (field->nullable() && isVecNull()) {
+    return NullType::__NULL__;
+  }
+
+  switch (field->type()) {
+    case PropertyType::VECTOR: {
+      int32_t vecOffset;
+      int32_t vecLen;
+      memcpy(reinterpret_cast<void*>(&vecOffset), &data_[offset], sizeof(int32_t));
+      memcpy(reinterpret_cast<void*>(&vecLen), &data_[offset + sizeof(int32_t)], sizeof(int32_t));
+
+      CHECK_LT(vecOffset, data_.size());
+      return Value{Vector(
+          std::vector<float>(reinterpret_cast<const float*>(data_.data() + vecOffset),
+                             reinterpret_cast<const float*>(data_.data() + vecOffset + vecLen)))};
+    }
+    case PropertyType::UNKNOWN:
+      break;
+    default: {
+      LOG(ERROR) << "Unsupported vector property type: " << static_cast<int>(field->type());
+    }
   }
   LOG(FATAL) << "Should not reach here, illegal property type: " << static_cast<int>(field->type());
   return Value::kNullBadType;
--- a/src/graph/validator/MutateValidator.cpp
+++ b/src/graph/validator/MutateValidator.cpp
@@ -77,13 +77,24 @@ Status InsertVerticesValidator::check() {
         names.emplace_back(propName);
       }
       propSize_ += numFields;
+      if (schema->hasVectorCol()) {
+        size_t numVecFields = schema->getVectorNumFields();
+        for (size_t i = 0; i < numVecFields; ++i) {
+          const char *vecPropName = schema->getVectorFieldName(i);
+          names.emplace_back(vecPropName);
+        }
+        propSize_ += numVecFields;
+      }
     } else {
       auto props = item->properties();
       // Check prop name is in schema
       for (auto *it : props) {
         if (schema->getFieldIndex(*it) < 0) {
-          LOG(ERROR) << "Unknown column `" << *it << "' in schema";
-          return Status::SemanticError("Unknown column `%s' in schema", it->c_str());
+          // check vector property
+          if (schema->getVectorFieldIndex(*it) < 0) {
+            LOG(ERROR) << "Unknown column `" << *it << "' in schema";
+            return Status::SemanticError("Unknown column `%s' in schema", it->c_str());
+          }
         }
         propSize_++;
         names.emplace_back(*it);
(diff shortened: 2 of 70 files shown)


vesoft-third-party-vecotr-3.3-x86_64-libc-2.31-gcc-9.4.0-abi-11.sh
czpmango, line 1: remove this

tom-jerr: OK

src/codec/RowReaderV2.cpp
```diff
@@ -301,11 +320,49 @@ Value RowReaderV2::getValueByIndex(const int64_t index) const {
...
+
+      CHECK_LT(vecOffset, data_.size());
+      return Value{Vector(
+          std::vector<float>(reinterpret_cast<const float*>(data_.data() + vecOffset),
+                             reinterpret_cast<const float*>(data_.data() + vecOffset + vecLen)))};
+    }
     case PropertyType::UNKNOWN:
       break;
+    default: {
+      LOG(ERROR) << "Unsupported vector property type: " << static_cast<int>(field->type());
```
czpmango, line 364: Log fatal here if this function getVectorValueByIndex can only handle vector property type

src/codec/RowReaderV2.cpp
```diff
@@ -301,11 +320,49 @@ Value RowReaderV2::getValueByIndex(const int64_t index) const {
...
+    case PropertyType::VECTOR:
+    case PropertyType::UNKNOWN:
       break;
+  }
+  LOG(FATAL) << "Should not reach here, illegal property type: " << static_cast<int>(field->type());
+  return Value::kNullBadType;
+}
+
+Value RowReaderV2::getVectorValueByName(const std::string& prop) const {
+  int64_t index = schema_->getVectorFieldIndex(prop);
```
czpmango, line 332: Will the SchemaProvider be updated if I execute the alter property name statement for the properties of vector type

src/codec/RowReaderV2.cpp
```diff
@@ -77,6 +90,12 @@ bool RowReaderV2::isNull(size_t pos) const {
   return flag != 0;
 }
 
+bool RowReaderV2::isVecNull() const {
+  size_t offset = headerLen_;
+  int8_t flag = data_[offset] & 0x80;
+  return flag != 0;
+}
```
czpmango, line 97: explain this code

src/codec/RowReaderV2.cpp
tom-jerr: After altering, the new schema will exist in metad. The meta client will periodically fetch the new schema from metad and cache it as SchemaProvider. I handled the vector type in the meta client.

src/codec/RowReaderV2.cpp
tom-jerr: Because each vector attribute is stored separately, distinguished by PropId, each RowReader essentially contains only one vector attribute. The way scalar attributes handle nulls is not suitable for vector handling, so a new function was built that detected whether this attribute is null or not.

czpmango:
Please add tck tests after u impl queries for vector type

src/graph/validator/MutateValidator.cpp
```diff
@@ -97,6 +108,7 @@ Status InsertVerticesValidator::check() {
 
 // Check validity of vertices data.
 // Check vid type, check properties value, fill to NewVertex structure.
+// The inserted value must follow the order of first other property, then vector property.
```
czpmango, line 111: Why is there such a limitation

src/graph/validator/MutateValidator.cpp
```diff
@@ -731,6 +786,8 @@ Status UpdateValidator::getUpdateProps() {
     std::string encodeStr;
     auto copyValueExpr = valueExpr->clone();
     NG_LOG_AND_RETURN_IF_ERROR(checkAndResetSymExpr(copyValueExpr, symName, encodeStr));
+    LOG(ERROR) << "lzy Update prop fieldName: " << fieldName
```
czpmango, line 789: remove these debug logs later

src/graph/validator/test/MutateValidatorTest.cpp
```diff
@@ -11,6 +11,14 @@ namespace graph {
 using PK = nebula::graph::PlanNode::Kind;
 class MutateValidatorTest : public ValidatorTestBase {};
 
+TEST_F(MutateValidatorTest, InsertVertexWithVectorTest) {
+  // vid use function call
+  {
+    auto cmd = "INSERT VERTEX person(name, vec) VALUES lower(\"TOM\"):(\"a\", (0.1,0.2,0.3));";
```
czpmango, line 17: support syntax for vector literal vector(1,2,3) instead of (1,2,3)

src/kvstore/RocksEngine.cpp
```diff
@@ -214,38 +214,17 @@ nebula::cpp2::ErrorCode RocksEngine::get(const std::string& key,
...
-    VLOG(4) << "Get: " << key << " Not Found";
-    return nebula::cpp2::ErrorCode::E_KEY_NOT_FOUND;
+  rocksdb::Status status;
+  if (NebulaKeyUtils::isVector(key)) {
+    auto it = cfHandleMap_.find(NebulaKeyUtils::kVectorColumnFamilyName);
+    if (it == cfHandleMap_.end()) {
+      LOG(ERROR) << "Column family " << NebulaKeyUtils::kVectorColumnFamilyName << " not found";
+      return nebula::cpp2::ErrorCode::E_UNKNOWN;
+    }
+    status = db_->Get(options, it->second, rocksdb::Slice(key), value);
```
czpmango, line 224: DCHECK here

src/kvstore/RocksEngine.cpp
```diff
@@ -411,62 +314,27 @@ nebula::cpp2::ErrorCode RocksEngine::prefixWithExtractor(const std::string& cfNa
...
-  rocksdb::ReadOptions options;
-  options.iterate_upper_bound = dynamic_cast<RocksPrefixIter*>(storageIter->get())->upperBound();
-  if (snapshot != nullptr) {
-    options.snapshot = reinterpret_cast<const rocksdb::Snapshot*>(snapshot);
-  }
-  // prefix_same_as_start is false by default
-  options.total_order_seek = FLAGS_enable_rocksdb_prefix_filtering;
-  auto it = cfHandleMap_.find(cfName);
-  if (it == cfHandleMap_.end()) {
-    LOG(ERROR) << "Column family " << cfName << " not found";
```
czpmango, line 444: Why remove these codes?

src/mock/MockCluster.cpp
```diff
@@ -266,20 +269,21 @@ std::unique_ptr<meta::SchemaManager> MockCluster::memSchemaMan(SchemaVer schemaV
...
+  //   indexMan->addEdgeIndex(spaceId, 102, 102, MockData::mockTeammateEdgeIndexColumns());
+  // }
 
-  indexMan->addTagIndex(spaceId, 1, 4, {});
-  indexMan->addTagIndex(spaceId, 2, 5, {});
-  indexMan->addEdgeIndex(spaceId, 101, 103, {});
-  indexMan->addEdgeIndex(spaceId, 102, 104, {});
+  // indexMan->addTagIndex(spaceId, 1, 4, {});
+  // indexMan->addTagIndex(spaceId, 2, 5, {});
+  // indexMan->addEdgeIndex(spaceId, 101, 103, {});
```
czpmango, line 285: do not break the previous tests

src/graph/validator/MutateValidator.cpp
tom-jerr: Actually, there is no such restriction; I forgot to remove this comment

src/graph/validator/MutateValidator.cpp
tom-jerr: Ok

src/mock/MockCluster.cpp
tom-jerr: Okay, I will resume this test.

src/kvstore/RocksEngine.cpp
tom-jerr: Now we will uniformly handle writes and deletions for different column families in the **raft commitLogs**. We can determine this using prefix keys, so there is no need to add extra member functions to the KVEngine.

src/storage/exec/EdgeNode.h
```diff
@@ -115,6 +115,30 @@ class FetchEdgeNode final : public EdgeNode<cpp2::EdgeKey> {
...
+    }
+    return ret;
+  }
+
+  std::vector<RowReaderWrapper*> vectorReaders() const {
+    std::vector<RowReaderWrapper*> ret;
+    for (auto& reader : vectorReaders_) {
+      ret.emplace_back(reader.get());
+    }
+    return ret;
```
czpmango, line 139: Can we read multiple vector properties in single RowReaderWrapper

src/storage/exec/QueryUtils.h
```diff
@@ -196,6 +243,20 @@ class QueryUtils final {
...
+      case PropContext::PropInKeyType::NONE: {
+        LOG(ERROR) << "LZY Read vertex vector prop, prop: " << prop.name_;
+        return readVectorValue(reader, prop.name_, prop.field_);
+      }
+      default:
+        LOG(FATAL) << "Should not read here";
+    }
+    return Status::Error(folly::stringPrintf("Invalid property %s", prop.name_.c_str()));
+  }
+
```
czpmango, line 259: Some bad smell here

src/storage/exec/EdgeNode.h
tom-jerr: I will try.

src/storage/mutate/AddVerticesProcessor.cpp
```diff
@@ -199,8 +237,19 @@ void AddVerticesProcessor::doProcessWithIndex(const cpp2::AddVerticesRequest& re
...
+        // Separate property names into regular and vector properties
+        std::vector<std::string> regularPropNames;
+        std::vector<std::string> vectorPropNames;
+        separatePropertyNames(propNames, schema, regularPropNames, vectorPropNames);
+
+        // Separate property values based on separated property names
+        std::vector<Value> regularProps;
+        std::vector<Value> vectorProps;
+        separatePropertyValues(props, regularProps, vectorProps);
+
```
czpmango, line 249: these codes increase the performance overhead if no vector property exists

czpmango: handle it in some init code path

tom-jerr: I will modify it.

src/storage/transaction/ChainAddEdgesLocalProcessor.cpp
```diff
@@ -8,6 +8,7 @@
 #include <thrift/lib/cpp/util/EnumUtils.h>
 
 #include "common/utils/DefaultValueContext.h"
+#include "common/utils/NebulaKeyUtils.h"
```
czpmango, line 11: remove unnecessary includes

Issue #6088, opened 2025-08-07
Design and implement a dedicated statement for Approximate Nearest Neighbor (ANN) search.


tom-jerr:
- [ ] #6089
- [ ] #6095
- [ ] DML for Vector Index
- [ ] ANN Benchmark

PR #6104, opened 2025-09-25
Ann search

tom-jerr:
## What type of PR is this?
- [ ] bug
- [x] feature
- [ ] enhancement

#### Issue(s) number: 
Design and implement a dedicated statement for Approximate Nearest Neighbor (ANN) search. #6088
#### Description:
Implement ANN search capability: if the current tag/edge has an ANN Index, the execution plan will first scan the ANN INDEX to obtain limit * K data and then filter with other data

## How do you solve it?
If the current tag/edge has an ANN Index, the execution plan will first scan the ANN INDEX to obtain limit * K data and then filter with other data. So we must modify the optimized rule and graphd physical plan and add new AnnIndexScan node in storaged.

### Overview
The overall execution process is as follows: 
1. The physical plan optimized by the optimization rules executes TAGANNindexscan; after storaged receives the request, it generates an internal plan to search on the ivf index and returns the results to graphd;。
2. Graphd calls storaged's getprop operator to generate a plan to obtain the non-vector properties of the nodes, and finally performs Limit and Project to obtain the final result

image

###  Graphd
- Do not return vector distance
`match (v:tag5:tag6) return v order by euclidean(vector(1.0,2.0,3.0), v.vec)  approximate limit 3 options {ANNINDEX_TYPE:'IVF', METRIC_TYPE:L2, NPROBE:3} `

Optimization rules as follows:

image

- Return vector distance
` match (v:tag5:tag6) return v, euclidean(vector(1.0,2.0,3.0), v.vec) order by euclidean(vector(1.0,2.0,3.0), v.vec)  approximate limit 1 options {ANNINDEX_TYPE:'IVF', METRIC_TYPE:L2, NPROBE:3}`

Optimization rules as follows:
image

### Storaged
Graphd's each executor will make remote calls to services on storaged via rpc, in fact, storaged starts its own executor, within which storaged plan is generated.

image

In fact, in storaged, the annindexscan process is very similar to the Indexscan process, which needs to be initialized first, then execute `doExecute`. The actual data acquisition is done by calling `resetIter,` where the results are obtained through ann index (here, t**he results are just vector IDs, not the actual vertex IDs**); the subsequent `doNext` simply iterates over the results, then calls `AnnIndexVertexScan::getVidByVectorid()` to obtain the actual vid data from rocksdb, and returns the vid and vector distance data together to graphd.

image

## Checklist:
Tests:
- [x] Unit test(positive cases)
- [ ] Function test
- [ ] Performance test
- [ ] N/A
## TODO
- tck test
Affects:

## Release notes:

> ex. Fixed the bug .....


src/common/expression/ApproximateLimitExpression.h
```diff
@@ -0,0 +1,76 @@
...
+
+class ApproximateLimitExpression final : public Expression {
+ public:
+  static ApproximateLimitExpression* make(ObjectPool* pool,
+                                          Expression* limitExpr,
+                                          AnnIndexQueryParamItem* queryParam) {
+    return pool->makeAndAdd<ApproximateLimitExpression>(pool, limitExpr, queryParam);
+  }
+
+  const Expression* limitExpr() const {
```
czpmango, line 23: Why implement the approx limit clause in the form of expression?

src/common/utils/NebulaKeyUtils.h
```diff
@@ -210,32 +214,35 @@ class NebulaKeyUtils final {
...
     constexpr int32_t len = static_cast<int32_t>(sizeof(NebulaKeyType));
     auto type = readInt<uint32_t>(rawKey.data(), len) & kTypeMask;
     return static_cast<NebulaKeyType>(type) == NebulaKeyType::kTag_;
   }
 
   static VertexIDSlice getVertexId(size_t vIdLen, const folly::StringPiece& rawKey) {
-    if (rawKey.size() != kTagLen + vIdLen) {
-      dumpBadKey(rawKey, kTagLen + vIdLen, vIdLen);
-    } else if (rawKey.size() > kTagLen + vIdLen) {
+    if (rawKey.size() > kTagLen + vIdLen) {
```
czpmango, line 226: don‘t change it.
plz use another interface for vector key if needed.

src/common/vectorIndex/IVFIndex.cpp
```diff
@@ -154,19 +154,25 @@ IVFIndex::IVFIndex(GraphSpaceID graphID,
...
   if (params == nullptr || res == nullptr) {
+    LOG(ERROR) << "IVFIndex::search failed: params or res is null";
     return Status::Error("SearchParams or SearchResult is null");
   }
   if (ivfIndex_ == nullptr) {
+    LOG(ERROR) << "IVF index is not initialized";
     return Status::Error("IVF index is not initialized");
   }
   if (params->topK < 1 || params->query == nullptr || params->queryDim != dim_) {
+    LOG(ERROR) << "Invalid SearchParams: topK=" << params->topK << ", queryDim=" << params->queryDim
```
czpmango, line 165: validate these errors at compile time rather than at runtime

src/common/vectorIndex/IVFIndex.cpp
```diff
@@ -154,19 +154,25 @@ IVFIndex::IVFIndex(GraphSpaceID graphID,
...
+               << ", expectedDim=" << dim_ << ", query is null: " << (params->query == nullptr);
     return Status::Error("Invalid SearchParams");
   }
   const auto *ivfParams = dynamic_cast<const SearchParamsIVF *>(params);
   if (ivfParams == nullptr) {
+    LOG(ERROR) << "SearchParams is not SearchParamsIVF";
```
czpmango, line 171: DCHECK

src/common/vectorIndex/IVFIndex.cpp
```diff
@@ -154,19 +154,25 @@ IVFIndex::IVFIndex(GraphSpaceID graphID,
...
     return Status::Error("SearchParams is not SearchParamsIVF");
   }
   if (ivfParams->nprobe < 1) {
+    LOG(ERROR) << "Invalid nprobe: " << ivfParams->nprobe;
```
czpmango, line 175: ditto

src/graph/context/ast/CypherAstContext.h
```diff
@@ -140,13 +141,19 @@ struct OrderByClauseContext final : CypherClauseContextBase {
...
 
 struct PaginationContext final : CypherClauseContextBase {
   PaginationContext() : CypherClauseContextBase(CypherClauseKind::kPagination) {}
 
   int64_t skip{0};
   int64_t limit{std::numeric_limits<int64_t>::max()};
+  bool isApproximateLimit{false};
+  AnnIndexType annIndexType{AnnIndexType::ANN_INVALID};
+  MetricType metricType{MetricType::METRIC_INVALID};
+  int64_t param{0};
```
czpmango, line 156: what if this `param` for?

czpmango:
Add some tck test cases.

src/graph/context/ast/CypherAstContext.h
tom-jerr: This is the ann search param, like ef_construction or nprobe

PR #6141, opened 2026-03-25
fix: fulfill snapshot promise on leadership change

Qian-Cheng-nju:
Hi, thanks for nebula!

We noticed an issue in SnapshotManager::sendSnapshot() and put together a fix — would appreciate a look when you get a chance.

When sendSnapshot() detects that leadership has changed (line 42-46), it returns early without fulfilling the promise `p` created at line 33. The promise destructor then triggers a BrokenPromise on the future side, but Host's cleanup code (sendingSnapshot_ = false at line 372-373) is only registered via .thenValue(), so it never runs on error. The Host ends up permanently stuck with sendingSnapshot_=true and can't send anything to that peer anymore.

The fix just adds p.setValue(Status::Error("Leader changed")) before the early return.

We also included a standalone reproduction (tests/nb5_snapshot_promise_wedge.cpp) that shows the wedging behavior using std::promise/std::future.


Qian-Cheng-nju:
recheck

PR #6163, opened 2026-08-01
fix #6138: fold constant unary expressions (e.g. IS NULL) in WHERE clause

### LINKED ISSUE #6138: Nebula crashes when running query with `WHERE 1 IS NULL` (control `WHERE false` works)
### Configuration
Nebula Version: 2025.10.28-nightly

Operating System: WSL2 – Ubuntu 24.04.1 LTS

Installation Method: Docker

---

### Steps to reproduce
1. Start Nebula via Docker
```
docker exec -it nebula-docker-compose-console-1 sh
```

2. Connect with `nebula-console`:
```
nebula-console -u root -p nebula -addr graphd -port 9669
```

3. Run the following query:

Create a test space
```
CREATE SPACE IF NOT EXISTS test(partition_num=1, replica_factor=1, vid_type=FIXED_STRING(8));
```

Select the space
```
use test
```

The buggy query:
```
(root@nebula) [test]> UNWIND range(1,10000) AS p WITH p LIMIT 1000 UNWIND range(1,2000000) AS qid WITH p, qid, {age: qid % 50 + 18, score: qid % 100} AS q WHERE 1 IS NULL RETURN p
2025/12/25 14:47:24 Loop error, failed to reconnect, failed to open transport, error: dial tcp 172.19.0.10:9669: connect: connection refused
```
A control Query:
```
(root@nebula) [test]> UNWIND range(1,10000) AS p WITH p LIMIT 1000 UNWIND range(1,2000000) AS qid WITH p, qid, {age: qid % 50 + 18, score: qid % 100} AS q WHERE false RETURN p
+---+
| p |
+---+
+---+
Empty set (time spent 5.615ms/6.29917ms)
```
---

### Expected behaviour
Both queries should return an empty result set, because both predicates are constant-false (`1 IS NULL` and `false`).
### Actual behaviour
The buggy query causes the client to lose the connection; The control query succeeds and returns an empty set.

lxcxjxhx:
## Motivation

The graphd service crashes when running a query whose `WHERE` clause contains a constant `IS NULL` predicate, e.g.:

```ngql
UNWIND range(1,10000) AS p WITH p LIMIT 1000
UNWIND range(1,2000000) AS qid WITH p, qid, {age: qid % 50 + 18, score: qid % 100} AS q
WHERE 1 IS NULL RETURN p
```

The control query `WHERE false` succeeds and returns an empty set, while `WHERE 1 IS NULL` is not constant-folded, so the filter cannot be optimized away and the whole 2M-row pipeline is materialized, eventually crashing the service (reported in #6138).

**Root cause:** `FoldConstantExprVisitor::visit(UnaryExpression*)` only treats a unary expression whose operand is already a constant as foldable for `kUnaryNegate` and `kUnaryPlus`. All the other pure unary kinds (`kIsNull`, `kIsNotNull`, `kIsEmpty`, `kIsNotEmpty`, `kUnaryNot`) are incorrectly treated as non-foldable, so `1 IS NULL` is never folded into `false` and the query does not behave like `WHERE false`.

## Changes

- `src/graph/visitor/FoldConstantExprVisitor.cpp`: allow folding any pure unary expression whose operand is a constant, excluding only the variable-modifying `kUnaryIncr`/`kUnaryDecr`. Now `1 IS NULL` folds to `false`, so it behaves exactly like `WHERE false`.
- `src/graph/visitor/test/FoldConstantExprVisitorTest.cpp`: add regression unit tests `TestIsNullExpr` and `TestIsNotNullExpr` verifying that `1 IS NULL => false`, `NULL IS NULL => true` and `1 IS NOT NULL => true` are folded into constants.
- `src/graph/visitor/test/VisitorTestBase.h`: add `isNullExpr`/`isNotNullExpr` test helpers.
- `tests/tck/features/bugfix/ConstantIsNullInWhere.feature`: end-to-end regression test for the crash.

## Testing

A full local build could not be run in the contributing environment (no C++ toolchain). Verification is provided by:

- New unit tests in `src/graph/visitor/test/FoldConstantExprVisitorTest.cpp`, runnable via `ctest -R FoldConstantExprVisitorTest` with `-DENABLE_TESTING=ON`.
- New TCK feature test `tests/tck/features/bugfix/ConstantIsNullInWhere.feature` asserting the query executes successfully.
- CI checks on this PR.

## Checklist

- [x] Code follows the project's coding standards
- [x] Tests have been added/updated
- [ ] Documentation has been updated (if applicable)
- [x] No new dependencies introduced (or justified)
- [x] Changes are backward compatible

### DIFF
--- a/src/graph/visitor/FoldConstantExprVisitor.cpp
+++ b/src/graph/visitor/FoldConstantExprVisitor.cpp
@@ -24,8 +24,12 @@ void FoldConstantExprVisitor::visit(UnaryExpression *expr) {
       if (!ok()) return;
     }
   } else {
-    canBeFolded_ = expr->kind() == Expression::Kind::kUnaryNegate ||
-                   expr->kind() == Expression::Kind::kUnaryPlus;
+    // When the operand is a constant, all pure unary expressions can be
+    // folded into constants, e.g. `1 IS NULL` should be folded into `false`.
+    // kUnaryIncr/kUnaryDecr modify variables, so they must not be folded
+    // even if the operand happens to be a constant expression.
+    canBeFolded_ = expr->kind() != Expression::Kind::kUnaryIncr &&
+                   expr->kind() != Expression::Kind::kUnaryDecr;
   }
 }
 
--- a/src/graph/visitor/test/FoldConstantExprVisitorTest.cpp
+++ b/src/graph/visitor/test/FoldConstantExprVisitorTest.cpp
@@ -45,6 +45,40 @@ TEST_F(FoldConstantExprVisitorTest, TestRelationExpr) {
   ASSERT_EQ(*root, *rootExpected) << root->toString() << " vs. " << rootExpected->toString();
 }
 
+TEST_F(FoldConstantExprVisitorTest, TestIsNullExpr) {
+  // 1 IS NULL => false
+  auto expr = isNullExpr(constantExpr(1));
+  FoldConstantExprVisitor visitor(pool);
+  expr->accept(&visitor);
+  ASSERT(visitor.canBeFolded());
+
+  auto root = visitor.fold(expr);
+  auto rootExpected = constantExpr(false);
+  ASSERT_EQ(*root, *rootExpected) << root->toString() << " vs. " << rootExpected->toString();
+
+  // NULL IS NULL => true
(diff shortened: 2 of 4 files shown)


lxcxjxhx:
自动关闭：PR 超过两周未活动

PR #6165, opened 2026-09-08
Set fixed GID/UID in Dockerfile for consistent host-container permissions

HarrisChu:
## What

Set fixed UID and GID in Dockerfile to ensure consistent file permissions between container and host.

## Changes

- Set user/group with specified IDs with `10001:10001`

## Why

Prevents permission errors when mounting volumes, especially on Linux hosts where UID/GID mismatch causes "permission denied" issues.


xiajingchun (approved):
LGTM

yuhongwei380 (approved):

